PR #10, opened 2016-09-09
Add first CSS-in-JS support

mxstbr:
```JS
styled.div({
  backgroundColor: 'blue'
})
```

is aliased to

```JS
css`${{
  backgroundColor: 'blue'
}}`
```

which works now!

So you can also do more complex, mixed things like
this:

```JS
styled.div`
border: none;
${{
  backgroundColor: 'blue'
}}
@media (max-width: 500px) {
  border: 1px solid awesome;
}
`
```


mxstbr:
This is done @geelen, moved the css-in-js logic to `css`. Good call on that one, it's so much nicer! This should be good to go now.

src/constructors/css.js
```diff
@@ -1,4 +1,5 @@
 import camelize from 'fbjs/lib/camelizeStyleName'
+import isObject from "lodash/isObject"
```
geelen: Should this be something about "simple object" or "flat object" if you're just iterating over keys & values? That way we don't end up setting `font-size` to `[object Object]` :)

geelen:
Left a comment on one line otherwise 👍  good to go

src/constructors/css.js
mxstbr: Good catch, totally forgot – should be `isPlainObject`! Fix coming, will merge, cut a new release and start working on integrating the PostCSS parser afterwards!

PR #11, opened 2016-09-09
Add AWESOME media query parsing support

### LINKED ISSUE #5: Media queries
Media queries should not require a user to understand partial imports and `css`. This should work out of the box:

```JS
const Component = styled.div`
  display: block;
  @media (max-width: 500px) {
    display: inline-block;
  }
`;
```

mxstbr:
This currently only handles the most simple of cases, I'm a bit stumped where to add support for more complex media queries with logical operators. Any pointers?

Closes #5

### DIFF
--- a/src/constructors/css.js
+++ b/src/constructors/css.js
@@ -1,13 +1,15 @@
 import camelize from 'fbjs/lib/camelizeStyleName'
 
 import rule from "./rule"
+import MediaQuery from "../models/MediaQuery"
 import RuleSet from "../models/RuleSet";
 import NestedSelector from "../models/NestedSelector";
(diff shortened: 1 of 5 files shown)


geelen:
Check out https://github.com/css-components/styled-components/blob/47d3a59bdd30a0564ccaed29aacca9c0eee57dff/src/models/MediaQuery.js#L12

It turns out you don't need to strip the brackets off the regexp. I'll push a commit and fix it.

geelen:
Wanna check this on a real project and merge it if you don't hit any problems?

mxstbr:
Did not hit any problems whatsoever, works perfectly fine for my usecase! Awesome!

PR #18, opened 2016-09-13
Integrate CI

mxstbr:


package.json
```diff
@@ -33,6 +33,7 @@
     "react": "^15.3.0"
   },
   "devDependencies": {
+    "babel-cli": "^6.14.0",
```
mxstbr: This was missing, so the `prepublish` step was failing on CI

package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "build": "babel --out-dir lib src",
     "build:watch": "npm run build -- --watch",
-    "test": "mocha 'src/**/*.test.js' --require babel-core/register",
+    "test": "mocha './src/**/*.test.js' --require babel-core/register",
```
mxstbr: Needed for Windows support

package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "build": "babel --out-dir lib src",
     "build:watch": "npm run build -- --watch",
-    "test": "mocha 'src/**/*.test.js' --require babel-core/register",
+    "test": "mocha \"./src/**/*.test.js\" --require babel-core/register",
```
mxstbr: This is needed for Windows support

PR #19, opened 2016-09-13
Use createElement directly in dev

### LINKED ISSUE #7: Components
Currently, doing this won't work (React throws an error that you have to return an element, null or etc.):

```JS
const Component = (props) => {
  return styled.div`
    background-color: ${props.bg};
  `;
}
```

You have to do this to make it work:

```JS
const Component = (props) => {
  const Primitive = styled.div`
    background-color: ${props.bg};
  `;
  return <Primitive />;
}
```

Which seems unnecessary, verbose and confusing. I'm not sure if we can do something against this though?

mxstbr:
Closes #7

**THIS SHOULD NOT BE MERGED**

This is very inconsistent behaviour. This will work in dev:

```JS
const Component = (props) => {
  return styled.div`
    background-color: ${props.bg};
  `;
}
```

**But it will break in production because we're returning a stateless functional component**

Imo this is the wrong solution, we cannot ship like this. Let's revert b2cc50d288d483a56e4b07977fb80e5f4397da2b and just add the option to tell us that you're using SSR, in which case the behaviour changes consistently for both prod and dev, and you'll just have to live with wrapper components.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,8 @@
     "eslint-plugin-jsx-a11y": "^2.2.2",
     "eslint-plugin-react": "^6.2.1",
     "expect": "^1.20.2",
-    "mocha": "^3.0.2"
+    "mocha": "^3.0.2",
+    "proxyquire": "^1.7.10"
   },
   "babel": {
     "presets": [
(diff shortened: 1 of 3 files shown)


mxstbr:
Closing since this is a bad idea. See #23 for some further dev.

PR #22, opened 2016-09-13
Ignore src directory on npm

mxstbr:
This'll make the installation a lot smaller because we're not pushing
the ES6 version of the code that people won't use anyway to npm.

Also ignores `.travis.yml` and `appveyor.yml`, those don't need to be on npm and come with each installation either!


geelen:
Nah I always prefer shipping ES6 to NPM. That way when you want to read the source of a project you have it right there in your `node_modules` directory.

Other than that, ignoring CI stuff is good.

geelen:
Let's get rid of the `src` exclusion and then merge this imo.

mxstbr:
Superseded by #32

PR #24, opened 2016-09-14
Dynamic values (proof of concept)

geelen:
Messssssssy atm but this is what I've been wanting to demonstrate the whole time. I think this is why we need tagged template literals, rather than just strings that are concatenated-at-definition-time

This now works:

```jsx
const A = styled.div`
  margin-top: 2rem;
  ${({hasBorder}) => hasBorder && 'border: 1px solid blue;'}
  color: ${props => props.color || 'black'};
  ${css`font-weight: bold;`}
  &:hover {
    text-transform: uppercase;
  }
`
```

```jsx
<A color="red">Omfg</A>
<A color="orange">Look</A>
<A hasBorder>At this shit</A>
```

image

Now... that `props` that we pass in... what if that merged both React context _and_ props. Looks for, say, a `theme` object on context then `Object.apply`s everything in props over the top. Basically turning the whole theming/contextual overrides into one simple conventional mechanism.

The implementation needs work but **I like this an alot**

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,8 @@
     "eslint-plugin-jsx-a11y": "^2.2.2",
     "eslint-plugin-react": "^6.2.1",
     "expect": "^1.20.2",
-    "mocha": "^3.0.2"
+    "mocha": "^3.0.2",
+    "proxyquire": "^1.7.10"
   },
   "babel": {
     "presets": [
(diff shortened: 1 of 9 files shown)


geelen:
The spirit of this lives on in #29 but with a much simpler implementation!

PR #25, opened 2016-09-16
Setup Flow Types

mxstbr:
I setup Flow, and typed most of the files.

The one I'm really struggling with is `css.js` (surprisingly, typing the meat of the library is hard!), would appreciate help!

Would love some help with this!

/cc @ryyppy

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,9 @@
 {
+  "parser": "babel-eslint",
   "extends": "airbnb",
   "rules": {
     "semi": [2, "never"],
+    "arrow-parens": 0,
     "class-methods-use-this": 0,
     "symbol-description": 0,
     "no-unused-vars": [2, { "varsIgnorePattern": "^_+$" }],
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,8 @@
+[ignore]
+.*/node_modules/fbjs/.*
+
+[include]
+
+[libs]
+
+[options]
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "babel-core": "^6.13.2",
     "babel-eslint": "^6.1.2",
     "babel-plugin-add-module-exports": "^0.2.1",
+    "babel-plugin-transform-flow-strip-types": "^6.14.0",
     "babel-preset-es2015": "^6.13.2",
     "babel-preset-es2016": "^6.11.3",
     "babel-preset-react": "^6.11.1",
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "build:watch": "npm run build -- --watch",
     "test": "mocha \"./src/**/*.test.js\" --require babel-core/register",
     "test:watch": "npm run test -- --watch",
+    "flow": "flow; test $? -eq 0 -o $? -eq 2",
     "lint": "eslint src",
     "prepublish": "rm -rf lib/* && npm run build"
   },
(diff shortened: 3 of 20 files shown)


.eslintrc
```diff
@@ -1,7 +1,9 @@
 {
+  "parser": "babel-eslint",
```
mxstbr: Needed for Flow types support

.flowconfig
```diff
@@ -0,0 +1,8 @@
+[ignore]
+.*/node_modules/fbjs/.*
```
mxstbr: For some reason flow cought errors in those files, so I had to ignore them

package.json
```diff
@@ -8,6 +8,7 @@
     "build:watch": "npm run build -- --watch",
     "test": "mocha \"./src/**/*.test.js\" --require babel-core/register",
     "test:watch": "npm run test -- --watch",
+    "flow": "flow; test $? -eq 0 -o $? -eq 2",
```
mxstbr: No idea what the whole `test $? -eq 0 -o $? -eq 2` thing does, just straight copied that from the flow website "Getting started" guide

src/constructors/styled.js
```diff
@@ -1,8 +1,9 @@
+// @flow
 import isPlainObject from 'lodash/isPlainObject'
 import css from './css'
 import element from './element'
 
-const styled = tagName => (...args) => {
+const styled = (tagName: string) => (...args: Array<any>) => {
```
mxstbr: This should probably not be `any`

src/constructors/toggle.js
```diff
@@ -13,8 +14,12 @@ import concat from './concat'
 * flex('inline vertical') => `display: flex; display: inline-flex; flex-direction: column;`
 * */
 
-export default (name, options) => (valueString = '') => {
-  const throwUnknown = (unknownValue) => {
+type Options = {
+  [name: string]: any,
```
mxstbr: This should maybe not be `any`

src/constructors/trait.js
```diff
@@ -1,8 +1,15 @@
...
 import concat from './concat'
 
-export default (name, definitions, cb) => {
+type Definitions = {
+  [category: string]: {
+    [option: string]: string
+  }
+}
+
+export default (name: string, definitions: Definitions, cb: Function): Function => {
```
mxstbr: This should _maybe_ not be `Function`? Could be more specific possibly?

src/constructors/element.js
```diff
@@ -1,8 +1,10 @@
+// @flow
 import { createElement } from 'react'
+import ValidRuleSetChild from '../models/ValidRuleSetChild'
 
 import Root from '../models/Root'
 
-const element = (tagName, ...rules) => {
+const element = (tagName: string, ...rules: Array<typeof ValidRuleSetChild>): Function => {
```
mxstbr: This should probably be `: createElement` instead of `: Function` I just realized

src/constructors/toggle.js
mxstbr: Fixed in d268930

src/constructors/styled.js
mxstbr: Fixed in d268930

src/constructors/element.js
mxstbr: Fixed in d268930

mxstbr:
I think this should be good to go for now, `constructors/css` is still missing but I'll try and land #24 and `@keyframes` support to get this ready for release.

package.json
ryyppy: `if $? == 0 || $? == 2` => pass... otherwise flow check threw an error...

src/constructors/trait.js
ryyppy: what are the parameters and return value of `cb`? I found something in the code like `cb(obj)` but I have no idea what `obj` is... soooo `cb: (obj: Object) => void`

ryyppy:
We should probably really go over this in person in our next meetup or so, since I have no idea what the library tries to achieve :-)

mxstbr:
Let's do that!

PR #26, opened 2016-09-20
[WIP] Implement @keyframes

### LINKED ISSUE #14: Keyframes
We need to get keyframes to work, maybe something like this?

```JS
styled.div`
@keyframes {
  0% { color: transparent; }
  100% { color: awesome; }
}
`;

mxstbr:
Implements `@keyframes` support. Since we aim to support all of CSS (contrary to Radium, Aphrodite,...) I figured having `@keyframes` support built-in without any helpers is what we should be doing.

The name mangling is very very rudimentary at the moment and will break if you put two different `@keyframes` into one `styled.something`. Need to find a better way to do this.

**TODO**:

- [ ] Proper name mangling
- [ ] Support `animation: some-name 150ms;` (currently only support `animation-name: bla;`)
- [ ] Way more tests for complex cases

Closes #14

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "aphrodite": "^0.5.0",
     "fbjs": "^0.8.3",
     "lodash": "^4.15.0",
+    "node-uuid": "^1.4.7",
     "react": "^15.3.0",
     "string-hash": "^1.1.0"
   },
--- a/package.json
+++ b/package.json
@@ -31,7 +31,8 @@
     "aphrodite": "^0.5.0",
     "fbjs": "^0.8.3",
     "lodash": "^4.15.0",
-    "react": "^15.3.0"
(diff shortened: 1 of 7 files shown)


geelen:
Hmm, yeah the name mangling stuff doesn't thrill me. We should adopt the same approach as we do with "normal" CSS, which is deferring the name generation until the contents are parsed, then hashing the contents to get the name. That way the same exact `@keyframes` definition produces the same hash twice.

Problem is, of course, you need that hashed name to inject into later declarations. And so you'd have to... somehow track the local keyframes names that you've defined and match them up later... hmm.

I think our implementation is hard because our design isn't good enough yet. I'm gonna jump back into #14 and jot some thoughts down.

geelen:
This needs to be redone against master now #29 is in

Issue #27, opened 2016-09-21
Parser implementation


geelen:
So I _think_ I've got a vague sense of the scope of the syntax we support.

```js
/* 
  Primary entry point, wraps everything in a hashed classname
  and returns a smart React component
*/
styled.tagName`
  simple: rules;
  @media (query) {
  }
  &:pseudo,
  &:selectors {}
  > nesting {
    as: well;
  }
`

/*
  Same syntax but can inject global CSS when you need it.
  (This could be a top-level export `global` instead of `styled.global`)
*/
styled.global`
html {
  font-size: 16px;
}
@font-face {
  probably: useful too;
}
`

/* 
  Special syntax for keyframes because they're global and need
  to be hashed like class names but aren't tied to a tag.
  Again, this syntax could totes change.
*/
styled.keyframes`
  0% { some: thing; }
  100% { something: else; }
`
```

Basically, at this point, that's totally valid CSS so an existing parser should work. But, it's not the whole story...

**Interpolations**

```js
const size = 100;
const sizeWithUnits = '1rem';
const wholeRule = 'color: black;'
const multipleRules = `display: flex; flex-direction;`
const withNesting = `
  &:hover,&:active {
    text-decoration: underline;
  }
`

styled.tagName`
  width: ${size}px;
  margin: ${sizeWithUnits};
  ${wholeRule}
  ${multipleRules}
  ${withNesting}
`
```

These are all static interpolations. We can just call `toString()` on all these before the parser runs. So they're fine.

```js
let ready = false;
const Tag = styled.tagName`
  ${ready ? 'display: block' : 'display: none'}
  ${props => `font-size: ${props.size}px;`}
  ${({bold}) => bold && 'font-weight: 900'}
`

export default () => {
  ready = true
  return <Tag size={18} bold/>
}
```

This is the problem, it follows on from my work in #24. I really think makes our approach _vastly_ more powerful than the CSS-in-JS approaches I've seen, so I want to preserve this "delayed" rendering of these properties. As far as I can tell, the way I've used `ready` will _never_ work because each interpolation is evaluated straight away. The only way to defer things is to interpolate a function:

```js
let ready = false;
const Tag = styled.tagName`
  ${() => ready ? 'display: block' : 'display: none'}
`
```

That seems ok to me. If you interpolate a function, we defer it to `render` time and evaluate it every time. If you interpolate a value, we `toString` it and continue the parser.

[...]

PR #29, opened 2016-09-23
Rebuild number... four?

geelen:
I've lost count how many times I've torn this down and started again. But I've had to build a lot less this time around than previously, which is a good thing.

I think we need a good way of testing this end-to-end. Maybe just test `renderToString()` or should we use Jest snapshots or something? Anyway, I'm not 100% that there weren't regressions so port your test projects to `@next` and try it out.

API as described in #27. Using a modified PostCSS parser and postcss nested to do the actual parsing, and _deferring all parsing_ until render time. `ComponentStyle` is where the action is, it almost certainly needs breaking up into a few different utils so we can implement `global` and `keyframes`.

Anyway, if your projects still work with this now we should get this merged and on `latest` and build up from here. Apart from perf issues because of the delayed parsing (which might actually not be a real problem in the end) this is going to be a WAY better place to start from.


geelen:
OH BTW I just implemented context so you can do this

```js
const Half = styled.div`
  padding: 1rem;
  > h2 {
    margin: 0;
    padding: 1rem;
  }
`

const SecondHalf = styled(Half)`
  background: #ccc;
  ${define({
    fg: 'white',
    bg: '#444'
  })}
`

const Button = styled.div`
  color: ${inherit('fg', 'black')}
  background: ${inherit('bg', 'none')}
  cursor: pointer;
  padding: 0.5em 1em;
  border: 1px solid;
`

export default () => (
  <Container>
    <Half>
      <h3>This is normal</h3>
      <Button>Ok</Button>
    </Half>
    <SecondHalf>
      <h3>This is inverted</h3>
      <Button>Ok</Button>
    </SecondHalf>
  </Container>
)
```



`define` and `inherit` are simple aliases and I'm definitely not sure about them yet. To do it manually you'd do:

```js
const SecondHalf = styled(Half)`
  background: #ccc;
  ${(props, context) => {
    context.fg = 'white'
    context.bg = '#444'
  }}
`

const Button = styled.div`
  color: ${(props, context) => context.fg || 'black'}
  background: ${(props, context) => context.bg || 'none'}
  cursor: pointer;
  padding: 0.5em 1em;
  border: 1px solid;
`
```

I did originally think about merging props and context when these functions get called but idk.

Anyway, I think context-passing is going to be the big thing that unlocks theming.

geelen:
Published at `styled-components@next` if you wanna play.

geelen:
Changed up the syntax for these function interpolations:

```js
const SecondHalf = styled(Half)`
  background: #ccc;
  ${props => props.updateTheme({
    fg: 'white',
    bg: '#444'
  })}
`

const Button = styled.div`
  color: ${props => props.theme.fg || 'black'}
  background: ${props => props.theme.bg || 'none'}
  cursor: pointer;
  padding: 0.5em 1em;
  border: 1px solid;
`
```

Better than the alternatives, published as `v0.1.6` or `@next`. I think this now PRETTY GREAT and I'm keen to get it shipped FOR REAL so we can start working on DOCUMENTATION AND OTHER THINGS

mxstbr:
Shipit

geelen:
@mxstbr can you look into this `appveyor` thing and merge this if it's all good?

geelen:
Version `0.1.9` published with tag `next` btw

mxstbr:
:tada: :tada: :tada: :tada:

PR #31, opened 2016-09-28
Preserving declaration order

### LINKED ISSUE #1: Order of CSS injection
Just hit a problem:

```js
const SVG = styled.svg`
  line {
    stroke-width: 1px;
  }
`

const Horizontal = styled.g`
  line {
    shape-rendering:crispEdges;
  }
`

const Vertical = styled.g`
  line {
    stroke-width: 1.5px;
  }
`

export default () => (
  <SVG>
    <Horizontal>
      <line x1={0} y1={0} x2={width} y2={0}/>
    </Horizontal>
    <Vertical>
      <line x1={0} y1={0} x2={width} y2={height}/>
    </Vertical>
  </SVG>
)
```

Because `Horizontal` and `Vertical` get rendered before `SVG`, and I'm only injecting the styles on render, `._compile_from_Vertical line {}` appears _before_ `._compiled_from_SVG line {}`. They have the same specificity of course and so `SVG` wins.

Injected source should be in the same order it is written, not in the order it's rendered.

geelen:
Inserting a null rule on ComponentStyle creation, then reusing that on each insert. It means that

```js
const X = styled.div` .. `
const Y = styled.div` ... `
const Z = styled.div` ... `
```

will be _guaranteed_ to be injected in `X`, `Y` and `Z` order, **regardless of which order they're rendered in**. This is a much more predictable behaviour, but I'm a bit worried about the performance of replacing styles at arbitrary points in the injected styles (since it could potentially invalidate a lot of the existing CSSOM).

Still, I can't think of a better way around it for now. Better to be correct and slow than fast & wrong, at least until we have hard data about performance to base an informed decision.

### DIFF
--- a/src/models/ComponentStyle.js
+++ b/src/models/ComponentStyle.js
@@ -12,6 +12,9 @@ const styleSheet = new StyleSheet({ speedy: false, maxLength: 40 })
 const generated = {}
 const inserted = {}
 
+interface UpdateableRule {
+}
+
 /*
  ComponentStyle is all the CSS-specific stuff, not
  the React-specific stuff.
@@ -19,11 +22,13 @@ const inserted = {}
 export default class ComponentStyle {
   rules: RuleSet
   insertedRule: Object
+  builtState: string
 
   constructor(rules: RuleSet) {
     this.rules = rules
     if (!styleSheet.injected) styleSheet.inject()
     this.insertedRule = styleSheet.insert('')
+    this.builtState = ''
   }
 
   /*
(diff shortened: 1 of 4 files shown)


geelen:
See #1 for the steps leading here.

mxstbr:
Yeah let's do it like this for now, let's worry about performance later.

threepointone:
why does preserving declaration order matter? especially when they're hashed based on contents? 

besides, because you're deduping, this won't guarantee preserving the order of similar rules declared across modules, yes?

threepointone:
reading up on #1

PR #33, opened 2016-09-28
Implement { global }

mxstbr:
- Rename "emoji" to "selector" – those aren't staying emojis :wink: (though we definitely need an opt-in "emoji mode"…)
- Add { global }

**NOTE: Haven't been able to actually test this in a project because linking locally isn't working for me. PLEASE TEST**


geelen:
Tested locally and it works fine. The injection code should probably be abstracted between this, `ComponentStyle` and the keyframe stuff, but that can come later.

PR #34, opened 2016-09-29
Docs

mxstbr:
*See the README live here: https://github.com/styled-components/styled-components/blob/docs/README.md*


README.md
```diff
@@ -1,62 +1,107 @@
 # Styled components
 
-**This is a work in progress** based off of [this demo](https://github.com/geelen/css-components-demo).
+The best way to style react apps!
```
mxstbr: Need a catchy tagline…

mxstbr:
Added lots more docs, including API documentation. Please review @geelen!

README.md
geelen: A true CSS–JS hybrid approach to styling

README.md
```diff
@@ -1,62 +1,272 @@
...
+### Keyframes
+
+Keyframes in CSS don't make sense to be scoped to a single component, they are meant to be global. This is why we export a `keyframes` helper which will generate a unique name for your keyframes. You can then use that unique name throughout your app.
+
+This way, you get all the benefits of using JavaScript, are avoiding name clashes and get your keyframes like always:
+
+```JS
+import { keyframes } from 'styled-components';
+
+const keyframes = {
```
geelen: having the word `keyframes` here is confusing because it's the same name you imported. Just `export const fadeIn` is better

docs/tips-and-tricks.md
```diff
@@ -0,0 +1,76 @@
...
+
+To adjust the placeholder color of an input, we need to use the <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder">`::placeholder`</a> pseudo element.
+
+> To attach pseudo elements to your styled component, you need to use `&`. For example, to style the placeholder:
+
+```JSX
+const Input = styled.input`
+  // Some other styles here…
+
+  // Style the placeholder
```
geelen: JS comments again ;)

docs/tips-and-tricks.md
```diff
@@ -0,0 +1,76 @@
...
+Which would be used like this:
+
+```JSX
+import placeholder from '../placeholder';
+
+const Input = styled.input`
+  // Old styles here…
+
+  // Style the placeholder
+  ${placeholder(`
```
geelen: no need for a `(` and `` ` `` here btw. Better to make `interpolate` a tagged template literal and forward the `rules` to `css` i.e. `&::placeholder { ${css(rules)} }`

README.md
geelen: Everybody chill. Styling is cool now 💅

http://i.imgur.com/pKckDkr.gifv

mxstbr:
TODO:

- [ ] `css`
- [ ] Theme stuff
- [ ] Badges (Travis, AppVeyor, Thinkmill)

mxstbr:
This is pretty much ready, please read through one more time, leave comments if necessary otherwise merge! (don't squash merge though! I'd like to preserve the history here.)

README.md
```diff
@@ -1,62 +1,397 @@
...
 
-**This is a work in progress** based off of [this demo](https://github.com/geelen/css-components-demo).
+Visual primitives for the component age. Use the best bits of ES6 and CSS to style your apps without stress 💅
 
-### Usage
+```
+npm install --save styled-components
+```
+
+[![Travis Build Status](https://travis-ci.org/styled-components/styled-components.svg?branch=master)](https://travis-ci.org/styled-components/styled-components) [![AppVeyor Build Status](https://ci.appveyor.com/api/projects/status/github/gruntjs/grunt?branch=master&svg=true)](https://ci.appveyor.com/project/mxstbr/styled-components) [![Supported by Thinkmill](https://thinkmill.github.io/badge/heart.svg)](http://thinkmill.com.au/?utm_source=github&utm_medium=badge&utm_campaign=styled-components)
```
geelen: Should we add a `npm v1.0.0` badge here too?

mxstbr: Not a fan of those badges, what do they add? ¯\_(ツ)_/¯ I'd much rather add one with the download count once we have a few downloads

geelen: Yeah fair enough. The latest version number is always on the releases tab anyway.

geelen:
Still super confused as to why we need namespacing inside the `theme` property. Why should a `Button` only listen to `myApp.fg` and `myApp.bg` instead of just `fg` and `bg`. It's inherited, not global, so if you need to translate components between naming conventions you can. The current implementation isn't preventing us but the docs always have this `myApp` nesting that I don't think we need.

geelen:
This hasn't really cleared things up tbh: https://github.com/styled-components/styled-components/blob/docs/docs/shared-component-libraries.md

I kinda feel like you're treating "theming" as actually something that could be handled better by a factory function. I.e. instead of doing

```js
<CustomThemeProvider>
  <Button/>
</CustomThemeProvider>
```

You'd be better off doing

```js
import { ButtonFactory } from 'some-framework'

const Button = ButtonFactory({ ... theme ... })
```

That's this kind of outside-in theming that Sass projects have been doing, which I think is really limited. The whole point for me for contextual styling is that I can publish a `Button` and it has 3 theming variables `fg`, `bg` and `roundedness`. It also has props of `primary` and `size` and a few others. I don't want the "theming" variables namespaced because _I'm not supposed to know anything about them_. Anything I have control over should be a prop.

This is also why I think a `ThemeProvider` should accept a `theme` prop that takes the outer theme as a function. If components have this contextual API then you need some way of translating keys & values from the conventions your site is already using. Namespacing doesn't seem to solve that problem, it just means you have a `theme` variable that ends up hundreds of lines long...

geelen:
In particular, the concept of "exporting a custom ThemeProvider" seems to be antithetical to the idea of contextual theming. Components should respond to the absence of theme variables, not require a base theme to be valid.

PR #35, opened 2016-09-29
Added lint-staged with autofix eslint config

okonet:
It will automatically fix eslint errors in pre-commit hook. It will also not allow commit code with eslint errors.


mxstbr:
Sure, LGTM

PR #37, opened 2016-09-30
e2e testing

mxstbr:
Added e2e testing, verifying that our public API works as expect.

Some tests don't pass, even though they are verifying that certain should be happening in the way we think they are – @geelen please take a look at all the `it.skip`s here!


src/models/ComponentStyle.js
```diff
@@ -32,21 +31,17 @@ export default class ComponentStyle {
...
+    if (!inserted[hash]) {
+      const selector = toEmoji(hash)
+      inserted[hash] = selector
       const root = parse(`.${selector} { ${flatCSS} }`)
       postcssNested(root)
-      generated[selector] = root.toResult().css
+      this.insertedRule.appendRule(root.toResult().css)
     }
-    return selector
+    return inserted[hash]
```
mxstbr: Need to merge these changes before I can land #14

PR #38, opened 2016-09-30
Add keyframes support

mxstbr:
Adds first support for

```JS
import { keyframes } from 'styled-components'

const fadeIn = keyframes`
  0% {
    opacity: 0;
  }
  100% {
    opacity: 1;
  }
`
```

We might need to use a different name generator (UUIDs aren't very _nice_) and we might need to optionally let people pass in a name that we prepend before our unique stuff for easier debugging.

**NOTE: UNTESTED IN REAL PROJECT, PLEASE VERIFY** (still can't `npm link`)

### DIFF
--- a/src/constructors/keyframes.js
+++ b/src/constructors/keyframes.js
@@ -1,0 +1,13 @@
+// @flow
+import uuid from 'node-uuid'
+import css from './css'
+import GlobalStyle from '../models/GlobalStyle'
+import type { Interpolation } from '../types'
+
+export default (strings: Array<string>, ...interpolations: Array<Interpolation>) => {
+  const name = uuid.v4()
+  const rules = css(strings, ...interpolations)
+  const keyframes = new GlobalStyle(rules, `@keyframes ${name}`)
+  keyframes.generateAndInject()
+  return name
+}
--- a/src/constructors/keyframes.js
+++ b/src/constructors/keyframes.js
@@ -8,7 +8,7 @@ const replaceWhitespace = (str: string): string => str.replace(/\s|\\n/g, '')
 
 export default (strings: Array<string>, ...interpolations: Array<Interpolation>): Number => {
   const rules = css(strings, ...interpolations)
-  const name = hashStr(replaceWhitespace(JSON.stringify(rules)))
+  const name = `k${hashStr(replaceWhitespace(JSON.stringify(rules)))}`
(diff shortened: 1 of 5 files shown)


src/constructors/keyframes.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+import uuid from 'node-uuid'
+import css from './css'
+import GlobalStyle from '../models/GlobalStyle'
+import type { Interpolation } from '../types'
+
+export default (strings: Array<string>, ...interpolations: Array<Interpolation>) => {
+  const name = uuid.v4()
+  const rules = css(strings, ...interpolations)
+  const keyframes = new GlobalStyle(rules, `@keyframes ${name}`)
```
geelen: We could make `ComponentStyle` build on top of `GlobalStyle` in this exact same way, right?

src/constructors/test/keyframes.test.js
```diff
@@ -0,0 +1,84 @@
...
+const stubbedSheet = {
+  StyleSheet: StubStylesheet,
+  '@global': true,
+}
+
+const stubbedKeyframes = proxyquire('../keyframes', {
+  '../vendor/glamor/sheet': stubbedSheet,
+  './vendor/glamor/sheet': stubbedSheet,
+  'glamor/lib/sheet': stubbedSheet,
+})
```
geelen: I think `../vendor/glamor/sheet` is the only one we really need.

geelen:
Rather than generating UUIDs we should hash the keyframes' contents just like `ComponentStyle` does. Then, whatever strategy we can come up with for generating nice names from a hash (tip: emoji 24/7) can be used consistently across everything.

src/constructors/keyframes.js
mxstbr: Yes very true, good point.

src/constructors/test/keyframes.test.js
mxstbr: Yeah, copy and pasted from my e2e testing setup. Will update!

mxstbr:
Updated naming to use a hash of the contents – can you test this, merge and publish if it works and we'll do the `GlobalStyles` and `ComponentStyle` merge in another PR?

*Note: The `JSON.stringify` of the object is literally what aphrodite does for all of their styles and name generation, so we should be good in case you're worried about that.*

PR #42, opened 2016-10-06
Fix flow

mxstbr:
- Component statics need to have a semicolon
- Added return type to interleave
- Fixed return type for keyframes
- Added eslint-plugin-flowtype-errors so flow errors are visible inline


amilajack:
Thanks for letting me know about the test failure! The reason why they're failing on windows is because windows 32bit OS's don't support flow. You have to [configure appveyor to run only on 64bit machines](https://github.com/amilajack/eslint-plugin-flowtype-errors/blob/master/appveyor.yml#L18). Also I believe that that travis tests are failing because of the npm version. I force mine to [run against npm@latest](https://github.com/amilajack/eslint-plugin-flowtype-errors/blob/master/.travis.yml#L17-L19).

Hope this helps! Also I'll add a section to the `eslint-plugin-flowtype-errors` readme section. Thanks for reporting this!

PR #53, opened 2016-10-10
React Native support (closes #28)

### LINKED ISSUE #28: ReactNative compatibility
This might be the killer feature. Making it possible to reuse CSS across platforms would be epic!

geelen:
But I was able to get this going. Import using

```js
import styled from 'styled-components/native'
```

This explodes horribly if you call it from your browser but I have no idea how to provide a better error message.

Tested (manually) that props, theming, and overriding using normal RN styles all work. Given these components:

```js
const Container = styled.View`
  flex: 1;
  justify-content: center;
  align-items: center;
  background-color: papayawhip;
  ${ props => props.theme.update({
    fg: 'palevioletred'
  })}
`
const Welcome = styled.Text`
  font-size: 20;
  text-align: center;
  margin: 10;
  color: tomato;
`
const Instructions = styled.Text`
  text-align: center;
  color: ${ props => props.theme.fg };
  margin-bottom: 5;
  ${ props => props.underlined && 'text-decoration-line: underline' }
`
const styles = StyleSheet.create({
  overrides: {
    fontWeight: 'bold'
  }
})
```

All the "Components" that React Native exports are aliased on `styled.X` (`styled.WebView`, `styled.Image` etc). I have no idea whether all of them accept styles but the error messaging is pretty good if you get it wrong.

When you render using this component:

```js
class NewNative extends Component {
  render() {
    return (
      <Container>
        <Welcome>
          Welcome to React Native!
        </Welcome>
        <Instructions style={styles.overrides}>
          To get started, edit index.ios.js
        </Instructions>
        <Instructions>
          Press Cmd+R to reload,{'\n'}
          Cmd+D or shake for dev menu
        </Instructions>
        <Instructions underlined>
          This should be underlined
        </Instructions>
      </Container>
    );
  }
}
```

The output is 😗 👌 

image

We're going to need some help from someone who knows their way around RN to make this really good but at least we've proven that it does work.


mxstbr:
This is so epic, holy cow

PR #54, opened 2016-10-10
Autoprefixer

geelen:
I'm just building on what Glamor uses, which isn't perfect but isn't bad. The only postcss one I could find was autoprefixer but it's way too heavy for our needs. Aphrodite have their own which looks ok but was a bit harder to work with than Glamor's, so I went with it.

### DIFF
--- a/src/models/ComponentStyle.js
+++ b/src/models/ComponentStyle.js
@@ -7,6 +7,7 @@ import flatten from '../utils/flatten'
 import parse from '../vendor/postcss-safe-parser/parse'
 import postcssNested from '../vendor/postcss-nested'
 import toEmoji from '../utils/toEmoji'
+import autoprefix from '../utils/autoprefix'
 
 const styleSheet = new StyleSheet({ speedy: false, maxLength: 40 })
 const inserted = {}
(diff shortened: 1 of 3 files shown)


somghosh:
Hello @geelen @mxstbr - thanks for this amazing new tool `styled-components`. However, I am using it with `react-boilerplate-3.4` and it seems like in react-boilerplate I no longer have the flexilbity of putting a line like specifying `autoprefixer` configurations. Is this correct? I believe this PR adds autoprefixer to styled-components, but how can I pass down the params from the react-boilerplate code into here? Thanks

    cssnext({ // Allow future CSS features to be used, also auto-prefixes the CSS...
              browsers: ['last 5 versions', 'IE > 8', 'Safari > 5', 'Firefox >= 25', 'iOS > 7'], // ...based on this browser list
              features: cssSettings,
            }),

mxstbr:
You don't need to do that, as this PR adds support for `inline-style-prefixer`, which prefixes everything. Since we are only applying prefixes at runtime, there's no harm done if we prefix too much, and since we're thusly prefixing everything it'll work in all old browsers!

somghosh:
Ah, thats good to know. Thanks @mxstbr

gaoqiang19514:
so cool!

PR #56, opened 2016-10-11
Theming v2

mxstbr:
This currently adds a (completely detached) `ThemeProvider` component and removes `theme.update`. (as per @gaearon's suggestion)

Discuss!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -8,7 +8,8 @@ import toggle from './constructors/toggle'
 import styled from './constructors/styled'
 import injectGlobal from './constructors/injectGlobal'
 import keyframes from './constructors/keyframes'
+import ThemeProvider from './models/ThemeProvider'
 
-export { css, toggle, injectGlobal, keyframes }
+export { css, toggle, injectGlobal, keyframes, ThemeProvider }
 
 export default styled
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -25,23 +25,13 @@ const createStyledComponent = (target: Target, rules: RuleSet) => {
     static rules: RuleSet
     static target: Target
 
-    getChildContext() {
-      return { theme: this.theme }
-    }
-
     componentWillMount() {
       this.componentWillReceiveProps(this.props, this.context)
     }
 
     componentWillReceiveProps(newProps: Object, newContext: ?any) {
       // Always pass down a theme, even if it's empty
-      this.theme = (newContext && newContext.theme) || {}
-      // Local copy for this instance with an update() method
-      const theme = Object.assign({}, this.theme, {
-        update: values => {
-          this.theme = Object.assign({}, this.theme, values)
-        },
-      })
+      const theme = (newContext && newContext.theme) || {}
 
       /* Generate and inject the styles and potentially update theme */
       const executionContext = Object.assign({}, newProps, { theme })
@@ -70,9 +60,6 @@ const createStyledComponent = (target: Target, rules: RuleSet) => {
   StyledComponent.target = target
 
   StyledComponent.displayName = isTag ? `styled.${target}` : `Styled(${target.displayName})`
-  StyledComponent.childContextTypes = {
-    theme: PropTypes.object,
-  }
   StyledComponent.contextTypes = {
     theme: PropTypes.object,
   }
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -1,9 +1,9 @@
-// @flow
+// @noflow
 import { Component, createElement, PropTypes } from 'react'
-import { Subscriber } from 'react-broadcast'
 
 import ComponentStyle from '../models/ComponentStyle'
 import validAttr from '../utils/validAttr'
+import { CHANNEL } from './ThemeProvider'
 
 import type { RuleSet, Target } from '../types'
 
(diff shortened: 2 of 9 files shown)


src/index.js
```diff
@@ -8,7 +8,8 @@ import toggle from './constructors/toggle'
 import styled from './constructors/styled'
 import injectGlobal from './constructors/injectGlobal'
 import keyframes from './constructors/keyframes'
+import ThemeProvider from './models/ThemeProvider'
```
gaearon: Can we just call this `<Theme>`?

src/models/StyledComponent.js
```diff
@@ -25,23 +25,13 @@ const createStyledComponent = (target: Target, rules: RuleSet) => {
...
     componentWillReceiveProps(newProps: Object, newContext: ?any) {
       // Always pass down a theme, even if it's empty
-      this.theme = (newContext && newContext.theme) || {}
-      // Local copy for this instance with an update() method
-      const theme = Object.assign({}, this.theme, {
-        update: values => {
-          this.theme = Object.assign({}, this.theme, values)
-        },
-      })
+      const theme = (newContext && newContext.theme) || {}
```
gaearon: This won't handle theme prop updates at the top if there is any pure component in the middle.
I think you want to pass an event emitter down instead.

src/index.js
mxstbr: Hmm, this would be a bit weird though:

```JSX
<Theme theme={{ bg: 'black' }} />
```

src/models/StyledComponent.js
mxstbr: Huh I didn't know pure components blocked context updates. How'd you mean pass an event emitter down?

mxstbr: Just switched to using `react-broadcast`, would this implementation now avoid that flaw?

gaearon: Probably, but as I noted:

>it's best to avoid using it so you don't add too many wrappers.

It's bad for perf if every styled div is actually two composite React components. We need to keep nesting as flat as possible. So it's better to reimplement that logic by hand (it's pretty simple).

mxstbr: Will do and test that too!

mxstbr:
Inlined the `react-broadcast` implementation, manually verified it works in simple cases.

Right now it's still `<ThemeProvider theme={{ bla: 'black' }} />` and `${props => props.theme}`, but this is already _much, much_ nicer than `props.theme.update`. What do we think about a separate `themed(mapThemeToProps)(StyledComp)` react-redux-style HoC?

geelen:
I'm ok if we wanna make `themed(obj)(Component)` available but I would prefer the default method is something that can move around your render tree a bit more easily. Remember, theming isn't just for site-wide changes, it's wherever you want a component to change its rendering. I.e.

```js
import { Base, Inverted } from "../themes" // both <ThemeProvider> components
import Header "./components/Header"
import Body from "./components/Body"

export default () => (
 <Base>
    <Inverted>
      <Header/>
    </Inverted>
    <main>
      <Body/>
    </main>
  </Base>
)
```

This way `Base` and `Inverted` can move around your document. Remember, react-redux is doing `context => props` whereas we're doing `props => context` so the `themed(obj)(Component)` has very little to do with `Component`. Still, I can imagine a file like:

```js
import { base, inverted } from "../themes" // both objects
const Component = //...

export const BaseComponent = themed(base)(Component)
export const InvertedComponent = themed(inverted)(Component)
```

But that seems to be missing the point somewhat.

mxstbr:
I used the implicit `props.theme` and it's quite nice, and the documentation . (also updated the documentation, see the recent changes in #34)

mxstbr:
I used the implicit `props.theme` and it's quite a nice DX! The API and thus the documentation is also a lot less confusing this way. (to see the updated documentation, check the recent changes in #34)

I added unit tests and e2e tests for the `ThemeProvider` component and tested it a bit in a real project and it works as expect. I haven't yet been able to test the native implementation since that doesn't work with `npm link` (😕) so please test that before merging, but other than that this should be good to go!

src/models/test/ThemeProvider.test.js
```diff
@@ -0,0 +1,117 @@
...
+    const child = (<p>Child!</p>)
+    const renderedComp = shallow(
+      <ThemeProvider theme={{ main: 'black' }}>
+        { child }
+      </ThemeProvider>
+    )
+    expect(renderedComp.contains(child)).toEqual(true)
+  })
+
+  it('should merge it\'s theme with an outer theme', (done) => {
```
geelen: `its` not `it's`

src/models/test/ThemeProvider.test.js
```diff
@@ -0,0 +1,117 @@
...
+    render(
+      <ThemeProvider theme={outerTheme}>
+        <ThemeProvider theme={innerTheme}>
+          <Child />
+        </ThemeProvider>
+      </ThemeProvider>
+    )
+  })
+
+  it('should merge it\'s theme with multiple outer themes', (done) => {
```
geelen: here too.

src/models/test/ThemeProvider.test.js
```diff
@@ -0,0 +1,117 @@
...
+  })
+
+  it('should merge it\'s theme with an outer theme', (done) => {
+    const outerTheme = { main: 'black' }
+    const innerTheme = { secondary: 'black' }
+    // Setup Child
+    class Child extends React.Component {
+      componentWillMount() {
+        this.context.broadcasts[CHANNEL](theme => {
+          expect(theme).toEqual(Object.assign({}, outerTheme, innerTheme))
```
geelen: bit confused about this here. Why isn't the assertion on the result of the `render`?

mxstbr: Because what do you want to test on the result of render? What we want to verify is that the themes merge, and we get the merged theme from the broadcast. That has nothing to do with render, I just render null so React doesn't error.

PR #63, opened 2016-10-13
docs(readme): Fix links on the table of content

antoinerey:
Fix the `theming.md` link

Also, the `shared-component-libraries.md` file does not exist. Do you still want to keep the link ?


mxstbr:
Ah we need to remove that. Would you mind submitting another PR removing the styling library link? Thanks very much!

antoinerey:
The other PR has just been submitted ;)

PR #78, opened 2016-10-14
Tighten dependencies

geelen:
Down from 65K to 23K min gzipped now.

@mxstbr can you merge, cut a `v1.0.1` release and just make sure I didn't break anything with react-native? (Or take a look at some automated testing so we don't have to do this manually?)


mxstbr:
Took two beta and one real publish but this is working on ReactNative too 🔥 shipped it!

PR #80, opened 2016-10-14
Fix#75 subscribe undefined 

### LINKED ISSUE #75: Styled-components break with react-router v4
Trying to make use of a styled-component with react-router v4 breaks with the following error

    StyledComponent.js:71 Uncaught TypeError: subscribe is not a function
    componentWillMount @ StyledComponent.js:71

Confirmed that this isn't an issue with react-router because
This works

    const Li = props => <li style={{display: "inline-block"}} {...props}/>

But this fails and throws the above error when used with react-router v4

    const Li = styled.li`
    display: inline-block`

madeinfree:
fixed #75


mxstbr:
Superseded by #86 – thank you so much for kicking that off though, much appreciated!

PR #84, opened 2016-10-15
Add ARc to README

diegohaz:
Hey, guys. Nice job you've done here.

I've just migrated a project from Radium to styled-components and added it to README.

Repo: https://github.com/diegohaz/arc
Demo: https://diegohaz.github.io/arc

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,6 +392,7 @@ We could use your help to get syntax highlighting support to other editors! If y
 ## Built with `styled-components`
 
 - [`grid-styled`](https://github.com/jxnblk/grid-styled): Responsive grid system ([demo](http://jxnblk.com/grid-styled/))
+- [ARc](https://github.com/diegohaz/arc): Atomic React App boilerplate with styled components ([demo](https://diegohaz.github.io/arc))
 
 *Built something with `styled-components`? Submit a PR and add it to this list!*
 


mxstbr:
Awesome, looks great! I'll share it on Twitter later today 👍

geelen:
Cool project!! I hadn't seen that trick with webpack's alias and having a bunch of `export * from './file', might give that a try in my next project!

Btw, how does this syntax feel for you (picked from [here](https://github.com/diegohaz/arc/blob/master/src/components/organisms/Hero/Hero.js)):

```js
/* Hero.js */

const Hero = styled((props) => {
  return (
    <div>{* ... *}</div>
  )
})`
  all: the;
  css: here;
`

export default Hero
```

I've tended to use this instead, which to me feels better cause I like my css above my renderer, so I thought I'd share it.

```js
/* Hero.js */

const Hero = styled.div`
  all: the;
  css: here;
`

export default props => (
  <Hero>{* ... *}</Hero>
)
```

Either works though 👍

diegohaz:
@geelen I'm still looking for the best way to do that 😄 

I will try what you said later, looks nicer and probably will solve a problem I had with tests:
```js
it('renders button by default', () => {
  // needs to shallow twice because my Button is wrapped inside StyledComponent
  const wrapper = shallow(<Button />).shallow()
  expect(wrapper.find('button').length).toBeGreaterThan(0)
})
```

diegohaz:
@geelen How would you deal with a component that renders different element based on the props passed? example:
```js
const Button = ({ kind, disabled, type, ...props, to, href }) => {
  if (href || to) {
    return <Link {...props} />
  }
  return <button {...props} type={type} disabled={disabled} />
}
```

I need to pass the styles to both of them.

PR #86, opened 2016-10-15
Fix react-router v4

### LINKED ISSUE #75: Styled-components break with react-router v4

mxstbr:
react-router uses react-broadcast to handle context updates. We do too,
except I inlined the implementation to avoid multiple wrapper
components. That would totally be fine, except I also slightly adapted
the implementation – and that broke anything else using react-broadcast,
including react-router v4.

This PR fixes that by instead attaching our theme to this.context
directly, not to this.context.broadcasts. This avoids overlap and fixes
the bug. (manually verified)

Closes #75

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "styled-components",
-  "version": "1.0.3",
+  "version": "1.0.4",
   "description": "Visual primitives for the component age. Use the best bits of ES6 and CSS to style your apps without stress 💅",
   "main": "lib/index.js",
   "scripts": {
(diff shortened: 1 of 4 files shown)


sheepsteak:
👍 LGTM

gbozee:
:+1:

madeinfree:
👍🏻

mxstbr:
Awesome, thanks for testing everyone! Publishing a new version now!

Issue #88, opened 2016-10-15
When a styled component is styled again it looses its defaultProps


diegohaz:
Reproduction: https://jsfiddle.net/zeoqxm8p/

mxstbr:
Wow, great catch! I'll see what we can do to mitigate that, very good catch!

diegohaz:
I'll start a PR with a failing test and try to fix it.

mxstbr:
Done by the amazing @diegohaz in #90! 👏

PR #90, opened 2016-10-15
Fix default props when extending styled components

diegohaz:
WIP

Reference: #88

### DIFF
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -13,10 +13,12 @@ class AbstractStyledComponent extends Component {
 }
 
 export default (ComponentStyle: any) => {
-  const createStyledComponent = (target: Target, rules: RuleSet) => {
+  const createStyledComponent = (target: Target, rules: RuleSet, parent?: Target) => {
     /* Handle styled(OtherStyledComponent) differently */
     const isStyledComponent = AbstractStyledComponent.isPrototypeOf(target)
-    if (isStyledComponent) return createStyledComponent(target.target, target.rules.concat(rules))
+    if (isStyledComponent) {
+      return createStyledComponent(target.target, target.rules.concat(rules), target)
+    }
 
     const isTag = typeof target === 'string'
     const componentStyle = new ComponentStyle(rules)
@@ -78,6 +80,7 @@ export default (ComponentStyle: any) => {
     /* Used for inheritance */
     StyledComponent.rules = rules
     StyledComponent.target = target
+    StyledComponent.defaultProps = parent && parent.defaultProps
 
     StyledComponent.displayName = isTag ? `styled.${target}` : `Styled(${target.displayName})`
     StyledComponent.contextTypes = {
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -22,8 +22,9 @@ export default (ComponentStyle: any) => {
 
     const isTag = typeof target === 'string'
     const componentStyle = new ComponentStyle(rules)
+    const ParentComponent = parent || AbstractStyledComponent
 
-    class StyledComponent extends AbstractStyledComponent {
+    class StyledComponent extends ParentComponent {
       static rules: RuleSet
       static target: Target
       state: {
@@ -80,7 +81,6 @@ export default (ComponentStyle: any) => {
     /* Used for inheritance */
     StyledComponent.rules = rules
     StyledComponent.target = target
-    StyledComponent.defaultProps = parent && parent.defaultProps
 
     StyledComponent.displayName = isTag ? `styled.${target}` : `Styled(${target.displayName})`
     StyledComponent.contextTypes = {
--- a/src/models/StyledNativeComponent.js
+++ b/src/models/StyledNativeComponent.js
@@ -11,16 +11,17 @@ class AbstractStyledNativeComponent extends Component {
   state: any
 }
 
-const createStyledNativeComponent = (target: Target, rules: RuleSet) => {
+const createStyledNativeComponent = (target: Target, rules: RuleSet, parent?: Target) => {
   /* Handle styled(OtherStyledNativeComponent) differently */
   const isStyledNativeComponent = AbstractStyledNativeComponent.isPrototypeOf(target)
   if (isStyledNativeComponent) {
-    return createStyledNativeComponent(target.target, target.rules.concat(rules))
+    return createStyledNativeComponent(target.target, target.rules.concat(rules), target)
   }
 
   const inlineStyle = new InlineStyle(rules)
+  const ParentComponent = parent || AbstractStyledNativeComponent
 
-  class StyledNativeComponent extends AbstractStyledNativeComponent {
+  class StyledNativeComponent extends ParentComponent {
     static rules: RuleSet
     static target: Target
     unsubscribe: Function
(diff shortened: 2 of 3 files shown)


mxstbr:
Can I tell you that you're one of the best contributors I've ever seen? Provides both a reproduction and a failing test – 👏👏👏 – I'm a fan.

Are you looking into fixing this? I think the correct way would be to use the check if it's a previous styled component and then doing `class StyledComponent extends PreviousStyledComponent` – but I'm not 100% sure if that works.

diegohaz:
I think the problem is [this line](https://github.com/styled-components/styled-components/blob/96d6bd547a8a9cd83b215736b1470a3bb0af4d22/src/models/StyledComponent.js#L19). When you create `StyledComponent` passing the `target.target` of the parent styled component, it looses other properties.

mxstbr:
Yeah exactly. I think instead of doing that, we should do 

```JS
let Extended
if (isStyledComponent) {
  Extended = target
} else {
  Extended = React.Component
}

class StyledComponent extends Extended {
  // etc.
}
```

or something like that maybe?

geelen:
Gotta merge those `RuleSet`s tho

diegohaz:
The `Extended` solution makes other tests to fail. :(

I managed to make it work in a simple way, but I'm not very used to these typings so maybe it needs some review.

diegohaz:
Yeah, I've tried to do that way, but I think I would need to understand better the whole program flow.

First, I've updated the StyledComponent.js file to this, but all "extending" tests fail, even the one I'm trying to fix:
```js
// @flow
import { Component, createElement, PropTypes } from 'react'

import validAttr from '../utils/validAttr'
import { CHANNEL } from './ThemeProvider'

import type { RuleSet, Target } from '../types'

export default (ComponentStyle: any) => {
  const createStyledComponent = (target: Target, rules: RuleSet, parent?: Target) => {
    const Extended = parent || Component
    /* eslint-disable react/prefer-stateless-function */
    class AbstractStyledComponent extends Extended {
      static isPrototypeOf: Function
      state: any
    }

    /* Handle styled(OtherStyledComponent) differently */
    const isStyledComponent = AbstractStyledComponent.isPrototypeOf(target)
    if (isStyledComponent) {
      return createStyledComponent(target.target, target.rules.concat(rules), target)
    }

    const isTag = typeof target === 'string'
    const componentStyle = new ComponentStyle(rules)

    class StyledComponent extends AbstractStyledComponent {
      static rules: RuleSet
      static target: Target
      state: {
        theme: any,
      }
      unsubscribe: Function

      constructor() {
        super()
        this.state = {
          theme: null,
        }
      }

      componentWillMount() {
        // If there is a theme in the context, subscribe to the event emitter. This
        // is necessary due to pure components blocking context updates, this circumvents
        // that by updating when an event is emitted
        if (this.context[CHANNEL]) {
          const subscribe = this.context[CHANNEL]
          this.unsubscribe = subscribe(theme => {
            // This will be called once immediately
            this.setState({ theme })
          })
        }
      }

      componentWillUnmount() {
        if (this.unsubscribe) {
          this.unsubscribe()
        }
      }

      /* eslint-disable react/prop-types */
      render() {
        const { className, children } = this.props
        const theme = this.state.theme || {}
        const executionContext = Object.assign({}, this.props, { theme })

        const generatedClassName = componentStyle.generateAndInjectStyles(executionContext)
        const propsForElement = {}
        /* Don't pass through non HTML tags through to HTML elements */
        Object.keys(this.props)
          .filter(propName => !isTag || validAttr(propName))
          .forEach(propName => {
            propsForElement[propName] = this.props[propName]
          })
        propsForElement.className = [className, generatedClassName].filter(x => x).join(' ')

        return createElement(target, propsForElement, children)
      }
    }

    /* Used for inheritance */
    StyledComponent.rules = rules
    StyledComponent.target = target

    StyledComponent.displayName = isTag ? `styled.${target}` : `Styled(${target.displayName})`
    StyledComponent.contextTypes = {
      [CHANNEL]: PropTypes.func,
    }
    return StyledComponent
  }

  return createStyledComponent
}
```

The test output is:
```
1) e2e extending should generate a single class if only parent has styles:

      Error: Expected '.a { color: blue; } .b { }' to equal '.a { color: blue; }'
      + expected - actual

      -.a { color: blue; } .b { }
      +.a { color: blue; }

      at assert (node_modules/expect/lib/assert.js:29:9)
      at Expectation.toEqual (node_modules/expect/lib/Expectation.js:81:30)
      at Context.<anonymous> (src/test/e2e.test.js:178:61)

  2) e2e extending should generate a single class if only child has styles:

      Error: Expected '.a { color: blue; } .b { }' to equal '.a { color: blue; }'
      + expected - actual

      -.a { color: blue; } .b { }
      +.a { color: blue; }

      at assert (node_modules/expect/lib/assert.js:29:9)
      at Expectation.toEqual (node_modules/expect/lib/Expectation.js:81:30)
      at Context.<anonymous> (src/test/e2e.test.js:188:61)

  3) e2e extending should generate a class for the child with the rules of the parent:

      Error: Expected '.a { color: red; }' to equal '.a { color: blue;color: red; }'
      + expected - actual

      -.a { color: red; }
      +.a { color: blue;color: red; }

      at assert (node_modules/expect/lib/assert.js:29:9)
      at Expectation.toEqual (node_modules/expect/lib/Expectation.js:81:30)
      at Context.<anonymous> (src/test/e2e.test.js:197:61)

  4) e2e extending should generate different classes for both parent and child:

      Error: Expected '.a { color: blue; } .b { color: red; }' to equal '.a { color: blue; } .b { color: blue;color: red; }'
      + expected - actual

      -.a { color: blue; } .b { color: red; }
      +.a { color: blue; } .b { color: blue;color: red; }

      at assert (node_modules/expect/lib/assert.js:29:9)
      at Expectation.toEqual (node_modules/expect/lib/Expectation.js:81:30)
      at Context.<anonymous> (src/test/e2e.test.js:207:61)

  5) e2e extending should copy nested rules to the child:

      Error: Expected '.a { color: blue; } .a > h1 { font-size: 4rem; } .b { color: red; }' to equal '.a { color: blue; } .a > h1 { font-size: 4rem; } .b { color: blue; color: red; } .b > h1 { font-size: 4rem; }'
      + expected - actual

      -.a { color: blue; } .a > h1 { font-size: 4rem; } .b { color: red; }
      +.a { color: blue; } .a > h1 { font-size: 4rem; } .b { color: blue; color: red; } .b > h1 { font-size: 4rem; }

      at assert (node_modules/expect/lib/assert.js:29:9)
      at Expectation.toEqual (node_modules/expect/lib/Expectation.js:81:30)
      at Context.<anonymous> (src/test/e2e.test.js:220:61)

  6) e2e extending should keep default props from parent:

      Error: Expected '.a { color: red; } .b { background-color: green; }' to equal '.a { color: red; } .b { color: red; background-color: green; }'
      + expected - actual

      -.a { color: red; } .b { background-color: green; }
      +.a { color: red; } .b { color: red; background-color: green; }

      at assert (node_modules/expect/lib/assert.js:29:9)
      at Expectation.toEqual (node_modules/expect/lib/Expectation.js:81:30)
      at Context.<anonymous> (src/test/e2e.test.js:241:61)
```

Do you have any idea?

diegohaz:
Actually, I think I figured it out.

mxstbr:
I was just about to take this, but if you figured it out feel free to push!

geelen:
@diegohaz you hero!!

PR #93, opened 2016-10-15
Move React from dependencies to peerDependencies

sheepsteak:
At the moment React is in `dependencies` which means `15.3.2` is being forced on everyone who uses `styled-components`. They'll end up having two versions of React in their `node_modules` if they aren't on the same version. I believe this can cause problems with Webpack (this may be fixed now).

This change moves React into `peerDependencies` like [most](https://github.com/reactjs/react-redux/blob/master/package.json#L98) [other](https://github.com/yahoo/react-intl/blob/master/package.json#L56) libraries that need React do. I also allowed React 0.14.x to be used and apart from some failing tests in `validAttr` (I've commented them out so you can see which ones) everything seemed fine. I believe the failing tests are just because React 15 has a lot better support for SVG and a few newer HTML attributes.

React has also been added to `devDependencies` to be more convenient during development and to avoid peer dependency warnings.

So the main benefit of all this is more people can use `styled-components` even if they're on an older version of React.


src/utils/test/validAttr.test.js
```diff
@@ -164,392 +164,392 @@ describe('validAttr', () => {
...
+  //   expect(validAttr('onLoad')).toEqual(true)
+  //   expect(validAttr('onLoadCapture')).toEqual(true)
+  //   expect(validAttr('onAnimationStart')).toEqual(true)
+  //   expect(validAttr('onAnimationStartCapture')).toEqual(true)
+  //   expect(validAttr('onAnimationEnd')).toEqual(true)
+  //   expect(validAttr('onAnimationEndCapture')).toEqual(true)
+  //   expect(validAttr('onAnimationIteration')).toEqual(true)
+  //   expect(validAttr('onAnimationIterationCapture')).toEqual(true)
+  //   expect(validAttr('onTransitionEnd')).toEqual(true)
+  //   expect(validAttr('onTransitionEndCapture')).toEqual(true)
```
sheepsteak: React 0.14 [does support](https://github.com/facebook/react/blob/0.14-stable/src/renderers/dom/shared/SVGDOMPropertyConfig.js#L24) some of these but it was really fiddly working out which ones so I took the nuclear option for now.

mxstbr:
I've been wondering if this works as expected, even in the UMD build. Can you test that?

sheepsteak:
It works as expected for me. React is in the `devDependencies` so you have it when you generate the UMD build.

src/utils/test/validAttr.test.js
```diff
@@ -164,245 +164,245 @@ describe('validAttr', () => {
...
+  //   expect(validAttr('xmlns')).toEqual(true)
+  //   expect(validAttr('xmlnsXlink')).toEqual(true)
+  //   expect(validAttr('xmlLang')).toEqual(true)
+  //   expect(validAttr('xmlSpace')).toEqual(true)
+  //   expect(validAttr('y')).toEqual(true)
+  //   expect(validAttr('y1')).toEqual(true)
+  //   expect(validAttr('y2')).toEqual(true)
+  //   expect(validAttr('yChannelSelector')).toEqual(true)
+  //   expect(validAttr('z')).toEqual(true)
+  //   expect(validAttr('zoomAndPan')).toEqual(true)
```
wmertens: Maybe these can be made conditional on `React.version`?

sheepsteak: I was just thinking about this again and wondered if we need to do anything? If we leave all the tests in and someone clones the repo and runs them then they'll pass because at the moment they get `react@15.3.2` from the `devDependencies`. It's only if they go back to an old React 0.14 version that the tests will fail.

@mxstbr, let me know what you'd prefer to have. I can put these failing tests in a conditional like @wmertens suggested (`if (React.version.indexOf('0.14') !== 0) {…}` maybe?) or just leave them as they were and they'll still pass on current versions of React 😄

JamieDixon:
I think it's a good idea to move the react dependency to `peerDependencies`. As for the tests, what are they really testing? It seems like they test React itself more than they test this framework. Thoughts?

mxstbr:
> As for the tests, what are they really testing?

They are testing that we allow all valid attributes through to the DOM node. That these tests break on react 0.14 means that people will have a horrible experience using styled-components with it, because we won't allow attributes to be attached to the DOM node underneath.

Can we maybe inline the react 15.0 list? We should do that anyway, and it'd mean nothing breaks on 0.14!

sheepsteak:
OK, I've inlined the list from React 15. I think it's the best thing to do, especially after [this comment](https://github.com/facebook/react/issues/7770#issuecomment-253899837). It should also reduce file size as other things were being pulled in from the imported files that weren't being used.

Original tests pass with both React 0.14 and 15.

mxstbr:
Awesome, thanks!

gaearon:
Thank you!

Issue #87, opened 2016-10-15
Super expression is undefined when using rollup


NicolasParada:
When used with rollup it throws an error at runtime:
```
Uncaught TypeError: Super expression must either be null or a function, not undefined
    at _inherits (root.js:25)
    at root.js:38
    at root.js:161
    at createCommonjsModule (index.js:3)
    at processor.js:254
```

With webpack works fine. And the app works fine until I add `styled-components`.
I wasn't sure if issue it here since it can be a problem with `rollup-plugin-commonjs` or even `react`...
I left a reproduction repo here: [NicolasParada/styled-components](https://github.com/NicolasParada/styled-components)

mxstbr:
This very much sounds like it's not an issue with `styled-components` but with `rollup-plugin-commonjs` – @Rich-Harris might have some input here?

Rich-Harris:
Okay, I've done some digging. It looks like this is one of this incredibly frustrating interop headaches caused by back-and-forth conversion between CommonJS and ES modules. Specifically, it looks like the `add-module-exports` Babel plugin is causing havoc, because it violates the expectation in rollup-plugin-commonjs that a module with `__esModule: true` will have `exports.default` rather than `module.exports`.

The good news is that it's easy to fix – just use the dist file rather than bundling the library yourself:

```js
import styled from 'styled-components/dist/styled-components.js'
```

<details>
  <summary>Or, use the existing import but update your Rollup config:</summary>

```js
import { resolve } from 'path'
const styledComponentsDist = resolve(__dirname, 'node_modules/styled-components/dist/styled-components.js')

export default {
  entry: 'src/main.js',
  dest: 'dist/bundle.js',
  preferConst: true,
  plugins: [
    {
      resolveId (importee) {
        if (importee === 'styled-components') return styledComponentsDist;
      }
    },
    replace({
      'process.env.NODE_ENV': JSON.stringify('development')
    }),
    ...
```
</details>

Slightly off-topic, but as part of my digging I experimented with building this library with Rollup rather than Webpack (so that it could export an ES module bundle for the benefit of people using Webpack 2 or Rollup like @NicolasParada). The UMD build is 13% smaller (20,549 bytes gzipped down from 23,553) and parses roughly 60% quicker (~8ms in Chrome down from ~20ms, similar increase in Safari, less in Firefox). Any interest in a PR? Totally cool if not but the offer's there 😀

mxstbr:
Yes yes yes please submit a PR!!

[...]

PR #96, opened 2016-10-15
Bundle with Rollup

Rich-Harris:
As discussed in #87. This change only affects people who are using the `dist` files (their app will get smaller and boot faster) and people using Webpack 2 or Rollup (allows them to consume the library as an ES module).


rollup.config.js
```diff
@@ -0,0 +1,70 @@
...
+import json from 'rollup-plugin-json'
+import flow from 'rollup-plugin-flow'
+import uglify from 'rollup-plugin-uglify'
+
+const processShim = '\0process-shim'
+
+const prod = process.env.PRODUCTION
+const mode = prod ? 'production' : 'development'
+
+console.error(`creating ${mode} bundle...`)
```
mxstbr: Why is this a `console.error`? (can we also capitalize the first `c`?)

rollup.config.js
```diff
@@ -0,0 +1,70 @@
...
+  {
+    resolveId(importee) {
+      if (importee === processShim) return importee
+      return null
+    },
+    load(id) {
+      if (id === processShim) return 'export default { argv: [], env: {} }'
+      return null
+    },
+  },
```
mxstbr: Can we get some comments in the file what this whole block is needed for?

rollup.config.js
Rich-Harris: Just because that prints to `stderr` (i.e. metadata) rather than `stdout` (i.e. data). No actual difference, just force of habit! Can change to `console.log` (or remove altogether) if you prefer

mxstbr: It makes me feel weird to `console.error` on every build, I'd prefer it to be `console.log`! 😉

Rich-Harris:
> Can we have some before/after stats? (or are they the same as in the original comment, i.e. 13% smaller UMD build and 60% quicker parse?)

Yep, nothing has changed from earlier

rollup.config.js
Rich-Harris: done

rollup.config.js
Rich-Harris: fixed

PR #98, opened 2016-10-16
More documentation

### LINKED ISSUE #85: CSS overriding, className vs StyledComponent
Passing a `className` prop to a StyledComponent works fine, but if a specific CSS property (e.g. `background-color`) is passed in via a `className` prop and also defined in the StyledComponent, the StyledComponent will win.

In the example below `<div>` will have a green background. Should the CSS properties sent via the `className` prop override any matching CSS properties in the StyledComponent? Either way, it would be great to have this documented, happy to submit a PR.

```javascript
// StyledUpYo.js
const StyledUpYo = styled.div`background-color: green;`;
```
```javascript
// styles.css
.styledUpYo {
  background-color: red;
}
```
```javascript
import StyledUpYo from './StyledUpYo';
import styles from './styles.css';

const Contrived = (props) => (
  <StyledUpYo className={styles.styledUpYo}>
    {props.children}
  </StyledUpYo>
);
```

mxstbr:
- Add section about the Sublime Text PR to the README
- Add "Using styled-components with existing CSS" doc (closes #85)
- Updated css-we-support to be more explicit about the fact that you can use all of CSS
- Move made-and-supported-by section to the top of the README

/cc @geelen and @samit4me for review ([live preview](https://github.com/styled-components/styled-components/blob/more-docs/README.md))

### DIFF
--- a/docs/existing-css.md
+++ b/docs/existing-css.md
@@ -1,0 +1,73 @@
+# Using `styled-components` with existing CSS
+
+`styled-components` generates an actual stylesheet with classes, and attaches those classes to the DOM nodes of styled components via the `className` prop. It injects the generated stylesheet at the end of the HEAD of the document.
+
+## Styling normal React components
+
+If you use the `styled(MyNormalComponent)` notation with an existing, non-styled-components react component that doesn't use the passed-in `className` prop the styles won't be applied.
+
+To avoid this issue, make sure your component (in this case `MyNormalComponent`) attaches the passed-in `className` to a DOM node:
+
+```JSX
+class MyNormalComponent extends React.Component {
+  render() {
+    return (
+      // Attach the passed-in className to the DOM node
+      <div className={this.props.className}></div>
+    );
+  }
+}
+```
+
+If you have pre-existing styles with a class, you can combine the global class with the `styled-components` one:
+
+```JSX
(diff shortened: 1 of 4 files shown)


docs/existing-css.md
```diff
@@ -0,0 +1,73 @@
...
+      // allowing this component to be styled with styled-components
+      <div className={`some-global-class ${this.props.className}`}></div>
+    );
+  }
+}
+```
+
+## Global class always overridden
+
+The other way around might happen too. You have a styled component, and a standard, global CSS class you're trying to override some styles with. For some reason though, the `styled-components` styles always end up overriding the styles of the global class, no matter what you do!
```
samit4me: You can apply a global CSS class to a styled component by adding a `className` prop. This will work, but if a specific CSS property (e.g. `background-color`) is defined in both, the global CSS and the styled component, the results may not be what your expecting!

^ I think the point to get across is that you can use global css, but it has caveats

mxstbr:
I love all of these suggestions, thanks so much! Will get the docs updated asap!

mxstbr:
Updated, I used your suggestions and slightly adapted them here and there – please give it another read!

mxstbr:
Awesome, merged!

PR #100, opened 2016-10-16
order import statements to break cyclical dependency deadlock (#87)

Rich-Harris:
Follow-up to https://github.com/styled-components/styled-components/issues/87#issuecomment-254062595. This PR replaces nested `require` statements in files in `vendor/postcss` with `import` declarations, so that they can be properly resolved at build time.

Doing so also means adding empty import statements to guarantee correct order of evaluation of modules (in ES modules, evaluation order is determined statically by the order of `import` declarations, whereas in CommonJS – and environments that simulate CommonJS by shipping a module loader, i.e. Browserify and Webpack – evaluation order is determined at runtime by the order in which `require` statements are encountered).

Specifically, since `Root`, `Rule` and `AtRule` all extend `Container`, it's essential that `Container` is evaluated (and therefore, in the context of a Rollup bundle, included) first. In order to do this, `input.js` (which is the 'gateway' to all the PostCSS stuff) must import `root.js`, `root.js` must import `rule.js` *before* it imports `container.js`, and `rule.js` must import `at-rule.js` before *it* imports `container.js`. Having those imports ensures that `container.js` doesn't then try to place `Root`, `Rule` or `AtRule` ahead of itself in the bundle.

I realise all this sounds a bit brittle and hairy, so would understand if you decided to call time on this whole experiment! Unfortunately something along these lines will always be necessary when handling these sorts of pathological cyclical dependency cases without using `require`.

Apologies again for creating this mess.


Rich-Harris:
👍 done

PR #104, opened 2016-10-17
Add Custom Property example to docs on "CSS we support"

anselmh:


mxstbr:
Hmm, I don't want to give off the impression that we transpile them to normal variables (like CSSnext does) – this is currently only supported in 65% of users browsers according to caniuse 😕 Can we maybe add a note stating that somehow?

anselmh:
Added a note.

mxstbr:
Awesome, LGTM! Thanks!

PR #113, opened 2016-10-18
[Proposal] Theme adapters

geelen:
The basic idea is that each `styled-component` uses a function called `themeAdapter` to translate the `theme` variable that's passed down the hierarchy into `this.props.theme` for the interpolations. By default that `themeAdapter` is `x => x` or identity.

Lets you do this:

``` js
/* material-design.js */

export const Button = styled.button`
  font-family: inherit;
  font-size: 1em;
  padding: ${ props => props.theme.buttonPadding || '0.5em' };
  border-radius: ${ props => props.theme.borderRadius || '3px' };
`
```

``` js
/* boostrap-two-point-oh.js */

export const Hero = styled.div`
  font-weight: 900;
  height: ${ props => props.theme.heroSize || '20rem' };
`
```

``` js
/* app.js */
import { Hero } from 'boostrap-two-point-oh.js'
import { Button } from 'material-design.js'
import { ThemeProvider } from 'styled-components'

const theme = {
  grid: 0.5
}

Hero.themeAdapter = theme => ({
  heroSize: `${theme.grid * 10}rem`
})
Button.themeAdapter = theme => ({
  buttonPadding: `${theme.grid * 2}em ${theme.grid * 4}em`,
  borderRadius: '0px'
})

const App = () => (
  <ThemeProvider theme={{ grid }}>
    <Hero>
       <Button>Go!</Button>
    </Hero>
  </ThemeProvider>
)
```

So each component can (but doesn't have to) define a mapping from the _inherited_ theme to their _local_ theme. Since each component can't update the theme that's passed down (you need a `ThemeAdapter` for that) you have a clear delineation from components that consume theme from components that pass it down. Check the test cases for a more complex example.

This way you can treat each imported component individually, although if a series of them use the same naming scheme they could certainly share a `themeAdapter`. But it means you can concentrate on making your `theme` only pass down the values that you care about. And use `themeAdapter` to inject default variables that a component has made themable but you don't care about (like `borderRadius` above)

Thoughts?


mxstbr:
I get where you're coming from here, but I'm not 100% sure I like it.

If we went with encouraging namespaces in the themes, nobody would need to have those weird functions, all you had to do was pass certain variables someplace:

```JS
const theme = {
  main: 'red',
}

<ThemeProvider
  theme={{
    ...theme,
    bootstrap: {
      primary: theme.main,
    },
  }}
>
  <BootstrapButton>I'm now red!</BootstrapButton>
</ThemeProvider>
```

_All_ components that we use from Bootstrap now recognize and use theme. If we went with what you see below, only the button would recognize the theme:

```JS
BootstrapButton.themeAdapter = theme => ({
  primary: theme.main,
})
```

Of course we could do this, but at that point I don't see what benefit we gain from that:

```JS
const bootstrapAdapter = theme => ({
  primary: theme.main,
})

BootstrapButton.themeAdapter = bootstrapAdapter
BootstrapTabs.themeAdapter = bootstrapAdapter
BootstrapTab.themeAdapter = bootstrapAdapter
BootstrapHero.themeAdapter = bootstrapAdapter
BootstrapEverything.themeAdapter = bootstrapAdapter
```

bhough:
@mxstbr I've gone back and forth on this, but the more I look at it the more I like the idea @geelen has proposed. The point you make is valid, that you could end up just tying a bunch of components to the same adapter. But in practice I think that would be more rare. 

If you were to encourage adapters that only modified the portions of the theme that the component uses, then those adapters will often be unique to that component only. IMO this would be the preferred approach, as too many shared adapters feels once step closer to shared styles between components. I feel like this idea is more about how it is used versus how it is implemented.

mxstbr:
Sorry, I don't quite get that last part. What do you mean with 

> I feel like this idea is more about how it is used versus how it is implemented.

bhough:
@mxstbr Whether you do it with the theme adapter as proposed by @geelen or you just do it as an override in props as you proposed, it is more about how much of theme you choose to map/override than how you do it. Discouraging blanket remapping of the theme in a single adapter in favor of using small one off adapters per component.

mxstbr:
> Discouraging blanket remapping of the theme in a single adapter in favor of using small one off adapters per component.

What's the benefit of doing that? That's just even more work, what do I gain for that?

bhough:
Let me ask this, so I make sure I understand the purpose behind the proposal. 

Is the adapter something that "ships" with the component so it should only handle what the component needs? 

Or is it something that is part of the greater app meant to just do this once for all components?

In @geelen's example it appear as if the proposal is the former, but maybe I'm wrong about that. And you are proposing that it acts a single interpreter for the entire theme?

robrkerr:
@geelen awesome idea! Could this be all done in one file, where third-party components are imported and remapped if needed? Then components to be used in the application could be all imported from this single file. This might be a nice way to manage it. 

@mxstbr the namespace idea is interesting too but where does the namespace come from: component author or component consumer? The former would be a problem because you could get clashes or inconsistent usages of the same namespace, the latter would start to look a bit like what @geelen is suggesting where you have to specify it for each component. I guess maybe you could automatically namespace components based on directory structure (so it would be consumer defined but without so much hassle). That might work nicely. 

@geelen so the remapped theme doesn't propagate down to all of the children inside that component? That's good because otherwise components that you pass in as children, where these children are from different authors and may need a different theme mappings. And this would require mappings to and from many more combinations rather than always just mapping from the single top-level theme specification for the application. Having looked at your example again though, it looks like you took this into consideration 👍 

@bhough my impression was that the component consumer defined the adaptors, but I'll let @geelen answer 🙂

bhough:
@robrkerr that actually helps quite a bit. I was originally linked here from another issue about providing standard "templating" rules for themes. So I viewed it from the perspective of adapting your component to an existing theme by giving generic properties like `theme.primary` more context like `button.bg`. Sorry if I caused thrash with my interpretation.

Given that context one thing @robrkerr said really hits home. Relying on all authors to properly name space could get sticky without at least an escape hatch to override that name space in the case of collisions or it not being provided.

geelen:
@robrkerr Yeah, my suggestion is that you'd put all the adapters in one file like

```js
/* externals.js */
import { Hero } from 'boostrap-two-point-oh'
import { Button } from 'material-design'

Hero.themeAdapter = theme => ({
  heroSize: `${theme.grid * 10}rem`
})

Button.themeAdapter = theme => ({
  buttonPadding: `${theme.grid * 2}em ${theme.grid * 4}em`,
  borderRadius: '0px'
})

export { Hero, Button }
```

Then whenever you `import { Hero } from '../externals'` you'd get the one with an adapter attached. 

It would be nicer maybe to have a different syntax that doesn't mutate the base class

```js
/* externals.js */
import * as Bootstrap from 'boostrap-two-point-oh'
import * as Material from 'material-design'
import { adaptTheme } from 'styled-components'

const Hero = adaptTheme(Bootstrap.Hero, theme => ({
  heroSize: `${theme.grid * 10}rem`
}))

const Button = adaptTheme(Material.Button, theme => ({
  buttonPadding: `${theme.grid * 2}em ${theme.grid * 4}em`,
  borderRadius: '0px'
}))

export { Hero, Button }
```

My problem with namespacing is that you end up having to pass down all the information for all the components from the top-level down. Whereas, the only information that should be being inherited down the render tree should be the _actual_ theme variables that make sense for your app. You don't want to be trying to figure out which values are 'real' and which ones are derived at random points in the hierarchy.

Also, I think you _do_ want to be using the same adapters for multiple components (this is yet another potential syntax btw):

```js
/* externals.js */
import * as Bootstrap from 'boostrap-two-point-oh'
import * as Material from 'material-design'
import { themeWrapper } from 'styled-components'

const wrapBootstrap = themeWrapper(theme => ({
  heroSize: `${theme.grid * 10}rem`,
  jumbotronDisplay: 'none',
  primary: theme.primary,
  secondary: theme.alt
}))

const Hero = wrapBootstrap(Bootstrap.Hero)
const Jumbotron = wrapBootstrap(Bootstrap.Jumbotron)
const Alert = wrapBootstrap(Bootstrap.Alert)

const wrapMaterial = themeWrapper(theme => ({
  buttonPadding: `${theme.grid * 2}em ${theme.grid * 4}em`,
  borderRadius: '0px'
}))

const Button = wrapMaterial(Material.Button)
const Nav = wrapMaterial(Material.Nav)

export { Hero, Jumbotron, Alert, Button, Nav }
```

This turns out feeling a lot more like namespacing, since you write a single adapter for a whole namespace of components. But its using function composition instead, so I think it's a lot cleaner.

robrkerr:
@geelen nice! Not sure about the name, but I really like the themeWrapper idea. Seems like a really nice way to manage it.

mxstbr:
I definitely like `themeWrapper` more than the `.themeAdapter`, but it's still very verbose (same thing for every component) for no visible benefit. What about an API like this?

```JS
// external.js
import { adaptTheme } from 'styled-components'
import { Hero, Jumbotron, Alert } from 'boostrap'

const bootstrap = adaptTheme(theme => {
  heroSize: `${theme.grid * 10}rem`,
  jumbotronDisplay: 'none',
  primary: theme.primary,
  secondary: theme.alt,
})

export bootstrap(Hero, Jumbotron, Alert)
```

```JS
// main.js

import { Hero, Alert } from './external';
```

It's a bit opaque because you don't visibly export things (I don't even know if this works tbh), but I like it more than wrapping each component.

*What I'm trying to do: I get where you're coming from with defining a theme at that level, but I don't see the value in wrapping every. single. component. with the same function from an API perspective. Surely every component in a third party library will respond the same variables, so they're going to use the same adapter anyway – why then wrap every single one individually?*

robrkerr:
@mxstbr ideally, you'd like to not even refer to the individual components, right? 

```js
// external.js
import { adaptCollectionTheme } from 'styled-components'
import bootstrapCollection from 'bootstrap'

const adaptedBootstrapCollection = adaptCollectionTheme(bootstrapCollection, theme => {
  heroSize: `${theme.grid * 10}rem`,
  jumbotronDisplay: 'none',
  primary: theme.primary,
  secondary: theme.alt,
})

export adaptedBootstrapCollection
```

```js
// main.js
import { Hero, Alert } from './external'
```

But then perhaps you're now adapting and including components that you're not going to use. I'm not sure if that matters.

geelen:
> `export bootstrap(Hero, Jumbotron, Alert)`

This won't work I'm pretty sure, neither will `export adaptedBootstrapCollection`. I do get where you're coming from, especially if each component is going to use the same adapter. I had the idea to inherit down the adapters, but I don't think it makes sense:

```js
import { Hero } from 'boostrap-two-point-oh'
import { Button } from 'material-design'

const wrapBootstrap = theme => ({
  heroSize: `${theme.grid * 10}rem`,
  jumbotronDisplay: 'none',
  primary: theme.primary,
  secondary: theme.alt
})

const wrapMaterial = theme => ({
  buttonPadding: `${theme.grid * 2}em ${theme.grid * 4}em`,
  borderRadius: '0px'
})

const theme = {
  grid: 0.5,
  primary: 'papayawhip',
  alt: 'midnightblue',
  wrapBoostrap,
  wrapMaterial
}

const App = () => (
  <ThemeProvider theme={theme}>
    <Hero>
       <Button>Go!</Button>
    </Hero>
  </ThemeProvider>
)
```

It's kinda weird for a component to trust this inherited property called `wrapMY_FRAMEWORK_NAME`. 

To be honest, I can't think of any way to have to type less in `externals.js` than my last submission, that's also valid ES6. And I think I'd prefer to have valid ES6.

geelen:
Credit to @robrkerr for this idea:

```js
/* external-components.js */

import { adaptTheme } from 'styled-components'

import * as bootstrapComponents from 'bootstrap'
import * as materialComponents from 'material-design'

const wrapBoostrap = theme => ({
  heroSize: `${theme.grid * 10}rem`,
  jumbotronDisplay: 'none',
  primary: theme.primary,
  secondary: theme.alt
})

const wrapMaterial = theme => ({
  buttonPadding: `${theme.grid * 2}em ${theme.grid * 4}em`,
  borderRadius: '0px'
})

export const {
  Hero,
  Jumbotron,
  Alert
} = adaptTheme(wrapBoostrap, bootstrapComponents)

export const {
  Button
} = adaptTheme(wrapMaterial, materialComponents)
```

Inside `adaptTheme` we'd have to do some rad `Object.defineProperty` action to lazy-wrap these components on `get` but this is kinda neat.

robrkerr:
So far the example cases we've been talking about has been adapting third-party components (e.g. material-design) so that they can respond to attributes of a theme we've defined for our application. Does it make sense to do this adapting for non-third party components (i.e. any components we build ourselves for the application)? 

You wouldn't have to. You could just write your components such that they look for the attributes  you've defined in your theme. But you could also write your components more generically, which expect different theme attributes, and then create an adaptor for them as well. The latter would be more work but maybe it would be a better way of doing it? That component would be perhaps more easily sharable? Maybe it's just a good habit to get in? 

The other thing that this makes me wonder, is how differently you would write components in this case. How specific would themes tend to be to an application? 

Anyway just a thought I had. What do people think: adaptors only for external/third party or adaptors for all components?

nikgraf:
100% agree on @robrkerr's idea and the API looks nice as well. @mxstbr this resembles what I'm talking about since beginning of the year.

Here my thoughts:
I'm a bit worried/scared of introducing context for theming. If we solely rely on props then components theming has this functional character: "calling a function f twice with the same value for an argument x will produce the same result f(x) each time". Once you introduce context the component styling can be influence "globally". I assume in large code-bases it will be hard to predictable from where theming is coming from. In addition this opens up the question if you merge theme properties or overwrite the whole thing.
see also https://github.com/nikgraf/future-react-ui/issues/1#issuecomment-178318507

geelen:
@nikgraf I think this is a big misconception about context, actually. There's a huge difference between things that are inherited and things that are global. Context can only flow down a render tree, and your point about idempotence is fine as long as you consider the "context" another argument to the function. That's pretty much all it is, a prop that's automatically applied.

To me, defining a component as _independent_ of these contextual styling values is the real gain here. So, following on from @robrkerr's point it feels like the really interesting effect is when you start to go hard-line on that idea: a component cannot `import` any common styling information, it must inherit it from `props.theme`. I do wonder what that'll feel like.

I'll update this PR to use the non-mutating `adaptTheme` I suggested last and give it a spin on some projects. Will report back.

nikgraf:
@geelen Looking forward to the outcome. Totally agree as seeing context as a prop that's automatically applied. The automatically is what I believe can cause troubles in large codebases.

Let me try to explain via an example:

App.js
```js
import Dashboard from './Dashboard';
import Navigation from './Navigation';
import DetailPage from './DetailPage';
import { ThemeProvider } from 'styled-components';
import lightTheme from '../themes/lightTheme';

export default () => ((
  <div>
    <ThemeProvider theme={lightTheme}>
      <Navigation />
      <Dashboard />
      <DetailPage />
    </Theme>
  </div>
));
```

Nav.js
```js
import Link from 'router/Link';
import { ThemeProvider } from 'styled-components'
import darkTheme from '../themes/darkTheme';

export default () => ((
  <div>
    <ThemeProvider theme={darkTheme}>
      <Link to="/dashboard">Dashboard</Link>
      <Link to="/about">About</Link>
    </ThemeProvider>
  </div>
));
```

Dashboard.js
```js
import Revenue from './Revenue';
import Button from 'elemental/Button';
import { ThemeProvider } from 'styled-components'
import dashboardTheme from '../themes/dashboardTheme';

export default () => ((
  <div>
    <ThemeProvider theme={dashboardTheme}>
      <Button>Save</Button>
      <Revenue />
    </ThemeProvider>
  </div>
));
```

DetailPage.js
```js
import Button from 'elemental/Button';

export default () => ((
  <div>
      <Button>Open</Button>
  </div>
));
```

There are two issues I see:

- You need to look at the whole tree to understand which theme is applied to for example Button in DetailPage.
- If you change/remove the ThemeProvider in App.js you have no idea which sub-components are affected without an understanding of the whole Tree -> only the Button in DetailPage, but not in Dashboard.

Maybe it's just me, but I like the concept of being very explicit and create different components for different themes + using having props as only input. 😄

mxstbr:
> having props as only input

I think the whole point of theming is to be global, or not? I don't want to redefine every single component of a third-party UI library, I just want to change the primary color, right? Why would I have to wrap every. single. third party component. in a custom one to pass in the right props?

geelen:
Ok implemented this design. You can do the following:

```js
import { adaptTheme } from 'styled-components'

const BaseComponent = styled.div`
  ${ props => props.theme.abc }
`

const WrappedComponent = adaptTheme(theme => ({
  abc: theme.xyz
}), BaseComponent)

export default WrappedComponent
```

Or you can do an object full of components

```js
import * as components from 'some-library'
import { adaptTheme } from 'styled-components'

const themeAdapter = theme => ({
  some: theme.thing
})

export const {
  ComponentA,
  ComponentB,
  ComponentC
} from adaptTheme(themeAdapter, components)
```

@robrkerr wanna test this out?

robrkerr:
Sure! I'll check it out tonight.

nikgraf:
@mxstbr yep, but with what this approach the components are also "globally" accessible via imports, but only take props as inputs:

```
export const {
  ComponentA,
  ComponentB,
  ComponentC
} from adaptTheme(themeAdapter, components)
```

mxstbr:
> @mxstbr yep, but with what this approach the components are also "globally" accessible via imports, but only take props as inputs:

Ah sorry, misunderstood what you were saying – yeah, totally agree that this is the right direction

robrkerr:
So in trying this out, it occurred to me that a "component" (either your's or a third-party's), which you import and want to adapt, may not technically be a single component - it may be two or more nested inside each other. Here's an example of what I mean:

```js
// Header.js
import React, { Component } from 'react';

const Container = styled.div`
  width: 100%;
  background-color: ${props => props.theme.bg || 'lightblue'};
`;

const Inner = styled.div`
  width: 100%;
  max-width: ${props => props.theme.maxWidth || 500}px;
  margin: 0 auto;
  padding: 1rem 2rem;
`;

class Header extends Component {
  render() {
    return (
      <Container>
        <Inner>{this.props.children}</Inner>
      </Container>
    );
  }
}

export default Header;
```

To me this seems like a reasonably normal thing to want to do and a likely way that third-party "components" might be build. But maybe others would disagree? The problem I'm running into is that adapting the theme for a "component" like this doesn't work:

```js
// components.js
import { adaptTheme } from 'styled-components'
import Color from 'color';
import BaseHeader from './Header';

const Header = adaptTheme(theme => ({
  bg: Color(theme.bg || 'white').darken(0.1).hslString()
}), BaseHeader);

export { Header }
```

In this case, `theme.bg` in the `Container` of the `Header` is still what is was before the theme was adapted. This is because the theme only gets adapted to a single level down - not to the "children" `Container` or `Inner`. 

Is this a legitimate use case or should I be doing this differently? Is there way to make the theme adapting work so that it applies to a conceptual component rather than only one level deep? What are your thoughts?

mxstbr:
> In this case, theme.bg in the Container of the Header is still what is was before the theme was adapted. This is because the theme only gets adapted to a single level down - not to the "children" Container or Inner.

What happens here is that the `theme` is passed in as a property, so you could pass it forward? Something like this:

```JSX
class Header extends Component {
  render() {
    return (
      <Container theme={this.props.theme.container}>
        <Inner theme={this.props.theme.inner}>{this.props.children}</Inner>
      </Container>
    );
  }
}
```

Now you have a leaky abstraction though, because you require users to understand internal setup, and if you ever change that it'll break, so something like this would be even better:

```JSX
class Header extends Component {
  render() {
    const containerTheme = {
      sidebar: this.props.theme.primaryColor,
      main: this.props.theme.secondaryColor,
    }
    const innerTheme = {
      maxWidth: this.props.theme.width,
    }
    return (
      <Container theme={this.props.containerTheme}>
        <Inner theme={this.props.innerTheme}>{this.props.children}</Inner>
      </Container>
    );
  }
}
```

geelen:
Yeah, along the lines of what @mxstbr said, I think the fact that you have multiple elements but expose a single component should be wrapped up inside your component like this:

```js
class Header extends Component {
  render() {
    const { theme, children } = this.props
    return (
      <Container theme={theme}>
        <Inner theme={theme}>{children}</Inner>
      </Container>
    );
  }
}
```

I think that should work as-is.

The problem will then be that `adaptTheme` expects a `StyledComponent` (at least the way it's currently implemented). So we'd have to add a wrapper component like this:

```js
const adaptTheme = (adapter, Base) => {
  return class ThemeAdapterWrapper extends Component {
    render() {
      return <Container theme={adapter(this.context.theme)} {...this.props}/>
    }
  }
}
```

Unless, of course, we could do a trick with subclassing and setting `defaultProps` so we don't need a wrapping element...

Hmm, I'll have a bit more of a think and do some experiments.

giuseppeg:
Long thread – I hope I didn't miss something.
If I can share my two cents: when in doubt just look at the CSS spec or existing features and try to come up with future proof solution is the best way to go. In this case it means that `namespaces` are not the best idea. I like @geelen's proposal because in a way it is similar to how you'd do that in regular css:

```css
.MyComponent {
  color: var(--MyComponent-color, #000);
}

/** theme.css */

:root {
  --MyComponent-color: var(--theme-color);
}

/* or limit the scope / similar to ThemeProvider */

.ThemeProvider {
  --MyComponent-color: var(--theme-color);
}
```

mxstbr:
How are we looking here implementation-wise @geelen? Is this finished?

Could we get an example of who would use this how? Some sort of made-up real-worldy usecase would be appreciated!

wmertens:
question about this proposal: what about components inheriting from others? Do they inherit the theme adapters too?

xcoderzach:
@geelen
I'm way late to the party here, but I really like this idea, but not necessarily this implementation.  I think all `styled-components` should be forced to provide a `mapThemeToProps` function, and not provide a `theme` prop at all.  

  It's big violation of the Interface Segregation Principal to pass in literally everything to every styled component.  To give an analogy, it would be terrible if redux's default mode of operation was to pass the entire state object as a prop to every `connect`ed component, but that's what we're doing here with themes.

wmertens:
@xcoderzach I am having a hard time understanding why a globally accessible theme is evil. It seems to me that your example instead makes things more complex?

xcoderzach:
It's not that having a globally accessible theme is evil, but coupling the component to the global theme is.

A button styled-component should only rely on a `buttonBackgroundColor`, and `buttonTextColor`, not the application's `primaryUIColor` and `primaryTextColor`, and it definitely shouldn't need to know about all of the other theme vars that are completely unrelated.  And I think that it's important enough to make it the default.

wmertens:
So you would map from the `theme` to `buttonBackgroundColor` before
defining the inline css? Wouldn't that be done by the same programmer that
writes the inline css?

>
In our case, we post-process the theme colors to have 4 states defined in
the theme: normal, hover, accent, accent-hover. In the components, we then
use whatever is appropriate, using the `accent` prop as a key if needed.

It's not like we'll have random people from the internet writing our S-C
css and multiplying the inputHeight with formBorderRadius…

mxstbr:
I don't quite get what the issue here is either @xcoderzach. What's the benefit of this

```JS
const Button = styled.button`
  background: ${props => props.theme.bg};
`

export default withTheme(theme => {
  bg: theme.bg,
})(Button)
```

over this?

```JS
const Button = styled.button`
  background: ${props => props.theme.bg};
`
```

What am I missing here?

giuseppeg:
@mxstbr he's basically saying that a styled-component should only get the `theme` object instead of `props` which could be a bigger one and contain unrelated (if not sensible) data.

giuseppeg:
I believe

```js
generateAndInjectStyles(theme: any) {
   return componentStyle.generateAndInjectStyles(theme)
}
```

https://github.com/styled-components/styled-components/blob/23ae59a7fe9e7ad894594c562439263727fb4bfe/src/models/StyledComponent.js#L39

xcoderzach:
@giuseppeg I wasn't saying we should not pass props 😄 I think we definitely need props along with the theme, I just want to be explicit about which theme variables are used.  I see it as the distinction between a function with explicitly named arguments, vs a function that takes in one big object.

Most of my themed components look more like this.  In really simple cases, it doesn't matter, however I this level of complexity (or worse) is fairly common when working with css, and would be a better case to optimize for.

```javascript
const Button = styled.button`
  height: ${get('theme.button.height')}px;
  line-height: ${get('theme.button.height')}px;
  min-width: ${get('theme.button.minWidth')}px;
  background-color: ${({ disabled, primary, secondary, theme }) =>   
    disabled  ? theme.colors.disabledTextColor:
    primary   ? theme.colors.primaryColor1:
    secondary ? theme.colors.secondaryColor1:
                theme.colors.alternateTextColor};
  color: ${get('theme.colors.textColor')};
  border-radius: ${get('theme.button.borderRadius')};
  user-select: none;
  overflow: hidden;
  text-align: center;
  opacity: 1;
  font-size: ${get('theme.type.fontSize')}px;
  letter-spacing: 0;
  text-transform: uppercase;
  font-weight: ${get('theme.type.fontWeight')};
  margin: 0;
  padding: 0 ${get('theme.spacing.noGutter')}px;
`
```
vs
```javascript
let Button = styled.button`
  height: ${get('height')}px;
  lineHeight: ${get('height')}px;
  minWidth: ${get('minWidth')}px;
  background-color: ${get('backgroundColor')};
  color: ${get('textColor')};
  border-radius: 10;
  user-select: none;
  overflow: hidden;
  text-align: center;
  opacity: 1;
  font-size: ${get('fontSize')}px;
  letter-spacing: 0;
  text-transform: uppercase;
  font-weight: ${get('fontWeight')};
  margin: 0;
  padding: 0 ${get('sidePadding')}px;
`

Button = withTheme(({ colors, type, spacing }, { disabled, primary, secondary }) => {
  return {
    color: colors.textColor,
    height: button.height,
    minWidth: button.minWidth,
    borderRadius: button.borderRadius,
    fontSize: type.fontSize,
    fontWeight: type.fontWeight,
    sidePadding: spacing.noGutter,
    backgroundColor:
      disabled  ? colors.disabledTextColor:
      primary   ? colors.primaryColor1:
      secondary ? colors.secondaryColor1:
                  colors.alternateTextColor
  }
})(Button)
```

giuseppeg:
gotcha my bad I misunderstood :)

xcoderzach:
Maybe we could extend #365's `attrs` to take a function, and it would become a `withProps` or `mapProps` type function :-/  Although, the more I look at it, the more I think mapping all the theme vars is gross, haha

```javascript
styled.button.attrs(
  ({ 
    theme: { colors, type, spacing }, 
    disabled, 
    primary, 
    secondary 
  }) => ({
    color: colors.textColor,
    height: button.height,
    minWidth: button.minWidth,
    borderRadius: button.borderRadius,
    fontSize: type.fontSize,
    fontWeight: type.fontWeight,
    sidePadding: spacing.noGutter,
    backgroundColor:
      disabled  ? colors.disabledTextColor:
      primary   ? colors.primaryColor1:
      secondary ? colors.secondaryColor1:
                  colors.alternateTextColor
  })
)`
  height: ${get('height')}px;
  lineHeight: ${get('height')}px;
  minWidth: ${get('minWidth')}px;
  background-color: ${get('backgroundColor')};
  color: ${get('textColor')};
  border-radius: 10;
  user-select: none;
  overflow: hidden;
  text-align: center;
  opacity: 1;
  font-size: ${get('fontSize')}px;
  letter-spacing: 0;
  text-transform: uppercase;
  font-weight: ${get('fontWeight')};
  margin: 0;
  padding: 0 ${get('sidePadding')}px;
`
```

geelen:
A couple of points. Theming isn't general-purpose data-passing. It's about a component deferring certain implementation details to its environment. In that sense, it _must_ be a single global object for it to have any value. Even in your example, the input to the `mapThemeToProps` is the same global object, so I don't see the benefit. If you're trying to move _logic_ about how props and themes interact out of the CSS, might I suggest:

```js
import { get, fromTheme, backgroundColors } from '../utils'

const Button = styled.button`
  ${ fromTheme('button', 'height line-height min-width border-radius') }
  ${ fromTheme('type', 'font-size font-weight') }
  ${ backgroundColors }
  
  color: ${get('theme.colors.textColor')};
  user-select: none;
  overflow: hidden;
  text-align: center;
  opacity: 1;
  letter-spacing: 0;
  text-transform: uppercase;
  margin: 0;
  padding: 0 ${get('theme.spacing.noGutter')}px;
`
```

...or whatever DSL makes sense for your components. I think `get` is a really nice idea already, but I think you can go further.

But yeah, theming should be about reducing the burden on components to be explicit about every one of their properties. With the goal of making a component visually "just work" when you drop it into a different part of the application.

xcoderzach:
Yeah, I think I was just getting weird about trying to give everything from the theme a "context specific" name, after thinking about it, it's not really necessary at all.

mxstbr:
Since `styled-theming` by @thejameskyle is pretty widely used by now and since a year ago not many people have asked for something like this, I'll close this PR.

If renewed interest appears in this we can always reopen later 👌

PR #117, opened 2016-10-18
add an example page

stephenkingsley:
add an example page, let the test and bug fix more and more convenient and fast.

when run `npm run dev`, it can open the `localhost:3000` to see the page. and when you change some file, they will show the new page.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "lint": "eslint src",
     "prepublish": "npm run build",
     "lint-staged": "lint-staged",
-    "dev": "node devServer.js"
+    "dev": "node example/devServer.js"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
     "build:lib": "babel --out-dir lib src",
     "prebuild:umd": "rm -rf dist/*",
     "prebuild:dist": "rm -rf dist/*",
-    "build:dev": "rollup -c",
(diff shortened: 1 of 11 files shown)


stephenkingsley:
yeah~ I will fix it

stephenkingsley:
may u review my code?

stephenkingsley:
ok, I will fix it now!

package.json
```diff
@@ -11,14 +11,16 @@
     "build:lib": "babel --out-dir lib src",
     "prebuild:umd": "rm -rf dist/*",
     "prebuild:dist": "rm -rf dist/*",
+    "build:dev": "rollup -c",
```
mxstbr: I'm not sure we need a separate command – with this example, I'd expect to be able to switch out `styled-components.js` with `styled-components.min.js` in the `index.html` and to test my changes in minfied mode, right?

stephenkingsley: yeah, I agree. I think output `styled-components.js` and `styled-components.min.js` is cost more time. so I want just output one file. I will change it.

stephenkingsley:
may you review again please?

mxstbr:
Let's fix the no-tests thing to not include `example/`, same for the changelog thing!

mxstbr:
How awesome is that! Danger doesn't comment again it edits the previous comment! 😱 😱 😱

orta:
Hah - yep, fav feature of mine - if you have no warnings ( or stickies ) she'll also delete the message

Issue #102, opened 2016-10-17
refs are attached to the styled components wrapper


mxstbr:
## Version

All of them

## Reproduction

http://www.webpackbin.com/NJKccFpC-

## Steps to reproduce

Attach a ref to a styled component

## Expected Behavior

The ref is attached to the DOM node rendered, e.g. in the case of a `styled.input` the `<input>`

## Actual Behavior

The ref is attached to the styled-components wrapper, making it impossible to e.g. focus a `styled.input` programmatically.

mxstbr:
Turns out, React doesn't support "forwarding" refs. The core team wants to add support for that, but it's not high priority so it'll be a while.

I think the way we need to work around this is have a special prop maybe called `innerRef` that attaches to the inner component:

```JS
const StyledInput = styled.input`
  some: styles;
`;

<StyledInput innerRef={comp => this.input = comp} />

// this.input.focus() works 🎉
```

Something to be aware of is that this won't work with string refs, only with callback refs, since react cannot guess which component to attach the ref to. That's something we'll need to mention in the docs.

Thoughts?

evan-scott-zocdoc:
Could this information be surfaced on the main README instead of buried in the "tips & tricks" section?

AshCoolman:
@mxstbr bumping for https://reactjs.org/docs/forwarding-refs.html

kitten:
@AshCoolman we have new issues for that and it’s unrelated to the solution that was introduced for this: the innerRef prop

PR #122, opened 2016-10-19
Add `innerRef` to get ref to inner DOM node

freiksenet:
Fixes https://github.com/styled-components/styled-components/issues/102

Not sure if those tests are good, I don't know any other way to test refs, other than simulating a browser :(


README.md
```diff
@@ -265,6 +265,31 @@ This will now rotate it's children over and over again, for example our logo:
   </a>
 </div>
 
+### Refs to DOM nodes
```
mxstbr: Let's move this to `docs/tips-and-tricks.md`

README.md
```diff
@@ -265,6 +265,31 @@ This will now rotate it's children over and over again, for example our logo:
...
+
+Passing `ref` to styled component will give a ref to the `StyledComponent`
+wrapper, not to DOM node. So it's not possible to call DOM methods, like focus
+on that wrapper. To get a `ref` to wrapped DOM node, pass `innerRef` prop. Note
+that this only supports callback-based refs, not string-based ones.
```
mxstbr: Make this note look like all our other notes like so:

```Markdown
…pass `innerRef` prop.


> **Note:** `innerRef` only supports callback refs (i.e. `ref={comp => this.bla = comp}`), string refs (i.e. `ref="bla"`) won't work. Since string based refs will be deprecated in the future anyway, don't worry about it too much and just use the callback pattern.
```

README.md
```diff
@@ -265,6 +265,31 @@ This will now rotate it's children over and over again, for example our logo:
...
+
+```JSX
+const StyledInput = styled.input`
+  color: red;
```
mxstbr: Let's make this `palevioletred` to stick with the other colors in the docs 😉

freiksenet:
> Except for a few nitpicks this is ready – I'd like to avoid jsdom rendering though, can you try and see if you can access the ref if you use enzymes render function?

Unfortunately it doesn't work, it doesn't seem to even have React elements in full renderer, just full rendered DOM. :( I understand that jsdom is a big dependency, but I think it's better to have a test and a big dependency rather than no test.

I've fixed the rest of the comments.

mxstbr:
This looks great, can you do one more thing and add an entry to the `CHANGELOG.md` in the `Added` section? Then we're good to ship this :rocket:

freiksenet:
@mxstbr Done :)

mxstbr:
Amazing, thank you! 🎉

PR #127, opened 2016-10-21
fix #115

### LINKED ISSUE #115: Bug [Export default]
## Version

1.0.6, 1.0.7
## Reproduction

Update from 1.0.5 to 1.0.6 or 1.0.7.
## Steps to reproduce

Normal use
## Expected Behavior

Render div or another component
## Actual Behavior

I hope these screenshots enough to understand the problem =)

stephenkingsley:
fix #115

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -18,6 +18,19 @@ import ThemeProvider from './models/ThemeProvider'
 const keyframes = _keyframes(generateAlphabeticName)
 const styled = _styled(_styledComponent(_ComponentStyle(generateAlphabeticName)))
 
+// before fix Bug [Export default] #115
 /* Export everything */
-export default styled
-export { css, keyframes, injectGlobal, ThemeProvider }
+// export default styled
+// export { css, keyframes, injectGlobal, ThemeProvider }
+
+// fix Bug [Export default] #115
+const styledComponent = {}
+Object.keys(styled)
+  .map(ele => (styledComponent[ele] = styled[ele]))
+
+styledComponent.css = css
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -18,6 +18,19 @@ import ThemeProvider from './models/ThemeProvider'
...
+const styledComponent = {}
+Object.keys(styled)
+  .map(ele => (styledComponent[ele] = styled[ele]))
+
+styledComponent.css = css
+styledComponent.keyframes = keyframes
+styledComponent.injectGlobal = injectGlobal
+styledComponent.ThemeProvider = ThemeProvider
+// export default
+export default styledComponent
```
k15a: With that you're breaking named exports.

```js
import {
  css,
  keyframes,
  injectGlobal,
  ThemProvider,
} from 'styled-components'
```

You now have to write

```js
const {
  css,
  keyframes,
  injectGlobal,
  ThemProvider,
} = styled
```

Note that this is not the same. For further information look at http://stackoverflow.com/questions/33524696/es6-destructuring-and-module-imports

stephenkingsley:
oh, I think this is not my want. I will modify it as soon as possible.

stephenkingsley:
oh, I think this is not my want. I will modify it as soon as possible.

k15a:
You probably don't have to because that is a bug from webpack. See https://github.com/webpack/webpack/issues/3168 for further information. ;)

stephenkingsley:
I think this bug just show in use <script />, not in commonJS. so I close

Issue #123, opened 2016-10-19
defaultProps doesn't work as described in the documentation


bhough:
## Version
1.0.7

## Reproduction
http://www.webpackbin.com/4JDvTTx1z

## Steps to reproduce
I was trying this with a more complex implementation, but created a simplified one based on the docs to make sure it wasn't to do with my specific implementation. Using the example given in the docs it appears as as if the `defaultProps` are not being taken into account when generating the styles. So the part of the docs describing how to handle fallbacks for theming is not working as described.

## Expected Behavior
Top button in the above example should be colored `palevioletred`.

## Actual Behavior
Top button has a blank `background` value and it inherits it from the default browser styles instead.

mxstbr:
Ohh good catch, it only works for actual props, it doesn't work for theme! We should add support for that from `StyledComponent.js` (and `StyledNativeComponent.js`) – instead of making `{}` the default for theme, check if `props.theme` exists first and pass that in!

bhough:
@mxstbr I will work to put a PR together to fix this over the weekend.

bhough:
I submitted a PR to address this: https://github.com/styled-components/styled-components/pull/132

However, I wanted to bring up one caveat to this approach. For `defaultProps` to work properly for `theme` we have to remove the empty object default. This means anyone who isn't properly setting default props would get an error when checking `props.theme.property` in the ternary's since `theme` won't exist at all on the styled component. 

If this is the desired behavior I can go through and update the examples/docs to reflect this change. If not, we should discuss the approach further, as I think it would require some heavy lifting to have the rendered component check itself somehow after `defaultProps` are applied and add it's own default...default? Would need to sort out a non-hacky way of doing this :) 

Any feedback is appreciated.

kopax:
Thanks!

I wanted to test it so I added to my package.json

        "styled-components": "git+https://git@github.com:styled-components/styled-components.git#8759e17a9f3e66742bd5bc4e8f718b17ab7f53c7",

I got this error when trying to import the module : 

    ERROR in ./app/app.js
    Module not found: Error: Can't resolve 'styled-components'

It appear the dist folder is missing.

I have tried to build but the file `rollup.config.js` is missing.

Could we please release a new version for this ?

mxstbr:
Done, just released `styled-components@1.0.9`!

kopax:
just thanks!

PR #132, opened 2016-10-24
Allows defaultProps to properly work for theme

bhough:
Fixes https://github.com/styled-components/styled-components/issues/123

Includes updated tests.


src/models/StyledComponent.js
```diff
@@ -58,11 +58,16 @@ export default (ComponentStyle: any) => {
...
+        if (!theme) return Object.assign({}, this.props)
+        return Object.assign({}, this.props, { theme })
+      }
+
       /* eslint-disable react/prop-types */
       render() {
         const { className, children } = this.props
-        const theme = this.state.theme || {}
-        const executionContext = Object.assign({}, this.props, { theme })
+        const theme = this.state.theme
```
mxstbr: Why not do `this.state.theme || this.props.theme || {}`?

bhough: This is what I get what I get for coding so late. This was actually what I attempted first and it wasn't fixing my test case. I'm guessing I didn't have the build/test suite running correctly at that point, sorry. Will update the PR with the much simpler version.

bhough:
AppVeyor was stuck in failed on this PR. Closed in favor of https://github.com/styled-components/styled-components/pull/136

PR #137, opened 2016-10-24
Possible performance improvement moving componentStyle.generateAndInjectStyles out of render

JamieDixon:
My performance testing experience is quite limited and this is the first time I've used `react-addons-perf`.

While we were perusing the styled-components code we got to wondering about the use of `generateAndInjectStyles` in the `render` method of `StyledComponent`. Our first thought was that a side-effecty method-call probably shouldn't happen in the `render` method but rather in one of the lifecycle hooks such as `componentWillReceiveProps`, therefor keeping `render` a pure method.

I wanted to add some benchmarks to see if the change made any difference as well as trying out a few other things that might have improved performance (playing around with `shouldComponentUpdate` etc).

After putting together a benchmark based on the existing example it looks like there may be a performance improvement made by moving `generateAndInjectStyles` out of `render` and into `componentWillReceiveProps`.

Like I say, my perf experience and knowledge is very limited but I ran these stats over and over to make sure the results were consistent and it seems to be the case. It's also very likely that I'm completely missing something so bare with me on this one.

Here's some numbers from before and after the adjustment. The initial times are the first render. The next two metrics are re-renders after the container hosting the `<Example />` component gets re-rendered due to a state change.

**Before:**
 
with-code-in-render

You can see here that the inclusive render time of `Example > styled.section` (index 2) is `6.54ms` and `Example > styled.h1` (index 3) is `0.68ms`.

**After**

with-code-moved

After I moved `generateAndInjectStyles` and the other code it's reliant on, out of `render` the inclusive render time of `Example > styled.section` (index 2) has dropped to `0.32ms` and `Example > styled.h1` (index 3) is `0.04ms`.

This seems to suggest a fair performance improvement by shifting the code from `render`.

I've included the benchmark example file that I used to create these results. If I've missed something let me know and feel free to ignore this. If this is of some use then great!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ All notable changes to this project will be documented in this file. If a contri
 
 - N/A
 
+### Changed
+Performance enhancer: Modified the `StyleComponent.js` so that `generateAndInjectStyles` is called outside of `render`.
+
 ## [v1.0.8] - 2016-10-18
 
 ### Added
--- a/example/devServer.js
+++ b/example/devServer.js
@@ -25,6 +25,10 @@ const port = 3000
 
 app.use(Express.static('dist'))
 
+app.get('/with-perf.html', (req, res) => {
+  res.sendFile(path.join(__dirname, 'with-perf.html'))
+})
+
 app.get('/*', (req, res) => {
   res.sendFile(path.join(__dirname, 'index.html'))
 })
--- a/example/with-perf.html
+++ b/example/with-perf.html
@@ -1,0 +1,84 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Example with Perf logging</title>
+  </head>
+  <body>
+    <h1>Example with Perf logging</h1>
+    <div id="container"></div>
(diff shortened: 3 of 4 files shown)


mxstbr:
Thank you so much for digging into this! I'm not too well-versed in the tradeoffs of this part of the codebase, but @geelen might have something to say here.

JamieDixon:
I see from the failing tests that I did miss something so I'm having a go at solving that problem whilst retaining the perf numbers outlined.

JamieDixon:
Ok between myself and @sheepsteak we've got the tests passing and the perf is still looking good. Let me know what feedback you have. I see the travis `danger` task is failing and that might be because we would also need to make changes to the native version also?? Cheers.

mxstbr:
I don't see any reason not to ship this, so let's bring it up to date and ship it if @geelen hasn't said anything differently until tomorrow morning!

JamieDixon:
Excellent! Thanks @mxstbr !

mxstbr:
Released as `styled-components@1.0.10`, twitter announcement coming soon! So many thanks @JamieDixon!

Issue #89, opened 2016-10-15
Add Typescript type definitions


schickling:
I'm almost sorry to bring this up since `styled-components` is using Flow (which is great!). Would you consider adding type definitions for Typescript (2) as well? That would be awesome for the other half of the community which went for Typescript instead of Flow for `undefined` reasons. 😆

mxstbr:
I'd love to add libdefs for TypeScript – except I have no experience with TS and our flow types suck currently. (working on that soon though, see #51) If somebody wants to pick that up it'd be much appreciated!

patrick91:
Typescript support has been releases in `v1.4.0` :)

PR #152, opened 2016-10-31
Typescript type definitions

patrick91:
I've started working on the typescript type definition. I still need to write the test and, well, try it using the package (I've only tried referencing it directly).

Right now the `styled.element` functions' return type is a Stateless component, since I wasn't able to include the ReactDom types (they aren't exported :( )

I need to include the definitions for:

- [x] native
- [x] keyframes
- [x] ThemeProvider
- [x] css function
- [x] injectGlobal

let me know if I missed something :)

Other todos:

- [x] add var element (I commented that out while testing)
- [x] tests

Related: #89 

Happy Halloween! 🎃


mxstbr:
Wow that's amazing, thank you so much! Can't wait to land this!

patrick91:
@mxstbr what do you think of adding tslint to lint the "tests"?

mxstbr:
Sounds good to me!

jonaskello:
@patrick91 Perhaps you want to consider making the typings non-ambient (ie. not hard coding the module name with `declare module "styled-components"`). From what I gather using ambient declarations is not recommended, at least not in TS >2.0. For reference see the [template](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html) Microsoft proposes.

patrick91:
@jonaskello will have a look at it this weekend! danke!

jonaskello:
@patrick91 Nice! I'm also interested in getting typings so let me know if I can help.

patrick91:
@jonaskello thanks! a review would be really appreciated, it is my first type definition with typescript so it might not be the best one :)

AntonAderum:
Fantastic work!
I have one question though,
can and how will this handle this example?:
```
(in my component classs)
---
const Button = styled.button`
    background-color:  ${(props: any)  => props.negative ? props.theme.negative : props.theme.action}
`;
---
(in render function of component class)
---
<Button negative>Cancel</Button>
---
```
TypeScript wont allow you to set the "negative" prop since it isnt defined in "React.HTMLFactory\<HTMLButtonElement\>"

Do you have any idea on how to solve this?

patrick91:
@mxstbr I had to move the `native.js` to `native/index.js` to have typings for the native library as well. Not sure if it is the best way to achieve this, but since we can't add multiple `typings` to the package.json I don't see any other alternative :)

mxstbr:
> @mxstbr I had to move the native.js to native/index.js to have typings for the native library as well. Not sure if it is the best way to achieve this, but since we can't add multiple typings to the package.json I don't see any other alternative :)

That doesn't work. We need the `native.js` to be in the root of the file so people can do `import styled from 'styled-components/native'`, the `/native` here refers to the `native.js`.

patrick91:
Uhm, that should work anyway. It is not written in the spec but looks like all the module loaders allow load the index.js file when requiring/importing a folder.

patrick91:
@mxstbr I just tried and it works actually:

patrick91:
@AntonAderum sorry! I missed your comment :(

I would do a custom component for that, like I've done in the tests: https://github.com/patrick91/styled-components/blob/d7d8e31af8ee1e6324385ff89ad2ca4505115718/typings/styled-components-test.tsx#L55

Does that help?

mxstbr:
Ah sorry, you're right of course! Is this ready to be reviewed? Then I'll post on Twitter asking for reviewers!

patrick91:
@mxstbr yes! I just rebased it, a review would definitively be appreciated!

native/index.d.ts
```diff
@@ -0,0 +1,68 @@
+// <reference path="../react/react-native.d.ts" />
+
+declare module "styled-components/native" {
+    import * as ReactNative from "react-native";
+    import { StatelessComponent } from "react";
+
+    interface StyledFunction<T> {
+      (values: TemplateStringsArray): T;
+      (...values: any[]): T;
```
RReverser: Doesn't this one invalidate value of previous? (meaning that any arguments will be still accepted)

patrick91: uhm, I did this because I was getting an error when trying to use a template literal that had functions in it, like this:



Not sure if there's a better way for it!

DanielRosenwasser: I'm only glancing at the picture you've posted, but I think you want something like

```ts
(strs: TemplateStringsArray, ...fns: Array<(props: any) => string>): T;
```

DanielRosenwasser: On that note, maybe `StyledFunction` should have a type parameter for the props of each function?

patrick91: @DanielRosenwasser thanks! I've just changed that, I'll have a look at passing the types to the props functions, if you have any idea please shout! I'm still a newbie with TS :)

native/index.d.ts
```diff
@@ -0,0 +1,68 @@
+// <reference path="../react/react-native.d.ts" />
```
DanielRosenwasser: In general declaration files should create dependencies between each other through `import`s, not `reference`s.

native/index.d.ts
```diff
@@ -0,0 +1,68 @@
...
+
+declare module "styled-components/native" {
```
DanielRosenwasser: Declaration files published within packages should be at the top level, and not use the `declare module` syntax. Each `.d.ts` file's top level should align to the exports of the `.js` file.

The reason is that if you had multiple files that wrote out `declare module "foo"`, then TypeScript wouldn't know how to pick the "correct" one. Being a module with a unique path on disk means that importing `"foo"` can be determined by the resolution process.

native/styled-components-native-test.tsx
```diff
@@ -0,0 +1,20 @@
+/// <reference path="index.d.ts" />
```
DanielRosenwasser: If you follow my above advice about moving the module out of the `declare module` block, you should just be able to use imports instead of a `/// <reference ... />` comment.

native/index.d.ts
DanielRosenwasser: If you don't do this, you're going to cause problems if two different TypeScript libraries ever depend on `styled-components` and are consumed. For instance:

```
styled-components
     /   \
     A    B
     \   /
       C
```

native/index.d.ts
patrick91: @DanielRosenwasser I just changed the signature to have also the PropTypes, but this won't work with the examples provided, like:



Do you think that's fine for an end user? I'm not sure if there could be a way to pass a custom PropTypeClass without creating a new component

native/index.d.ts
DanielRosenwasser: Sure, check out the `lib` folder in the `typestyle` package (cc @basarat). The thing is that if you're publishing a package with multiple importable things, you should generally include the `.d.ts` files in directories adjacent to the source themselves. This is to mirror the exact path so that they can be imported the same way.

Since I'm looking at the package now, it looks like there should be a `native.d.ts`, not an `index.d.ts` within the `native/` directory.

DanielRosenwasser:
@patrick91 `typings/styled-components.d.ts` should probably be `lib/index.d.ts` to align with the directory layouts, as mentioned above. I'm don't think that should be too intrusive, given that the `.flow` files are in `lib` as well.

native/index.d.ts
patrick91: I removed the declare module blocks, I haven't moved yet the library outside the "typings" folder since the lib folder is created at build time.

Can you have a look at this comment? https://github.com/styled-components/styled-components/pull/152#discussion_r89686267

Should be the last step :) 
Thank you very much for the reviews, really appreciated!

arrayjam:
I just thought I'd share how I've been using these typings. (From an older version, not sure if they're compatible now, or if the current types are more streamlined).

First I defined a StyledElement type: ```type StyledElement<NewProps, HTMLType extends Element> = React.StatelessComponent<React.HTMLAttributes<HTMLType> & NewProps>;```

Then I use it like so:
```
<FancyAnchorElement primary={true} href="http://example.com">Fancy Anchor Elements just want to have fun and have both anchor and fancy props</FancyAnchorElement>

interface FancyAnchorElementProps {
    primary: boolean;
}

const FancyAnchorElement = styled.a`
    color: ${(props: FancyAnchorElementProps) => props.primary ? "steelblue" : "black"};
    text-decoration: none;
` as StyledElement<FancyAnchorElementProps, HTMLAnchorElement>;
```

This gives me typechecking for both normal anchor props and my new props.

Zshazz:
@patrick91 ... This is the first time I've messed with typescript definitions, but here's an idea for something that's really close to working as I think you want:

    type GeneralComponent<P> = React.ComponentClass<P> | React.StatelessComponent<P>;
    type ConstrainedProps<C, P> = C & ({ defaultProps?: P } | { new(props?: P, context?: any): C});
    interface StyledInterface {
      <C extends GeneralComponent<P>, P>(component: ConstrainedProps<C, P>): StyledFunction<C, P>;
    }

I created a (non-exported) combination of the `ComponentClass` and `StatelessComponent` (so it works with both) and used the fact that `StatelessComponent`s always have a `defaultProps` and `ComponentClass`es must have a contructor that takes a props as the first argument. Thus,

    interface BtnProps {
      color: string;
    }
    const HappyButton: React.SFC<BtnProps> = () => (<button>A happy button</button>);
    const CustomizableBtn = styled(HappyButton)`color: ${props => props.color}`

`CustomizableBtn` typechecks and is a `React.StatelessComponent<BtnProps>`

Similarly,

    class MyHappyButton extends React.Component<BtnProps, {}> {
      public render() {
        return <button>my happy button</button>;
      }
    }
    const CustomizableBtn = styled(MyHappyButton)`background-color: ${props => props.color}`;

Also typechecks, but as a `React.ComponentClass<BtnProps> | React.StatelessComponent<BtnProps>`.

Unfortunately, I'm in the process of learning styled-components, so I don't want to say that it works exactly as expected since I'm not comfortable enough with styled-components. But these definitions have helped tremendously, so I hope these will be completed soon. Hope this helps!

Zshazz:
Changing `ConstrainedProps` to

    type ConstrainedProps<C, P> = C & ({ defaultProps?: P } | { new(props?: P, context?: any): any });

(changed the constructor return to `any`) makes `styled(MyHappyButton)...` infer the type to `typeof MyHappyButton` (the SFC version remains working the same as before) which is closer to what I expected from the defined types, so that might be better.

Sorry for double-posting. 😄 I'm working on something that involves this, so I'm learning/iterating as I go.

mxstbr:
How are we looking here? Would love to finally land this!

patrick91:
@mxstbr just found out sometime to work on it! @Zshazz suggestion is working perfectly (I've just pushed the updated code).

@Zshazz thank you very much! Do you have any documentation links for that? :)

I think we can land this, finally! @mxstbr do you want me to make a demo "app" using TS to show how it works with Styled Components?

PS. I haven't updated the yarn lock file yet

mxstbr:
> I think we can land this, finally! @mxstbr do you want me to make a demo "app" using TS to show how it works with Styled Components?

Yes pleeeaseeee that'd be super awesome!!

patrick91:
@mxstbr Sorry for the long wait! Here's the example repo: https://github.com/patrick91/Styled-Components-Typescript-Example

I've also rebased the commits :)

patrick91:
@Zshazz I was trying styled components on a side project and I found that there is another improvement we could make is to let know typescript that the className is passed from the styled function, eg. if I do this:

```
import * as React from "react";

interface LogoProps {
    className: string,
}

class Logo extends React.Component<LogoProps, {}> {
    render() {
        return <div className={this.props.className}>
            Hello World
        </div>;
    }
}

const LogoStyled = styled(Logo)`
    color: white;
    background: linear-gradient(to left, #0575E6 , #021B79);
`;

export default LogoStyled;
```

And then I use the LogoStyled, typescript complains that the property `className` isn't passed:

```
error TS2324: Property 'className' is missing in type 'IntrinsicAttributes & IntrinsicClassAttributes<Logo> & { children?: ReactNode; } & LogoProps'
```
any idea on how to do this?

patrick91:
Just added support to the theme prop, as TS was throwing an error when trying to use the theme prop.

I guess I need to add some docs for using Styled Components with TS. Will try to do over the next few days!

patrick91:
I've improved support for stateless components and now it should get the correct props inside the css. I've added some docs as well, just explaining a couple of caveats. 

https://github.com/patrick91/styled-components/blob/feature/type-def/docs/typescript-support.md

docs/typescript-support.md
```diff
@@ -0,0 +1,67 @@
+# Typescript Support
+
+Styled-components has typescript definitions to allow the library to be
```
mxstbr: To adhere to the style of the docs generally, could you make all mentions of `styled-components` in text look like code by doing this:

```
`styled-components`
```

docs/typescript-support.md
```diff
@@ -0,0 +1,67 @@
...
+used in any Typescript project.
+
+To use the library you can import it as you normally do with any TS dependency.
+
+```
+import styled from "styled-components";
```
mxstbr: Could you remove the semicolon and change from double quotes to single quotes to have consistent style across the docs? Thanks! (this applies to all code snippets, not just this one)

patrick91:
@mxstbr done :)

mxstbr:
Oh and we need to resolve the merge conflicts 🙈

patrick91:
Will do a rebase later on!

patrick91:
@mxstbr should be ok now!

native/index.d.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  ActivityIndicator: StyledFunction<ReactNative.ActivityIndicator, StyledProps<ReactNative.ActivityIndicatorProperties>>;
+  ActivityIndicatorIOS: StyledFunction<ReactNative.ActivityIndicatorIOS, StyledProps<ReactNative.ActivityIndicatorProperties>>;
+
+  // ART: StyledFunction<ReactNative.ART>;
+  DatePickerIOS: StyledFunction<ReactNative.DatePickerIOS, StyledProps<ReactNative.DatePickerIOSProperties>>;
+  DrawerLayoutAndroid: StyledFunction<ReactNative.DrawerLayoutAndroid, StyledProps<ReactNative.DrawerLayoutAndroidProperties>>;
+  Image: StyledFunction<ReactNative.Image, StyledProps<ReactNative.ImageProperties>>;
+
+  KeyboardAvoidingView: StyledFunction<ReactNative.KeyboardAvoidingView, StyledProps<ReactNative.KeyboardAvoidingViewProps>>;
```
mxstbr: I think Button, introduced in #322, is missing here maybe?

patrick91: Good catch! Added now :)

minamorl:
I sent PR to @patrick91 for fixing issue with existing props like event handlers.

patrick91:
@mxstbr hold on! just got a PR on my repo by @minamorl
Will let you know when I merge it :)

patrick91:
@mxstbr we are good to go :)

typings/styled-components.d.ts
```diff
@@ -0,0 +1,162 @@
+import * as React from "react";
+import { StatelessComponent } from "react";
+
+type ConstrainedProps<C, P> = C & ({ defaultProps?: P } | { new(props?: P, context?: any): any });
+type StyledProps<P> = P & { theme: any };
+
+interface StyledFunction<T, P> {
+  (strs: TemplateStringsArray, ...fns: Array<(props: P) => string>): T;
```
Igorbek: I would recommend `fns` to be of type `ReadonlyArray<Interpolation>`.
`export type Interpolation = ((executionContext: StyledProps<P>) => string) | string | number`
- it aligns with flow type
- it allows to pass strings and numbers in addition to functions

Igorbek: Also, if I want to augment props there's no way to do this.
What about adding an overload that accepts additional props type which will be added to styled component?

Igorbek: TypeScript doesn't seem to be allowing syntax for tagged string interpolation with generic type arguments ```styled.a<{ newProp: number; }>` ... ${ (props) => props.newProp } ... `; ```
Filed [an issue there](https://github.com/Microsoft/TypeScript/issues/13397)

Igorbek: Ah, @patrick91 you already filed such issue :) https://github.com/Microsoft/TypeScript/issues/11947 sorry

package.json
```diff
@@ -45,9 +47,14 @@
     "inline-style-prefixer": "^2.0.5",
     "is-function": "^1.0.1",
     "is-plain-object": "^2.0.1",
-    "supports-color": "^3.1.2"
+    "lodash": "^4.15.0",
+    "supports-color": "^3.1.2",
+    "tslint": "^3.15.1",
+    "typescript": "^2.0.7"
```
mxstbr: Duplicate dependencies, these should be under `devDependencies`.

patrick91:
@mxstbr fantastic! Oh, and I'll have a look at @Igorbek comments tonight or tomorrow :)

typings/styled-components.d.ts
```diff
@@ -0,0 +1,162 @@
...
+  tr: StyledFunction<React.ComponentClass<React.HTMLProps<HTMLTableRowElement>>, StyledProps<React.HTMLAttributes<HTMLTableRowElement>>>;
+  track: StyledFunction<React.ComponentClass<React.HTMLProps<HTMLTrackElement>>, StyledProps<React.HTMLAttributes<HTMLTrackElement>>>;
+  u: StyledFunction<React.ComponentClass<React.HTMLProps<HTMLElement>>, StyledProps<React.HTMLAttributes<HTMLElement>>>;
+  ul: StyledFunction<React.ComponentClass<React.HTMLProps<HTMLUListElement>>, StyledProps<React.HTMLAttributes<HTMLUListElement>>>;
+  "var": StyledFunction<React.ComponentClass<React.HTMLProps<HTMLElement>>, StyledProps<React.HTMLAttributes<HTMLElement>>>;
+  video: StyledFunction<React.ComponentClass<React.HTMLProps<HTMLVideoElement>>, StyledProps<React.HTMLAttributes<HTMLVideoElement>>>;
+  wbr: StyledFunction<React.ComponentClass<React.HTMLProps<HTMLElement>>, StyledProps<React.HTMLAttributes<HTMLElement>>>;
+
+  // SVG
+  circle: StyledFunction<React.ComponentClass<React.SVGProps>, React.SVGAttributes<SVGCircleElement>>;
```
Igorbek: Here in all SVG elements `StyledFunction` is used without `StyledProps`. Is that svg-specific where it cannot accept `theme` prop?

typings/styled-components.d.ts
```diff
@@ -0,0 +1,162 @@
+import * as React from "react";
+import { StatelessComponent } from "react";
+
+type ConstrainedProps<C, P> = C & ({ defaultProps?: P } | { new(props?: P, context?: any): any });
+type StyledProps<P> = P & { theme: any };
```
Igorbek: there's `ThemeProps` interface defined below. reuse?

typings/styled-components.d.ts
```diff
@@ -0,0 +1,162 @@
...
+
+interface StyledFunction<T, P> {
+  (strs: TemplateStringsArray, ...fns: Array<(props: P) => string>): T;
+}
+
+interface StyledInterface {
+  <C extends React.ComponentClass<P>, P, ThemeInterface>(component: ConstrainedProps<C, StyledProps<P>>): StyledFunction<C, P>;
+  <C extends React.StatelessComponent<P>, P, ThemeInterface>(component: ConstrainedProps<C, P>): StyledFunction<C, P>;
```
Igorbek: should `component` be of type `ConstrainedProps<C, StyledProps<P>>`? that would mean it could use `theme` prop.
Also, it should return component with `StyledProps<P>` because resulting component allows to pass `theme` as a prop.

minamorl:
Just suggestions:
- Adding tsconfig in typings folder is more helpuful when testing it in another editor.  I mean like what definitelytyped did. 
- Adding testing for SVGComponent.
Personally I think those can be sent as separated PR. Shipping the code first, then accept another PR on typings seems better approach than discussing more about on this.
At least typings by @patrick91 works fine on normal React element.

.travis.yml
```diff
@@ -5,8 +5,8 @@ node_js:
   - "7"
 script:
   - yarn run danger run -- --verbose
-  - yarn run flow
-  - yarn run lint && npm run test
+  - npm run flow
+  - npm run lint && npm run tslint && npm run test
```
geelen: Probably can still be yarn, right?

docs/typescript-support.md
```diff
@@ -0,0 +1,67 @@
...
+A very basic example can be found [here](https://github.com/patrick91/Styled-Components-Typescript-Example).
+
+## Caveats
+
+### Class Name
+
+When defining a component you'd need to have `className` marked as optional
+in the propTypes interface:
+
+```
```
geelen: Add language identifier ```` ```ts ````

geelen:
This is great. Happy to see this ship as a `1.x` version if it's ready in time.

mxstbr:
Any update @patrick91? Really want to ship this as another 1.x version!

patrick91:
Sorry for the radio silence, I've been busy! I should be able to work on this during the weekend :)

mxstbr:
No worries, glad to hear! Can't wait to ship this 😊

typings/styled-components.d.ts
patrick91: You mean doing something like changing this:

```ts
interface StyledFunction<T, P> {
  (strs: TemplateStringsArray, ...fns: Array<Interpolation<P>>): T;
}
```
to this?
```ts
interface StyledFunction<T, P> {
  (strs: TemplateStringsArray, ...fns: Array<Interpolation<P>>): T & { props: StyledProps<P> };
}
```

.travis.yml
patrick91: Yup, I forgot to change when I did a rebase :)

patrick91:
@mxstbr found some time today :)

@geelen @Igorbek I should have addressed all your points :)

@minamorl yup, I think we can refactor a bit the tests when this lands and people start using it.

typings/styled-components.d.ts
patrick91: I did it, seems fine :)

typings/styled-components.d.ts
```diff
@@ -0,0 +1,163 @@
+import * as React from "react";
+import { StatelessComponent } from "react";
+
+interface ThemeProps {
+  theme: Object;
+}
+
+type ConstrainedProps<C, P> = C & ({ defaultProps?: P } | { new(props?: P, context?: any): any });
+type StyledProps<P> = P & ThemeProps;
+type Interpolation<P> = ((executionContext: StyledProps<P>) => string) | string | number
```
Igorbek: Sorry for being pedantic, an interpolation function can also return a number:
```ts
type Interpolation<P> = ((executionContext: StyledProps<P>) => string | number) | string | number
```

typings/styled-components.d.ts
```diff
@@ -0,0 +1,163 @@
...
+
+interface StyledFunction<T, P> {
+  (strs: TemplateStringsArray, ...fns: Array<Interpolation<P>>): T & { props: StyledProps<P> };
+}
+
+interface StyledInterface {
+  <C extends React.ComponentClass<P>, P, ThemeInterface>(component: ConstrainedProps<C, P>): StyledFunction<C, P>;
+  <C extends React.StatelessComponent<P>, P, ThemeInterface>(component: ConstrainedProps<C, P>): StyledFunction<C, P>;
```
Igorbek: That's not really a concern, I'm just curious. Why are you trying to preserve the original component type?
Here in the case of stateless component, a resulting _styled_ component will no longer be a stateless component.
It will be wrapped in a separate component class and it's original type will be just hidden inside.
So my understanding is that whatever was before style been applied, only its props would be really significant and need to be passed down.
So assuming this, my suggestion would be to slightly simplify and remove `C` from `Styled Function`.
If you don't mind, I can make a change as a proof of concept and send you a PR to your branch.

Igorbek: ok, @patrick91 I've created a PR to yours: patrick91#3
`native' module hasn't been done yet. Let me know if would like to incorporate my changes.

Igorbek:
Just a heads up, guys. I'm waiting a review on [my PR](patrick91#3) by @patrick91. If you want to take a look/review/comment before it gets merged there, you're welcome.

Igorbek:
@mxstbr @geelen please review. I think it's ready to be merged.

mxstbr:
Amazing work everybody!! :tada:

Do we need to publish these under `@types` too or does it work out of the box?

Igorbek:
@mxstbr it works out of the box. `@types` are used in case an official package doesn't have typings or need to override them.

patrick91:
Yes! @mxstbr thanks for merging and huge thanks to @Igorbek @geelen @minamorl @Zshazz @DanielRosenwasser and everyone who commented and helped with this PR :)

mxstbr:
Published as v1.4.0! :tada:

https://twitter.com/mxstbr/status/824182042149654529

bhovhannes:
@patrick91, thanks for your work!

I have an issue with new typings. It appear that the syntax
```
const NavItem = styled.span`
    color: ${props => props.isSelected 'red' : 'black' });
`
```
 results in

`error TS2339: Property 'isSelected' does not exist on type 'StyledProps<HTMLProps<HTMLSpanElement>>'.`

Did I missed something, or it is a valid issue?

Igorbek:
@bhovhannes Unfortunately, TypeScript cannot infer what type of props you're adding. You have to explicitly define the type of additional props:

```ts
interface NavItemProps { isSelected?: boolean; }
const NavItem = styled.span`
    color: ${(props: NavItemProps) => props.isSelected 'red' : 'black' });
`
```

bhovhannes:
@Igorbek, thank you. I'll proceed with the syntax you recommended.

However I wonder if the following syntax can be made possible:
```typescript
interface NavItemProps { isSelected?: boolean; }
const NavItem = styled<NavItemProps>.span`
    color: ${(props) => props.isSelected 'red' : 'black' });
    font-weight: ${(props) => props.isSelected 'bold' : 'normal' }); 
`
```

That way we will avoid writing `props: NavItemProps` every time.

patrick91:
@bhovhannes unfortunately that's not (yet) possible, see: https://github.com/Microsoft/TypeScript/issues/11947

An alternative could be this: 

```ts
interface NavItemProps { isSelected?: boolean; }

const NavItem:React.StatelessComponent<NavItemProps> = ({children, ...props}) => <span {...props}>{children}</span>

const StyledNavItem = styled(NavItem)`
    color: ${(props) => props.isSelected ? 'red' : 'black' });
    font-weight: ${(props) => props.isSelected ? 'bold' : 'normal' }); 
`
```

PR #153, opened 2016-10-31
Add display name to generated class

JamieDixon:
I'm opening this PR more for discussion than to actually be included at this time.

I wanted to have the ability to see the `displayName` of the rendered component in the css `class` when using dev tools in the browser. Sometimes it can be tough to figure out which component generated which class name, especially when switching from custom `className="my-descriptive-foo` to `class="hbfd5"`.

screen shot 2016-10-31 at 21 26 53

This current PR adds the `displayName` all the time however I'm sure there's a way of doing this so it only happens in Dev mode. I'll need a pointer in the right direction to do that.

Here's what I have so far. Please have a look and let me know if there's a different approach which might work better. Thanks.

- [x] Suffix `displayName` onto generated class when `process.env.NODE_ENV === 'development'`
- [x] Strip out CSS selector characters from `displayName`


mxstbr:
I like it, but I really thought we were doing something like this already!

dvkndn:
is there any way to apply this for development only?

vdanchenkov:
Great addition! I'm so glad that it works smoothly with my babel plugin. Although, there is minor problem.

I might miss something, but in my tests two components with equal rule set displays this way.
```jsx
const Arguments = styled.div`
  color: red;
`

const Condition = styled.div`
  color: red;
`
``` 

```jsx
export default () => (
  <div>
    <Arguments>A</Arguments>
    <Condition>B</Condition>
  </div>
)
```

```jsx
<div data-reactid="19">
  <div class="jldpnK-Arguments" data-reactid="20">A</div>
  <div class="jldpnK-Arguments" data-reactid="21">B</div>
</div>
```

I'd expect either different classes (jldpnK-Arguments, jldpnK-Condition), or one jldpnK-Arguments-Condition. I guess latter is not possible.

JamieDixon:
Thanks @vdanchenkov. We've got some ideas to fix that problem. Looking into it now.

JamieDixon:
@vdanchenkov Your observation should be fixed in the latest commit. I've changed the way we add the `displayName` to the class and it's no longer combined with the hash.

Let me know what you guys think.

JamieDixon:
Now, instead of appending something to the generated class hash, we're just inserting the display name as a separate class value. Here's how it looks in dev given these recent changes:

screen shot 2016-11-01 at 16 46 27

JamieDixon:
Replaced by https://github.com/styled-components/styled-components/pull/160

PR #160, opened 2016-11-02
Add component display name along with generated class

JamieDixon:
I'm raising this PR to replace https://github.com/styled-components/styled-components/pull/153 which I'll close. The commit log got into a bit of a mess and I couldn't pick it apart. This PR contains only the commits necessary for the final outcome of this change.

Add `displayName` as a class value in the output HTML when in development mode to make it easier to debug from the browser console

This PR adds the component `displayName` as a class when `process.env.NODE_ENV !== 'production'`

screen shot 2016-11-01 at 16 46 27


ehd:
@JamieDixon This is great! I’ve also had a discussion with @mxstbr and @geelen on Twitter about this. I think this would pair very well with a babel-transform that adds `displayName`s to styled components in development (or also production).

JamieDixon:
@ehd Yes! We've tested this locally with the babel-transform (https://github.com/vdanchenkov/babel-plugin-styled-components-named) and it works great.

ehd:
Hot damn, I want this 😄 Let me know if I can assist with this PR in any way. It would help in two projects I’m working on with a team where people had trouble finding components by name.

mxstbr:
What's the reason for not prepending the name to the hash? I read #153 but couldn't quite make out why that was changed.

That would also guarantee that the name doesn't exist in the global namespace, without having to have tons of emojis in the DevTools. (not a fan of that)

Let's get this landed asap!

ehd:
> That would also guarantee that the name doesn't exist in the global namespace, without having to have tons of emojis in the DevTools. (not a fan of that)

Yeah, that’s a good point. If there’s no objection to concatenating I’d prefer that, too. Bonus: People are discouraged from writing external CSS matching on 💅FooBar💅 when it’s concatenated with the random string.

JamieDixon:
One of the problems I had was that the check for whether the className had been generated already was based on the hash (`nameGenerator`) and the hash has to be generated from a number as per `generateAlphabeticName = (code: number)`. This meant we couldn't suffix with the `displayName` until after the hash was created but because of that, the check for a cashed hash wasn't taking into account the `displayName`. To get around this I thought it would be easier to just subvert the problem and append the `displayName` as a separate class name.

However, you're absolutely right that this isn't necessary and might cause other problems like name clashes. We can get around this by reverting to a similar proposal as earlier and have a `hash` and a `hashSuffixed` which we use for the cache. I'm about to submit the code that makes this change.

As for the emojis in the DevTools, I thought it was cute but I've removed it now :)

JamieDixon:
I think this PR is now ready and inline with your comments above. Let me know if there's more to be done here. 👍

ehd:
Btw, I’d be happy to contribute a test case for this, but not sure where to push 😄

mxstbr:
@ehd push a new branch based on this branch to your fork, and submit a PR against JamieDixon's fork!

ehd:
@mxstbr @JedWatson I shall!

ehd:
@mxstbr @JamieDixon Good morning/evening/`timeZoneAppropriateGreeting`! So the test case is fairly simple:

``` js
it('should add displayName as className suffix', () => {
  const Comp = styled.div``
  Comp.displayName = 'Comp'
  shallow(<Comp />)
  expectCSSMatches('.a-Comp { }')
})
```

However, something came to my attention:

The current implementation only attaches the displayName suffix when `NODE_ENV=development`, which I agree is a good idea.

However, when running the tests with `NODE_ENV=development`, the current test cases which assert on generated CSS all fail because each className is now suffixed with `-styled-div`.

This probably also means that when using styled-components in development, all current styled-components classNames will have a suffix of `-styled-something` unless you explicitly set `displayName`s or add the babel transform. This might actually add value to the current class names by hinting at where they are coming from and thereby help devs with finding definitions.

Two questions:

1. Should we set `NODE_ENV=development` when running tests?
2. Should we change the suffix generation to only suffix for custom `displayName`s? Or adjust the current tests?

mxstbr:
> Should we set NODE_ENV=development when running tests?

Can we somehow test that it does the appending in dev, but doesn't do it prod? (i.e. test both cases?)

> Should we change the suffix generation to only suffix for custom displayNames? Or adjust the current tests?

I like this. `styled-div` seems a bit pointless, since one sees `<div class="xYhfd-styled-div" />` in the DevTools – well, who would've thought it's a div! 😱

ehd:
> Can we somehow test that it does the appending in dev, but doesn't do it prod? (i.e. test both cases?)

1. I wouldn’t want to add an extra parameter to `ComponentStyle` and make it part of the public API.
2. Run mocha twice with different NODE_ENV settings.
3. Override process.env.NODE_ENV before and after that test.

I think 3 is the smallest change but could break, 2 is the cleanest, 1 doesn’t seem worth it.

Also, we should check for `NODE_ENV !== 'production'` instead of checking for dev like React does.

> I like this. styled-div seems a bit pointless, since one sees <div class="xYhfd-styled-div" /> in the DevTools – well, who would've thought it's a div! 😱

That was my first impression, but then I thought it might hint at where these otherwise random classNames are originating from 🤔

JamieDixon:
Thanks for this feedback @ehd. Loving it :)

I switched `NODE_ENV !== production` in one of the latest commits to fix the broken tests, otherwise I agree that it's the better approach.

For the cases where `xYhfd-styled-div` is an issue, I'd suggest using the https://github.com/vdanchenkov/babel-plugin-styled-components-named plugin to ensure `displayName` is correctly added. This is how we're doing it on our project.

RE the environment during a test, I'm a big fan of tests being clear and easy to understand at a glance. This is one of the reasons I never use `beforeTest` and `afterTest` hooks. I like the look of option 3 the most "Override process.env.NODE_ENV before and after that test" where we set the `NODE_ENV` twice as part of a single atomic test. What do you think?

I'm happy to change the test to `!== production` and to double up on the current tests so we check that the `displayName` is appended when `NODE_ENV=test` and not appended when 'production'.

Thoughts?

ehd:
> Thanks for this feedback @ehd. Loving it :)

Thank you, I appreciate that! I love this PR! 😄

> I switched NODE_ENV !== production in one of the latest commits to fix the broken tests, otherwise I agree that it's the better approach.

Gotcha!

> For the cases where xYhfd-styled-div is an issue, I'd suggest using the https://github.com/vdanchenkov/babel-plugin-styled-components-named plugin to ensure displayName is correctly added. This is how we're doing it on our project.

I was planning on adding the transform to the projects where I’m using styled-components. I reckon @mxstbr was thinking of cases when the transform hasn’t been added. I’m personally still leaning towards adding display name suffixes even if they’re just going to result int `-styled-something`, but leaving them out seems perfectly fine, too.

> RE the environment during a test, I'm a big fan of tests being clear and easy to understand at a glance. This is one of the reasons I never use beforeTest and afterTest hooks. I like the look of option 3 the most "Override process.env.NODE_ENV before and after that test" where we set the NODE_ENV twice as part of a single atomic test. What do you think?

Yeah, that sounds good! Also allows you to see the expected behavior around NODE_ENV at a glance instead of having to look at two test runs in `package.json` or something.

> I'm happy to change the test to !== production and to double up on the current tests so we check that the displayName is appended when NODE_ENV=test and not appended when 'production'.

Btw, NODE_ENV isn’t even set to `test` as far as I know, currently it’s just not set at all during testing. So if we change it back to matching `!== `production'` that should work out for tests, too! Adjusting current tests was as simple as replacing ".a " and ".b " with ".a-styled-div " and ".b-styled-div " respectively.

Only need to agree on classNames generated from default displayNames now :)

mxstbr:
> I’m personally still leaning towards adding display name suffixes even if they’re just going to result int -styled-something, but leaving them out seems perfectly fine, too.

I don't have a hugely strict opinion on this, if you prefer it so be it! 👍

mxstbr:
Is there anything left to be done here?

JamieDixon:
Everything works as it is however if we want to change the check to `NODE_ENV !== production` then I'll need some time to change all the tests and write new tests for the condition where `NODE_ENV ===  production`. 

I'm happy to do that and could even do it as a separate PR once this is merged in? Might be best if people would get a lot of value from this right now. What do you think?

mxstbr:
Let's do that fix first, I don't want to ship a version just to ship another one straight afterwards that works differently!

JamieDixon:
The fix is now done. I've changed to using `NODE_ENV !== 'production'` and updates all test cases.

The next step is to add new tests that check that the display name is not added when we're in production mode. Happy to do that as part of this PR or as a separate update.

ehd:
Looking great 😄 I’m also happy to contribute the additional tests then!

mxstbr:
Please add that test so we got it, then we can merge and ship this! 🎉

ehd:
@JamieDixon Will send a PR your way tonight!

ehd:
🎉 → https://github.com/JamieDixon/styled-components/pull/1/files

JamieDixon:
Thanks @ehd. Awesome stuff! Looking forward to landing this one. The people on our team will be so happy 💃

geelen:
Hey just caught up on this issue, had a couple of thoughts.

I don't think the `-styled-div` should be added unless the user has set the `displayName` (i.e. with the babel plugin). Given that most people are starting without that, it seems noisy to add it to all their classnames.

Also, be careful with anything that affects caching behaviour between dev and prod. There could be some _spectacular_ bugs introduced that way, which people will only see when pushing to prod. Rare, sure, but high pain.

Personally I'd inject an unselectable classname instead of concatenating the hashed class. Like, starting with a number or a zero-width space character or something. TBH the 💅 would be sufficient.

Looks great otherwise. Very keen to try to make the dev env here as good as possible 👍

mxstbr:
> Also, be careful with anything that affects caching behaviour between dev and prod. There could be some spectacular bugs introduced that way, which people will only see when pushing to prod. Rare, sure, but high pain.

How does this affect caching behaviour between dev and prod? When would there ever be an overlap between those two?

ehd:
Any update here?

mxstbr:
This will superseded by #227 which solves this problem in a more thorough way. Together with the new `babel-plugin-styled-components` (which builds upon the original `styled-components-named`) we'll have comprehensively solved this, and a bunch of other, issues. I'm working on getting #227 shipped literally all of today!

Sorry for letting you go through all of this effort @JamieDixon just to then not merge this PR, but it set us off in the right direction!

ehd:
@mxstbr Oh hey this is even more awesome, thank you so much!

JamieDixon:
Aha! Well great, glad that there's a better solution to this problem. Being new to OSS it feels hard when a PR doesn't get merged but the learning from feeling bad is always good. New personal discoveries made. 💃

mxstbr:
Really sorry for not merging it @JamieDixon, seriously thanks so much for taking the time to make this happen!

PR #161, opened 2016-11-02
Use `innerRef` to get DOM node in the example code

TaylorBriggs:
The doc says to use `innerRef` on a styled component to get its DOM node, but the code example uses `ref`. This change updates the code example to be consistent with the docs.


mxstbr:
Ah, thank you!

Issue #128, opened 2016-10-21
Web Components support


mxstbr:
From a cursory glance at web components, supporting them should be possible. This should probably live at another entry point (like `import styled from 'styled-components/web'`), and the only thing that would need to happen is instead of returning a react component, we return a web component.

_The entire react-specific part of `styled-components` lives in a single, tiny file: `src/models/StyledComponent.js`. (and `StyledNativeComponent.js`)_

I won't be able to really dig into this until after #120 and #59, so if somebody wants to pick this up it'd be greatly appreciated!

stephenkingsley:
you want to return a web component, not just react component?

mxstbr:
Yes, exactly, return a web component. Not sure if the API makes any sense, but love to try!

stephenkingsley:
Yeah! I think first we can try to return some common web components like `div`, `h1`-`h6`, `section`, `input`. and then observe some people's feedback

stephenkingsley:
may I have this challenge? @mxstbr

mxstbr:
Sure, please go ahead @stephenkingsley!

stephenkingsley:
@mxstbr those day, my work is too busy, so give me more time please~

stephenkingsley:
just now, I try a demo for suppert web components. I want to have your opinions.

in `src/models`, I create a file `StyledWebComponents.js`

```js
// @flow

import type { RuleSet, Target } from '../types'

export default (ComponentStyle: any) => {
  const createStyledComponent = (target: Target, rules: RuleSet) => {
    /* Handle styled(OtherStyledComponent) differently */
    const isTag = typeof target === 'string'
    let element
    const componentStyle = new ComponentStyle(rules)
    const generatedClassName = componentStyle.generateAndInjectStyles({})
    if (isTag) {
      element = document.createElement(target)
      element.className = generatedClassName
    }
    return element
  }

  return createStyledComponent
}

```

and how to use? the example like this

```js
const Title = styled.styledWeb.h1`
  font-size: 1.5em;
  text-align: center;
  color: palevioletred;
`;
Title.textContent = 'hello style-components'
// Create a <Wrapper> react component that renders a <section> with
// some padding and a papayawhip background
const Wrapper = styled.styledWeb.section`
  padding: 4em;
  background: papayawhip;
`;

Wrapper.appendChild(Title);

document.querySelector('#container').appendChild(Wrapper);
```

[...]

PR #164, opened 2016-11-03
add a feature aboute web components

stephenkingsley:
add a feature aboute web components  #128 

how to use? 

```js
const Title = styled.styledWeb.h1`
  font-size: 1.5em;
  text-align: center;
  color: palevioletred;
`;
Title.textContent = 'hello style-components'
// Create a <Wrapper> react component that renders a <section> with
// some padding and a papayawhip background
const Wrapper = styled.styledWeb.section`
  padding: 4em;
  background: papayawhip;
`;

Wrapper.appendChild(Title);

document.querySelector('#container').appendChild(Wrapper);
```


philpl:
It's a year later and I don't think we'll provide this functionality any time, or any time soon.
I think it would fit in more with something like styled-elements: https://github.com/styled-components/styled-elements

Every surface area that we add we'll have to maintain and as this is a really different platform, we should think about separating any efforts from the main styled-components library.

cc @mxstbr

PR #166, opened 2016-11-04
Use Proxy instead of adding 130+ helper lines

KevinMartin:
By using Proxy, you also inherently support any elements added by future specs, any missing elements, and HTML5 Custom Elements.

On the other hand, projects using this would need to add a Proxy polyfill like:
- https://github.com/krzkaczor/babel-plugin-proxy
- https://github.com/GoogleChrome/proxy-polyfill

or other since [IE and Safari <= 9 don't support Proxy](http://caniuse.com/#search=proxy).


KevinMartin:
Not sure what's up with the failure in `InlineStyle.js`... :/

mxstbr:
Huh, this looks like a really smart solution! How much is the proxy polyfill? Not a fan of having that if it meaningfully increases package size, even though I see the value of having this proxy here...

KevinMartin:
> How much is the proxy polyfill?

Free.

It actually depends which route you go in... If you go the babel route, [it's about 30 short lines of code](https://github.com/krzkaczor/babel-plugin-proxy/blob/master/src/runtime.js).

mxstbr:
> it's about 30 short lines of code.

That's less lines of code than our helpers are currently! I'm a fan, let's do it. Can you add that polyfill and fix the CI error?

> TypeError: src/models/InlineStyle.js: Property value expected type of string but got null

JamieDixon:
I love this change. Nice work! 🍾

KevinMartin:
@mxstbr Don't mean to sound like a noob, but I'm not sure what's wrong with `InlineStyles.js`. I'm not too familiar with `flow` or `babel-types`. Can you help?

geelen:
Umm, from [babel-plugin-proxy](https://github.com/krzkaczor/babel-plugin-proxy)

> It is not suitable for production environments because performance impact is huge.

Also, from my understanding, the polyfill only works across call-sites. So it would mean that every user trying to use `styled.div` would need the Babel polyfill installed and some how to know that `styled` was a Proxy object, which would mean compiling against the un-transpiled Styled Components source, right?

Am I missing something here? I looked into Proxies right at the beginning of this project and concluded that there was no reasonable way to use them until they were implemented natively. Has that changed?

mxstbr:
I don't see anything in the https://github.com/GoogleChrome/proxy-polyfill/ that points towards a "Don't use this in production"?

KevinMartin:
I've used the `babel-plugin-proxy` in production before. No issues for me.

But yes, in order for other people to use the helpers in their projects (with support for IE and older browsers), they would need to also install some sort of proxy polyfill or babel plugin.

geelen:
@mxstbr from https://github.com/GoogleChrome/proxy-polyfill/

> This means that the properties you want to proxy must be known at creation time.

I.e. the `get` trap will only fire for _properties already defined on the object_. Which defeats the whole purpose.

geelen:
> But yes, in order for other people to use the helpers in their projects (with support for IE and older browsers), they would need to also install some sort of proxy polyfill or babel plugin.

That, to me, is a dealbreaker.

mxstbr:
Seeing as this has too many downsides I wasn't aware of, let's not merge this for now :cry: thanks for kicking off the discussion @KevinMartin!

Issue #59, opened 2016-10-13
Babel transform


mxstbr:
~~We need a way (probably babel transform?) to extract and precompile the static styles of our components!~~

Extracting static styles is actually not what we want, we want to pre-parse the CSS in a build step but leave it in the JS: See [this comment](https://github.com/styled-components/styled-components/issues/59#issuecomment-257092965)

Has anybody done this before? I know I haven't…

tizmagik:
Take a look at CSJS, similar template strings construct and has a pipeline to statically extract styles

mxstbr:
Good tip, this doesn't look _too_ hard! https://github.com/rtsao/csjs-extractify/blob/master/index.js (browserify though…)

oliviertassinari:
I'm curious about the motivation behind.
Shouldn't we aim for the opposite? What I mean is, injecting as small bits of CSS as posible in the document in order to:
- reduce network usage when inlining CSS for SSR
- reduce bootstrap JS execution
- reduce bootstrap browser CSSDOM processing

geelen:
@oliviertassinari the reason is that once certain styles are defined they can never change. Some component `X` will always have some chunk of styles applied. So we can basically parse over the files and extract that into a static mapping, which is effectively all a CSS file is anyway :)

If you're doing SSR, you _might_ prefer not to do this, but if you have a fairly common set of components across the entire site it could be better to extract all _those_ into a static, separately-cached CSS file. But not everyone is doing SSR.

As for bootstrap JS & CSSDOM parsing, serving raw CSS is always going to be faster than serving JS, executing it and dynamically injecting styles. Browsers have been optimising their CSS parsing since... CSS became a thing. So I'd like to make use of as much of that as is possible.

ndelangen:
With the content of the stylesheet classes changing between renders (because they are based on props) how is this going to work?

Whatever CSS is extracted, is only for the initial state? Looking at how it works it seems each different render wich results in a different style output will also result in a new classname.. 
I don't see an easy way of pre-generating all possible classes.

take this style:
```
styled.button`
  background: ${(props) => props.theme.main || 'palevioletred'};
  border: ${(props) => (props.width || 2) + 'px'} solid ${(props) => props.theme.main || 'palevioletred'};
`;
```

Potentially `props.width` could be any Integer, resulting in infinite classes to generate?

As @tizmagik pointed out, CSJS seems rather similar and also support interpolation inside the template strings, so might deal with the same issue?

[...]

PR #172, opened 2016-11-08
Extract DOM shorthands into a separate file

philpl:
A really, *really* simple PR that extracts the DOM shorthands to a separate file for readability :)

I definitely want to get more into styled-components development, since I'm totally stoked by #59. I was even considering starting my own styling library for getting a babel transform for styling working, but I'm really into styled-components, so I'd love to help out here :)


mxstbr:
This is a great change, but a bit dependent on how #166 plays out. If we decide to go the proxy route we won't merge this, but if we don't I'll definitely land this!

Thanks for taking the time to do this refactor, appreciated!

(Also, any help with the babel transform would also be much appreciated)

philpl:
@mxstbr Cheers for the quick reply :)

For what it's worth, pushing the proxy overhead to the user sounds like it could frustrate a lot of people. I don't see why `styled('unsupported-element')` would be bad in any way.
These are after all just shortcuts ;)

mxstbr:
The proxy polyfill is 30 lines of code, so that's even less than all of our helpers, plus it allows you to use whatever custom element you want! 🎉

I'm going to close this in favor of #166, thank you so much for taking the time to contribute!

mxstbr:
So as it turns out, we're not going to merge #166 after all, so I'm reopening this PR! Let's get this landed!

mxstbr:
@styled-components/typers this broke flow, as now `styled.div`, `styled.a`,... are dynamically defined, so flow complains that `styled.x` is undefined. Any ideas how we can fix that?

relekang:
@mxstbr Looks like it works with adding `// @flow` to the domElements file. PR coming up 🙂

mxstbr:
Oh man, how'd I miss that! Ugh.

philpl:
Sorry for not being active on this :) have yet to stay on top of my Github notifications ^^

PR #186, opened 2016-11-12
Add a new test with composition in mind

oliviertassinari:
I was wondering how the lib would behave in such circumstance.
I couldn't find where we the order logic is in the source code, but it's working :). Any clue?
Thanks

### DIFF
--- a/src/test/extending.test.js
+++ b/src/test/extending.test.js
@@ -53,6 +53,18 @@ describe('extending', () => {
     expectCSSMatches('.a { color: blue;color: red; }')
   })
 
+  it('should add the child class before the parent', () => {
+    const pure = (BaseComponent) => (props) => <BaseComponent {...props} />
+    const Div = styled.div`color: green;`
+    const Parent = pure(styled(Div)`color: blue;`)
+    const Child = styled(Parent)`color: red;`
+
+    const wrapper = shallow(<Child />)
+    wrapper.shallow().shallow()
+
+    expectCSSMatches('.b { color: green;color: blue; } .a { color: red; }')
+  })
+
   it('should generate different classes for both parent and child', () => {
     const Parent = styled.div`color: blue;`
     const Child = styled(Parent)`color: red;`


mxstbr:
I'm not 100% sure which part you're looking for, is this it? https://github.com/styled-components/styled-components/blob/b69e494638df7427a6c963d55b119e918f0c1ef6/src/models/StyledComponent.js#L16-L22

oliviertassinari:
> I'm not 100% sure which part you're looking for, is this it?

Oh exactly, that's what I was looking fro. Thanks!

So the style is injected in the DOM at the component's creation time.
That mean we have to be cautious with asynchronous js evaluation.
That's probably a rare enough case to be ignored.

E.g.
```js
import React from 'react';
import { render } from 'react-dom';
import defer from './defer';

const Div1 = styled.div`color: green;`
const Div2 = defer(() => (
  styled(Div1)`color: blue;`
))
const Div3 = styled(Div2)`color: red;` //  ---> output blue, expect red

const App = () => (
  <Div3>Click me!</Div3>
);

render(<App />, document.getElementById('content'));
```

<details>
  <summary>defer.js</summary>
```jsx
// @flow weak

import React, { Component } from 'react'

export default function defer(baseComponentGenerator) {
  class Defer extends Component {
    constructor(props, context) {
      super(props, context)

      this.state = {
        show: false,
      }
    }

    componentDidMount() {
      /* eslint-disable react/no-did-mount-set-state */
      this.timer = setTimeout(() => {
        this.setState({
          show: true,
        })
      }, 0)
      /* eslint-enable react/no-did-mount-set-state */
    }

    componentWillUnMount() {
      clearTimeout(this.timer)
    }

    render() {
      if (this.state.show) {
        const BaseComponent = baseComponentGenerator();
        return <BaseComponent {...this.props} />
      }

      return null
    }
  }

  return Defer
}
```
</detail>

mxstbr:
No, it's actually injected at render time – I don't see how that's an issue, even with the above case? I feel like it works exactly how I would expect it to.

oliviertassinari:
@mxstbr You are right sorry. I meant the rules declaration is injected at the component's creating time. It's not injected yet in the DOM.

> it works exactly how I would expect it to.

Right, depend on the use case.
I'm still digging through this rabbit hole. I'm getting confused 😬 .

oliviertassinari:
@mxstbr I have finally figured it out 🎋 >< .
The style injection order is determined at the Component creation time with this [**line**](https://github.com/styled-components/styled-components/blob/master/src/models/StyledComponent.js#L21) that inject a placeholder `styleSheet`. That can be later hydrated at the render time.
That's smart. 
So anything differing the component creation in the HoC chain is going to break the composition order. I better understand the tradeoff now with this approach.

Issue #182, opened 2016-11-11
Media Template Trick


erikras:
Hi all. An issue might not be the best place for this, but I wanted to share a little trick I discovered with the `styled-components` community. It's really too simple to be a feature or utility function in the library itself, but I am finding it extremely useful and wanted to give back.

```jsx
const sizes = {
  giant: 1170,
  desktop: 992,
  tablet: 768,
  phablet: 572,
  phone: 376
}

export const media = Object.keys(sizes).reduce((accumulator, label) => {
  accumulator[label] = (...args) => css`
    @media (max-width: ${sizes[label]}px) {
      ${css(...args)}
    }
  `
  return accumulator
}, {})
```

This creates a media template for each of the sizes, so you can do _this_ in your components:

```jsx
const Container = styled.div`
  color: #333;
  ${media.desktop`padding: 0 18px;`}
  ${media.tablet`padding: 0 18px;`}
  ${media.phablet`padding: 0 10px;`}
  ${media.phone`padding: 0 3px;`}
`
```

Thanks for all the work you are doing. `styled-components` is the most elegant styling solution I've seen so far.

mxstbr:
Would love for you to add this to the Tips and Tricks doc! 👏

wmertens:
You know, that's already nice, but if you use the JS mediaquery API, these values can be read from JS and you can render a completely different UI depending on viewport.

I do this in https://www.npmjs.com/package/redux-mediaquery for Redux (so the app redraws on change) but having these values available in styled would also be really nice. Plus, still not sure if it should be in the store and not in context (but then it doesn't force redraw).

valscion:
Would you be OK with changing the breakpoint example values from `px` unit to `em`? It would work properly cross-browser and support users changing their browsers `font-size` accordingly.

Source: https://zellwk.com/blog/media-query-units/

mxstbr:
Yeah sure, I don't mind that.

erikras:
@valscion Thanks for that link. With CSS it's literally possible to learn some new quirk every single day of your life. 😂

PR #189, opened 2016-11-13
Use em in Media Templates example

valscion:
As per https://github.com/styled-components/styled-components/issues/182#issuecomment-260122678, this changes the Media Templates example to use `em` as the breakpoint unit to work properly cross-browser whilst supporting users who change their browser's `font-size`.

I'm a bit torn by this change, though, as it does add some complexity to the example. I glanced at the other examples that used an example of a breakpoint of `420px`, and I didn't want to go touching them, at least not yet. It is arguably a lot more difficult to think of what `26.25em` means in real world, compared to `420px`.

Source for choosing `em` over `px`: https://zellwk.com/blog/media-query-units/

### DIFF
--- a/docs/tips-and-tricks.md
+++ b/docs/tips-and-tricks.md
@@ -150,8 +150,11 @@ const sizes = {
 
 // iterate through the sizes and create a media template
 export const media = Object.keys(sizes).reduce((accumulator, label) => {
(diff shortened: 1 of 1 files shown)


mxstbr:
I think this is good, we should be encouraging best practices. Thanks!

Issue #138, opened 2016-10-25
Conditional styling not working in React Native


paulxuca:
## Version

1.0.8
## Code

Styling code  

```
const LinkText = styled.Text`
  font-family: Avenir;
  opacity: ${ props => props.active ? 1 : 0.3 }
`;
```

Component Render Code  

```
<LinkText active>Hello</LinkText>
```
## Expected Behavior

Link text should have full opacity (1)
## Actual Behavior

Link text has the opacity 0.3, as if it is not active

suevalov:
Looks like the props are not being passed at all.

mxstbr:
Fixed by the amazing @haikyuu in #190! Will be released soon!

jh3y:
Is there an ETA on that release?  Or best to pull in `master` for now?

Keen to use `styled-components` 👍

mxstbr:
Released as 1.0.11! 🎉

jh3y:
Hero!

jh3y:
Unfortunately I'm not getting any joy with this :cry:

I'm getting `can't find variable: props` error.

My implementation is;
```js
import styled from 'styled-components/native'

const StyledEl = styled.View`
  background-color: ${(props.light) ? 'white' : 'black' };
`
```

Anything I'm missing?

jh3y:
To add, previously it was as if the props were just being ignored, there was no error.

jh3y:
Seems this [PR](https://github.com/styled-components/styled-components/pull/171) could fix the issue but awaiting a rebase 👍 

Hopefully this can get pushed in sharp. I'd be happy to do the rebase but I don't think that's logistically possible :joy:

mxstbr:
You just have a typo there, it needs to be a function:

`${props => props.light ? 'white' : 'black'}` should work!

jh3y:
Ah brilliant! Thanks, typical to miss something simple 😄  :joy:

PR #190, opened 2016-11-13
pass props to interpolations in StyledNativeComponents

haikyuu:
In react native, interpolations had access to the theme object only. This pull request fixes that. #138

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file. If a contri
 
 ### Added
 
-- N/A
+- Fixed props that are passed to interpolations in `StyledNativeComponent`. Now, they get `theme` and component props, thanks to [@haikyuu][https://github.com/haikyuu]. (see [#190](https://github.com/styled-components/styled-components/pull/190))
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


haikyuu:
@mxstbr after putting it in a function, i think we should take it outside `styledNativeComponent.js` and`styledComponent.js`. Where would be a good place for it?

mxstbr:
Let's do that refactor as the next step, this LGTM code-wise. Have you tried running this and verified it fixes the bug? Could you post a screenshot?

haikyuu:
Here is the code:
```javascript
const isOdd = (props)=> {
  console.log("props: ", props);
  return props.number % 2 ? 'steelblue': 'papayawhip'
}

  background-color: ${props=> isOdd(props)};
`;
const StyledText = styled.Text`
  background-color: ${props=> isOdd(props)};
`;
...

render(){
    return(
        <StyledView number={1} style={{flex: 1, marginVertical: 15}}>
          <StyledText number={2}>I am working</StyledText>
        </StyledView>
      )
}
```
And a screenshot from the console, and the device:
screenshot 2016-11-14 10 48 27


screenshot 2016-11-14 10 46 35

PR #201, opened 2016-11-14
fixes #196 Converts Object.assign to spread operator

### LINKED ISSUE #196: IE11 and Object.assign
## Version
1.0.10

IE11 barfs an error because the use of `Object.assign`:

```
Uncaught (in promise) TypeError: Object doesn't support property or method 'assign'
```

Might be worth mentioning that it needs a polyfill, if it can't be solved some other way?

thisguychris:
- switch Object.assign to spread operator
- rationale: babel transforms this to Object assign / or a polyfill equivalent
this way we won't need to add another module like object-assign
- please take note I don't have experience with flow annotations,
so if this is valid `return Object.assign({}, this.outerTheme, (theme: Object))`
then this PR works.

### DIFF
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -34,7 +34,7 @@ export default (ComponentStyle: Function) => {
       }
 
       generateAndInjectStyles(theme: any, props: any) {
-        const executionContext = Object.assign({}, props, { theme })
+        const executionContext = { ...props, ...{ theme } }
         return componentStyle.generateAndInjectStyles(executionContext)
       }
 
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -34,7 +34,7 @@ export default (ComponentStyle: Function) => {
       }
 
       generateAndInjectStyles(theme: any, props: any) {
-        const executionContext = { ...props, ...{ theme } }
+        const executionContext = { ...props, theme }
         return componentStyle.generateAndInjectStyles(executionContext)
       }
 
--- a/src/models/ThemeProvider.js
+++ b/src/models/ThemeProvider.js
@@ -43,7 +43,7 @@ class ThemeProvider extends Component {
   }
 
   getChildContext() {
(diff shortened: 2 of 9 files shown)


thisguychris:
@mxstbr / @geelen  let me know if you want me to convert Object assign inside `ThemeProvider.test.js` too.

mxstbr:
Could you convert `ThemeProvider.test.js` too? Then we can ship this today! Closes #196!

thisguychris:
@mxstbr on it

src/models/StyledComponent.js
```diff
@@ -34,7 +34,7 @@ export default (ComponentStyle: Function) => {
       }
 
       generateAndInjectStyles(theme: any, props: any) {
-        const executionContext = Object.assign({}, props, { theme })
+        const executionContext = { ...props, ...{ theme } }
```
mxstbr: You're right, I think this does the same thing?

thisguychris:
updated PR 👍

thisguychris:
@mxstbr / @JamieDixon anything more to add here? or are we good?

src/models/ThemeProvider.js
```diff
@@ -43,7 +43,7 @@ class ThemeProvider extends Component {
   }
 
   getChildContext() {
-    return Object.assign({}, this.context, { [CHANNEL]: this.broadcast.subscribe })
+    return { ...this.context, ...{ [CHANNEL]: this.broadcast.subscribe } }
```
JamieDixon: I think we can skip the spread here too:
```
{ 
   ...this.context, 
   [CHANNEL]: this.broadcast.subscribe 
}
```

JamieDixon:
Just one minor comment from me about an additional spread that's not needed. Other than that I think this looks great! 👍

thisguychris:
@JamieDixon updated 👍 💯

mxstbr:
Thanks so much man, love it!

Issue #94, opened 2016-10-15
Test ReactNative


mxstbr:
Currently we have some [e2e tests](https://github.com/styled-components/styled-components/blob/master/src/test/e2e.test.js) (and some unit tests, though not enough yet) in place to ensure nothing breaks unexpectedly with changes.

These only test the web version of react though, we somehow need to make sure the changes between `StyledNativeComponent` and `StyledComponent` (and any other possible components) in sync. Any ideas how we could do that?

jacobp100:
I could pick this up. Apart from unit tests (we'll probably want to use jest—#325), we might want to do device snapshot tests.

For snapshot tests, we'll need a RN project—should that be in the core styled-components repo, or should we make a new repo for it? I'm inclined to say new repo because it's easier, and because `css-to-react-native` will also affect the snapshot tests.

I have no idea how to make the CI work either way—if it'll even work with RN 😂

mxstbr:
We test native now! :tada: https://github.com/styled-components/styled-components/blob/master/src/native/test/native.test.js

PR #211, opened 2016-11-17
Use css-to-react-native to parse ReactNative stylesheets

jacobp100:
I couldn’t find any tests for this, and don’t have any example code I can verify it’s working. If somebody could just double check this first.


mxstbr:
There are none, ref #94 😕

Will try to get RN working locally!

mxstbr:
I published this as `styled-components@dev`, this is working! 🎉 Amazing work @jacobp100, thanks so so so much!

screen shot 2016-11-20 at 08 48 10

Issue #134, opened 2016-10-24
Managing highly dynamic styles


JamieDixon:
Example component:

```
const Comp = styled.div`
     /*lots of styles here*/
     animation-delay: ${ props => props.index * 0.1 };
`;
```

With this component rendered over a collection of items, we'll generate a new style tag with an associated class for each one.

Is there a recommended way of working with these kinds of styles so as not to thrash the head with  a new `style` tag everytime `render` is called on `Comp`?
- One solution could be to move the `/* lots of styles here */` into another component that doesn't include the dynamic properties that cause huge amounts of CSS to be rendered into the `head`. This way the generated `style` tag only contains the `animation-delay` property. Still a lot of `style` tags being injected into the head though and since the injection only happens at render time, this could cause some trashing.
- Another option is to put the `animationDelay` into the `style` prop of the component as we compose the components and not let styled-components handle this.

How would you do this? Is there a recommended approach?

I saw there might be some plans to extract static styles which, as a solution, is similar to our first option above.

JamieDixon:
On the other hand, it could be that we're trying to do things with `styled-components` that it's just not designed for.

Another option would be to use something like `Velocity` to handle our animations and `styled-components` for the rest.

mxstbr:
Have you noticed any visible performance impact on the animation with the current way? If so, could you provide a reproduction in a webpack bin?

mikepombal:
Hi @mxstbr,

No performance impact but duplication of the same style again and again which increases the bundle size. Following @JamieDixon example if we have a list of `Comp` components the style created will be the following:
```css
.[generated class name for 1st instance of comp] {
    /*lots of styles here*/
    animation-delay: 0s;
}
.[generated class name for 2nd instance of comp] {
    /*lots of styles here*/
    animation-delay: 0.1s;
}
.[generated class name for 3rd instance of comp] {
    /*lots of styles here*/
    animation-delay: 0.2s;
}
/* and so on */
```
As you see the `/*lots of style here*/`, although it is the same style, appears n times the number of items in the list.

I'm wondering if #59 could actually fix this issue.

vdanchenkov:
Comparison of animation done directly with style prop and with styled-components. http://www.webpackbin.com/VJ3-tBukf  In this example animation with props is about 2 times slower on first run.

JamieDixon:
I came across an example today where styled-components couldn't nicely handle the requirement I had.

[...]

PR #214, opened 2016-11-18
Expose Server Rendering API

### LINKED ISSUE #124: Server Side Rendering
Thought I'd open an issue to kick off the discussion. It's already possible because of the way we build on top of Glamor, but we haven't exposed it as an API. But basically, if you did this:

``` js
import styleSheet from 'styled-components/lib/models/StyleSheet'

/* before each render */
styleSheet.flush()

/* after each render */
styleSheet.rules().map(rule => rule.cssText).join('\n')
```

Then you should get the chunk of CSS you need for each request. Can someone with a server-rendered setup take a look and confirm this works, and maybe show how they'd be invoking it? I'd be happy enough to export something like:

``` js
import { serverStylesheet } from 'styled-components'

/* before each render */
serverStylesheet.reset()

/* after each render */
serverStylesheet.getCSS()
```

What do people think?

thisguychris:
- fixes #124
- remap flush to reset()
- convenience function getCSS() for SSR


thisguychris:
pending tests.

src/index.js
```diff
@@ -20,4 +21,4 @@ const styled = _styled(_styledComponent(_ComponentStyle(generateAlphabeticName))
 
 /* Export everything */
 export default styled
-export { css, keyframes, injectGlobal, ThemeProvider }
+export { css, keyframes, injectGlobal, ThemeProvider, serverStyleSheet }
```
diegohaz: Isn't it better to export it as `ServerStyleSheet` (pascal case)? What do you think?

thisguychris: maybe we need a 3rd opinion? /cc @mxstbr / @geelen

thisguychris:
@diegohaz so I changed it to PascalCase as per your advice. Let me know what else do we need for this PR.  /cc @mxstbr and @geelen

mxstbr:
This LGTM code-wise, but maybe @geelen has something to say about the general approach?

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file. If a contri
 
 ### Added
 
-- N/A
+- Expose API for Server Side rendering: `serverStyleSheet.reset()` and `serverStyleSheet.getCSS()`, thanks to [@thisguychris](https://github.com/thisguychris), (see [#214](https://github.com/styled-components/styled-components/pull/214)) fixes [#124](https://github.com/styled-components/styled-components/issues/124)
```
geelen: Needs capitalisation here

src/models/ServerStyleSheet.js
```diff
@@ -0,0 +1,8 @@
+import styleSheet from './StyleSheet'
+
+styleSheet.reset = styleSheet.flush
```
geelen: Hmm, I'd prefer not mutating things here. I'd suggest this:

```js
class ServerStyleSheet {
  reset() {
    styleSheet.flush()
  }
  getCSS() {
    return styleSheet.rules().map(rule => rule.cssText).join('\n')
  }
}

export default new ServerStyleSheet()
```

That way if `styleSheet` changes we don't have potentially public-facing changes to worry about here.

src/index.js
```diff
@@ -20,4 +21,4 @@ const styled = _styled(_styledComponent(_ComponentStyle(generateAlphabeticName))
 
 /* Export everything */
 export default styled
-export { css, keyframes, injectGlobal, ThemeProvider }
+export { css, keyframes, injectGlobal, ThemeProvider, ServerStyleSheet }
```
geelen: You're exporting a singleton instance, so I'd go with a name like `serverStyles`

thisguychris: @geelen you do realize this comment contradicts the one in the CHANGELOG.md comment right?  You are wanting to remove the capitalization in the export even though you want capitalization in the CHANGELOG?

mxstbr: I think it's more about being consistent 😉 renaming both to `serverStyles` should be fine!

mxstbr:
I think we need to bikeshed on the name a bit more, I'm not super happy with `serverStyles`. (sorry for the drawn out PR @thisguychris!)

This is a bit weird:

```JS
import { serverStyles } from 'styled-components'

serverStyles.getCSS()
```

I don't like the name, because you're not getting the CSS of some styles that are used on the server. (whatever that would mean) You're getting styles you want to have on the client. `serverStyles.getCSS()` kind of doesn't fit that, even `client.getCSS()` is better in that regard. (don't take that, it's not a good name)

Why don't we just call it `styleSheet` and make it part of our public API that way? This makes much more sense imo:

```JS
import { styleSheet } from 'styled-components'

styleSheet.getCSS()
```

From a users perspective it makes sense that `styled-components` has a style sheet, and that you can get the CSS of that style sheet as a string. This is a logical and fitting public API imo.

> There shouldn't be any confusion with users thinking it's the `glamor` `styleSheet`, since nobody knows what that even is except for people who contribute, and those can also check what `styleSheet` is anyway.

thisguychris:
> You're getting styles you want to have on the **client**. serverStyles.getCSS() kind of doesn't fit that

@mxstbr 
Correct me if I'm wrong but we are actually getting styles to have on the **server** right?  Hence the name seems appropriate?  I don't really have a strong opinion if you want a more generic name like `styleSheet`.  I can update this PR :)

thisguychris:
@mxstbr / @geelen cleaned up the commits, and just styleSheet as per your recommendation. Let me know if this works 👍

CHANGELOG.md
```diff
@@ -8,9 +8,12 @@ All notable changes to this project will be documented in this file. If a contri
...
 - Added support for deeply nested styles in ReactNative (e.g. `transform`), thanks [@jacobp100](https://github.com/jacobp100). (see [#139](https://github.com/styled-components/styled-components/pull/139))
 - Added support for camelized style properties in ReactNative (e.g. `fontWeight`), thanks [@jacobp100](https://github.com/jacobp100). (see [#145](https://github.com/styled-components/styled-components/pull/145))
 - Properly expose `flow` typings by adding a `flow:build` step and `flow` support docs, thanks to [@ryyppy](https://github.com/ryyppy). (see [#219](https://github.com/styled-components/styled-components/pull/219))
+- Added support for deeply nested styles in ReactNative (e.g. `transform`), thanks [@jacobp100][https://github.com/jacobp100]. (see [#139](https://github.com/styled-components/styled-components/pull/139))
+- Added support for camelized style properties in ReactNative (e.g. `fontWeight`), thanks [@jacobp100][https://github.com/jacobp100]. (see [#145](https://github.com/styled-components/styled-components/pull/145))
```
mxstbr: Duplicate changelog entries?

mxstbr:
I much prefer the `styleSheet.getCSS()` API. @geelen?

src/models/StyleSheet.js
```diff
@@ -5,4 +5,13 @@
 
 import { StyleSheet } from '../vendor/glamor/sheet'
 
-export default new StyleSheet({ speedy: false, maxLength: 40 })
+class StyleSheetExtended extends StyleSheet {
+  reset() {
+    return super.flush()
+  }
+  getCSS() {
+    return super.rules().map(rule => rule.cssText).join('\n')
```
mxstbr: Why are we including linebreaks by default? Shouldn't we just do `.join('')` for performance reasons, so that the code the developers send down to the client is as small as possible?


If we want to allow devs to get non-minified CSS (for debugging purposes),  maybe we could have an option called `min` which is set to true by default?

```JS
getCSS({ min = true }) {
  return super.rules().map(rule => rule.cssText).join(min ? '' : '\n')
}
```

That way devs can do normal mode, which is minified, or non-minified mode if they want to debug:

```
// minified
styleSheet.getCSS()

// Not minified
styleSheet.getCSS({ min: false })
```

CHANGELOG.md
thisguychris: whoopsie, fixed

src/models/StyleSheet.js
thisguychris: oooh, I like this. 👏  Adding this option now.

mxstbr:
Can we also add tests for this btw? Just render some things and make sure `styleSheet.reset()`, `styleSheet.getCSS()` and `styleSheet.getCSS({min: false})` do the correct things 👌

(Sorry for the bikeshedding, but this'll be a huge part of our public API so we need to make sure we're solid here)

thisguychris:
on it @mxstbr

geelen:
This looks good to me. I'm gonna refactor `StyleSheet.js` a bit because it's growing in to our _own_ thing, using Glamor underneath is an implementation detail. But it won't change the public API so feel free to release without that. (I'll PR separately)

geelen:
Pushed my changes as a new branch and have PR-ed them against your project @thisguychris: https://github.com/thisguychris/styled-components/pull/1

thisguychris:
@geelen 👋  merged your commit from my fork and pushed it here in this PR.
@mxstbr let me know if this tests suffice.

thisguychris:
cleaned up commits as well 🎉

thisguychris:
It's worth to note that I wanted to write this exactly how @geelen wrote the refactor :D forgot the `as` keyword :D.  TIL.

lucianlature:
Somebody please merge this :) 👍

mxstbr:
@lucianlature we still need to figure out how we're going to do #227 before we can ship this, sorry!

Kuirak:
Is there any ETA when this will be ready?

mxstbr:
@Kuirak we'll try to land #227 asap, so in around a week this should be ready I hope!

satya164:
Can you give an example of how to use the API? I'm assuming something like this,

```js
serverStyles.reset();
const html = ReactDOMServer.renderToString(<App />);
const css = serverStyles.getCSS();
```

Am I right?

I think something like this is a much nicer API though,

```js
const { css, html } = StyleSheet.collect(() =>
  ReactDOMServer.renderToString(<App />)
)
```

mxstbr:
I like that API! We can bikeshed around this a bit more until #227 finally lands, what are the APIs of other CSS-in-JS libraries like?

satya164:
Ok, I haven't seen this, but aphrodite's API looks quite similar

```js
var {html, css} = StyleSheetServer.renderStatic(() => {
    return ReactDOMServer.renderToString(<App/>);
});
```

And glamor's API,

```
let { html, css, ids } = renderStatic(() =>
  ReactDOMServer.renderToString(<App/>)) // or `renderToStaticMarkup`
```

thisguychris:
@mxstbr any idea why in appveyor it shows a reverse vendor prefixed styles?

```css
  -.a { transition: opacity 0.3s; -webkit-transition: opacity 0.3s; }
  +.a { -webkit-transition: opacity 0.3s; transition: opacity 0.3s; }
```

code-punkt:
@thisguychris How do you stop styled-components from re-injecting all those server rendered styles at the server again? is there some sort of client-side hydration i couldn't find in the PR's sources?

thisguychris:
@mxstbr what @code-punkt wants is a rehydration function, where client will 'skip' injecting style tags twice, (one in ssr and and in client side).  

On this PR, we are injecting the styles twice.  While getting the actual ids is pretty trivial in SSR.  I need some direction where to put the expensive check in client side style injection.  DM'd you.

mxstbr:
Let's keep the discussion here so everybody can chime in! I don't actually know what the best place to put it would be, what are your ideas?

code-punkt:
@thisguychris @code-punkt
Actually, what i'd want is a multitude of things. One is a rehydration function. Another thing i'd like is a function that analyses the server-rendered HTML and only server renders the appropriate styles, instead of server-rendering all styles. Look at `glamor` for reference.

Speaking of rehydration, i'm not sure how i'd get the actual ids in SSR. This is what i've understood so far:
- numeric hashes are generated by hash method imported from glamor, based on css string
- alphanumeric names are created based on the numeric hashes
- alphanumeric names for ComponentStyle are stored in `inserted` object

Are those all we're interested in? What about the numeric hashes for InlineStyles stored in `generated` object there? What about keyframe numeric hashes?

What i also don't understand is which client-side check would be expensive exactly and why. Once `styled-components` is initialized on the client, it could read a global variable which has an array of the server-side rendered hashes and store this information. Afterwards, whenever a style is going to be added/injected on the client-side, that is only done when the hash of the style doesn't exist in that array.

mxstbr:
> Another thing i'd like is a function that analyses the server-rendered HTML and only server renders the appropriate styles, instead of server-rendering all styles. Look at glamor for reference.

We already do this. Since we inject at render time, only CSS you actually use on the server-side rendered page is injected. 😉

All the hashing things are actually about to change as of #227 and/or `babel-plugin-styled-components`, so many of those facts will be wrong very soon.

code-punkt:
That's a pity. Sounds like too many open construction works on the inner parts of `styled-components` for me to be of any help atm.

thisguychris:
@code-punkt I just used the word `expensive` since we'll add these checks on each component render.  It is certainly doable of course!  We might just need to revisit this after #227 lands.

terribleplan:
The only thing stopping me from using styled components is the lack of isomorphic/universal/SSR support, which this solves. I would love for this PR to land before worrying about the harder/finer parts of rehydration if possible.

diegohaz:
@terribleplan SSR already works. I'm using it here: https://github.com/diegohaz/arc/tree/universal-redux with the following code on [`src/server.js`](https://github.com/diegohaz/arc/blob/f556753f30ad40a38162703acef51021ffc607d5/src/server.js)
```js
import styleSheet from 'styled-components/lib/models/StyleSheet'
...
const styles = styleSheet.rules().map(rule => rule.cssText).join('\n')
...
const markup = <Html {...{ styles, assets, state, content }} />
```

I'm also using this on another bigger project. The only thing is the client is re-generating styles.

terribleplan:
@diegohaz thanks for the example, I'll probably start using that now.

Generally I prefer using public APIs, so would love to see anything not involving reaching inside the library like that. It always hurts a bit when I do something like that, then the implementation detail I am relying on changes.

mxstbr:
> Generally I prefer using public APIs, so would love to see anything not involving reaching inside the library like that. It always hurts a bit when I do something like that, then the implementation detail I am relying on changes.

You can rely on that API for now until we flesh out the public API here, I pinkie promise we won't break it ❤️

nicohartto:
@diegohaz I've implemented the styles as you described, but this does not prevent duplicate styles i.e. styles are injected correctly on server-side, but they are added once again when the client bundle is loaded. I can see that styles are injected twice.

Am I doing something the wrong way here (must be :-) )? If people do get this to work without getting duplicate styles.

codepunkt:
@nicohartto
The server rendering functionality here works, there is just no such thing as rehydration that stops the server rendered styles from being injected by the client again.

@geelen @mxstbr 
I think client-side rehydration of server-rendered styles should definitely be a (albeit optional) part of the API used to server-render `styled-components`. I'd love to help with this, but i'm not sure where to start:

- which branch do i base this on?
- which PR should i wait for (such as #227) before i start?
- is anyone else actively working on this?

Maybe we can and should talk about the API surface for server side rendering in the mean time.

### API
#### `renderStyles`
Most methods that do server side rendering in the react ecosystem by returning a string of whatever it is that needs to be rendered on the server start with `render` - so i'd propose a method called `renderStyles` that basically does what the `getCSS` method in this PR does - it returns a string of styles to be rendered. In contrast to aphrodite or glamor, styled-components only renders styles used in the current component tree - as @mxstbr pointed out earlier. Therefore, we don't need a second (optimized) version of this like glamor does.

#### `rehydrateSelectors`
As @diegohaz and @nicohartto pointed out amongst others, we should probably also have a method that rehydrates or injects the server-rendered styles to the client-side code so existing ones are not re-injected on the client. On the one hand, this could be done manually by informing the client-side `styled-components` with a method like `rehydrateSelectors`, which is given a list of the hashes/selector-identifiers of server-rendered styles. On the other hand, this could be done "automagically" when a special attribute is added to the `HTMLStyleElement` that hosts the server-rendered styles - which would need parsing of server-rendered css on the client and is most likely a maintenance as well as a performance nightmare, but would slightly reduce the amount of data sent to the client.

thisguychris:
@codepunkt just waiting on @mxstbr and @geelen if we can merge this.  I think the fleshed out the api for #227 .

My idea for rehydration part is just automagically checks if there's SSR.  (probably a global variable, window._glamorStyles --> contains an array of hashes)   If there is then it skips on the styles to be rendered if the id matches.  I do believe we can just save it to an array and just do a check there.

codepunkt:
@thisguychris Some people might have problems with that name (e.g. eslint no-underscore-dangle). Also, where would you get the hashes from?

thisguychris:
@codepunkt oh that's not final ( I'll leave bike shedding to @mxstbr and @geelen ), also that's something we won't be exposing.  I'm just describing how would SSR communicate with the client side.  Which is thru that global variable.  

>  where would you get the hashes from? 

from `getCSS()`

mxstbr:
I'm really unsure what to do here right now to be honest 😁 need to let this simmer in my head, meanwhile @geelen you got any input?

Lordnibbler:
any updates on this?

codepunkt:
The server-side rendering itself is done. Maintainers and collaborators just need to agree on an API and then publish that.

What is missing is the rehydration of server-rendered styles on the client. I sketched possible APIs for both above.

satya164:
I'd really love to have a nice API like other libraries instead of the current awkward API in the PR.

teonik:
Is anyone else having issues calling `styleSheet.flush()` at the moment? I'm getting some strange behaviour,everything works ok on the first render request but after the first `flush()` call i get an empty array from `styleSheet.rules()`.

wmertens:
For rehydrating, perhaps the SSR CSS can be stored as a JS global in head and then immediately injected, so that it only needs to be transferred once?

OTOH, that breaks CSS for non-JS clients. Also, why bother with rehydrating, really? Won't generating everything from scratch just give the same results? Then it's just a matter of replacing the injected style element?

codepunkt:
@wmertens Performance.

To replace the server side injected style element you would need a way to identify it amongst possible others. When simply re-injecting on client, the time this takes is wasted because the styles are already applied by the server. In addition, when not replacing the server styles all styles that have been rendered on the server before are now duplicate, which is not great for performance either.

geelen:
Performance isn't actually going to be that bad having two copies of everything, the CSS starts to really chug once you have _a lot_ of rules (i.e. #134) and a server is only going to be rendering the CSS for a single state of the DOM. It _is_ unnecessary though, so we should get rid of it when we can.

I say we should ship this as-is and treat rehydration as a performance optimisation that can come in a point release.

codepunkt:
I'm not too keen on seeing performance as an afterthought. "Not going to be that bad" is not what i like to see from libraries i was hoping to be able to use one day. Nevertheless, i'd rather release this as is before keeping this open any longer.

@geelen @mxstbr You commented on Dec, 11th that a lot of the hashing things is subject to change as of #227 and/or `babel-plugin-styled-components`. What's the current state of that? I'd like to have a starting point to add rehydration features!

geelen:
> I'm not too keen on seeing performance as an afterthought. "Not going to be that bad" is not what i like to see from libraries i was hoping to be able to use one day.

@codepunkt this is me politely asking you to take your entitled tone and swallow it. Everyone else on this thread (and on the issue tracker generally) has given their feedback without needing to mischaracterise our position and our priorities for the project. You can too.

I understand rehydration is a significant issue for you, and appears to be a blocking you from using Styled Components. That's fine, that's up to you. But as I said in the previous comment:

> the CSS starts to really chug once you have a lot of rules and a server is only going to be rendering the CSS for a single state of the DOM. It is unnecessary though, so we should get rid of it when we can.

I wonder, have you actually _measured_ any of the performance you consider us to be so ignorant of? How many CSS rules are you shipping on each server-rendered page? How many DOM elements are rendered by the server? What is the total weight of your pre-rendered HTML? How does the time taken to regenerate the CSS on the client compare with the time taken for React to generate a DOM of that size? How do they grow in relation to one another?

"Performance" is not a mantra. It is not some box to be ticked. It is data. If you have some, I'm more than willing to engage with it. But if all you have are opinions, [well...](http://bryce.vc/post/26886353878/if-we-have-data-lets-look-at-data-if-all-we)

thisguychris:
@mxstbr would you mind reviewing the flow annotation fix (last 2 commits).  I added those since CI was failing.

thisguychris:
To everyone.  We'll be merging this PR very soon, just need @mxstbr to review the flow annotation part since I don't use flow.  I'll be doing a new PR for the rehydration feature (with tests).  Don't worry this is under our radar.

src/models/StyleSheet.js
```diff
@@ -3,6 +3,28 @@
...
-import { StyleSheet } from '../vendor/glamor/sheet'
+import { StyleSheet as GlamorStyle } from '../vendor/glamor/sheet'
 
-export default new StyleSheet({ speedy: false, maxLength: 40 })
+class StyleSheet {
+  styleSheet: any
+  constructor() {
+    this.styleSheet = new GlamorStyle({ speedy: false, maxLength: 40 })
+  }
+  get injected(): any {
```
relekang: Please change the return type here to boolean.

src/models/StyleSheet.js
```diff
@@ -3,6 +3,28 @@
 /* Wraps glamor's stylesheet and exports a singleton for the rest
 *  of the app to use. */
 
-import { StyleSheet } from '../vendor/glamor/sheet'
+import { StyleSheet as GlamorStyle } from '../vendor/glamor/sheet'
 
-export default new StyleSheet({ speedy: false, maxLength: 40 })
+class StyleSheet {
+  styleSheet: any
```
relekang: I think this can be `typeof GlamorStyle` or maybe just `GlamorStyle`?

thisguychris: @relekang what should be the whole line be?  Not really familiar with flow.  The `as` in this line is importing it as a different variable, aka `var b = a`.

thisguychris: @relekang also reading the flow documentation it seems that typeof statement in `import` is actually short ciruciting `as` which is a valid es2015 statement.  It's inferring it as type instead of a variable.

satya164: `StyleSheet: GlamorStyle` if `StyleSheet` is an instance of `GlamorStyle`.

thisguychris: @satya164 the line that @relekang commented is this, 
`import { StyleSheet as GlamorStyle } from '../vendor/glamor/sheet'`

from your suggestion it should be like this?
`import { StyleSheet as GlamorStyle: StyleSheet } from '../vendor/glamor/sheet'`  I'm getting token errors 😱

mxstbr: @thisguychris I think he means on line 9 change

```JS
class StyleSheet {
  styleSheet: any;
}
```

to 


```JS
class StyleSheet {
  styleSheet: GlamorStyle;
}
```

thisguychris: thanks @mxstbr ! changed.

satya164:
@mxstbr what about a nicer API?

mxstbr:
Ah sorry, what did I miss here?

satya164:
@mxstbr the current API feels awkward. I love the API other libs like glamor and aphrodite have - https://github.com/styled-components/styled-components/pull/214#issuecomment-263479828

wmertens:
@satya164 the example you propose actually needs to do a render run to know which CSS to build, and then another render run to generate the string. This is expensive.

S-C generates out-of-band data, so it is reasonable to receive it out-of-band too… Redux store collection works in the same way.

I'm happy with the current API.

wmertens:
@mxstbr is there a reason this can't go in v1? It only extends the API, right?

How can I try v2, is there breaking API?

mxstbr:
This won't work without the [babel plugin](https://github.com/styled-components/babel-plugin-styled-components), which won't work without #227, which is why this is v2.

> the example you propose actually needs to do a render run to know which CSS to build, and then another render run to generate the string. This is expensive.

Does it? I wasn't aware of that!

threepointone:
Glamor /aphrodite don't do two passes. Glamor works on raw HTML as well, so you don't even have to run the renderToString inside the callback.

satya164:
> needs to do a render run to know which CSS to build, and then another render run to generate the string.

Can't it use the same render to get the CSS that it uses to get the HTML string? Why would it need two passes? How is it diffrent from calling these 3 functions manually?

> Glamor /aphrodite don't do two passes

I'll trust @threepointone on this one

wmertens:
Very strange, I thought I left a comment yesterday but now I don't see it.

I was wrong in my comment about the API, there are no two passes, the callback is for proper lifecycling of the style generation.

With the API @satya164 proposes, you don't need to call `reset()` on the stylesheet, and it is more obvious that one depends on the other. So I agree that

```js
const { css, html } = StyleSheet.collect(() =>
  ReactDOMServer.renderToString(<App />)
);
```

is a nicer API.

wmertens:
@satya164 @mxstbr I found a problem with that nicer API: it is not composable. If you have other out-of-band data that is generated (like HEAD metadata or SVG symbols), you can't nest this.

Unless… if `collect(cb)` accepts either a string or an object from the cb, it would work. If it's a string, it presumes that to be the `html`, and if it's an object, it just adds `css` to the object and returns it. Would that work? Then you could have something like

```js
const {css, html, svg, headers} = StyleSheet.collect(SvgSprites.collect(Helmet.collect(
  () => ReactDOMServer.renderToString(<App />)
)))
```

all those other projects have to do the same, of course.

satya164:
It will work just fine.

```js
let { css, html, svg };

{ svg } => Svg.collect(() => {
  { css, html } = StyleSheet.collect(() => ReactDOMServer.renderToString(<App />);
  return html; // don't need this return if `Svg.collect` doesn't need the html string
});
```

Even though it completely depends on what kind of API other libs have, it will work fine in any kind of API.

wmertens:
Hmmm yeah I suppose. I dislike that I have to use `let` though :P

wmertens:
@thisguychris I am using the v2 branch and I can't get SSR to work - I just get `''` from `getCSS()`. Thoughts?

wmertens:
oh, scratch that, it works, but only once. That's due to #378

PR #215, opened 2016-11-18
Automatically scope keyframes when defined within a component

jacobp100:
It’s a really ugly way of doing this! Potentially something like [this](https://github.com/jacobp100/cssta/blob/a74d40309a7817f4c25367fe612466ec5a2317a4/cssta/src/util/getRoot.js#L7-L46) might be cleaner.


mxstbr:
We actually already had it working like that, but decided against doing it this way.

Animations are almost never specific to a single component, and while it is possible to have "animation-components" (e.g. `<FadeIn>`) it's not a common thing to do and restricts the possibilities of the animations.

That being said, I'd love to hear your use case for this! Why do you think this is a good idea?

jacobp100:
There isn't too much of a use case past moving style-related code from JS to CSS.

You could in theory remove the keyframe constructor by defining all shared animations in `injectGlobal`, and one-off animations in the individual components.

I'm not sure about the higher order components—but I do know it isn't uncommon to have specific animations for specific components.

Feel free to close if you don't think this is the right direction for the project though!

mxstbr:
Do you know we have the `keyframes` helper? I think you might've missed that in the docs! https://github.com/styled-components/styled-components/blob/c68c3355fa7b83c19e719326fa195377d334acbc/docs/api.md#keyframes

```JS
const fadeIn = keyframes`
  from {
    opacity: 0;
  }

  to {
    opacity: 1;
  }
`
```

jacobp100:
No I saw that! I was talking about moving JS styling code to CSS.

> You could in theory remove the keyframe constructor

Either way, feel free to close.

mxstbr:
Ah sorry, I missed that. My bad for not reading properly!

> Either way, feel free to close.

I don't want to close this until I understand why you think this is a good idea to do! What do you mean with "moving JS styling code to CSS"?

jacobp100:
The keyframe constructor is defining keyframes using JS rather than CSS—that's all I mean. The benefits mostly come down to consistency. It might also make static analysis easier in some cases.

mxstbr:
It doesn't seem like there's a lot of interest in shipping this, but I appreciate you kicking off this discussion! Thanks for taking the time to submit a PR!

Issue #177, opened 2016-11-08
Flow errors when using react-create-app


msmfsd:
## Version
1.0.10

Guys love styled-components so thanks!
Am on v1.0.10 and still getting flow-type errors. Note I am using create-react-app.

My flow config is as per create-react-app guidelines:
```
[ignore]
<PROJECT_ROOT>/node_modules/fbjs/.*
```

Mu deps:
```
"devDependencies": {
    "babel-eslint": "6.1.2",
    "eslint": "3.5.0",
    "eslint-plugin-flowtype": "2.18.1",
    "eslint-plugin-import": "1.12.0",
    "eslint-plugin-jsx-a11y": "2.2.2",
    "eslint-plugin-react": "5.2.2",
    "flow-bin": "^0.34.0",
    "react-scripts": "0.7.0"
  },
  "dependencies": {
    "react": "^15.3.2",
    "react-dom": "^15.3.2",
    "recompose": "^0.20.2",
    "styled-components": "^1.0.10"
  }
```

Here are the errors:
```
node_modules/styled-components/src/models/InlineStyle.js:3
  3: import camelizeStyleName from 'fbjs/lib/camelizeStyleName'
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ fbjs/lib/camelizeStyleName. Required module not found

node_modules/styled-components/src/models/InlineStyle.js:5
  5: import { StyleSheet } from 'react-native'
                                ^^^^^^^^^^^^^^ react-native. Required module not found

node_modules/styled-components/src/native/index.js:4
  4: import reactNative from 'react-native'
                             ^^^^^^^^^^^^^^ react-native. Required module not found

node_modules/styled-components/src/utils/flatten.js:2
  2: import hyphenate from 'fbjs/lib/hyphenateStyleName'
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ fbjs/lib/hyphenateStyleName. Required module not found
```

mxstbr:
Handing over to @relekang @ryyppy, they should know what's going on here! The thing about `react-native` is that we require it in the code, but it's not in our dependencies to avoid bundling it in. (you'll have it in your dependencies anyway when using `styled-components` with it)

relekang:
I think this will be resolved by #151. The problem is that in the flowconfig fbjs is ignored, but fbjs is imported in styled-components. 

When we are able to generate the flow imports will not be read by flow since they will not be a part of the flow files.

[...]

Issue #151, opened 2016-10-31
Expose flowtype interface


ryyppy:
We need to add flow definitions to the dist directory on release... need to check if `flow gen-flow-files` will be enough for this regard :-) ... will wait for #148 to merge before doing that

mxstbr:
Done in #219, thanks to the amazing @ryyppy!

PR #219, opened 2016-11-19
Expose flow types & add detailed usage docs

ryyppy:
This commit introduces a few flow-related things (and also some misc
stuff):

* Fix some typing errors with glamor / nameGenerator
* Move adapted `flow-typed/npm` libdefs to `flow-typed`
* Add some flow related ignore rules for eslintrc / flowconfig etc.
* Add a `build:flow` script and add it to the build step
* Add docs for making `styled-components` and `flow` work for the consumer

MISC:
* Add test for generateAlphabeticName

As soon as this feature is merged, there will be `flow.js` files
vendored along with transpiled `lib` files on every build.

Related issues: #177 and #151


relekang:
The flow files generated by build:flow contains regular javascript, but as far as I understand from [the flowtype docs](https://flowtype.org/docs/declarations.html#declaration-files) it should contain declarations similar to the libdefs in the flow-typed folder?

ryyppy:
@relekang Shadow files can also contain implementing code... that's the whole point why they wanted to introduce the `flow gen-flow-files` command... to strip out the implementation code and make the delivery size much more efficient... you can imagine it like this... code located in `lib` is transpiled source without any annotations... if flow sees a `[file].js.flow`, it will parse this code instead... this has nothing to do with libdef syntax

ryyppy:
@mxstbr @geelen @relekang ... Everything done on my side... I even made it Windows compatible ... time to >ship it< 🚀

relekang:
> Shadow files can also contain implementing code... that's the whole point why they wanted to introduce the flow gen-flow-files command... to strip out the implementation code and make the delivery size much more efficient... you can imagine it like this... code located in lib is transpiled source without any annotations... if flow sees a [file].js.flow, it will parse this code instead... this has nothing to do with libdef syntax

Ah, okay 😊

Great job :v:

relekang:
This looks good to me. Tested it on a few projects of mine it works great except there seems to be some problems with the ThemeProvider typings, but I guess it would be better to fix that outside this PR as it is not related to the bundling of flow types. 🙂

ryyppy:
@relekang That's most likely because you use a different flow version... I am 80% sure :-) ... yep we can do that

relekang:
> That's most likely because you use a different flow version

Yeah, it was 0.33.0. Sorry about that.

PR #222, opened 2016-11-20
Travis: Add node 7 and cache

siddharthkp:
- [x] Add node 7
- [x] Use latest version of `4.x` and `6.x` instead of `4.0` and `6.0`
- [x] Cache bundler and `node_modules`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,3 +21,7 @@ script:
 notifications:
   email:
     on_failure: change
+cache:
+  bundler: true
+  directories:
+    - node_modules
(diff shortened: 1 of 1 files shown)


siddharthkp:
@mxstbr Appveyor installs node v5. That's not intentional, right? :thinking:

siddharthkp:
@mxstbr What's the additional benefit of using appveyor along with travis?

mxstbr:
Appveyor tests on Windows, Travis on Linux. That way we avoid any annoying build step issues.

PR #223, opened 2016-11-20
Appveyor: Changing node versions to 6

siddharthkp:
Was using node 5 earlier.


siddharthkp:
@mxstbr The build times with appveyor have shot up to `5mins 41sec`!! 

Should probably test LTS here.

mxstbr:
Let's only test LTS then, sounds good to me!

mxstbr:
Thanks!

PR #224, opened 2016-11-20
Travis: Removed sudo for faster builds

siddharthkp:
1. You don't need sudo
2. Builds will start faster with [travis's container based infra](https://docs.travis-ci.com/user/trusty-ci-environment/#Container-based-with-sudo%3A-false)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,3 +21,7 @@ script:
 notifications:
   email:
     on_failure: change
+cache:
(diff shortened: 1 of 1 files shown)


mxstbr:
Should we only test LTS on Travis too?

siddharthkp:
Nope, builds are parallel and pretty fast. Can test 4 and latest too.

PR #225, opened 2016-11-20
Add flow comment to files missing it

relekang:
* Add `// @flow` to all files missing it
* Fix typing of ThemeProvider
* Add babel plugin to generate proptypes from flowtypes
* Add a flow suppresser `// $FlowInvalidInputTest`
* Upgrade babel to 0.35.0


relekang:
@ryyppy The tests of ThemeProvider gets a bunch of these errors:
```
src/test/theme.test.js:34
 34:       <ThemeProvider theme={theme}>
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ React element `ThemeProvider`
 12: type ThemeProviderProps = {|
                               ^ property `children`. Property not found in. See: src/models/ThemeProvider.js:12
 34:       <ThemeProvider theme={theme}>
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ props of React element `ThemeProvider`
```

It seems to be a bug with flow or should children be typed differently? 

[A smaller example that reproduces the issue](https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoAlgWwA5wE4AuYASgKYCGAxkVHnJmAOR4XWMY75EDeLAdgBNSeAL5ha9JiyoEAtAPrtUBAJ7ZSYAAp1sAZzABeMN1RgwlABboYA-gC4w5PitQjUqSjHK79AMThwYKQAHgSkgvpkMgB0AML0uHzhPKZg2Dq6Dtpweqmp-EJ4ABQAlMapYACQLAQArnh8YAA8AugAbgB83ARWutHpOX2W1rbhIk3ArZ1gqW5uqAXCRakt7R0Vzf5wHQAS6BNb62ZmTVvmVjb8BgBEe9dgwEfNk2uoJUA)

relekang:
What is allowed in the first argument of `flatten`? Currently it is typed as `Array<Interpolation>` which is `Array<((executionContext: Object) => string) | string | number>`, but the tests sends in more different types. Are all the values in the tests allowed or is that checks to prevent runtime errors of bad input data?

mxstbr:
> should children be typed differently

We totally forgot to add `propTypes` to that! Would you mind adding `children` and `theme` as prop types? I reckon that fixes the error?

I think most of the things in the `flatten` test suite are just for runtime, I think with `Function | string | number` we've covered everything that doesn't get dropped?

relekang:
>We totally forgot to add propTypes to that! Would you mind adding children and theme as prop types? I reckon that fixes the error?

That should not be necessary, typing `prop` should be enough. However, we should probably add a babel transform that adds proptypes at compile time so that it is not necessary to maintain both the flowtype and proptypes.

relekang:
The types of ThemeProvider is fixed now.

mxstbr:
Is this ready to be reviewed?

relekang:
@mxstbr I think so :)

relekang:
Thanks, fixed the changelog. Also rebased it on master.

mxstbr:
Awesome, thanks so much!

Issue #142, opened 2016-10-27
RFC: Component Selector


mxstbr:
I was chatting to @markdalgleish at Reactive conf and he had this great idea to implement a other-styled-component-selector.

The API would look like this:

``` JS
const Button = styled.button`some: styles;`

const Box = styled.box`
  > ${Button} {
    // A button component nested within a box will have these styles
    other: styles;
  }
`
```

This would allow you to override styles of children, and from children react to parents, too! What do you think?

geelen:
I like it.

stephenkingsley:
just override the `Button` within a box, not change the `Button`?

I like it

jonaskello:
I have some "complex" reusable components in a library. With "complex" I mean they contain multiple `<select>`, `<input>` etc. I would like importers of these component to be able to override each part of the complex component (eg. override the styles of one of the `<select>` inside the component). This suggestion seems like it would fit my needs. Just wanted to check if there is some way to handle this type of "complex" components already or is styled-component only usable for primitives like `<Button>`, `<Input>` etc at the moment?

geelen:
After a bit more thought I... don't think this is really going to be possible. Or at least, easy enough to justify its addition.

Consider this:

```js
const X = styled.div`
  one: rule;
  ${ props => props.alt && 'then: another;' }
`

const Y = styled.div`
  > ${ X } {
    over: ride;
  }
`
```

If you construct a `<X/>` you get `<div className='random-hash-123'/>`, whereas if you construct a `<X alt/>` you get `<div className='random-hash=456'/>`. So you can't use `X`'s generated classnames inside `Y`'s rules.

You'd have to generate a globally unique classname that was attached to _all instances of X_ that applied no rules by itself. But then, how do you generate that classname? Elsewhere we hash the contents of the CSS to produce a deterministic, order-independent hashed classname. Order-independence is important because we don't want the order of `import` statements in a project to affect the generated classnames. That's particularly important in a server-side rendered context where the generated classnames need to match on the client.

So, if someone can come up with a deterministic, order-independent random classname generator that can distinguish between `const X = styled.div` and `const Y = styled.div` we can do this. Otherwise, we'll have to drop the idea.

Also, any solution can't depend on running a babel transform or other build step. You can do [cool things with babel transforms](https://github.com/vdanchenkov/babel-plugin-styled-components-named) I just don't want to _need_ them for anything in this project. Basically, my threshold is that styled-components needs to be able to work with Create React App without ejecting.

[...]

PR #227, opened 2016-11-21
[WIP] Per-component classnames

geelen:
Super hacky proof of concept, but something's been bugging me for a while about Styled Components — we don't have any classnames that are _always_ going to be attached on a component. This is the key problem with supporting #142 but it also causes a few DX annoyances, prevents an optimisation, and presents one nasty af potential bug that nobody has hit yet...

One annoyance is that, in dev tools, you don't have a single class hook to tweak settings if you have several instances of a component with different props. `<Button primary>` and `<Button>` will generate totally different classnames, so you can't jump in to dev tools and _tweak all the buttons_. If they share a classname, you can.

It also means that #160 is harder than it should be. If there's a unique class for every Styled Component then in dev you can make that whatever you like. 5 internet high fives for the person who implements BEM-style `.app_components_Logo_Inner` classnames using the directory structure 👍 , ping @markdalgleish.

The performance optimisation has to do with #134. If we had a single shared classname for all instances of a SC then we could potentially lift all the static styles into them. Then, when props change, we're injecting just the styles that depend on props. I'm not convinced this will be a _massive_ gain but it certainly would map better to how people think about their styles.

And finally the nasty af bug. Consider this setup:

```js
const A = styled.div`
  color: blue;
`

injectGlobal`
  /* contrived example, but all you need is a rule with equal specificity
     so source order becomes important */
  .override {
    color: red;
  }
`

const B = styled.div`
  color: ${props => props.selected ? 'blue' : 'green'};
`

<A className="override"/> // red, since .override appears later
<B className="override"/> // green, since it is rendered afterwards
<B selected className="override"/> // red, not blue !!
```

The problem arises from the internal cache about rendered styles. If two components end up producing the exact same CSS, we'll only inject a single class. Which means in cases where source-order is important (two rules applying with equal specificity) the result can be unpredictable. But, without a legitimate per-component identifier, the alternative (render the same selector twice) is actually kinda worse, since it would break components you're not looking at. In the above example, rendering a single  `<B selected>` would change all the  `<A className="override">` on the page to blue!! 😱

So, it's something I've wanted for a while, but the problem is there's no deterministic information from which to derive a per-component identifier. This PR uses a simple counter, so the 5th time you call `styled.div` you'll get a classname based on a hash of `div` and `5` (`.sc-ioLzpW` at the time of writing). If you have two components with `displayName = 'Inner'`, the first will get a classname of `.Inner-jaWsLN` and the second `.Inner-dBMIyz`). Which works really _really_ well, replaces #160 and made #142 a one-line change 😎

image

Note the empty `.Inner-jaWsLN` tag in dev tools. Makes it really to tweak all instances at once 👍 

The only issue is SSR, which we're about to land support for in #214. On the server, components for different pages might all be loaded, and so relying on order-of-instantiation will cause a checksum mismatch on the client. Not good.

So I've built this with a potential babel transform in mind, which would go through all StyledComponents instances and generates a deterministic `identifier`, much the way [babel-plugin-styled-components-named](https://github.com/vdanchenkov/babel-plugin-styled-components-named) generates `displayName`s. If that bundle is sent down to the client the counter won't be used and the checksum will match.

I don't really have any time before CSS Conf AU next week to go any further with this, but as a proof of concept it really does prove the concept. We can bikeshed the generated classnames but I'm pretty convinced per-component classnames are the way to go. Anyone keen to take this and run with it?

### DIFF
--- a/src/models/ComponentStyle.js
+++ b/src/models/ComponentStyle.js
@@ -25,6 +25,10 @@ export default (nameGenerator: NameGenerator) => {
       this.insertedRule = styleSheet.insert('')
     }
 
+    generateName(str: string) {
+      return nameGenerator(hashStr(str))
+    }
+
     /*
      * Flattens a rule set into valid CSS
      * Hashes it, wraps the whole chunk in a ._hashName {}
--- a/src/models/ComponentStyle.js
+++ b/src/models/ComponentStyle.js
@@ -31,10 +31,10 @@ export default (nameGenerator: NameGenerator) => {
      * Parses that with PostCSS then runs PostCSS-Nested on it
      * Returns the hash to be injected on render()
      * */
-    generateAndInjectStyles(executionContext: Object) {
+    generateAndInjectStyles(executionContext: Object, identifier: string) {
       const flatCSS = flatten(this.rules, executionContext).join('')
         .replace(/^\s*\/\/.*$/gm, '') // replace JS comments
-      const hash = hashStr(flatCSS)
+      const hash = hashStr(identifier + flatCSS)
       if (!inserted[hash]) {
         const selector = nameGenerator(hash)
         inserted[hash] = selector
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -9,6 +9,14 @@ import type { RuleSet, Target } from '../types'
 import AbstractStyledComponent from './AbstractStyledComponent'
 import { CHANNEL } from './ThemeProvider'
 
+/* We depend on components having unique IDs */
+const identifiers = {}
+const safeDisplayName = (displayName: string) => (
+  displayName
+    .replace(/[[\].#*$><+~=|^:(),"'`]/g, '-') // Replace all possible CSS selectors
+    .replace(/--+/g, '-') // Replace multiple -- with single -
+)
+
 export default (ComponentStyle: Function) => {
   // eslint-disable-next-line no-undef
   const createStyledComponent = (target: Target, rules: RuleSet, parent?: ReactClass<*>) => {
@@ -24,6 +32,7 @@ export default (ComponentStyle: Function) => {
     class StyledComponent extends ParentComponent {
       static rules: RuleSet
       static target: Target
+      static identifier: string
 
       constructor() {
         super()
@@ -35,7 +44,7 @@ export default (ComponentStyle: Function) => {
 
       generateAndInjectStyles(theme: any, props: any) {
         const executionContext = { ...props, theme }
-        return componentStyle.generateAndInjectStyles(executionContext)
+        return componentStyle.generateAndInjectStyles(executionContext, StyledComponent.identifier)
       }
 
       componentWillMount() {
@@ -83,7 +92,7 @@ export default (ComponentStyle: Function) => {
           .forEach(propName => {
             propsForElement[propName] = this.props[propName]
           })
-        propsForElement.className = [className, generatedClassName].filter(x => x).join(' ')
+        propsForElement.className = [className, StyledComponent.identifier, generatedClassName].filter(x => x).join(' ')
         if (innerRef) {
           propsForElement.ref = innerRef
         }
@@ -92,10 +101,27 @@ export default (ComponentStyle: Function) => {
       }
     }
 
+    let displayName = isTag ? `styled.${target}` : `Styled(${target.displayName})`
+    const generateIdentifierFromDisplayName = () => {
+      const nr = (identifiers[displayName] || 0) + 1
+      identifiers[displayName] = nr
+      StyledComponent.identifier = safeDisplayName(`${displayName}-${nr}`)
+    }
+    generateIdentifierFromDisplayName()
+
+    Object.defineProperty(StyledComponent, 'displayName', {
+      get() {
+        return displayName
+      },
+      set(newName) {
+        console.log('SETTING DISPLAY NAME')
+        displayName = newName
+        generateIdentifierFromDisplayName()
+      },
+      enumerable: true,
+    })
     StyledComponent.target = target
     StyledComponent.rules = rules
-
-    StyledComponent.displayName = isTag(target) ? `styled.${target}` : `Styled(${target.displayName})`
 
     return StyledComponent
   }
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -102,12 +102,13 @@ export default (ComponentStyle: Function) => {
     }
 
     let displayName = isTag ? `styled.${target}` : `Styled(${target.displayName})`
-    const generateIdentifierFromDisplayName = () => {
+    const generateIdentifierFromDisplayName = (hashOnly: boolean) => {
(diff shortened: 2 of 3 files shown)


thisguychris:
@geelen couldn't we potentially create a unique identifier since we're creating a React Component anyways? (at StyledComponent.js)  I'm thinking of adding uuid in the component creation instead of adding babel-transform after everything is rendered.

Re: performance.  A fast fix for the performance is the ephermal helper for now.  But I do agree this is the better solution.

geelen:
Using something like a UUID would be like `Math.random` though, right? Which would definitely not match between client and server, breaking the checksum stuff...

mxstbr:
I've written a first babel plugin for this! https://github.com/styled-components/babel-plugin-styled-components-ssr

Go to `test/fixtures/example` to see what it does – essentially, it goes through and adds an incremental numeric `.identifier` to each styled component. This isn't perfect yet, but I think it's a good first start?

mxstbr:
For some reason this is throwing lots of these errors: `Error: Expected '.b { color: black; }' to equal '.a { color: black; }'` where the class is just off-by-one. Is that a problem in the code or is that a sideeffect of the changes and the class names need to be updated?

geelen:
Nah I didn't even run the tests while i was sketching this out, not surprised I broke everything :)

mxstbr:
So I think the last thing for this to be ready is that extending/inheritance is completely broken for some reason unknown to me. The tests fail in very weird ways, not applying the classes how you'd expect them to. Any ideas what's up with that?

mxstbr:
In more happy news, this is working perfectly with the babel plugin! 🎉 

Code:

screen shot 2016-11-28 at 09 31 02

DevTools:

screen shot 2016-11-28 at 09 30 44

😍

thisguychris:
@mxstbr babel-plugin = works with SSR?

mxstbr:
Yes, and better debugging (because it adds the `displayNames`), but it's not mandatory at all

clempat:
Do you mean by "not applying the classes how you'd expect them to" that the class name is always one extra time incremented.
I guess it is because the nameGenerator (`const classNames = () => String.fromCodePoint(97 + index++)`) passed to ComponentStyle is run twice. 

First in [StyledComponent.js line 115](https://github.com/styled-components/styled-components/blob/per-component-classname-generation/src/models/StyledComponent.js#L115)

Then in [componentWillMount line 65](https://github.com/styled-components/styled-components/blob/per-component-classname-generation/src/models/StyledComponent.js#L65)

Or I misunderstand something.

mxstbr:
No, I'm aware of that issue and fixed all tests that broke because of that. The issue is with the `extending.test.js` file – if you run the tests on this branch, you'll see that they fail because the generated CSS is completely wrong! (not just the class names incremented by one)

geelen:
So, I've come at this again with a different syntax that I think works a lot better, and deleted most of my original code.

Per-component classnames are great. But the existing syntax didn't work:

```js
const MyComponent = styled.div`
  ...
`

MyComponent.displayName = 'MyComponent' // for React dev tools
MyComponent.identifier = 'MyComponent-a1c2d3e4' // A unique component classname.
                                                                                          // Will be generated from displayName
                                                                                          // or a hashed counter if there is none.
```

This is clunky, and hard for a babel transform to target. Instead, I've gone with this:

```js
const MyComponent = styled.div.displayName('MyComponent')
                                                       .componentId('MyComponent-a1b2c3d4')`
  ...
`
```

This has a couple of advantages. The implementation is much simpler, since you never _create_ a component before it has a `displayName` and/or `componentId` set. It also means that the babel transform can blindly inject its suggested values, at the beginning of the chain but they'll get overridden:

```js
const X = styled.div.displayName('MyX')
                                 .componentId('💖'')`
  ...
`

// babel transform converts to:

const X = styled.div.displayName('X')                      // injected by
                                 .componentId('X-e9f8a7c6')    // babel transform
                                 .displayName('MyX')
                                 .componentId('💖'')`
  ...
`
```

The last values passed simply take precedence. I've already got a test for this too.

The good thing about this is that it gives us a nice way of _targeting the component itself_ at the point of creation. #291 was quite an influential suggestion here, because I can definitely see it enabling a really nice hybrid of atomic, immutable CSS and CSS-in-JS:

```js
const Title = styled.h1.classes('left p2 mr1 border')`
  font-size: 1.5em;
  text-align: center;
  color: palevioletred;
`
```

At the moment I've only defined `classes`, `displayName` and `componentId` but we could add more. In particular, we have special handling for the following case:

```js
const X = styled.div` ... `
const Y = styled(X)` ... `
```

Where because `X` is a StyledComponent, `Y` steals all its definitions and tag name and then adds customisation. If `X` was a normal React component, `Y` would wrap `X` and pass it a `className`. We could replace this to be more explicit:

```js
const X = styled.div` ... `
const Y = styled.div.extends(X)` ... `
```

That way you could change the tag underneath but apply all the same styles, which is something folks have asked for.

Anyway, let me know what you think!

src/constructors/styled.js
```diff
@@ -2,10 +2,21 @@
...
-    (strings: Array<string>, ...interpolations: Array<Interpolation>) =>
-      styledComponent(tag, css(strings, ...interpolations))
+  const expandedApi = (tag, options: Object = {}) => {
+    const templateFunction =
+      (strings: Array<string>, ...interpolations: Array<Interpolation>) =>
+        styledComponent(tag, options, css(strings, ...interpolations))
+    templateFunction.classes = classes => expandedApi(tag, { ...options, classes })
+    templateFunction.componentId = componentId => expandedApi(tag, { ...options, componentId })
+    templateFunction.displayName = displayName => expandedApi(tag, { ...options, displayName })
+    templateFunction.css = templateFunction // pure sugar
```
ticky: Tasty, tasty sugar! 😃

ticky:
This is really, really cool! I think it’d definitely fill the gap I’ve found which led to #291! 👍

geelen:
Ok, in the interests of getting this shipped, I've pulled this back to a single extra method `withConfig`. No `classes` support for the moment (reverted in [5fd2de5](https://github.com/styled-components/styled-components/pull/227/commits/5fd2de50ca09d665e08d48bcc2c6e604f422f9b3)) but `displayName` and `componentId` are in there. This isn't designed to be user-facing, but to be hooked in to by the babel transform:

```js
const X = styled.div`
  ...
`

/* transformed to */
const X = styled.div.withConfig({ displayName: 'X', componentId: 'X_a1b2c3d4' })`
  ...
`
```

This PR can now be merged and will close #142 and #160 and enable deterministic SSR, so it'll be good to get this shipped. I had hoped to solve `classes` and `extends` with the same API but this takes the babel-targeted properties and removes them from the equation. Let's move discussion on the others to #291.

thisguychris:
@geelen / @mxstbr do we need to change anything on this #214 in respect to this PR?  If not, can we merged it?   So that someone can add the rehydration part of SSR.

geelen:
Ok, kicking of the `v2` branch with this.

PR #169, opened 2016-11-07
Convert to Danger JS

orta:
Converts your Ruby Dangerfile into JavaScript, effectively removing ruby from the toolchain and keeping it inside Node.

The new Dangerfile does not use regexes, instead uses a bunch of smaller filter functions, as that was just easier for me to write.

```js
// @flow

import { danger, warn, fail, message } from 'danger'
import fs from 'fs'

const jsModifiedFiles = danger.git.modified_files.filter(path => path.startsWith('src') && path.endsWith('js'))
const vendorModifiedFiles = danger.git.modified_files.filter(path => path.startsWith('src/vendor') && path.startsWith('js'))

const hasAppChanges = jsModifiedFiles.length > 0
const jsTestChanges = jsModifiedFiles.filter(filepath => filepath.endsWith('test.js'))
const hasTestChanges = jsTestChanges.length > 0

// Congrats, version bump up!
const isVersionBump = danger.git.modified_files.includes('package.json')
const packageDiff = danger.git.diffForFile('package.json')

if (isVersionBump && packageDiff && packageDiff.includes('version')) { message(':tada: Version BUMP UP!') }

// Warn when there is a big PR
const bigPRThreshold = 500
if (danger.github.pr.additions + danger.github.pr.deletions > bigPRThreshold) {
  warn(':exclamation: Big PR')
}

// Fail if there are app changes without a CHANGELOG
if (!danger.git.modified_files.includes('CHANGELOG.md') && hasAppChanges) {
  const changelogLink = 'https://github.com/styled-components/styled-components/blob/master/CHANGELOG.md'
  fail(`Please include a CHANGELOG entry. You can find it at <a href='${changelogLink}'>CHANGELOG.md</a>`)
}

// Warn if there are library changes, but not tests (excluding vendor)
const libraryOnlyFiles = jsModifiedFiles.filter(file => !vendorModifiedFiles.includes(file))
if (libraryOnlyFiles.length > 0 && hasTestChanges) {
  warn("There are library changes, but not tests. That's OK as long as you're refactoring existing code")
}

// Warn if StyledComponent.js was edited but not StyledNativeComponent.js or viceversa
const hasStyledChanges = danger.git.modified_files.find(path => path.endsWith('StyledComponent.js')) !== null
const hasNativeStyledChanges = danger.git.modified_files.find(path => path.endsWith('StyledNativeComponent.js')) !== null
if (hasStyledChanges && !hasNativeStyledChanges) {
[...]

PR #242, opened 2016-11-23
Fix comments from Danger

orta:
This should move you to 4.0.1 - which doesn't get mangled comments from the new github markdown renderer

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ install:
   - npm install
   - rvm install 2.3.1
   - rvm use 2.3.1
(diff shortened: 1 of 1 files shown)


mxstbr:
Can we do this in #169?

orta:
They are two different things, #169 completely removes all of Danger Ruby - but until I find time to work on that, you're going to have broken comments due to GitHub updates, this PR fixes that

mxstbr:
Let's temporarily merge this then, thanks!

PR #250, opened 2016-11-24
React native example

cem2ran:
Had to lock down the react version as it otherwise leads to some incompatibility with `15.4.1`. 
`Unable to resolve module react/lib/ReactUpdates`:
image


.babelrc
```diff
@@ -1,7 +1,6 @@
 {
   "presets": [
-    "latest",
-    "react"
+    "react-native"
```
mxstbr: Doesn't this break our build process for the browser version?

cem2ran: It builds just fine and tests pass. Dev also works. I thinks it just a matter of picking out which transforms are missing in the react + latest preset: https://www.diffchecker.com/Sez4D019

I know that babel-plugin-transform-class-properties is needed for sure, after applying that the next error that appeared didn't really specify what else was missing: `babelhelpers.typeof is not a function`

cem2ran: First thing I tried. You'll get a syntax error because the packager needs the react-native preset to package react-native.

mxstbr:
Merging this for now and continuing work on the `react-native` branch, I should be able to get this done now. Thanks so so so much for digging into this, seriously appreciated! 🙏

mxstbr:
Hmm, I still can't get this to work. How did you start the process? It fails with `index.ios.js not found` for me..

cem2ran:
You need to have two processes running from example/native:
1. `npm start` to launch the packager with options that sets the projectRoot. 
`react-native start` from root of the project should also work and try with `--reset-cache` flag if you're having issues/irregularities.
2. `npm run ios` to launch simulator

mxstbr:
Oh that works, beautiful! Thanks!

cem2ran:
👍

.babelrc
mxstbr: Can't we just leave all of them in here? Why do we have to take the others out?

mxstbr:
I put the `react` and `latest` presets back into the `.babelrc` and everything works perfectly fine. Why did you take them out?

cem2ran:
As tests and build passed with only the react-native preset I deemed them unnecessary. So no major reason :)

mxstbr:
I'll just leave them in there so we don't mess anything up accidentally. Thanks so much!

PR #253, opened 2016-11-24
Fix defaultProps/props theme theming

### LINKED ISSUE #247: Dynamicly update theme on react-storybook with knob
## Version
1.0.5

## Reproduction
https://github.com/clempat/styled-component-theming

## Steps to reproduce
- See the repo above
- See the result there: http://piano-tuner-larry-61732.netlify.com/

## Expected Behavior
- When updating the color on react-storybook-knob. it should update the button color

## Actual Behavior
- It get out of sync and display the color previously selected

I am not sure it is a problem with styled-components but if I take a look into the react dev tool I see the theme there correctly updated.

### LINKED ISSUE #252: Bugs in handling theme from defaultProps / props
## Version
1.1.1

## Reproduction
Couldn't install styled-components on webpackbin

## Steps to reproduce

```js
import { mount as render } from 'enzyme';
import jsdom from 'mocha-jsdom';

describe('theming', () => {
  jsdom();

  beforeEach(() => {
    styled = resetStyled()
  })

  it('should properly render with the same theme from default props on re-render', () => {
    const Comp1 = styled.div`
      color: ${props => props.theme.color};
    `

    Comp1.defaultProps = {
      theme: {
        color: "purple"
      }
    }
    const wrapper = render(
      <Comp1 />
    )
    expectCSSMatches(`.a { color: purple; }`)

    wrapper.update();
    expectCSSMatches(`.a { color: purple; }`)
  })

  it('should properly update style if theme is changed', () => {
    const Comp1 = styled.div`
      color: ${props => props.theme.color};
    `

    Comp1.defaultProps = {
      theme: {
        color: "purple"
      }
    }
    const wrapper = render(
      <Comp1 />
    )
    expectCSSMatches(`.a { color: purple; }`)

    wrapper.setProps({ theme: { color: 'pink' } })
    expectCSSMatches(`.a { color: purple; }.b { color: pink; }`)
  })

});

```

## Expected Behavior

Should pass tests.

## Actual Behavior

Fails in both cases.

First case fails because it will render with `.a { color: ; }` in second render.

Second case fails because it will not react to new `theme` so it won't render `.b { color: pink; }`.

michalkvasnicak:
Fixes #252 
Fixes #247

### DIFF
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -2,6 +2,8 @@
 
 import { createElement } from 'react'
 
+import type { Theme } from './ThemeProvider'
+
 import validAttr from '../utils/validAttr'
 import isTag from '../utils/isTag'
 import type { RuleSet, Target } from '../types'
@@ -59,7 +61,7 @@ export default (ComponentStyle: Function) => {
         }
       }
 
-      componentWillReceiveProps({ theme }: { theme: ?{ [key: string]: mixed } }) {
+      componentWillReceiveProps({ theme }: { theme: ?Theme }) {
         if (!theme) {
           return
         }
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -59,22 +59,17 @@ export default (ComponentStyle: Function) => {
         }
       }
 
-      componentWillReceiveProps(nextProps: any) {
-        if (!nextProps.theme) {
+      componentWillReceiveProps({ theme }: { theme: ?{ [key: string]: mixed } }) {
+        if (!theme) {
           return
         }
 
         // compare if there is any change
-        if (JSON.stringify(nextProps.theme) === JSON.stringify(this.state.theme)) {
+        if (JSON.stringify(theme) === JSON.stringify(this.state.theme)) {
           return
         }
 
-        const { theme } = nextProps
-
-        const generatedClassName = this.generateAndInjectStyles(
-          theme,
-          nextProps
-        )
+        const generatedClassName = this.generateAndInjectStyles(theme)
         this.setState({ theme, generatedClassName })
       }
 
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -65,9 +65,11 @@ export default (ComponentStyle: Function) => {
         }
 
         // compare if there is any change
-        if (JSON.stringify(theme) === JSON.stringify(this.state.theme)) {
+        /* eslint-disable max-len */
+        if (theme === this.state.theme || JSON.stringify(theme) === JSON.stringify(this.state.theme)) {
           return
         }
+        /* eslint-enable max-len */
 
         const generatedClassName = this.generateAndInjectStyles(theme)
         this.setState({ theme, generatedClassName })
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -61,7 +61,7 @@ export default (ComponentStyle: Function) => {
         }
       }
 
-      componentWillReceiveProps({ theme }: { theme: ?Theme }) {
+      componentWillReceiveProps({ theme }: { theme?: Theme }) {
         if (!theme) {
           return
         }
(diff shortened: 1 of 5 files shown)


src/models/StyledComponent.js
```diff
@@ -50,20 +50,32 @@ export default (ComponentStyle: Function) => {
...
         }
       }
 
       componentWillReceiveProps(nextProps: any) {
+        if (!nextProps.theme) {
+          return
+        }
+
+        // compare if there is any change
+        if (JSON.stringify(nextProps.theme) === JSON.stringify(this.state.theme)) {
```
mxstbr: Should this be a deep equality check instead of `JSON.stringify`? What's the difference?

src/models/StyledComponent.js
```diff
@@ -50,20 +50,32 @@ export default (ComponentStyle: Function) => {
...
+          return
+        }
+
+        const { theme } = nextProps
```
mxstbr: Let's move this destructure to the function argument and use the `theme` variable throughout `componentWillReceiveProps({ theme })`.

src/test/theme.test.js
```diff
@@ -1,13 +1,15 @@
 // @flow
+import jsdom from 'mocha-jsdom';
 import React from 'react'
-import { render } from 'enzyme'
+import { mount as render } from 'enzyme'
```
mxstbr: Why do we rename this to `render`? That's a separate mode, let's rename all instances of `render` to `mount` instead, otherwise this could be confusing.

src/models/StyledComponent.js
mxstbr: Since the main difference between a deep equality check and the `JSON.stringify` method (which I find very smart btw, have never heard of that before) is that the deep equality check doesn't care about the ordering of the elements. ([see here for more info](http://www.mattzeunert.com/2016/01/28/javascript-deep-equal.html))

How big is the a dependency that can do deep equality?

michalkvasnicak: @mxstbr I don't know if there is need to do other equality check than `===`. Because if we pass the same theme to `generateAndInjectStyles()` we should get the same result right?

I am looking on this https://github.com/substack/node-deep-equal.

src/test/theme.test.js
michalkvasnicak: Yeah I forgot to refactor this :)

src/models/StyledComponent.js
michalkvasnicak: Done. I hope flowtype is correct.

src/models/StyledComponent.js
```diff
@@ -50,20 +50,27 @@ export default (ComponentStyle: Function) => {
...
         }
       }
 
-      componentWillReceiveProps(nextProps: any) {
-        const generatedClassName = this.generateAndInjectStyles(
-          this.state.theme || this.props.theme,
-          nextProps
-        )
-        this.setState({ generatedClassName })
+      componentWillReceiveProps({ theme }: { theme: ?{ [key: string]: mixed } }) {
```
michalkvasnicak: @mxstbr This annotation is the same as `Theme` in `ThemeProvider`. I can export `Theme` type from `ThemeProvider` component and use it here. Then it would be `{ theme: ?Theme }`.

src/models/StyledComponent.js
michalkvasnicak: Yeah this solution is better than introducing new dependency :)

mxstbr:
Oh sorry, I just realized that this will probably increase our rendering time since we're now doing a full `jsdom` render for every test. Do you mind adding a `describe('jsdom', () => { jsdom() })` section in which we can put all the tests that require a full DOM render (i.e. `mount`), but use enyzmes `render` for all the original ones? That should speed it up quite a bit I think.

Sorry for the troubles, thanks!

michalkvasnicak:
@mxstbr done 😉

michalkvasnicak:
@mxstbr Do not merge this yet, there is buf in `componentWillReceiveProps`. We can't bail out at all because it will not react to props change so for example. If you have a style that is using props to calculate something, then it won't propagate changes at all.

michalkvasnicak:
@mxstbr ok fixed

mxstbr:
Amazing work, thanks so much! Shipping a new version with these changes very soon!

PR #262, opened 2016-11-27
Fix theming when theme has been changed using ThemeProvider

### LINKED ISSUE #261: Theme stopped updating
## Version
1.1.2 on React Native

## Expected Behavior
All components that use { theme } update when the theme is changed.

## Actual Behavior
Most of components stopped updating the colors after migrating from v1.1.1 to v1.1.2.

michalkvasnicak:
Fixes #261

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file. If a contri
 
 ### Changed
 
+- Fixed theme changes using `<ThemeProvider />` when changes has not been propagated to components (still using the first known theme). [@michalkvasnicak](https://github.com/michalkvasnicak) (see [#262](https://github.com/styled-components/styled-components/pull/262))
+
 ## [v1.1.2]
 
 ### Added
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -62,10 +62,10 @@ export default (ComponentStyle: Function) => {
       }
 
       componentWillReceiveProps(nextProps: { theme?: Theme, [key: string]: any }) {
-        const theme = nextProps.theme || this.state.theme
-
-        const generatedClassName = this.generateAndInjectStyles(theme, nextProps)
-        this.setState({ theme, generatedClassName })
+        // set theme only if we are not using ThemeProvider
+        if (this.props.theme) {
+          this.setState({ theme: nextProps.theme })
+        }
       }
 
       componentWillUnmount() {
@@ -76,7 +76,7 @@ export default (ComponentStyle: Function) => {
 
       render() {
         const { className, children, innerRef } = this.props
-        const { generatedClassName } = this.state
+        const generatedClassName = this.generateAndInjectStyles(this.state.theme, this.props)
 
         const propsForElement = {}
         /* Don't pass through non HTML tags through to HTML elements */
--- a/src/models/StyledNativeComponent.js
+++ b/src/models/StyledNativeComponent.js
@@ -54,10 +54,10 @@ const createStyledNativeComponent = (target: Target, rules: RuleSet, parent?: Ta
     }
 
     componentWillReceiveProps(nextProps: { theme?: Theme, [key: string]: any }) {
-      const theme = nextProps.theme || this.state.theme
-
-      const generatedStyles = this.generateAndInjectStyles(theme, nextProps)
-      this.setState({ generatedStyles, theme })
+      // set theme only if we are not using ThemeProvider
+      if (this.props.theme) {
+        this.setState({ theme: nextProps.theme })
+      }
     }
 
     componentWillUnmount() {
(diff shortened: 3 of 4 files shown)


michalkvasnicak:
@mxstbr appveyor didn't pass because order of css properties generated by autoprefixer is reversed.

michalkvasnicak:
Added `className` checks to tests too so it won't happen again.

k15a:
Unfortunately this doesn't work for me (tested with React)

michalkvasnicak:
@k15a can you create an example so I can reproduce this? Because I added the test that reproduced the issue and it's fixed.

k15a:
I will try it.

michalkvasnicak:
If you pass theme as same reference but with changed property it won't update a theme.

```js


```

michalkvasnicak:
I am sorry for reopening, misclicked. :(

k15a:
> If you pass theme as same reference but with changed property it won't update a theme.

Is this the intended behaviour?

k15a:
Reproduction without fix: https://styledtheme-xmbczspacp.now.sh/
Reproduction with fix: https://styledtheme-prtnqsuxfk.now.sh/

You can see the src code if you append `_src` to the url.

michalkvasnicak:
@k15a I am looking at https://styledtheme-prtnqsuxfk.now.sh/_src/?f=package.json and you don't have correct `styled-components`. You need to set it like this:

```json
{
   "dependencies": {
       "styled-components": "git://github.com/michalkvasnicak/styled-components.git#fix/theming"
   }
}
```

k15a:
I've copied the styled-components.js from dist to the src and importing it with '../styled-components.js'. See https://styledtheme-prtnqsuxfk.now.sh/_src/?f=src%2FApp.js. I will upload another one with the `"styled-components": "git://github.com/michalkvasnicak/styled-components.git#fix/theming"`.

k15a:
How can I trigger the styled-components build script if I am using the git url? Otherwise webpack can't find styled-components. Sorry I haven't done this before.

michalkvasnicak:
@k15a I tried your example locally and this bug is still here.

michalkvasnicak:
@mxstbr this is caused by weird race condition. In @k15a example it'll generate css classes for all themes but `componentWillReceiveProps` will rewrite changes made by `setState` from change subscriber.

In examples from above links there is change in `theme`. This will call `setState` in subscriber but it will also trigger re-render so `componentWillReceiveProps` will be called too. Somehow `this.state.theme` in `componentWillReceiveProps` is not new value from subscriber, but previous one (in this case it is always the value from first render).

michalkvasnicak:
If we keep everything as is except `componentWillReceiveProps` and we change it to be like this:

```js
componentWillReceiveProps(nextProps: { theme?: Theme, [key: string]: any }) {
        const theme = nextProps.theme || this.state.theme;

        const generatedClassName = this.generateAndInjectStyles(theme, nextProps)
        this.setState({ generatedClassName })
}
```

it starts to work but first change will be weird because it will use second theme not after first click, but after second click.

brunolemos:
@michalkvasnicak the second argument of `componentWillReceiveProps` is `nextState`, maybe that could help

brunolemos:
Another idea is to use the callback of `this.setState` to prevent the race condition

michalkvasnicak:
@brunolemos second argument is not `nextState` or maybe I am missing something (https://facebook.github.io/react/docs/react-component.html#componentwillreceiveprops)

image

I don't see how I can use callback from `this.setState` because we are setting `theme` and `generatedClassName` to `state`. But maybe I am just tired. Do you have some idea?

michalkvasnicak:
Only thing that helped is if we `generateAndInjectStyled` in `render` method. This way we have always the correct state.

Component `willReceiveProps` will work only if `this.props.theme` is set otherwise it won't do anything.

This way it behaves correctly with `ThemeProvider` or without it. But I think we would need to incorporate some memoization to `generateAndInjectStyles`. 

Any ideas?

michalkvasnicak:
Sorry again for reopening, incorrect rebase. Now it should be just fine. Please someone test it with react-native. It works fine with examples from @k15a. Also there is a test case which is testing `ThemeProvider` changes as well as props changes.

brunolemos:
@michalkvasnicak Tested on React Native, it's working 👍  :shipit:

brunolemos:
Yeah, probably, because it's back on render method.

michalkvasnicak:
Maybe there is but I can't find one. If we revert back to 1.1.1 then theming does not work without `ThemeProvider`. 

I don't see how `componentWillReceiveProps` is performance optimization. It is called on every re-render except the one issued by `setState`. So if you are using `ThemeProvider` like the one in examples from @k15a there is `setState` called with `componentWillReceiveProps` too.

michalkvasnicak:
I don't know how #137 has been benchmarked but from what I see there is only speed up in `render` but slow down in `componentWillReceiveProps` so I think this affects reconciliation. But really I don't know :) 

Maybe some memoization would be better, we return current theme if theme or props are not changed.

mxstbr:
Closing in favor of #264 which solves this much more elegantly. Thanks for the effort put into this PR and digging into it @michalkvasnicak!

PR #264, opened 2016-11-28
Fix theming

k15a:
If the subscribe function triggers a state change the componentWillReceiveProps lifecycle will still have the old state. Setting a property called theme circumvents this. In my opinion this is a React bug because the subscribe function changes the state before the componentWillReceiveProps lifecycle will fire.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file. If a contri
 
 ### Changed
 
+- Fixed theme changes using `<ThemeProvider />` where classNames didn't changed at the component [@k15a](https://github.com/k15a)
+
 ## [v1.1.2]
 
 ### Added
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -48,6 +48,7 @@ export default (ComponentStyle: Function) => {
           const subscribe = this.context[CHANNEL]
           this.unsubscribe = subscribe(theme => {
             // This will be called once immediately
+            this.theme = theme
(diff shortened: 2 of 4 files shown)


k15a:
Example without fix: https://styledtheme-xmbczspacp.now.sh/
Example with fix: https://styledtheme-qoqdyhbyop.now.sh/

mxstbr:
Ohh, the issue is that `setState` executes asynchronously, meaning the component `render` is called before the `setState` in `componentWillReceiveProps` has set the new state? Do I understand that correctly?

k15a:
Yeah that's the issue. ;)

k15a:
Or better the setState from subscribe is executed after compononentWillReceiveProps.

k15a:
If i understand it correctly the setState from subscribe is called after componentWillReceiveProps. So at the moment componentWillReceiveProps executes the setState from subscribe isn't executed and the setState from componentWillReceiveProps will override the change from subscribe.

k15a:
I've changed the synchronous property set to a asynchronous setState. Now it's a fix and not a hack :p

mxstbr:
Ah, that looks reasonably like a fix! Please test it and let me know!

k15a:
@mxstbr I've added a test case. AppVeyor is still failing because of the changed vendor prefix order.

brunolemos:
@mxstbr could you release a patch to npm? thanks

mxstbr:
I haven't? So sorry, will do tomorrow!

brunolemos:
@mxstbr 🙃

mxstbr:
I'm so bad, sorry @brunolemos. Pushing a new release right now, so sorry for the delay!

PR #268, opened 2016-11-28
Warn if component generates too many classes

vdanchenkov:
demo is https://demo-bpsynvnxvp.now.sh/
Move mouse for several seconds to see a warning. It pinpoints inappropriate usage that may lead to poor performance.

I'll look into tests if general approach is accepted.


vdanchenkov:
Not sure if and how it applies to React Native as I never used it.

mxstbr:
This is an interesting idea! @geelen, what do you think? Should we ship this?

vdanchenkov:
I have another idea of runtime warning that could be useful in development.

A lot of people misses the fact that 3-rd party components should accept and use className property. We could check that generated class name actually been used.  
```JS
if (!document.querySelector('.' + className)) warn()
```

Main concern is that there might be cases when you actually do not want to render the element with passed className. I have no examples of it, but if there are some we need to have a method to mute this warning.

mxstbr:
That's a super cool idea @vdanchenkov, can you submit an issue so we can get some input about people who might not want this warning?

geelen:
Yep, this is a nice idea definitely. 200 seems like a good limit too.

src/utils/warnTooManyClasses.js
```diff
@@ -0,0 +1,27 @@
+import styleSheet from '../models/StyleSheet'
+
+const components = WeakMap ? new WeakMap() : undefined
+const limit = 200
```
mxstbr: Let's make this `const LIMIT` to stay consistent with constants

src/utils/warnTooManyClasses.js
```diff
@@ -0,0 +1,27 @@
+import styleSheet from '../models/StyleSheet'
+
+const components = WeakMap ? new WeakMap() : undefined
```
mxstbr: Do we need this to be a WeakMap? Does Babel transpile that to an object or something or would this hurt browser support?

vdanchenkov: This code will run only in development mode. I assumed usage of the modern browser in development. If WeakMap is not supported - no checks will be made.

It was the easiest way to keep changes in the main codebase minimal, but I'll take a look at removing it.

vdanchenkov:
@mxstbr WeakMap removed, tests added. Message tweaked a little. I think it makes more sense now. What do you think? 

Resulting bundles look fine, but there are error messages 
```sh
yarn run build
> rollup -c && rollup -c --environment PRODUCTION

Creating development bundle...
rollup-plugin-inject: failed to parse /Users/user/work/styled-components/src/models/StyledComponent.js. Consider restricting the plugin to particular files via options.include
Creating production bundle...
rollup-plugin-inject: failed to parse /Users/user/work/styled-components/src/models/StyledComponent.js. Consider restricting the plugin to particular files via options.include
```
I'm not sure what to do with it. Do we need rollup-plugin-inject at all?

src/utils/createWarnTooManyClasses.js
```diff
@@ -0,0 +1,27 @@
...
+  return (className: string, displayName: string) => {
+    if (!warningSeen) {
+      generatedClasses[className] = true
+      if (Object.keys(generatedClasses).length >= LIMIT) {
+        const latestRule = styleSheet.rules().find(s => s.selectorText && s.selectorText.indexOf(`.${className}`) !== -1)
+        // Unable to find latestRule in test environment.
+        /* eslint-disable no-console, prefer-template */
+        console.warn(`Over ${LIMIT} classes were generated for component ${displayName}. ` +
+           'Consider using style property for frequently changed styles.\n' +
+           'Example: const MyStyledComponent = styled(props => <div className={props.className} style={{color: props.color}}>{props.children}</div>)`width: 100px; height: 100px;`' +
```
mxstbr: I don't think we should advocate usage of the `styled(props => <div />)` pattern, what about changing this example to these two lines:

```JS
const StyledComp = styled.div`width: 100%;`
<StyledComp style={{ background: background }} />
```

PR #286, opened 2016-12-03
Use stylis instead of PostCSS (21kB -> 9kB 🎉)

mxstbr:
styled-components bundle size (min+gzip) with...

- PostCSS: 21kB
- stylis: 9kB :tada:

----

I hack around stylis API a bit, because we e.g. don't want to prefix keyframes with the generated classname. (something stylis does by default). Maybe @thysultan can chime in here if there's any chance we could change this, otherwise this (slightly ugly) works too!

This isn't done as we need to do dependency cleanup and fix the whitespace changes in the tests, but I wanted to kick off a discussion if we want to go down that route!

cc @geelen


thysultan:
I've updates stylis. You can now opt out of namespacing keyframes & animations, more details on the [readme](https://github.com/thysultan/stylis.js#api)

mxstbr:
Most of the test failures are whitespace related, but there's two actual regressions when using stylis:

1. Nesting with `>` doesn't work:

   ```JS
   styled.div`
     > h1 {
       color: red;
     }
   `
   ```

  ```CSS
  /* The above should generate this CSS.. */
  .somerandomhash { }
  .somerandomhash > h1 {
    color: red;
  }

  /* ...but generates this 😢 */
  .somerandomhash { }
  > h1 {
    color: red;
  }
  ```

2. The autoprefixer doesn't autoprefix everything – which I'm not sure it has to, but `inline-style-prefixer` prefixes a lot more. I wonder where @rofrischmann gets the prefixing information from?

rofrischmann:
I am using the caniuse-api to fetch browser versions and prefix support. I might not have all prefixes, but https://github.com/rofrischmann/inline-style-prefixer/blob/master/generator/searchMap.js shows pretty much all properties I am checking. The key represents the caniuse search name.
Additionally there are some special plugins such as different flexbox specifications or value prefixes (e.g. `calc()`). (see https://github.com/rofrischmann/inline-style-prefixer/tree/master/modules/static/plugins)

thysultan:
@mxstbr stylis has gone through some updates so it now supports everything listed in [css-we-support.md](https://github.com/styled-components/styled-components/blob/master/docs/css-we-support.md) and more.

mxstbr:
Awesome!!

Will update this PR soon – we'll release it as v2.0.0 (i.e. a breaking change to make sure people don't kill us) in January, together with #227!

geelen:
So this needs some work before it can proceed—converting objects isn't quite working right (see the failing tests). Can you take a look @mxstbr?

mxstbr:
While v0.9 of stylis is out, it introduces features we really really don't want to have. (Sass-style `@mixin`, variables with `~~var`) I've upgraded to 0.8, which is the last version that has bugfixes we need but no unnecessary features.

mxstbr:
Fixed all the unnecessary test failures which are related to whitespace and whatnot, but there are three which I think are bugs upstream in stylis: (cc @thysultan)

1. flexbox is not as comprehensively autoprefixed as before. According to https://autoprefixer.github.io/, the prefixes we test for are correct:

  ```CSS
  display: flex;
  flex-direction: column;
  align-items: center;

  /* should turn into */
  display: -webkit-box;
  display: -ms-flexbox;
  display: flex;
  -ms-flex-direction: column;
  -webkit-box-orient: vertical;
  -webkit-box-direction: normal;
  flex-direction: column;
  -ms-flex-align: center;
  -webkit-box-align: center;
  align-items: center;

  /* but turns into */
  display: -webkit-flex;
  display: flex;
  -webkit-flex-direction: column;
  flex-direction: column;
  align-items: center;
  ```

2. Contextual selectors are inverted:

  ```CSS
  html.something & { }

  /* should turn into */
  html.something .asdf123 { }

  /* but turns into */
   .asdf123 html.something { }
  ```

3. Messes up with flat media queries:

  ```CSS
  { 
    background-color: blue;
    @media screen and (min-width: 250px) {
      background-color: red;
    }
  }

  /* should turn into */
  .asdf123 {
    background-color: blue;
  }
  @media screen and (min-width: 250px) {
    .asdf123 {
      background-color: red;
    }
  }

  /* but turns into */
  .asdf123 {
    background-color: blue;
    @media screen and (min-width: 250px) {
      background-color: red;
    }
  }
  ```

> Please note the comment above, it'd be great to get these bugfixes in as another patch release of 0.8 @thysultan!

thysultan:
@mxstbr 

> html.something & { }

Should be `:globa(html.something) &` since you are bailing out of the namespace scope but that still produces a bug, new patch will fix this.

> @media screen and...

Yes nested `@media` is a bug was not handling that before, fixed it in the new patch, will push the patch today.

> Please note the comment above, it'd be great to get these bugfixes in as another patch release of 0.8

> While v0.9 of stylis is out, it introduces features we really really don't want to have. (Sass-style @mixin, variables with ~~var) I've upgraded to 0.8, which is the last version that has bugfixes we need but no unnecessary features.

I can add a flag to disable compact features `mixins & variables`.

mxstbr:
Awesome thanks for the fast response @thysultan!

> Should be :global(html.something) & since you are bailing out of the namespace scope but that still produces a bug, new patch will fix this.

I'm not sure I understand this—why do users need to use `:global` here for a selector that _can_ only be global? What should that transpile to if we didn't use `:global`? It's not nested (no `>`), so I don't quite see why the special flag to escape the namespace is needed?

thysultan:
> I'm not sure I understand this—why do users need to use :global here for a selector that can only be global? What should that transpile to if we didn't use :global? It's not nested (no >), so I don't quite see why the special flag to escape the namespace is needed?

`h1 {}`  will generate `.namespace h1 {}` so it follows that `html {}` will generate `.namespace html {}` so `html.something & { }` will generate `.namespace html.something & { }` i could whitespace the `html` and `body` tags as special cases that aliases the global scope but this may break the edge case of when an iframe is within an element and you only want to select the html/body tag in the iframe within that element `& html.something {}`, the same can be said for the `body`. 

Actually now that i think about it i can make this work by assuming a case that when the `&` token is not the first character of the trimmed selector `& h1 {}` that it is assume to be a global scope since the current patched version has `h1 & {}` generating `.namespace h1 .namespace {}` which seems like a circular reference that will never select anything.

mxstbr:
> I can make this work by assuming a case that when the & token is not the first character of the trimmed selector & h1 {} that it is assume to be a global scope

This sounds super reasonable to me!

thysultan:
@mxstbr added the prefixes with the caveat that the old flex box webkit syntax for `flex-direction` specifically is not ployfiled.

```
-webkit-box-orient: horizontal;
-webkit-box-direction: normal;
```

geelen:
@thysultan That's basically how postcss-nested and Sass work yeah:

```scss
.namespace {
  .foo { /* .namespace .foo */ }
  & .foo { /* .namespace .foo */ }
  &.foo { /* .namespace.foo */ }
  .foo & { /* .foo .namespace */
}
```

Basically, if there's no `&`, you can implicitly add `& ` (including space) to the front.

thysultan:
[0.10.0 change log](https://github.com/thysultan/stylis.js/blob/master/CHANGELOG.md)

You can disable compact features(variables, mixins), handles flexbox prefixes better, and `html &` is treaded as a reference to the global scope.

mxstbr:
Updated to `0.10`, there are now four test failures related to the fact that we expect empty classes to be emitted:

```CSS
.sc-a { } .b { }
```

But stylis emits this

```CSS
.sc-a { }
```

But I feel like that actually is the better behaviour? @geelen maybe I'm missing something here.

----

There is (still) one real bug, again with `html`. This is the input CSS:

```CSS
html.something & {
  some: styles;
}
```

This is the stylis output:

```CSS
.b html.something {
  some: styles;
}
```

😢

thysultan:
>But stylis emits this `.sc-a { }`

@mxstbr It should actually emmit nothing(empty string).

In the console [https://rawgit.com/thysultan/stylis.js/master/tests/index.html](https://rawgit.com/thysultan/stylis.js/master/tests/index.html)

```
stylis('.abc', `
html.something & {
  some: styles;
}
`)

// and

stylis('.abc', `
.sc-a { } .b { }
`)
```

returns the expected

```
html.something .abc {
  some: styles;
}

// nothing
```

What am i missing?

mxstbr:
> It should actually emmit nothing(empty string).

No those classes have to be there, they're always empty. Not sure how we actually add them though, @geelen?

> What am i missing?

Hmm, let me give it another try and figure out what the issue is then.

mxstbr:
You are absolutely correct @thysultan, that was my mistake. Now a bunch of test failures popped up because we expect this:

```CSS
background: red;
html.something & {
  background: blue;
}
```

To generate this

```CSS
.b {
  background: red;
}
html.something .b {
  background: blue;
}
```

But get this:

```CSS
html.something .b {
  background: blue;
}
.b {
  background: red;
}
```

(same thing also goes for pseudo-selectors, nesting,...)

Any reason why those are switched around in the generated CSS? They should stay as they are defined in the source, otherwise this might lead to ugly specificity bugs?

thysultan:
> Any reason why those are switched around in the generated CSS?

You can place flat css anywhere, so all of the flat css as a single selector block either has to be prepended or appended, i'm currently appending, but it could just as well use prepending, in fact pre-pending sounds better, but this(adding all flat css at one time) was done to reduce the payload size since if you add flat css as they come you could end up with selectors for every property, though this could be mitigated with tracking the index the bytes that implementation would add did not seem worth it. Example of flat css anywhere.

```
background: red;

html.something & {
  background: blue;
}

color: red;

h1 { color: blue; }

color: orange;
```

> pseudo-selectors

what was the issue exactly? nested pseudo-selectors?

mxstbr:
Wait sorry I don't quite understand this. When you add the block as it comes, I assume it'd be in the direction the source is in? Why is there a discrepancy between the source order and the generated CSS order?

thysultan:
@mxstbr 

> Wait sorry I don't quite understand this. When you add the block as it comes, I assume it'd be in the direction the source is in?

```
stylis('.abc', `
color red;
color: orange;
h1 {color: red;}
color: blue;
`)
```

When it sees `color: red;` 1. it can add it directly but that would mean that `color:blue;` will also get added as it's own selector `.abc {color:blue;}` and so would `.abc {color: orange;}` 2. what i'm currently doing is collecting all flat css and adding it at the end just before the output is returned. 3. Another approach could to to add them at every intersect, in this case `red` and `orange` would get added together whereas color and any proceeding(before another intersect) flat selector will also get added together. Where an intersect is something hat breaks out of the flat css context. Between the 3 approaches the current one .2 yields the best payload deliver but the third one 3. is the most semantic approach and yields a better payload than option 1.

> Pseudo-selectors just have the same issue, the ordering is different from the passed source.

This is what i get, can you post an example.

```
stylis(`.abc`, `
h1:before, h2 { color: red; }
`)
".abc h1:before,.abc h2 {color: red;}"
```

mxstbr:
Take a look at the test failures on Travis 😉 For this input:

```CSS
background-color: blue;
&:hover {
  text-decoration: underline;
}
```

```CSS
// Got:
.sc-a { } .b:hover { text-decoration: underline; } .b { background-color: blue; } 
// Expected:
.sc-a { } .b { background-color: blue; } .b:hover { text-decoration: underline; } 
```

thysultan:
yes, going with the third option for adding flat css will fix the order of that as well.

> 3. Another approach could to to add them at every intersect, in this case red and orange would get added together whereas color and any proceeding(before another intersect) flat selector will also get added together. Where an intersect is something that breaks out of the flat css context.

mxstbr:
Let's go with 3. then? Would that be a big effort?

thysultan:
@mxstbr no it's not, will push a patch later today.

thysultan:
@mxstbr pushed v0.11.0 - [CHANGELOG](https://github.com/thysultan/stylis.js/blob/master/CHANGELOG.md)

mxstbr:
Now I'm getting duplicates:

```CSS
// I get (notice the double .d)
.sc-a { } .c { color: blue; } .c > h1 { font-size: 4rem; } .sc-b { } .d { color: blue; } .d > h1 { font-size: 4rem; } .d { color: red; } 
// I want
.sc-a { } .c { color: blue; } .c > h1 { font-size: 4rem; } .sc-b { } .d { color: blue; color: red; } .d > h1 { font-size: 4rem; } 
```

Intended?

thysultan:
@mxstbr what is the input?

mxstbr:
```CSS
color: blue;
> h1 { font-size: 4rem; }
color: red;
```

thysultan:
@mxstbr Yes that's intend, this is the method that preserves the order in which they are declared... if it where

```
color: blue;
color: red;
> h1 { font-size: 4rem; }
```

They would be together, anything that is not also flat css will break the current context of flat css and append that, when flat css is encountered again a new flat css context is initiated.

geelen:
@thysultan you shouldn't (I don't think) have any specificity issues grouping all rules at the same level together. As long as the order within each CSS rule is preserved, you should be fine to pull `color: red` up into the first `.d`. It's how PostCSS nested and Sass would both output that block.

That said, what you're currently outputting will work fine too. It just might look a little weirder in devtools.

thysultan:
@geelen edge case:

```
color: red;

& {
color: blue;
}

color: orange;
```

`orange` color should win but if you group `orange` with `red` the `blue` color will win.

thysultan:
> Maybe PostCSS and Sass will group them all under one selector?

Though if that's the case then multiple selectors will break it.

```scss
color: red;

&, h1 {
    color: blue;
}

color: orange;
```

mxstbr:
In your first case, Sass outputs this:

```CSS
div {
  color: red;
  color: orange;
}
div {
  color: blue;
}
```

So yeah, that breaks it. That behaviour might even be more consistent than Sass?

geelen:
Hahah true. Just confirmed that Sass behaves like I expect:

image

image

I'm happy with either. Your way is more correct, let's go with that.

mxstbr:
> Your way is more correct, let's go with that.

Awesome, will update the PR!

mxstbr:
As soon as CI passes this is ready to go! 🎉

mxstbr:
Failed because of CHANGELOG entry, but we'll have to do that for v2 down the line anyway. SHIPPING!

k15a:
Just wanted to thank you @thysultan. Amazing work you did 💯

kopax:
I have massive change when updating a project from v1.4.3 to v2.0.0-4, I suspect this is because of this.

I have no experience with stylis, and I've learned postcss when I started to use styled-components.

I have now a whole module for making website like bootstrap that require a old version of styled components. Did I miss the discussion of postcss >< stylis ? What are the big change and what about retro compatibility ?

mxstbr:
How do you mean massive change @kopax, is everything broken?

kopax:
@mxstbr, the last semicolon of rules break the css. 

That can be corrected by it requires some cold blood and redbull, apparently if styled-component release a `2.0.0-5` with the latest stylis version, this would be fix.

By massive I meant that media query issues described in #453 (see http://www.webpackbin.com/4J0u-y_Ff) have a major bug which break massively our stylesheets with ignored css from theses broken components.

mxstbr:
> That can be corrected by it requires some cold blood and redbull, apparently if styled-component release a 2.0.0-5 with the latest stylis version, this would be fix.

Will do that now!

PR #287, opened 2016-12-04
Replace lodash with util packages (-0.5kB package size)

mxstbr:
By using the `is-plain-object` (one dep, `is-object`) and the `is-function` (no deps) packages instead of using lodash utils (even when requiring them directly) we knock our min+gzip bundle size down by ~0.5kB.

(I think this is mostly due to lodash pulling in some additional files which we don't need)

Also adds the `rollup-plugin-visualizer` to the rollup build workflow, which creates a `bundle-stats.html` file that looks like this when viewed in the browser:

screen shot 2016-12-04 at 12 11 46

Very useful for keeping track of our bundle size 👍  (the above graph is for the new `stylis` bundle, in case you're wondering why it's so small. Ref #286)

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,10 +39,10 @@
   "homepage": "https://styled-components.com",
   "dependencies": {
     "buffer": "^5.0.0",
-    "css-to-react-native": "^1.0.3",
+    "css-to-react-native": "^1.0.6",
     "fbjs": "^0.8.4",
     "glamor": "^2.15.5",
-    "inline-style-prefixer": "^2.0.4",
+    "inline-style-prefixer": "^2.0.5",
     "is-function": "^1.0.1",
     "is-plain-object": "^2.0.1",
     "supports-color": "^3.1.2"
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -43,7 +43,9 @@
     "fbjs": "^0.8.4",
     "glamor": "^2.15.5",
     "inline-style-prefixer": "^2.0.4",
-    "lodash": "^4.15.0",
+    "is-function": "^1.0.1",
+    "is-plain-object": "^2.0.1",
+    "stylis": "^0.6.1",
```
oscar-b: `stylis` belongs in https://github.com/styled-components/styled-components/pull/286 doesn't it?

mxstbr: Oh yeah, good catch, my bad! I was trying this over on the #286 branch and then extracted those changes, will remove!

mxstbr:
Fucking hell finally fixed the whole test situation too. Tests were breaking on Travis due to caching, but not on AppVeyor because that had the newest version. What a mess!

Need to update all PRs to this as soon as it's merged.

jdalton:
For projects that are webpack friendly you can also use [lodash-webpack-plugin](https://github.com/lodash/lodash-webpack-plugin) to further reduce Lodash bundles.
For your current usage it looks like you could replace `isFunction` with just a `typeof` check.

PR #295, opened 2016-12-07
Fix overriding theme with props

k15a:
At the initial render styled-components was taking the theme from ThemeProvider even the component had an own theme prop.


mxstbr:
Amazing, thanks so much!

PR #300, opened 2016-12-09
Improve readme styling! 💅🏼

kutyel:
OK, I know it's a *silly* PR, but **isn't it all about style?** 😋 

I just love the [Babel](https://github.com/babel/babel) readme and really think this little detail is an improvement 😉 

screen shot 2016-12-09 at 14 46 49


mxstbr:
I actually had it like this at first, but liked the left-aligned version better! 

What do you think @geelen?

mxstbr:
Closing since I like the left-aligned version better 😉 thanks for taking the time to submit a PR though, much appreciated!

Issue #305, opened 2016-12-10
"Warning: Unknown props" when wrapping an external component


kitze:
I have this styled-component:
```js
import FontAwesome from 'react-fontawesome';

export const Icon = styled(FontAwesome)`
  font-size: 25px;
  margin-bottom: 3px;
  ${props => props.active && `color: ${props.colors.brand}`}
`;
```

Which is wrapping ```FontAwesome``` which is an external component.

I'm passing ```active``` and ```colors``` as props but the component complains that they are unknown.

![](https://i.imgur.com/ByNSMM5.png)

vdanchenkov:
Styled components pass props to the underlying component by design. You would get same results with 
```jsx
<FontAwesome active={} colors={}/>
```

Problem is that FontAwesome passes all unknown props to the DOM element. 

One solution is to filter props manually
```jsx
export const Icon = styled(({ active, colors, ...rest }) => <FontAwesome {...rest} />)
```

k15a:
How about something similar to this: https://github.com/styled-components/styled-components/issues/291#issuecomment-265292990

```JS
import styled from 'styled-components'
import FontAwesome from 'react-fontawesome'

export const Icon = styled(FontAwesome)`
  font-size: 25px;
  margin-bottom: 3px;
  ${props => props.active && `color: ${props.colors.brand}`}
`

Icon.passProps = false
```

diegohaz:
@k15a 

I like it, but most third party components need some custom props. It should be something like this:
```js
export const Icon = styled(FontAwesome).passProps({ active: false, colors: false })``
```

k15a:
You could also combine both and allow false to pass no props and a object which will whitelist / blacklist (which one should be decided) props.

diegohaz:
That `passProps` would be very useful. I currently have some duplicated code with @vdanchenkov's solution

e.g. A `Button` that renders a `MenuButton` from [react-aria-menubutton](https://github.com/davidtheclark/react-aria-menubutton) or a `Link` from [react-router](https://github.com/ReactTraining/react-router)
```js
const StyledMenuButton = styled(({ disabled, transparent, light, kind, size, ...props }) =>
  <MenuButton {...props} />
)`${styles}`
const StyledLink = styled(({ disabled, transparent, light, kind, size, ...props }) =>
  <Link {...props} />
)`${styles}`
```

diegohaz:
Another issue complaining about it: #135

k15a:
[...]

Issue #291, opened 2016-12-05
Proposal: CSS Class Composition


ticky:
Here at Buildkite, we’ve got a slightly odd situation with our styles. We’ve traditionally used Basscss in an attempt to make sure we’re DRY and things are conceptually simple, and partly in an effort to keep engineers and CSS usually-far-apart.
We’ve now hit the point where we want some bespoke stuff, and I think Styled Components best fills the gap we have - we want simplicity on one end and explicity and specificity on the other. Basscss fills the former, Styled Components the latter.

I’ve found, with this, that I want some sort of composition functionality. Given that the current approach would seem to be, implement your styles in one place, and set the classes you expect to compose elsewhere, it feels like it’s a bit less than ideal!

After a bit of fiddling I’ve settled for now on the workaround of setting `defaultProps.className` on the styled component;

```javascript
const Title = styled.h1`
  font-size: 1.5em;
  text-align: center;
  color: palevioletred;
`;

Title.defaultProps = {
  className: 'meow nya'
};

// => <h1 className="meow nya kZQsGZ" …
```

That way, the composition is in the same “place” as the Styled Component definition – but I wonder if the Styled Components API _itself_ should explicitly support this!

Having thought about the API surface area, it feels like a JS approach isn’t perhaps the best, as I feel while it would probably be the simplest solution to engineer, it would probably produce a rather more confusing and ugly API;

```javascript
const Title = styled.h1.withClass("meow nya")`
  font-size: 1.5em;
  text-align: center;
  color: palevioletred;
`; // => <h1 className="meow nya kZQsGZ" …

// ostensibly the same as
styled('h1').withClass("meow nya")//…
```

Instead, **I’d propose** an approach making use of the ability to extend syntax with PostCSS, like this;

```javascript
const Title = styled.h1`
  @compose 'meow nya';
  @compose 'purr';

  font-size: 1.5em;
  text-align: center;
  color: palevioletred;
`; // => <h1 className="meow nya purr kZQsGZ" …
```

A few more things which would require consideration;
[...]

PR #306, opened 2016-12-10
Options

k15a:
This PR will add some options to the styled components which will solve #305 and #291.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file. If a contri
 ## [Unreleased]
 
 ### Added
+- Added passProps and className option to the styled component [@k15a](https://github.com/k15a).
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


k15a:
Will do the changes to StyledNativeComponent and add some tests once this is approved 👍

mxstbr:
This needs a lot more discussion.

Right now our API surface is very small, and that's a very good thing. Simply introducing options like this will make it much harder to contain the API surface we have and keep styled-components simple!

This doesn't mean we won't ever add these features, but the how and why need a lot more discussion and thought that I haven't put in yet. (also, `classes` was trail-implemented in #227, though I'm not a fan of that either.)

Thanks for taking the time to figure this out, let's discuss a bit more about this and then try to figure out which direction we want to go in terms of convention vs configuration!

avin-kavish:
lol

PR #310, opened 2016-12-12
Do not pass innerRef to the component

mkhazov:
Avoid
```Warning: Unknown prop `innerRef` on <textarea> tag. Remove this prop from the element. For details, see https://fb.me/react-unknown-prop```

### CODE REVIEW

File: src/models/StyledComponent.js [hunk 1]
```diff
@@ -84,7 +84,7 @@ export default (ComponentStyle: Function) => {
         const propsForElement = {}
         /* Don't pass through non HTML tags through to HTML elements */
         Object.keys(this.props)
```
mxstbr: I like that!

PR #312, opened 2016-12-12
withTheme Higher Order Component (fix #272)

### LINKED ISSUE #272: Provide easier way to access the theme outside 'styled' components
It's easy to get the `theme` from components that are created using the `styled` way:

```jsx
    color: ${({ theme }) => theme.color};
`;
```

But how to get from **normal components** or apply it for **different properties**? Example:

```jsx
<ThemeProvider theme={{ color: 'red' }}>
    <Main />
</ThemeProvider>
```
```jsx
<View>
    <Card aCustomColorProperty={PASS COLOR FROM THEME HERE} />
</View>
```
> Notice how the property that needs the theme is called `aCustomColorProperty`, not `style`

We can use this issue to discuss the topic and provide different solutions.
I'll post mine in the next comment.

_This was originally posted on [stackoverflow](http://stackoverflow.com/q/40796973/2228575) and [asked by @mxstbr](http://stackoverflow.com/questions/40796973/how-to-get-the-theme-outside-styled-components#comment68961671_40848211) to be posted here_

brunolemos:
Fix #272.

Add the possibility to get the current theme using `withTheme(component)` syntax. The theme will be passed as a `theme` prop to the child component.

### DIFF
--- a/docs/theming.md
+++ b/docs/theming.md
@@ -55,3 +55,22 @@ const InvertColors = ({children}) => (
   </ThemeProvider>
 )
 ```
+
+## Getting the theme outside styled components
+
+If you ever need to get the current `theme` outside styled components (e.g. inside big components), you can use the `withTheme` Higher Order Component:
+
+```JS
+import { withTheme } from 'styled-components'
+
+class MyComponent extends React.Component {
+  render() {
+    const { theme } = this.props
+
+    console.log('Current theme: ', theme);
+    // ...
+  }
+}
+
+export default withTheme(MyComponent)
+```
--- a/.babelrc
+++ b/.babelrc
@@ -7,6 +7,7 @@
     "flow-react-proptypes",
     "add-module-exports",
     "transform-flow-strip-types",
-    "transform-object-rest-spread"
+    "transform-object-rest-spread",
+    "transform-class-properties"
   ]
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file. If a contri
 
 ### Added
 
+- Added [`withTheme`](docs/api.md#withtheme) higher order component; thanks [@brunolemos](https://twitter.com/brunolemos). (see [#312] (https://github.com/styled-components/styled-components/pull/312))
 - Fixed prop changes not updating style on react native; thanks [@brunolemos](https://twitter.com/brunolemos). (see [#311](https://github.com/styled-components/styled-components/pull/311))
 
 ### Changed
--- a/docs/api.md
+++ b/docs/api.md
@@ -66,7 +66,7 @@ This can take any combination of
 
 > **Note**: You can also pass an Object as an interpolation, and we will treat it like inline styles. This is highly discouraged, as the CSS syntax has support for media queries, nesting, etc., which the object syntax doesn't.
 
-The properties that are passed into an interpolated function get attached a special property, `theme`, which is injected by a higher level [`ThemeProvider`](#theme-provider) component.
+The properties that are passed into an interpolated function get attached a special property, `theme`, which is injected by a higher level [`ThemeProvider`](#themeprovider) component.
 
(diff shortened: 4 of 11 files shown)


README.md
```diff
@@ -401,6 +401,25 @@ Button.defaultProps = {
...
+  render() {
+    const { theme } = this.props
+
+    console.log('Current theme: ', theme);
+    // ...
+  }
+}
+
+export default withTheme(Component)
+```
```
mxstbr: This can only live inside theming.md. No need for this to be in the README, it's not a core API!

docs/theming.md
```diff
@@ -55,3 +55,22 @@ const InvertColors = ({children}) => (
   </ThemeProvider>
 )
 ```
+
+## Getting the theme outside styled components
+
+If you ever need to get the current `theme` outside styled components (e.g. inside render functions), you can use the `withTheme` Higher Order Component:
```
mxstbr: "e.g. inside render functions" doesn't make much sense to me, what about "e.g. inside big components"?

k15a:
@mxstbr Isn't the ThemeProvider to provide the context and the withTheme to access the context? How is this "extendible"?

mxstbr:
Oops, I meant the `StyledComponent` and the `StyledNativeComponent` of course, brain fart on my part! Fixed my comment 😊

brunolemos:
@mxstbr About the extending, the only similarity is that `withTheme` and `StyledComponent` both subscribe to context, but they do different things in the subscription: one just get the theme, the other needs to handle style parsing. And it's important that this style parsing happens together with the new theme arriving, I remember it was made this way to fix some bugs, so I'm not sure it's a good idea to separate them.

Updated the commits!

geelen:
All good from my end! Nice work @brunolemos <3

brunolemos:
@mxstbr time for a 1.1.4 release? :)

geelen:
Oh btw I was getting some flow warnings locally, but they don't seem to be breaking the build:

```
yarn flow v0.15.1
$ flow; test $? -eq 0 -o $? -eq 2
src/hoc/withTheme.js:6
  6: export default Component => class extends React.Component {
                    ^^^^^^^^^ parameter `Component`. Missing annotation

src/hoc/withTheme.js:22
 22:     this.unsubscribe = subscribe(theme => {
              ^^^^^^^^^^^ property `unsubscribe`. Property not found in
 22:     this.unsubscribe = subscribe(theme => {
         ^^^^ class expr `<<anonymous class>>`


Found 2 errors
✨  Done in 0.26s.
```

@mxstbr thoughts on this?

brunolemos:
Tried to use `ReactClass<{}>`, fixes the `flow` command but `eslint` plugin does not understand it. 
Possibly related: gajus/eslint-plugin-flowtype#107

kopax:
So this `withTheme(component)` will it make us possible to have a defaultTheme if not `<ThemeProvider />` is provided, and in case  `<ThemeProvider />`, it will use the passed theme ?

Also, it will allow us to extend styled is that correct ? 

Where do you pass the css ?

It sounds the same as `Styled(Component)` to me.

I am not sure if I understand the benefits compare to the previously used method, would you mind explain to me?

mxstbr:
> Oh btw I was getting some flow warnings locally, but they don't seem to be breaking the build:

What, that's weird! We might need to tweak the CI settings and can we get that fixed @styled-components/typers?

> Where do you pass the css ?

You don't pass any CSS, that's the point 😉 `withTheme` allows you to get the global `theme` _in any component_, even ones that aren't styled components!

relekang:
> What, that's weird! We might need to tweak the CI settings and can we get that fixed @styled-components/typers?

Oh, looks like CI is not running flow at all. Will create a PR fixing that and the errors that have made it to master✌️

cellvia:
this seems awesome, plans to pull it?

eric-burel:
Hi, is this supposed to work with functionnal components ?

PR #317, opened 2016-12-17
Add support for numbers in objects

### LINKED ISSUE #283: Numbers in objects don't get a unit appended
I'm trying to migrate my app from aphrodite (which uses objects) to 💅, but I have some issues. 

For example [this object](https://i.imgur.com/4wfwMMa.png) gets compiled to [this](https://i.imgur.com/BSp2c3K.png).

The problem is it's not adding "px" where it's supposed to.

@mxstbr suggested [this tool](https://jsbin.com/gesaniseto/edit?js,output) for easier migration but it's also not converting the syntax [properly](https://i.imgur.com/gw6lMch.png).

mxstbr:
This is necessary for an easier transition from other styles in JS
libraries that let you write styles as objects. All of them support the
following:

```JS
const styles = {
  fontSize: 14,
}
```

This commit adds support for this in styled-components. There is a
handful of properties where appending `px` doesn't make sense, I stole
the list of non-pixel-properties from JSS (thanks @kof) and simply
append their unit.

Closes #283

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file. If a contri
 
 ### Added
 
-- Added support for media queries, pseudo selectors and nesting in styles-as-objects. (see [#280](https://github.com/styled-components/styled-components/pull/280))
+- Added support for media queries, pseudo selectors, numbers and nesting in styles-as-objects. (see [#280](https://github.com/styled-components/styled-components/pull/280) and [#317](https://github.com/styled-components/styled-components/pull/317))
 - Added [`withTheme`](docs/api.md#withtheme) higher order component; thanks [@brunolemos](https://twitter.com/brunolemos). (see [#312] (https://github.com/styled-components/styled-components/pull/312))
 - Fixed prop changes not updating style on react native; thanks [@brunolemos](https://twitter.com/brunolemos). (see [#311](https://github.com/styled-components/styled-components/pull/311))
 
--- a/src/utils/flatten.js
+++ b/src/utils/flatten.js
@@ -4,10 +4,9 @@ import isPlainObject from 'is-plain-object'
 
 import type { Interpolation } from '../types'
 
-// A list of properties that need a unit other than px appended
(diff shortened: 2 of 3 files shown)


mxstbr:
Hmm, I forgot unitless numbers. There's things like opacity, which shouldn't get `px` appended:

```JS
const styles = {
  opacity: 0,
}
```

This is more complex than I initially thought, is it worth it to add this?

mxstbr:
88fe81c adds support for unitless numbers.

I'm not 100% if we really want to merge this as it adds quite a bit of code and whitelists, but it'd be a nice addition. I'm slightly in favor, as making it easy to switch to styled-components is quite a big deal, but I'm leaving the final call with @geelen.

k15a:
How about moving this out in an extra package like `unitless`. This way only people moving their codebase to `styled-components` would have to add this.

I could imagine an API like:

```JS
unitless({
  flex: 1,
})
```

mxstbr:
> How about moving this out to an extra package. This way only people moving their codebase to styled-components would have to add this.

Hmm, interesting. I hadn't considered that. The thing is, if we want to add a proper API for objects down the line we'll have to have this anyway?

k15a:
> The thing is, if we want to add a proper API for objects down the line we'll have to have this anyway?

Yeah probably even if I don't understand the value of adding the unit automatically. In my opinion appending `px` or `em` manually is much more clear.

geelen:
Yeah.... I don't like this. Lots of CSS properties can have multiple units, and having `px` as a default is not really correct. Assuming that everything is driven from pixel measurements is a sure fire way to making brittle styles, since you're missing out on that whole part of CSS. The point of Styled Components is to not artificially limit the power of CSS, in fact the very choice of using template strings was to stop this artificial friction, where `14` is a first class JS symbol but `14px` or `14pt` requires being wrapped in quotes, the same way an object key `opacity` is fine but `font-size` needs to be wrapped in a string or camelized.

I get that it's only for objects but it's a design choice (default unit being `px`) that I don't want to help "standardise" by adding support for it. So it's a no from me.

PR #321, opened 2016-12-18
Add security doc

mxstbr:
Adds a documentation page to warn developers about CSS injection, a possible attack vector against their web applications.

This is in no way specifically styled-components related, you should always santize input, but since styled-components _does_ make it easy to interpolate arbitrary CSS declarations and also lets users add themes we should mention this.

*Full credit for notifying me of this goes to @jayphelps*


mxstbr:
(CI failure unrelated and due to a stale cache on AppVeyor)

docs/security.md
```diff
@@ -0,0 +1,62 @@
+# Security
+
+**Please do not forget to sanitise and validate input!**
```
k15a: `sanitise` should be `sanitize`

docs/security.md
```diff
@@ -0,0 +1,62 @@
...
+
+```CSS
+.YourApp__navbar__item--active {
+  background: url(/api/withdraw-funds);
+}
+```
+
+**⚠ OH NO ⚠** suddenly, every user that finds this random theme on Google and copy and pastes it to your app interface withdraws all their funds!
+
+### Constain user input
```
k15a: `Constain` should be `Constrain`?

PR #324, opened 2016-12-20
Updated all deps to latest semver

amilajack:
- Fixed linters
 - Updated deps


amilajack:
Sure thing. Also any idea why travis tests are failing? Because of cache?

amilajack:
Also I would recommend greenskeeper for this repo. Thoughts?

mxstbr:
> Also I would recommend greenskeeper for this repo. Thoughts?

I've not had great experiences with greenkeeper, it's quite painful to keep up. Updating these deps isn't really something we have to do very often, as contrary to e.g. `react-boilerplate` having up to date deps is unimportant for this package.

mxstbr:
Massive thanks for doing this though, very much appreciated! Just because we don't have to doesn't mean we don't want to update every so often 😊

amilajack:
> Updating these deps isn't really something we have to do very often, as contrary to e.g. react-boilerplate having up to date deps is unimportant for this package.

Makes sense. There was a concern with the `eslint-plugin-flowtype-errors` plugin, which originally locked down the version of flow. To update the version of flow, the plugin needed to be updated. This isnt the case with `eslint-plugin-flowtype-errors@2`. This was the reason for the update.

Issue #302, opened 2016-12-09
Flow errors with latest versions


msmfsd:
### v1.1.3
I am getting flow errors when using latest versions of styled-components with create-react-app. 

My flow config is as per create-react-app and styled-components docs and I have done a flow-typed install as well. My deps:
```
"react-scripts": "0.8.3"
"flow-bin": "^0.36.0",
"react": "^15.4.1",
"react-dom": "^15.4.1",
"styled-components": "^1.1.3"
```

Flow errors:
```
node_modules/styled-components/lib/utils/autoprefix.js.flow:2
  2: import camelizeStyleName from 'fbjs/lib/camelizeStyleName'
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ fbjs/lib/camelizeStyleName. Required module not found

node_modules/styled-components/lib/utils/autoprefix.js.flow:3
  3: import hyphenateStyleName from 'fbjs/lib/hyphenateStyleName'
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ fbjs/lib/hyphenateStyleName. Required module not found

node_modules/styled-components/lib/utils/flatten.js.flow:2
  2: import hyphenate from 'fbjs/lib/hyphenateStyleName'
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ fbjs/lib/hyphenateStyleName. Required module not found

node_modules/styled-components/lib/models/StyledNativeComponent.js.flow:24
 24:   class StyledNativeComponent extends ParentComponent {
                                           ^^^^^^^^^^^^^^^ identifier `ParentComponent`. Expected polymorphic type instead of
 24:   class StyledNativeComponent extends ParentComponent {
                                           ^^^^^^^^^^^^^^^ string
```

This below seems to fix it for now:
```
.*/styled-components/lib/utils/autoprefix.js.flow
.*/styled-components/lib/utils/flatten.js.flow
.*/styled-components/lib/models/StyledNativeComponent.js.flow
```

mxstbr:
@styled-components/typers any ideas?

relekang:
The problem here is that react-create-app documentation says to add the following to ignore sections of the flowconfig.

```
<PROJECT_ROOT>/node_modules/fbjs/.*
```

Since we import that and have not provided a libdef for it flow will give `Required module not found`. We can fix this two ways:
* Add to flow docs that one should only ignore problematic paths in fbjs.
[...]

PR #335, opened 2016-12-27
Fix flow ci config and use default flow supressor

relekang:
This fixes an issue where `npm run flow` always exited with `0` and thus the CI job never failed. It fixes a few errors that had sneaked in by ignoring react-native source and upgrading libdefs.

Furthermore, it changes the suppressor comment to `// $FlowFixMe` which is the default. This will make it easier for users to adopt flowtype. (Based on [tip on twitter](https://twitter.com/flowtype/status/813119893969240069))

cc @styled-components/typers


ryyppy:
I like it :-) ... what's happening with CI tho?

relekang:
What do you mean? Last commit is green. 🙂

mxstbr:
Awesome, thanks!!

Augustin82:
@ryyppy 
Like I asked in issue #302, are you sure this fixes the problems mentioned there? When can we expect to see this PR in a release?

Many thanks.

ryyppy:
I just tried the current commit on your example project,.. unfortunately the error still persists

Augustin82:
@ryyppy 
Thanks for trying! Shouldn't the issue be reopened, then? Clearly there's something wrong, with a basic clean styled-components install causing flow errors for the user project (thus technically preventing CI and flow gen-files and other stuff).

ryyppy:
@relekang I believe there is something wrong in general, since `import styled from 'styled-components` does not export anything relevant... I would really really like to see a libdef on `flow-typed` as a fix

PR #345, opened 2017-01-02
Use ThemeProvider theme instead of defaultProps theme

### LINKED ISSUE #344: defaultProps theme is overriding ThemeProvider theme
## Version
1.2.1

## Reproduction
http://www.webpackbin.com/EyaDeRzHG

## Expected Behavior

Text color to be green, which is the color defined in `ThemeProvider`

## Actual Behavior

Text color is red, which is the color defined in `Text.defaultProps`

If I remove `defaultProps`, text color becomes green.

diegohaz:
Fix #344


src/models/StyledComponent.js
```diff
@@ -48,7 +48,10 @@ export default (ComponentStyle: Function) => {
           const subscribe = this.context[CHANNEL]
           this.unsubscribe = subscribe(nextTheme => {
             // This will be called once immediately
-            const theme = this.props.theme || nextTheme
+            const { defaultProps } = this.constructor
+            const defaultTheme = defaultProps && defaultProps.theme
+            const isDefaultTheme = defaultTheme && this.props.theme === defaultTheme
```
k15a: Couldn't this be also `const isDefaultTheme = this.props.theme === defaultTheme`?

diegohaz: You are right. I'm doing some small changes right now.

diegohaz:
@k15a There's the test you wrote on #295 and I added [a test](https://github.com/styled-components/styled-components/pull/345/files#diff-e9e543c50e5cefa6f2c6469f6abccc5bR91) to ensure that `props` will take precedence over `ThemeProvider` even if its value is equal to `defaultProps`. Do you know some other use case?

diegohaz:
I did small changes on the code for better readability. Also added some comments (I'm not a big fan of it, but this `props` > `ThemeProvider` > `defaultProps` may be confusing).

ghalex:
This still dosen't work as it should be, check this example:
https://codesandbox.io/s/6w6nq15n2k

diegohaz:
@ghalex Once you pass the prop from a normal component to a styled component (`{...this.props}`), there's no way for the latter to know whether it came from props or default props.

But you can enhance component with `withTheme`: https://codesandbox.io/s/m3nrmxvjw9

ghalex:
Thanks @diegohaz  I understand now why it only works if you pass defaultProps only directly to styled component.

ghalex:
Another thing @diegohaz  your example crashes in production when you have no ThemeProvider. Check this https://codesandbox.io/s/vr3q97or5 I have deployed it here: https://csb-vr3q97or5-ylvgjfxbld.now.sh/ and it crash. I think is related to this #1709

PR #346, opened 2017-01-02
Update faq.md

SachaG:
Adding a section to the FAQ to clarify the fact that styled-components will generate duplicate styles.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -37,7 +37,7 @@ In a future version of webpack this bug should be fixed. See [issue#115](https:/
 
 ### My styles are being repeated multiple times
 
-You should be aware that generating styles based on dynamic props will result in repeated CSS declarations. In other words, in the following example:
+You might notice that generating styles based on dynamic props will result in repeated CSS declarations. In other words, in the following example:
 
 ```js
 const Button = styled.button`
@@ -48,7 +48,7 @@ const Button = styled.button`
 `;
 ```
 
-You would ultimately end up with two classes, both of which contain the same "more styles here" lines:
+You will ultimately end up with two classes, both of which contain the same "more styles here" lines:
 
 ```css
 .foo{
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -34,3 +34,44 @@ alias: {
...
+  padding: 0.25em 1em;
+  /* …more styles here… */
+}
+.bar{
+  padding: 0.5em 2em;
+  /* …more styles here… */
+}
+```
+
+You may decided that this is an acceptable drawback since increasing the size of a CSS file rarely has any noticeable impact on performance compared to adding a single image to a site, but if this is a concern you can use the `css` helper to factor our the common styles into their own declaration:
```
vdanchenkov: Usage of `css` helper has no effect on generated css rules. Second example will generate same exactly the same rules in style tag.

SachaG: Oh, I misunderstood what @mxstbr told me then. Is there any way to achieve what I'm talking about?

vdanchenkov: What is your concern? Regular usage or server-side rendering.

If it's about regular usage - number of rules in generated classes doesn't matter in practice.

If it's about SSR I'm not sure what to recommend, by SSR is not supported yet anyway.

SachaG: Regular usage. I'm not sure if I agree that the number of rules doesn't matter. Or at least, I think it matters enough to some people to be worth addressing somewhere in the docs.

vdanchenkov: Sorry, you are right that it worth mentioning. I meant that I'm not aware of any performance issues related to it. If you have some measurements indicating that extraction of common rules improves performance I'd love to discuss it.

SachaG: Wouldn't extracting common rules make the file smaller, and thus improve performance? The improvement would then strictly depend on the number of rules extracted. So I'm not sure what there is to measure? 

Again I know in *most* cases this is not a major worry, but I just think people should understand how styled-components works in order to make the decision themselves.

mxstbr: > Wouldn't extracting common rules make the file smaller, and thus improve performance?

No, since you're only sending down the JS file the _only difference that would make is that there would be less CSS injected into the HEAD of the DOM_.

We've tested this quite thoroughly (/cc @geelen) and not noticed any performance impact related to that. This will become relevant when we support SSR (#214), in which case you will send down slightly more CSS.

I'm inclined to include this note?

SachaG: Oh right, I was thinking about SSR. In a way, I think the fact that I got pretty much everything wrong about this is in itself a good reason to include a note about it in the docs :)

k15a: Why is this a problem with SSR? You are probably not changing the styles on the server when rendering because you don't have interactions from the user so no new styles have to be applied?
I am missing something important?

vdanchenkov: There might be several instances of one component in different states. But I think that gzip will cover it anyway.

mxstbr: Actually, you're right! `gzip` will cover that as far as I'm aware, so it shouldn't be an issue?

SachaG: You can only gzip individual CSS files though, or does that also help with CSS injected directly into the HTML file?

SachaG:
OK, let me know if this is better :)

docs/faq.md
k15a: Yes you can and should gzip all your files. So HTML with CSS injected should also be gzipped.

mxstbr:
Awesome, thanks so much @SachaG!

Issue #229, opened 2016-11-21
Add screenshot integration tests


mxstbr:
We should do automated browser testing for each PR to make sure we don't have regressions like #212 again and to make sure our injection mechanism works as expected.

~~I have never done that, so if somebody has tips on free services we could use to test on IE10+ and all major evergreen browser that'd be 👌 It basically just needs to make sure our example in the `example/` folder renders as expected in all browsers and doesn't throw errors.~~

The best way to go about this would be to use screenshot integration testing, similar to [what we do for the styled-components website](https://github.com/styled-components/styled-components-website/pull/168). Essentially, have a simple page that you render with `puppeteer`, then screenshot it and compare it against a stored screenshot in the repo. If there's any diversion between the screenshots (i.e. if any injected styles have changed), fail the tests.

This would be a great first task for anybody wanting to start contributing to styled-components. Take a look at the website PR (https://github.com/styled-components/styled-components-website/pull/168) and implement something based on that in this repo.

thisguychris:
@mxstbr take a look at saucelabs.

FezVrasta:
I can take this, I literally just finished tweaking the SauceLabs tests for one of my projects, I could simply copy-paste the setup.

We should add `karma` and `karma-sauce-launcher` to be able to run `mocha` in the SL browsers.

Would it be okay @mxstbr?

mxstbr:
I would love for you to do that @FezVrasta!

geelen:
@FezVrasta thanks so much for picking this up! I'm going to add some comments against your PR. ❤️

yanivefraim:
@FezVrasta @mxstbr - why Karma and not Selenium Webdriver? Selenium Webdriver is the standard way to write browser tests. It let you test on real browsers plus use service like [Applitools](https://applitools.com/) (we use them @ our company), which let you compare an image of your component before and after your change, so you have control over breaking UI changes... WDYT?

mxstbr:
@yanivefraim no strong opinion on either, if you want to go ahead and submit a PR that integrates Selenium I'd love you forever!

mxstbr:
We should do screenshot integration tests instead of relying on an external service, now that that's very easy to do with `puppeteer`. See the styled-components-website PR by @philpl https://github.com/styled-components/styled-components-website/pull/168, we'd love a PR that implements something similar!

maciej-ka:
I will be happy to work on this.

morajabi:
[...]

Issue #212, opened 2016-11-18
Internet explorer 10+ styled-components issue (JS error)


fatpc:
Hi Guys -

Our app works in chrome and firefox except for IE10+. We are getting below JS error:

![image](https://cloud.githubusercontent.com/assets/8001832/20417260/4a6a3b5a-ad98-11e6-9514-fb300f85bbf2.png)

Below is the specific line of code:

![image](https://cloud.githubusercontent.com/assets/8001832/20417282/63e6dfd4-ad98-11e6-9699-c71cfa4293ae.png)

Is this a known issue? Also tried to access the styled-components website in IE and it does not load (JS error).

Thanks,
Pat

## Version
1.0.10

## Steps to reproduce
Run app in IE10+

## Expected Behavior
App works

## Actual Behavior
JS Error

mxstbr:
Ugh that's annoying. Will fix this asap, sorry for the issues!

thisguychris:
@mxstbr / @fatpc the fix is just use babel-polyfill via `import 'babel-polyfill'`

@mxstbr - while including babel-polyfill on a library level would indeed fix this.  As per babel docs: This  
>...will emulate a full ES2015 environment and is **intended** to be used in an application rather than a library/tool.

So I'm in disagreement including the polyfill in this library, maybe we just add it in the README?  Another alternative is just polyfill what the library use, maybe Symbol, or Array functions?  Let me know what course of action we should take and I'll issue a PR.

mxstbr:
We'll just not use `.startsWith`, instead let's use a regex to check this!

`/$--/.test(decl.prop)` should do the trick I think?

thisguychris:
@mxstbr checking in ie11 though I'm seeing a different error:
http://take.ms/U3iTM

That regex doesn't work by the way.
how about `!!(decl.prop.indexOf('--') === 0)`

mxstbr:
I did mean `/^--/.test(decl.prop)`, my brain had a hanger there 😉 Would you mind submitting a PR to fix this? Thanks!

We fixed the Symbol thing a while ago, can you make sure you're running the newest version of `styled-components`?

thisguychris:
@mxstbr sure will issue a PR in awhile.  Will pull the latest as well.

thisguychris:
@mxstbr / @fatpc fixed at #217

fatpc:
@thisguychris & @mxstbr thanks! will test it as well :D

mxstbr:
[...]

PR #348, opened 2017-01-03
[NOT READY] added support for Karma + SauceLabs tests

FezVrasta:
I added the needed configuration files.

There are some problems when I try to run them:

1. We need to get rid of Enzyme and jsdom from tests
2. Even after removing jsdom and enzyme, I get `'Component' is not exported by node_modules/react/react.js (imported by src/test/basic.test.js).`

Leaving here in case someone know how to fix this error.

@mxstbr you will have to subscribe to SauceLabs and email them asking for an open source account for this repository. If you need help feel free to email/DM me.

This PR aims to implement #229


rollup.config.js
```diff
@@ -1,13 +1,13 @@
 /* eslint-disable flowtype/require-valid-file-annotation, no-console */
-import nodeResolve from 'rollup-plugin-node-resolve'
```
FezVrasta: I had to convert this to ES5 because karma doesn't support ES6 `import`s and `export`s and I had problems in the past using `babel-register` inside Karma.

rollup.config.js
```diff
@@ -61,13 +63,17 @@ const plugins = [
       'transform-object-rest-spread',
       'transform-class-properties',
     ],
+    exclude: [
+      'node_modules/**',
+      '*.json',
```
FezVrasta: Without this, the Karma build fails because babel complains about syntax errors in `package.json`

karma.conf.js
```diff
@@ -0,0 +1,60 @@
...
+      },
+      SLInternetExplorer11: {
+        base: 'SauceLabs',
+        browserName: 'internet explorer',
+        version: '11',
+        platform: 'Windows 10',
+      },
+    },
+    preprocessors: {
+      './src/**/*.test.js': ['rollup'],
```
FezVrasta: To make it easier to debug the problem, you can replace this with `./src/**/base.test.js` (also on line 52). In this way you can test just a single test.

mxstbr:
> @mxstbr you will have to subscribe to SauceLabs and email them asking for an open source account for this repository. If you need help feel free to email/DM me.

Done!

> We need to get rid of Enzyme and jsdom from tests

That's not optimal, as I remember we introduced `jsdom` for a reason that we couldn't otherwise test something, but I can't remember what that was...

package.json
```diff
@@ -14,7 +14,7 @@
     "prebuild:dist": "rm -rf dist/*",
     "build:dist": "rollup -c && rollup -c --environment PRODUCTION",
     "build:watch": "npm run build:lib -- --watch",
-    "test": "mocha \"./src/**/*.test.js\" --require babel-core/register --timeout 5000",
```
relekang: Could we keep running mocha in node as default and add `npm run test:browsers` for testing with browsers?

FezVrasta: If we find a way to disable enzyme and jsdom during the tests yes, otherwise they throw errors if ran inside a browser...

relekang: What about something like this?

```diff
diff --git a/src/test/theme.test.js b/src/test/theme.test.js
index 0516b64..e70c4c3 100644
--- a/src/test/theme.test.js
+++ b/src/test/theme.test.js
@@ -1,6 +1,5 @@
 // @flow
 import expect from 'expect'
-import jsdom from 'mocha-jsdom'
 import React from 'react'
 import { mount, render } from 'enzyme'
 
@@ -209,7 +208,10 @@ describe('theming', () => {
 })
 
 describe('theming (jsdom)', () => {
-  jsdom()
+  if (typeof document === 'undefined') {
+    require('mocha-jsdom')()
+  }
 
   beforeEach(() => {
     styled = resetStyled()
```

FezVrasta: I'll play with this idea tomorrow, thanks

geelen:
So, I've been thinking about the value of browser testing for this project a bit. I don't think it's particularly valuable to run our unit tests inside browsers — jsdom, Enzyme & full ES6 support make for happy, expressive testing that runs fast and everyone can run locally before check in. They're the things that are going to catch logic errors in the code.

Browser testing should be a replacement for manual verification that everything we say in the Readme remains possible on all the browsers we support. Bugs like #212, #344 and a bunch more, where we simply don't have the test coverage (or the time to do manual verification) for every new release we cut, and errors creep in.

So I think this PR can change a bit, and not try to run the existing tests in a new environment, but rather try to run the _built output_ against a test project and make a bunch of assertions as to an example project's output.

I had thought we could turn the `example` directory into a project that covered the whole public API, then build it with two webpack configs — one linking `styled-components` source directory as if it was installed from NPM, the other using the UMD build. Then, we'd have a list of assertions like:

```
- The webpack build succeeded
- There were no JS console errors
- Div X should have this exact CSS
- Div Y should have a blue background
```

That sort of thing. It's basically a "does our example project generate the exact output we expect on all browsers" test. We could potentially even wire it up to Jest's snapshots somehow. To me, that's really valuable, because it means we're not shipping broken builds, and if a PR needs to change the example or acceptance tests then maybe it's a breaking API change?

What do you think?

philpl:
This won't work anymore, since we're switching to Jest, I suspect.

Do we want to update this or think about it before, @geelen?

Sorry, @FezVrasta! I'll close this for now

Issue #149, opened 2016-10-29
Support for style-related props (not set via 'style') in React Native


adamterlson:
Presently in styled components, there is no way to affect other style-related props in React Native. 

This style-related-prop-not-defined-via-`style` can be found in many third party libraries (such as [React Native Navbar](https://github.com/react-native-community/react-native-navbar)'s `tintColor` prop), in custom components one might write themselves, and in React Native's core components, for example:
- [TouchableHighlight](https://facebook.github.io/react-native/docs/touchablehighlight.html) underlayColor
- [ActivityIndicator](https://facebook.github.io/react-native/docs/activityindicator.html) color and size
- [Slider](https://facebook.github.io/react-native/docs/slider.html) maximumTrackTintColor and minimumTrackTintColor
- [StatusBar](https://facebook.github.io/react-native/docs/statusbar.html) backgroundColor
- [TextInput](https://facebook.github.io/react-native/docs/textinput.html) placeholderTextColor, underlineColorAndroid and keyboardAppearance
- [Switch](https://facebook.github.io/react-native/docs/switch.html) onTintColor, thumbTintColor and tintColor
- ... and a lot more

It's of course possible to simply set these props manually, but styled-component's themes are unable to be leveraged to standardize them across the codebase.

irakli-janiashvili:
This definitely will be a great enhancement!

mxstbr:
Is there a list of the ones that exist somewhere? We should definitely add support for them!

adamterlson:
@mxstbr I'm unaware of any existing list.  The only option would be to go through every available component and manually catalog them

irakli-janiashvili:
I can volunteer to go through and get the list of those items.

mxstbr:
That would be amazing @irakli-janiashvili! 🙏

irakli-janiashvili:
@mxstbr @adamterlson

Here's the list of all possible attributes that can go in styles ->  https://gist.github.com/irakli-janiashvili/dac99f8ceb0c1a9bafa78aad0f19d861

I put a question marks in front of some attrs that I'm not sure we have to put in styles..

adamterlson:
It seems like a lot to maintain manually. 

Why not simply provide a way to generally set any prop desired?

mxstbr:
[...]

PR #365, opened 2017-01-11
"Attrs" constructor

### LINKED ISSUE #134: Managing highly dynamic styles

### LINKED ISSUE #184: Using styled-components with react-router
react-router's `<Link>` component and react-modal2's `<Modal>` component accept additional classNames via separate properties.  e.g. `<Link>` accepts `activeClassName` and `<Modal>` accepts `backdropClassName` and `modalClassName`.  It would be great to have a way to style these using styled-components.  Perhaps we could have something like:

```js
const StyledLink = styled(Link)`
  color: palevioletred;
  display: block;
  margin: 0.5em 0;
  font-family: Helvetica, Arial, sans-serif;

  &:hover {
    text-decoration: underline;
  }
  &.active {
    color: red;
  }
`;
```

where `&.active` gets passed as a separate class name to the `activeClassName` property.

### LINKED ISSUE #291: Proposal: CSS Class Composition

geelen:
Ok, so this is my proposal for #291, #134 and #184 (and probably others). I think it does what we need in a really nice way:

```js
const X = styled.div.attrs({
  /* add extra props here */
})`
  /* css here as normal */
`
```

So, @ticky's original code becomes:

```js
const Title = styled.h1.attrs({
  className: 'meow nya'
})`
  font-size: 1.5em;
  text-align: center;
  color: palevioletred;
`
```

And @ForbesLindesay's becomes:

```js
const activeClassName = btoa(Math.random())
const StyledLink = styled(Link).attrs({
  activeClassName
})`
  color: palevioletred;
  display: block;
  margin: 0.5em 0;
  font-family: Helvetica, Arial, sans-serif;

  &:hover {
    text-decoration: underline;
  }
  &.${activeClassName} {
    color: red;
  }
`
```

I've added a bit of smarts to make it more similar to how interpolations work:

```js
const Button = styled.button.attrs({
  type: props => props.submit ? 'submit' : 'button'
})`
  /* ... */
`

<Button> //#> <button type="button" class="...">
<Button submit> //#> <button type="submit" class="...">
```

It gets passed the same `{ ...this.props, theme }` object that interpolations do, and recalculates whenever the CSS snippet gets reevaluated. I.e. it works exactly the same, as much possible.

If you have a `prop` and an `attr` with the same key, the `prop` wins (in that sense they're like `defaultProps`). That is except for `className` which just adds everything together. I decided against trying to generalise it—adding extra classes is kind of a special case so I'm happy to keep it that way. If you want any other kind of weird manipulation of properties just make a stateless functional component.

I've also called it `attrs` because, well, I think `props` is too overloaded. And, at the end of the day you're rendering HTML elements most of the time. They have attributes, not props. And I see this sort of API most useful when working with these elements—if you have more complex objects you should define a real component. I'm not trying to reinvent React here, but being able to define attributes in the same way you generate and attach CSS turns out to work quite nicely.

The big exception to the previous paragraph is, of course, react-router's `Link`. It masquerades as a HTML element, but it's a component, so having a single syntax that preserved that facade was important to this. Still, I didn't think it was worth changing the name.

Oh also anything in `attrs` gets turned into a `prop` later. That's not very useful, except for some cases like this:

```js
const Image = styled.img.attrs({
  'alt': 'My Beautiful Face'
})`
  &::after {
    content: '${props => props.alt}';
  }
`
```

Happy to hear feedback, in particular if you can see this breaking down in certain circumstances (that don't just warrant adding an extra component). And everyone's reaction to the name, as well. Don't submit alternate syntaxes here though. That discussion is over at #291, and I'm _pretty_ sure we've considered every possible syntax already. If not, continue the conversation there.

Closes #291, closes #134 and closes #184.


mxstbr:
This is beautiful. Solves all of the concerns with one concise, lovely API! Let's see what everybody else has to say.

----

> I've added a bit of smarts to make it more similar to how interpolations work...

Please tell me that we can reuse that to make `props => {}` happen in interpolated objects?!

```JS
const Box = styled.div`${{
  background: props => props.bg,
}}`
```

geelen:
Oh shit just realised this solves #134 as well!

```js
const Box = styled.div.attrs({
  style: props => {
    background: `hsl(
      ${props.hue}, 
      ${ 50 + props.luminosity / 2}%, 
      ${props => props.luminosity}%
    )`
  }
})`
  margin: 2rem auto;
  width: 360px;
  height: 360px;
`
```

mxstbr:
Holy cow that's huge!

k15a:
Hey,

Do you have any solution how to solve https://github.com/styled-components/styled-components/issues/305 with `attrs`? Maybe don't pass `attrs` to the element which are false?

```JS
import FontAwesome from 'react-fontawesome';

export const Icon = styled(FontAwesome).attrs({
  active: false,
  color: false,
})`
  font-size: 25px;
  margin-bottom: 3px;
  ${props => props.active && `color: ${props.colors.brand}`}
`;
```

This won't work because of:
> Oh also anything in attrs gets turned into a prop later. That's not very useful, except for some cases like this:

brunolemos:
Interesting, I like it.

If it's not already, I suggest being able to pass string literals in the attributes like this:

```
const ScrollView = styled.ScrollView.attrs({
  contentContainerStyle: `
    background-color: red;
  `,
  style: `
    background-color: green;
  `
})
```

^ Also, I passed the `style` attribute, not sure if you handled this case.
I suggest merging the styles from the `style` attr with the styles from the default way to do it.

ForbesLindesay:
This is great.  The one thought I have is that:

```js
const activeClassName = btoa(Math.random())
const StyledLink = styled(Link).attrs({
  activeClassName
})`
  color: palevioletred;
  display: block;
  margin: 0.5em 0;
  font-family: Helvetica, Arial, sans-serif;

  &:hover {
    text-decoration: underline;
  }
  &.${activeClassName} {
    color: red;
  }
`;
```

still requires me to think a lot about how the internals work in order to cover this use case.  What I'd ideally like to get working is:

```js
const StyledLink = styled(Link).attrs({
  activeClassName: css`
    color: red;
  `
})`
  color: palevioletred;
  display: block;
  margin: 0.5em 0;
  font-family: Helvetica, Arial, sans-serif;

  &:hover {
    text-decoration: underline;
  }
`;
```

This gives me no need to think about how an active class name gets generated.  The only problem I see at the moment is that the return type of `css` is, I think, the actual string of `css`, so this would need to somehow know to convert that css into a `className`.

On an unrelated note, i wonder about calling it `withDefaultProps` rather than `attrs`.  I think overall we should prioritise clarity over succinctness.  A more descriptive name might make it more obvious what it does.

diegohaz:
@ForbesLindesay For this specific case, I don't think we need that complexity.
```js
const StyledLink = styled(Link).attrs({
  activeClassName: 'active'
})`
  color: palevioletred;
  display: block;
  margin: 0.5em 0;
  font-family: Helvetica, Arial, sans-serif;

  &:hover {
    text-decoration: underline;
  }
  &.active {
    color: red;
  }
`;
```

It would work perfectly (or am I missing something?)

diegohaz:
> On an unrelated note, i wonder about calling it withDefaultProps rather than attrs. I think overall we should prioritise clarity over succinctness. A more descriptive name might make it more obvious what it does.

Yeah, it seems too much like React's defaultProps. I didn't read everything yet, but the only difference I see is we are passing in actual props here (so you can set a prop based on actual props).

I'm afraid it stimulates people to use this instead of `defaultProps`, which should be used for most cases. The `image alt` case, for example, should use `defaultProps` instead of `attrs`, from my point of view.

ForbesLindesay:
> For this specific case, I don't think we need that complexity.

The advantage of my proposal is that it doesn't require me to know the internal structure of the `Link` component.  If I use `&.active { color: red; }`, I need to know that the `activeClassName` gets applied to the same element as the normal `className`.

If I can somehow tell styled-components to generate another, totally separate style, it doesn't matter what the internal structure is.

I'm not currently sure if this is worth trying to change/extend the API for though.  I may need to keep thinking about it/try out the existing API.

> I'm afraid it stimulates people to use this instead of `defaultProps`

I'm not sure if this is actually a problem.  Is there anything wrong with doing:

```js
const Image = styled.img.withDefaultProps({
  alt: 'My Beautiful Face',
})`
  &::after {
    content: '${props => props.alt}';
  }
`
```

ticky:
Nice!!! [Just had a play](https://github.com/buildkite/frontend/commit/0217415320fa96a4d27f0289fbccbe144975d1b9), this feels much, much cleaner!

diegohaz:
> The advantage of my proposal is that it doesn't require me to know the internal structure of the Link component. If I use &.active { color: red; }, I need to know that the activeClassName gets applied to the same element as the normal className.

If this is a problem, it's `Link`'s responsibility, not ours. This is how react-router works.

> I'm not sure if this is actually a problem.

What's the benefit of using a different API from another library (`styled-components`) if it does exactly the same as React's `defaultProps`? There're also some libraries that counts on that, such as [`react-docgen`](https://github.com/reactjs/react-docgen). I think we should not encourage it.

Actually, besides `className` (which I would prefer to have a separated `.className` constructor) and cases when you want to set the attribute based on props (like the `submit` example), I don't think `.attrs` should be used. Even the `Link` example should use React's `defaultProps`.

Just my 2c.

geelen:
Thanks for the discussion and the feedback. `withDefaultProps` is not a bad name, but it doesn't actually work the way `defaultProps` works—these are reevaluated every time `props` changes, and they can take functions. It's also kind of a react-specific name. Styled Components is wholly tied to React at the moment but only by implementation, not by design.

As for the `Link` discussion, I tend to agree with @diegohaz that it's part of `Link`'s API that it applies `activeClassName` in certain cases. I don't think it makes much sense to add css-generation-and-injection behaviour into the `attrs` API, they just feel like totally separate concerns.

However, I _could_ consider treating `css` blocks specially:

```js
/* React Native example */
const ScrollView = styled.ScrollView.attrs({
  contentContainerStyle: css`
    background-color: red;
  `
})`
    background-color: green;
`

/* inline style example */
const Box = styled.div.attrs({
  style: css`
    background: hsl(
      ${props => props.hue}, 
      ${props => 50 + props.luminosity / 2}%, 
      ${props => props.luminosity}%
    );
  `
})`
  margin: 2rem auto;
  width: 360px;
  height: 360px;
`
```

This would basically convert CSS snippets to objects, so no nesting or anything would be permitted, but it might be useful for these cases. Feels like this would be useful for #149 too.

geelen:
Ok so that previous comment isn't going to work just yet. It ends up needing to do something really similar to [css-to-react-native](https://github.com/styled-components/css-to-react-native) which at the moment depends on PostCSS, and we're just about to replace PostCSS with Stylis (#286).

I've opened #371 to track it. But otherwise I think we should merge this.

jason-henriksen:
Anything I could contribute / work on to help this along?
I'm getting buried in repetitive class names.  :)

kidwm:
will this feature be in v2?

mxstbr:
@kidwm yes.

bondz:
This should be available in `styled-components@2.0.0-4` right? I'm getting `_styledComponents2.default(...).attr is not a function`

mxstbr:
That's because @geelen forgot to export it 😉 PR?

bondz:
Oh, yes. 😀

mxstbr:
Wait sorry that's not right, this is working as intended. What's the code you're running?

bondz:
@mxstbr  that's weird, tried it in `webpackbin` now and it works. It was webpack using the old cache in my build, works now. Thanks.

bondz:
@mxstbr sorry to bug you, but it happened again, replicated it here http://www.webpackbin.com/V1_gYxlKf

k15a:
It's working for me at your link ;)

bondz:
@k15a that's weird, Firefox and Edge show the same thing. http://imgur.com/a/SIRFa

firefox
Edge

k15a:
Yeah I've noticed that I looked wrong that's why I've removed my comment :p

geelen:
Not sure what the problem is here but `attrs` shouldn't be being attached to the styled component itself—it's part of the constructor, not the component, so being attached [here](https://github.com/styled-components/styled-components/blob/v2/src/constructors/constructWithOptions.js#L16) is correct.

streamich:
@geelen Can it be done like this?

```jsx
const Button = styled(({className, type}) =>
  <input className={className} type={type ? 'submit' : 'button'} />
)`
  /* ... */
`
```

kitten:
@streamich this certainly works, but please don't comment on stale PRs and issues. Especially since this is a PR and you have a rather unrelated question. I don't mean to be rude, so my suggestion is to go to Spectrum, or Twitter, or at the last create an issue for your questions :wink:

PR #368, opened 2017-01-11
Fix #367 innerRef

### LINKED ISSUE #367: #310 breaks innerRef when used like styled(MyStyledComponent)
I believe #310 broke innerRef, as a lot of components now return undefined for the refs.

Example:

```
const StyledInput = styled.input`background: red`;
const Input = ({ ...otherProps }) => (
  <StyledInput {...otherProps} />
);
```

And using it in other file

```
import Input from '../components/Input';
const StyledInput = styled(Input)`background: blue`;

class InputGroup extends Component {
  render() {
    return (
      <div>
        <StyledInput innerRef={input => this.input1 = input} />
        <StyledInput innerRef={input => this.input2 = input} />
      </div>
    )
  }
}  
```

After the update `this.input1` and `this.input2` are undefined. Where as previously they returned input elements.

IljaDaderko:
Related issue #367 , hope this is correct way to determine these.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ All notable changes to this project will be documented in this file. If a contri
 - Fixed defaultProps theme overriding ThemeProvider theme, thanks to [@diegohaz](https://github.com/diegohaz). (see [#345](https://github.com/styled-components/styled-components/pull/345))
 - Removed custom flowtype supressor in favour of default $FlowFixMe [@relekang](https://github.com/relekang). (see [#335](https://github.com/styled-components/styled-components/pull/335))
 - Updated all demos to link to latest version [@relekang](https://github.com/relekang)
+- Only delete innerRef if it is being passed down to native / html elements.
+
 
 ## [v1.2.1]
 
--- a/README.md
+++ b/README.md
@@ -474,6 +474,7 @@ We could use your help to get syntax highlighting support to other editors! If y
 - [reshake](https://github.com/elrumordelaluz/reshake) : CSShake as a React Functional Component ([demo](https://elrumordelaluz.github.io/reshake/))
 - [space-exp](https://github.com/caspg/space-exp) : Brings you each day a stunning picture of our universe, Astronomy Picture of the Day (APOD). Website: [spaceexperience.club](https://spaceexperience.club/)
 - [sachagreif.com](http://sachagreif.com): personal homepage built with [Gatsby](https://github.com/gatsbyjs/gatsby) ([source](https://github.com/SachaG/sg2017)).
+- [last-draft](https://github.com/vacenz/last-draft): A Draft.js Editor ([demo](http://lastdraft.vace.nz))
 - [`Dirtyredz.com`](https://github.com/dirtyredz/dirtyredz.com): David McClain | Dirtyredz - About me, Latest projects and Contact ([Website](http://dirtyredz.com))
 
 *Built something with `styled-components`? Submit a PR and add it to this list!*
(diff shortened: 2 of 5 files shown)


IljaDaderko:
@diegohaz previous PR that this one fixes added in some: https://github.com/styled-components/styled-components/pull/310/files I'm not good with testing, but let me know if these are not enough and I'll try to wrap my head around them :)

IljaDaderko:
@mxstbr @diegohaz Just had a go at this, unfortunately I can't figure out tests for our specific use case, could you guys help out  with these? 2017 goal is to eventually get better with them 🌝

mxstbr:
So I tried writing a test verifying this behavior, but it doesn't work on `master` nor after this fix – so either my test is bad (possible) or the issue doesn't lay with `styled-components`.

In the React docs I found the following section for refs:

The react docs are saying "You may not use the ref attribute on functional components because they don't have instances."

*(source: https://facebook.github.io/react/docs/refs-and-the-dom.html)*

Are you sure that your example in the original issue worked before the breaking version? I think it might be a case of switching from class syntax to functional components, which refs don't get passed through on by React...

IljaDaderko:
@mxstbr We switched back to v1.1.3 (version prior to one that introduced this) and everything works fine there, no console warnings / errors from react either. This still could be the case of using functional component, true. I will test that latter in the evening today. However, innerRef's should still be passed down given that we are passing them to styled components no?

IljaDaderko:
@mxstbr Alright just gave this a test and it seems to still be an actual issue. Indeed that ref passing won't work in scenarios described in documentation link you provided even on 1.1.3, but that is not an issue in this case.

mxstbr:
> Alright just gave this a test and it seems to still be an actual issue.

Do you have some code that breaks? That'd make it much easier to write a test for it, because that means my test is wrong 😉

IljaDaderko:
@mxstbr one from the original issue #367 is essentially it, just stripped down.

mxstbr:
This is the test I have right now:

```
it('should handle inherited components correctly', () => {
	const StyledComp = styled.div``
	const WrappedStyledComp = (...props) => <StyledComp {...props} />
	const ChildComp = styled(WrappedStyledComp)``
	const WrapperComp = class extends Component {
		testRef: any;
		render() {
			return <ChildComp innerRef={(comp) => { this.testRef = comp }} />
		}
	}
	const wrapper = mount(<WrapperComp />)

	expect(wrapper.node.testRef).toExist()
	expect(wrapper.node.ref).toNotExist()
})
```

which fails both on `master` and with this patch because `wrapper.node.testRef` is `null`. Any ideas how we need to fix this test?

IljaDaderko:
Logic wise test looks correct to me, are we 100% sure that my check is correct in this patch and that we are deleting correct things?

mxstbr:
Oh I assumed you tried your patch in your project locally and confirmed it fixed the issue! Let me dig in further then.

mxstbr:
Now I know the issue, you almost had it. You also checked if the passed in component wasn't a styled component, but we actually also want to pass it to normal react components! Changing the checks to `isTag()` only fixes everything.

Thank god we finally got this, thanks for helping me debug this so far. Pushing up the test and my changes now, give me a minute...

IljaDaderko:
Awesome! Sorry for a bit off a mess created, this was actually my first time trying to contribute to an oss project 😳

mxstbr:
Oh wow, congratulations 🎉 You did really well! Sorry for the huge troubles in landing this, I wish your first experience contributing code was faster. Next time though!

> Nitpick: it definitely wasn't your first _contribution_. Your input in the discussions around the repo has been invaluable. Don't discredit those, just because you haven't written any code for the project doesn't mean you haven't contributed anything—the opposite, actually!

jesnil01:
Big thanks for this!

PR #371, opened 2017-01-12
Parse CSS blocks to objects in attrs

geelen:
Added a failing test case for this behaviour. If someone could pick this up, that would be ace.


mxstbr:
Welp, doesn't look like anybody picked this up. Going to close it in order to make the repo a bit more tidy—if you're reading this and you are interested in picking this up let us know and we'll reopen it!

PR #380, opened 2017-01-17
Updated deps

amilajack:


amilajack:
Should the `src/vendor` dir be ignored by eslint?

amilajack:
@mxstbr What default should be used to fix this error?

https://ci.appveyor.com/project/mxstbr/styled-components/build/963#L1420

mxstbr:
> /home/travis/build/styled-components/styled-components/src/models/ThemeProvider.js
>  14:3  error  propType "children" is not required, but has no corresponding defaultProp declaration  react/require-default-props

`children` should actually be required, a `ThemeProvider` without children doesn't make any sense, does it?

mxstbr:
Thanks for this PR btw, this is awesome!

mxstbr:
Wait why'd we never merge this? 😱

amilajack:
I think it would be better to wait until the jest migration has finished. There'll probably be a lot of merge conflicts if we don't

mxstbr:
Good call, as always 👌

PR #396, opened 2017-01-21
adds simple ssr example

GGAlanSmithee:
This adds a super simple ssr example. Not sure if this is something you want, but I was testing out so I figured i might as well send it as a PR since there is no docs/examples for this yet AFAICT.

_This needs to be run with `babel-node example/ssr`_


GGAlanSmithee:
@mxstbr done

PR #406, opened 2017-01-25
[Breaking] Update css-to-react-native

jacobp100:
Brings in a rewrite of `css-to-react-native`, which has a few breaking changes:

* You must use `px` for non-zero lengths (`margin: 50px 20px 0`)
* You must use quotes for `font-family` (`font-family: "Georgia"`)

We have a [codemod](https://github.com/styled-components/styled-components-native-code-mod) to do most of this, but you’ll have to fix interpolated values manually.

**Testing.** We need people to test for any major regressions, spec inconsistencies, and also the codemod!

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   "homepage": "https://styled-components.com",
   "dependencies": {
     "buffer": "^5.0.2",
-    "css-to-react-native": "^1.0.6",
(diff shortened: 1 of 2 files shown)


mxstbr:
Can you submit a new, non-conflicted PR so we can merge this and get some feedback? Sorry for the delay!

PR #408, opened 2017-01-25
Fix 1.4.0 breaking native require

IljaDaderko:
Took me a while to find, but I believe this will fix following errors in react native that occurred after updating to 1.4.0

```
Cannot read property 'indexOf' of undefined
Module AppRegistry is not a registered callable module (calling runApplication)
```

Note: it seems like with every new release something breaks in react-naive, is it not tested well compared to styled components for react?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file. If a contri
 ### Added
 
 ### Changed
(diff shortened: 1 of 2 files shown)


IljaDaderko:
Issue still persists after this, I was confused as lib folder wasn't present in master, but it seems to be fine in node modules,

Issue #400, opened 2017-01-23
Browser rerenders for every injection


saschagehlich:
Hey there,

we've been using styled-components in a new project for two weeks now, and so far it's a great experience. But as our application gets larger and more complex, we start running into performance issues, which I'd like to talk about here.

Please see the following video: https://vid.me/5rlI

### Problem 1: Re-rendering the whole application when rules are added

I've enabled paint flashing, so you see which parts of the screen are re-rendered. As soon as new rules are added to the <style> tag, the browser has to re-parse the whole CSS and re-render the whole screen. This happens for example when moving the cursor to the navigation bar, but it only happens the first time, when the CSS rule is created. Subsequent hovering works as expected.

### Problem 2: Reloading web-fonts, causing text to flicker

As you see in the networks tab, web fonts are re-loaded when new rules are added. This is happening, because styled-components adds new CSS rules to the same <style> tag as our global font-face rules. Reloading the font causes all texts to flicker and re-render.

Is this being worked on? Is this a problem with my application or is this actually a problem with styled-components? For me it looks like a styled-components issue and a generic issue with changing the contents of a global <style> tag.

diegohaz:
Related #238

mxstbr:
This is interesting, thanks for kicking off this discussion! Just to confirm, you're running into actual performance issues with your app where `styled-components` is the bottleneck and creates a worse user experience?

I remember @geelen did a bunch of benchmarking around the injection mechanism to make sure it's as performant as possible, I'd be interested to see if we can maybe fix this by editing that? Any ideas Glen?

saschagehlich:
Yes, `styled-component` is the reason why we're experiencing flickering and performance issues. It's probably due to styled-components' concept of dynamic styles inside <style> tags - updating rules that are at the beginning of a <style> tag might sometimes case following rules to re-apply and re-render the whole screen.

mxstbr:
That should definitely not be happening and sounds like a bug on our side. Let's get this resolved asap!

k15a:
I think a solution would be to inject every newly generated style in a seperate style tag?

xcoderzach:
Problem 1:
I think the solution here might be to enable the "speedy" option on the glamor sheet in "production", so that classes get added via `insertRule`, rather than string manipulation.

Problem 2: 
Given that `injectGlobal` should only be called at initialization time, the solution here is probably to create a single separate sheet for global styles.

geelen:
[...]

PR #415, opened 2017-01-26
Put global styles in a separate stylesheet.

xcoderzach:
Putting dynamic styled component css rules into the same stylesheet
as global styles can cause global rules such as `@font-face` to be
recomputed.  Instead, create two stylesheets, one for global
styles, and one for styled-components.

Fixes part of #400.

### DIFF
--- a/src/models/GlobalStyleSheet.js
+++ b/src/models/GlobalStyleSheet.js
@@ -1,0 +1,8 @@
+// @flow
+
+/* Wraps glamor's stylesheet and exports a singleton for global styles. */
+import { StyleSheet } from '../vendor/glamor/sheet'
+
+/* Don't specify a maxLength, since these rules are defined at initialization
+*  and should remain static after that */
+export default new StyleSheet({ speedy: false })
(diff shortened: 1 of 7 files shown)


src/models/GlobalStyleSheet.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+
+/* Wraps glamor's stylesheet and exports a singleton for global styles. */
+import { StyleSheet } from '../vendor/glamor/sheet'
+
+/* Don't specify a maxLength, since these rules are defined at initialization
+*  and should remain static after that */
+export default new StyleSheet({ speedy: false })
```
mxstbr: Can we set speedy to true here?

xcoderzach: I don't think we want to do that, because `speedy: true` apparently breaks the chrome dev tools.  

https://github.com/threepointone/glamor/blob/master/src/sheet.js#L70 

So if we do enable it, it should probably be an optional "production only" thing.

geelen: Yeah leave this false for now. Inserting rules is faster but they're not slow atm—the problem with #400 is invalidation, not injection. And not being able to debug CSS in production is irritating af.

mxstbr: Well then let's include a CHANGELOG entry and ship this!

PR #417, opened 2017-01-26
css`something` returns commas

kristojorg:
I was trying to replicate the example [here](https://github.com/styled-components/styled-components/blob/master/docs/tips-and-tricks.md#more-powerful-example) and it wasn’t working.

I wasn’t getting any error messages, but had empty styles. 

It looks like there is incorrect css being generated when you use one of these functions alongside other styles or when there are newlines in the template string. 

Not quite sure how to fix, but could look into it if you give me some direction.

### DIFF
--- a/src/constructors/test/css.test.js
+++ b/src/constructors/test/css.test.js
@@ -9,6 +9,14 @@ describe('css', () => {
     const expected = `There should not be commas around me`
     expect(cssResult).toEqual(expected)
   })
+  it('properly handles nested functions with returns', () => {
+    const func = (...args) => css`
+      ${css(...args)}
+    `;
+    const cssResult = `${func`There should not be commas around me`}`;
+    const expected = `There should not be commas around me`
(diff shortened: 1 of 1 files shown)


geelen:
Hmm, I'm not quite sure what the failure is here. Could you show an example of the commas on a webpackbin link? I _think_ you might be passing the output of `css` (which is an array) to a non-tagged template literal, which will call `toString` on the array and inject commas, but I won't know for sure unless I can see a failing example.

kristojorg:
Well after testing around with webpackbin for a while, I really don't know what my problem was! Sorry for the unnecessary issue!

pavel06081991:
I have the same issue. When I use mixin like this:
```javascript
breakpoints.sm`background: dodgerblue;`
```

I get such styles with comma at the beginning:

```css
`@media (max-width:500px){.bTEyzC{,background:dodgerblue;}}`
```

Issue #426, opened 2017-01-29
TS issue: cannot complete styled.div


minamorl:
I'm making PR to fix this. This is just for notification.

## Version
1.4.2

## Reproduction
http://www.webpackbin.com/Ek8DrVsRZ

## Steps to reproduce
Editor: latest VSCode
```
import * as styled from 'styled-components'
styled.div // error
```

## Expected Behavior
Completions work correctly.

## Actual Behavior
Only interfaces appears.

minamorl:
Plus current implementation has no namespace separation so i will fix up them too

minamorl:
I thought this was the typing issue but it's totally my fault.
Using
```
import styled ...
```
instead of
```
import * as styled 
```
just works. I close this issue

PR #427, opened 2017-01-29
[WIP] Fix typescript definition

minamorl:
Related to #426 

- [x] Fix typings issue
- [ ] Separate all into namespace
- [ ] Export namespace (need help)

I add some bug fixes and separate it into its own namespace.
Also, I created interface `StyledComponentStatic` extending `StyledInterface`. The reason is `styled` is defined as default export. But I think there is better solution for this. If there is better solution to handle `styled` function (which is exported as default), let me know.

### DIFF
--- a/typings/styled-components.d.ts
+++ b/typings/styled-components.d.ts
@@ -1,179 +1,182 @@
 import * as React from "react";
 import { StatelessComponent, ComponentClass } from "react";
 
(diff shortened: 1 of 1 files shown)


minamorl:
Sorry, now I think namespace is no necessary so I'll revert this.

minamorl:
I have to reconsider about a cause of the problem so now I temporary close this PR

PR #428, opened 2017-01-29
TypeScript: Improve typings of interpolations

Igorbek:
This PR slightly improves TypeScript support (typings) of interpolations.

### Rationale

There're two very different contexts where styles are being created: global (`injectGlobal`, `keyframes` functions) and component (`styled` function). In the global contexts, functions as interpolations are not accepted. Indeed, there's nothing to use - no props available.
So the typings of these functions already make this distinctions, `injectGlobal` and `keyframes` do not accept functions as interpolations.
The issue comes up when `css` needs to be used. Since it is meant to be composable and can be used inside both contexts, we need to know what types of interpolations are hidden inside.

### Changes

- an overload of `css` has been introduced which only takes simple interpolations
- introduced 'flatten' interpolation type to use as a result of `css` (which flattens passed array interpolations)
- also made possible to pass arrays as interpolations which wasn't working before
  - Note, that due to TypeScript's circular-reference restrictions, deeper nesting of arrays isn't allowed

Now using function interpolations that came from `css` is also restricted:
css-with-function-interpolation

### Upcoming

I'm going to make another PR with better theme support in TypeScript. So you could bundle all the changes in the next release.

### DIFF
--- a/typings/styled-components-test.tsx
+++ b/typings/styled-components-test.tsx
@@ -110,3 +110,42 @@ class Example extends React.Component<{}, {}> {
     </ThemeProvider>;
   }
 }
+
+// css which only uses simple interpolations without functions
+const cssWithValues1 = css`
+  font-size: ${14}${'pt'};
+`;
+// css which uses other simple interpolations without functions
+const cssWithValues2 = css`
+  ${cssWithValues1}
+  ${[cssWithValues1, cssWithValues1]}
+  font-weight: ${'bold'};
+`;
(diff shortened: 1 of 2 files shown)


mxstbr:
Wow this is awesome. I'll leave it up to @patrick91 to review the code, but this sounds like a useful change to me!

patrick91:
LGTM! :)

Igorbek:
@mxstbr yeah, let's bundle them together.

PR #432, opened 2017-01-30
TypeScript: Add support of typed themes

Igorbek:
Ok, this follows #428 but is a bit more complex.

The issue with themes in TypeScript is that they are not strongly typed. When a component is defined on the application level, it just assumes that theme exists with a specific structure:

```ts
const Button = styled.button`
  color: ${p => p.theme.primaryColor};
`;
```

Currently, the `theme` property just has type `any` which just allows any structure.
On other hand, allowing passing a theme type at the call-site is not really desirable:

```ts
//const Button = styled.button<MyTheme>` // currently is not a proper TS syntax
const Button = styled<{}, MyTheme>("button")`
  color: ${p => p.theme.primaryColor};
`;
```

Note, first `{}` is an augmented props type parameter, second is an actual theme type.
I think that is really cucumbersome.

My approach is to allow define an app-specific theme in app module and then re-export the `styled-components` with the type of app theme.

It's going to be like this:

##### themed-styled-components.ts
```ts
import * as styledComponents from "styled-components";
import { ThemedStyledComponentsModule } from "styled-components";

// Definition of the app-specific theme
export interface MyTheme {
    primaryColor: string;
}

const {
    default: styled,
    css,
    injectGlobal,
    keyframes,
    ThemeProvider
} = styledComponents as ThemedStyledComponentsModule<MyTheme>; // cast to "themed" module

// re-export
export default styled;
export { css, injectGlobal, keyframes, ThemeProvider };
```
##### components.ts

```ts
import styled from "./themed-styled-components"; // import app module

// now, type of the theme can be used easily
export const Button = styled.button`
  color: ${p => p.theme.primaryColor};
`;
```

### Intellisense

themed-intellisense

### Changes

- introduce `Themed*` interfaces with theme type parameter
- realias old interfaces to use `Themed*` and `any` as a theme type
- add definition of a themed module
- add tests/examples

### Breaking change

The `InterpolationFunction<P>` is changed to simply pass `props: P` as an argument instead of previously defined `props: StyledProps<P>`. Therefore, all `Interpolation*<P>` also change meaning. This made to unify usage of interpolations of themed and non-themed variations. This breaking change can only effects these users who import interpolation types and reuse them somewhere. I assess that the number of such users is extremely low :) Technically, that breaking change could be avoided by introducing other types of interpolations with theme type parameter, but this seems to me an overkill.

/cc @mxstbr @patrick91

### DIFF
--- a/typings/styled-components.d.ts
+++ b/typings/styled-components.d.ts
@@ -3,180 +3,205 @@ import { StatelessComponent, ComponentClass } from "react";
 
 type Component<P> = ComponentClass<P> | StatelessComponent<P>;
 
-export interface ThemeProps {
-  theme: any;
+export interface ThemeProps<T> {
+  theme: T;
 }
 
-export type StyledProps<P> = P & ThemeProps;
-export type OuterStyledProps<P> = P & {
-  theme?: Object;
+export type ThemedStyledProps<P, T> = P & ThemeProps<T>;
+export type StyledProps<P> = ThemedStyledProps<P, any>;
+
+export type ThemedOuterStyledProps<P, T> = P & {
+  theme?: T;
   innerRef?: (instance: any) => void;
 };
+export type OuterStyledProps<P> = ThemedOuterStyledProps<P, any>;
 
 export type Interpolation<P> = FlattenInterpolation<P> | ReadonlyArray<FlattenInterpolation<P> | ReadonlyArray<FlattenInterpolation<P>>>;
 export type FlattenInterpolation<P> = InterpolationValue | InterpolationFunction<P>;
 export type InterpolationValue = string | number;
 export type SimpleInterpolation = InterpolationValue | ReadonlyArray<InterpolationValue | ReadonlyArray<InterpolationValue>>;
 export interface InterpolationFunction<P> {
-  (props: StyledProps<P>): Interpolation<P>;
+  (props: P): Interpolation<P>;
 }
 
-export interface StyledFunction<P> {
-  (strings: TemplateStringsArray, ...interpolations: Interpolation<P>[]): ComponentClass<OuterStyledProps<P>>;
-  <U>(strings: TemplateStringsArray, ...interpolations: Interpolation<P & U>[]): ComponentClass<OuterStyledProps<P & U>>;
+export interface ThemedStyledFunction<P, T> {
+  (strings: TemplateStringsArray, ...interpolations: Interpolation<ThemedStyledProps<P, T>>[]): ComponentClass<ThemedOuterStyledProps<P, T>>;
+  <U>(strings: TemplateStringsArray, ...interpolations: Interpolation<ThemedStyledProps<P & U, T>>[]): ComponentClass<ThemedOuterStyledProps<P & U, T>>;
 }
+export type StyledFunction<P> = ThemedStyledFunction<P, any>;
 
-export type HtmlStyledFunction<E> = StyledFunction<React.HTMLProps<E>>;
-export type SvgStyledFunction<E extends SVGElement> = StyledFunction<React.SVGAttributes<E>>;
+export type ThemedHtmlStyledFunction<E, T> = ThemedStyledFunction<React.HTMLProps<E>, T>;
+export type HtmlStyledFunction<E> = ThemedHtmlStyledFunction<E, any>;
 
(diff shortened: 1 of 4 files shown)


patrick91:
That's really interesting, I think we can get some good feedback from @DanielRosenwasser :)

@mxstbr maybe we should add @Igorbek to the typers team :)

Igorbek:
@patrick91 yeah 👍  it'd be great to get some feedback from a TS member.

mxstbr:
> maybe we should add @Igorbek to the typers team :)

Oh yeah good idea, invited! :blush: Welcome to the team.

Igorbek:
Thank you @mxstbr @patrick91 !

Igorbek:
Formally requesting review from @styled-components/typers and from @DanielRosenwasser if possible :)

patrick91:
@Igorbek I feel like it is a bit verbose, but I can't find any other solution to it.

Maybe we can leverage [Mapped Types](https://github.com/Microsoft/TypeScript/pull/12114) somehow?

Igorbek:
@patrick91 I don't think Mapped Types are suitable here. Do you have something in your mind? What exactly seems verbose to you? Usage or definitions? Usage of unthemed components hasn't changed. Usage of themed components requires a bit boilerplate, but it seems acceptable. Definitions are still pretty compact without much duplicates, but now uses "themed" variant as a primary source of the types.

patrick91:
Yes, just the boilerplate :) I was wondering if we could have done something more magic,
like just extending the ThemeInterface, but I don't think that's something TS allows to do in the way we'd like to.

This could be helpful, but I didn't have time to read it properly: https://www.typescriptlang.org/docs/handbook/declaration-merging.html

Igorbek:
Declaration merging has nothing to do with this. So we have 2 different versions of `styled-components` where first is just relaxes `theme` type to `any`, other is strictly tied to a concrete theme type. They just cannot be merged, because `any` would poison the typing.

patrick91:
They idea was more to merge the declaration of the whole module, like in this example: http://stackoverflow.com/questions/37505881/extending-typescript-declaration-files

But we are not using declare/namespaces so that won't work :)

Igorbek:
@patrick91 do you or anyone have any concerns or I can just go ahead and merge?

typings/styled-components.d.ts
```diff
@@ -3,180 +3,205 @@ import { StatelessComponent, ComponentClass } from "react";
...
 };
+export type OuterStyledProps<P> = ThemedOuterStyledProps<P, any>;
 
 export type Interpolation<P> = FlattenInterpolation<P> | ReadonlyArray<FlattenInterpolation<P> | ReadonlyArray<FlattenInterpolation<P>>>;
 export type FlattenInterpolation<P> = InterpolationValue | InterpolationFunction<P>;
 export type InterpolationValue = string | number;
 export type SimpleInterpolation = InterpolationValue | ReadonlyArray<InterpolationValue | ReadonlyArray<InterpolationValue>>;
 export interface InterpolationFunction<P> {
-  (props: StyledProps<P>): Interpolation<P>;
+  (props: P): Interpolation<P>;
```
DanielRosenwasser: I don't know how this library works at all, but is the idea that an interpolation function can in turn return another interpolation function? Is that a common use case?

Igorbek: It is fully allowed. Here's repro: http://www.webpackbin.com/EJvhXeTDz
However, it doesn't seem to be a common use case.

patrick91:
@Igorbek fine to merge :)

Igorbek:
@mxstbr both this and #428 are ready to be shipped in the next release.

mxstbr:
Awesome, shipping!

goodmind:
@Igorbek why you can't use module augmentation here?

Igorbek:
@goodmind how would you suggest to use module augmentation here? From my understanding, it is used mostly when a library do monkey-patching of existing modules. I don't see that is the case here.

Igorbek:
The only possibility could be to define theme interface as `{}` and allow library consumer to augment it with new members. This would be to restrictive for other regular users when they want to see default theme as just `any`.

philpl:
@goodmind @Igorbek this probably belongs into a new issue if it's worth pursuing / discussing :)

Issue #433, opened 2017-01-30
[v2] data urls in css break parsing


wmertens:
## Version
@next, so recent v2.0. I can't post a reproduction, webpackbin just hangs when trying to add `styled-components@next` or `@2.0`.

## Steps to reproduce
```js
styled.div`
    background-image: url(data:image/png;etc); /* quoting the url doesn't help */
`
```
## Expected Behavior
it works like in v1

## Actual Behavior
it breaks:
```css
.gMVerY {
    background-image: url(data:image/png;}.other-component-1hm8ero {}.kWlwJm {position: absol…: 3px;left: -300px;width: 1290px;background-color: rgba(120, 120, 120, 0.8);
}
```

mxstbr:
This is exactly why we need people to get to try v2 before we release it. 👍  @thysultan this sounds like a bug in stylis, could you look into that?

wmertens:
@mxstbr:  @thysultan already fixed it in  v1.0.1 :) Can you upgrade?

mxstbr:
Ohh nice 👌 Sure, PR welcome!

wmertens:
@mxstbr I tried installing stylis v1.0.1 in SC's node_modules folder but that didn't fix matters - could there be something else at play?

wmertens:
Hmm, the es build actually already includes stylis, which is why it doesn't use the new one. Building a fresh one does indeed fix the problem, but perhaps the es build should not package deps

wmertens:
Closing since #436 fixed it.

Andreyco:
As I reached you in the morning at Twitter, I can confirm this. Currently, v2 does not parse style definition when data url is in place.

PR #434, opened 2017-01-30
Explain how to use v2

wmertens:


mxstbr:
I'm not sure this is worth putting in the README, as it'll be out soon anyway and it's currently too buggy (see #433) to tell everybody to use.

Let's leave this open until we've tried it internally on a few projects so we can find the gravest bugs. (like #433)

axelson:
@mxstbr is there a place to collect feedback about v2? I've used in on a small simple internal project and it appears to be working great.

GGAlanSmithee:
I am also using v2 with ssr and it's working great so far (with the exception of the babel transform, which made react re-render server -> client)

mxstbr:
> @mxstbr is there a place to collect feedback about v2?
> I am also using v2 with ssr and it's working great so far

Good idea, just opened #453, please comment!

garetmckinley:
Is there anything written that explains whats new/different in 2.0? I couldn't find release notes

k15a:
@garetmckinley The main differences are a new parser #286 and classnames are now based on a per component basis #227 which gives you support for a component selector #142. It's a major update because the new parser and classnames could potentially break styled-components for some users.

kopax:
> could potentially break styled-components for some users
@k15a Could you elaborate a bit please ? I am experiencing issue with the V2 and I have no clue. #453

philpl:
We're tracking v2 as a GH milestone here: https://github.com/styled-components/styled-components/milestone/5

It should be out soon. And by soon, this time we mean **soon** 😆

PR #436, opened 2017-01-31
v2: Upgrade stylis, don't bundle external dependencies

### LINKED ISSUE #433: [v2] data urls in css break parsing

wmertens:
I may be wrong on the bundling thing :)

Fixes #433 

Update: I removed a commit that made all dependencies externals for rollup, that was wrong. 

Instead, I need to convince Webpack to not use the ES build, since it includes bundled deps that should not be bundled when building an app. Perhaps the ES build should not be there at all, browser can use the normal build and for ES you just point to an ES6 transpilation?

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "inline-style-prefixer": "^2.0.5",
     "is-function": "^1.0.1",
     "is-plain-object": "^2.0.1",
-    "stylis": "0.11.0",
+    "stylis": "^1.0.1",
     "supports-color": "^3.1.2"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


mxstbr:
> Perhaps the ES build should not be there at all, browser can use the normal build and for ES you just point to an ES6 transpilation?

We need an ES build so people can use tree shaking! (doesn't work with CommonJS/UMD builds)

Thanks for the PR, will release a new pre-release!

PR #449, opened 2017-02-01
Add name property as possible name for component

### LINKED ISSUE #448: Styled(undefined) in React Developer Tools when wrapping Link component
## Version
2.0.0-1

## Reproduction
http://www.webpackbin.com/EygFtNoDf

## Steps to reproduce
Wrap a `<Link>` component (react-router 4.0.0-alpha.6) as `styled(Link)`.

## Expected Behavior
In React Developer Tools it should be named `Styled(Link)`.

## Actual Behavior
It is named `Styled(undefined)`.

k15a:
Closes #448

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ All notable changes to this project will be documented in this file. If a contri
 - Removed custom flowtype supressor in favour of default $FlowFixMe [@relekang](https://github.com/relekang). (see [#335](https://github.com/styled-components/styled-components/pull/335))
 - Updated all demos to link to latest version [@relekang](https://github.com/relekang)
 - Fixed SSR when no `styled-components` are rendered [@kristojorg](https://github.com/kristojorg). (see [#124](https://github.com/styled-components/styled-components/issues/124))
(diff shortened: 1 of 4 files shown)


k15a:
Hey @mxstbr,

I am happy to add some tests. Could you say me where the tests for this should be located? Couldn't find a similar test.

mxstbr:
What about `src/test/basic.test.js`? We don't verify this behaviour at all at the moment, it'd be great to get all of it tested!

k15a:
I've added some tests and changed the displayName to `Styled(Component)` instead of `Styled(undefined)`

mxstbr:
Oh wait, we still need a CHANGELOG entry 😊

k15a:
Mhh. That's not really a change because it was this way in v1. But I've written a little changelog entry to make your bot happy 😊

mxstbr:
> That's not really a change because it was this way in v1.

Wait, why do we have to fix this then? Is this already fixed on the `master` branch?

k15a:
Ohh sorry that was my fault. It's unfixed in master.

mxstbr:
Let's fix it on `master` then, and when we merge `master` back into `v2` (_cough_ @xcoderzach) this fix will propagate. That we avoid getting out of sync between the branches :blush:

PR #450, opened 2017-02-01
Fix defaultProps used instead ThemeProvider after initial render

### LINKED ISSUE #445: StyledComponent uses defaultProps for theme instead of theme provided by ThemeProvider after initial render
## Version
1.4.2

## Reproduction
http://www.webpackbin.com/4Jdl6JjwM

## Steps to reproduce

Click the button at least once and the click count text will turn from blue (the given value for `theme.color`) to red (the default prop for `theme.color`).

## Expected Behavior

`StyledComponent`s that are children of `ThemeProvider` should always use the provided `props.theme` instead of `defaultProps.theme`.

## Actual Behavior

On initial render `props.theme` is correct, every subsequent render `props.theme` is `defaultProps.theme`.

k15a:
Closes #445

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file. If a contri
 
 ### Changed
 
+- Fix defaultProps used instead of ThemeProvider on first render [@k15a](https://github.com/k15a). (see [#445](https://github.com/styled-components/styled-components/issues/445) and [#450](https://github.com/styled-components/styled-components/pull/450))
+
 ## [v1.4.2] - 2017-01-28
 
 ### Changed
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -49,9 +49,10 @@ export default (ComponentStyle: Function) => {
           const subscribe = this.context[CHANNEL]
           this.unsubscribe = subscribe(nextTheme => {
             // This will be called once immediately
-            const { defaultProps } = this.constructor
+
             // Props should take precedence over ThemeProvider, which should take precedence over
             // defaultProps, but React automatically puts defaultProps on props.
+            const { defaultProps } = this.constructor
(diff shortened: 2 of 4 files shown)


k15a:
@chadwatson Could you check out if this is working for you?

chadwatson:
Sure thing. Give me a few minutes.

chadwatson:
Yes, that fixes the problem! It is still double rendering the child styled component though. Is that a problem?

k15a:
It's double rendering because you're passing a new object every time and ThemeProvider is checking the reference. In your example they are different so ThemeProvider thinks it is a new Theme.

k15a:
@diegohaz Do you how chadwatsons example is different than the test you introduced in [#345](https://github.com/styled-components/styled-components/pull/345/files#diff-e9e543c50e5cefa6f2c6469f6abccc5b)? I am not seeing any difference.

diegohaz:
@k15a The difference is that he uses `setState` after the first render, right?

xcoderzach:
@k15a I think it's because the props change, in this specific case, it's the `children` prop (the count).

xcoderzach:
Any time the props on the styled component change without re-mounting this should happen.

xcoderzach:
Probably something fix to do after this gets landed, but it feels a bit precarious to have this same logic duplicated 4x. 😬

k15a:
@xcoderzach I wasn't thinking of the children prop. Added tests now.

k15a:
> Probably something to fix after this gets landed, but it feels a bit precarious to have this same logic duplicated 4x. 😬

Basically the whole StyledNativeComponent.js is a duplicate of StyledComponent.js. But it feels wrong anyway^^

diegohaz:
@k15a Can you confirm if that test fails without the changes on this PR?

k15a:
Yeah it's failing without the changes.

PR #457, opened 2017-02-05
[Typescript] Fix ThemedStyledInterface

patrick91:
This small fix allows to not declare the theme prop to custom components when reexporting the styled function with a custom theme interface.

```ts
import * as React from "react";

import styled, { ThemeInterface } from "my-themed-components";

interface BoxProps {
    borderBottom?: boolean;
    className?: string;
}

const Box:React.StatelessComponent<BoxProps> = (props) =>
    <div className={props.className}>{props.children}</div>;

const StyledBox = styled(Box)`
    padding: ${props => props.theme.lateralPadding}; // <-- this gets the correct interface now
    ${props => props.borderBottom && `
        border-bottom: 1px solid #ccc;
    `}
`;

export default StyledBox;
```

/cc @Igorbek 

PS. I'll add some docs for the re-exporting styled components thing after lunch :)

### DIFF
--- a/typings/styled-components-test.tsx
+++ b/typings/styled-components-test.tsx
@@ -113,17 +113,17 @@ class Example extends React.Component<{}, {}> {
 
 // css which only uses simple interpolations without functions
 const cssWithValues1 = css`
-  font-size: ${14}${'pt'};
+  font-size: ${14}${"pt"};
 `;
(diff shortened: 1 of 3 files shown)


patrick91:
No worries! Thanks for checking :)

PR #464, opened 2017-02-06
Exclude test files from npm publish

Amwam:
Noticed that test files were being published to npm, which seems unnecessary to me.


mxstbr:
I think there might have been a reason we published them. @geelen any ideas?

geelen:
Nah the only thing I wanted to see published was the ES6 source files, test files I'm happy to leave off NPM.

PR #473, opened 2017-02-08
Added two packages to the Built With section of the README

jameslnewell:
Hello! 👋 

I've been experimenting with `styled-components` in a few projects. None of the grids I looked into supported setting different prop values at specific breakpoints (other than width). So I've started porting my old `SASS` packages to `styled-componnents`. I've shared them on npm to make it easier for me to use across multiple projects, and in-case anyone else might find them useful.

I'd love to add them to the `Built With` list. I'd also love some feedback on the components themselves if anyone has any spare time.

https://github.com/jameslnewell/styled-components-breakpoint
https://github.com/jameslnewell/styled-components-grid

### DIFF
--- a/README.md
+++ b/README.md
@@ -463,8 +463,6 @@ We could use your help to get syntax highlighting support to other editors! If y
 
 ## Built with `styled-components`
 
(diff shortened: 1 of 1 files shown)


mxstbr:
Hey, that's awesome!!

Mind moving them to the bottom to be fair with all the other packages? Thanks 😊

Issue #325, opened 2016-12-20
PR: Migration to Jest


amilajack:
Would a migration to Jest be welcome?

mxstbr:
Yes, please!!

rolfea:
Hi @amilajack, are you still working on this migration? I'd be interested in helping out. 
I'm new to Jest and styled-components and interested in digging into the source code

amilajack:
I actually forgot to do this! Thanks for the reminder! And yea, sure. I'll create a branch on my fork that you can commit to

rolfea:
@amilajack is that your fix-ci-config branch?

amilajack:
I actually didn't create a branch yet because I wanted to merge the updates branch first.

https://github.com/styled-components/styled-components/pull/380

Turns out that there were a number of complications that prevent it from moving forward quickly. I'll create  new branch and let you know about it

jacobp100:
Happy to help with this if needed!

mxstbr:
Yes?! We want this!

amilajack:
@jacobp100 I'm running low on time these days so i haven't been able to implement this. Would be great if you could give it a go!

jacobp100:
Will do!

mxstbr:
v2 I guess?

PR #474, opened 2017-02-08
[DO NOT MERGE] Add benchmarks

jacobp100:
I used Jest because I need to mock react-native for this. Possibly we could switch all tests to Jest, or figure out how to switch the benchmark to Mocha at a later date.

Note that this is in dev mode. We need to test production build perf.

Screenshot is run on Mac Mini.

screen shot 2017-02-08 at 11 53 35

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "build:watch": "npm run build:lib -- --watch",
     "test": "mocha \"./src/**/*.test.js\" --require babel-core/register --timeout 5000",
     "test:watch": "npm run test -- --watch",
+    "benchmark": "jest",
     "flow": "flow check",
     "lint": "eslint src",
     "tslint": "tslint typings/*.ts native/*.ts",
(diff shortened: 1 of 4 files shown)


k15a:
For Jest migration take a look at: https://github.com/styled-components/styled-components/issues/325

jacobp100:
I've actually redone this with `benchmark.js` over on [my project](https://github.com/jacobp100/cssta/blob/master/benchmark/tests.js). The results aren't quite as dramatic as was shown with Jest, and a lot of this will be due to the fact that we're now running in production mode.

screen shot 2017-02-08 at 14 51 15

If you want to move this benchmarking into the styled-components repo, I'm more than happy for you to do so!

It does seem like the caching isn't working though. The last result is from running,

```js
const instance = render(<Component />);
instance.update(<Component danger />);
instance.update(<Component />);
instance.update(<Component danger />);
instance.update(<Component />);
instance.update(<Component danger />);
instance.update(<Component />);
instance.update(<Component danger />);
```

Which should only parse two stylesheets, and then reuse the results. 😕

PR #475, opened 2017-02-08
Allow resetting style cache for benchmarking

jacobp100:


jacobp100:
It seems that for browsers, each component definition (``styled.h1`…`;``) has its own cache rather than using a global cache. This is probably because each browser component definition has its own selector, whereas native does not.

philpl:
This must've been merged separately at some point. So we should be able to close this PR :wink: https://github.com/styled-components/styled-components/blob/master/src/models/InlineStyle.js#L12

PR #476, opened 2017-02-09
Add stateless functional component example to "Existing CSS" doc

tayiorbeii:
This example will hopefully save someone else some time 🍻

### DIFF
--- a/docs/existing-css.md
+++ b/docs/existing-css.md
@@ -33,6 +33,20 @@ class MyNormalComponent extends React.Component {
 }
 ```
 
+An example using a stateless functional component:
+```JSX
+const Button = styled(({href, children, className}) => {
+  return (
+    <a href={href} className={`some-global-class ${className}`}>
+      {children}
+    </a>
+  )
+})`
+  width: 200px;
+`
+```
+
+
 ## Global class always overridden
 
 You can apply a global CSS class to a styled component by adding a `className` prop. This will work, but if a specific CSS property (e.g. `background-color`) is defined in both the global CSS _and_ the styled component the results may not be what your expecting!


geelen:
Hmm, I think the more idiomatic approach would be this:

```js
const A = styled.a`
  width: 200px;
`

const Button = ({href, children}) => {
  return (
    <A href={href} className="some-global-class">
      {children}
    </a>
  )
})
```

Is that something that you considered and didn't work for you? I'd be interested to hear why.

tayiorbeii:
Thanks for the reply @geelen !

I actually didn't approach this particular example the way you have in your counterexample. I think in this particular example that yours is more clear, but for my project I have other instances where I'm working with [components that wrap other components](https://github.com/eggheadio/egghead-ui/blob/master/src/components/Card/CourseCard.js#L32):

```javascript
const CourseCard = ({response, expanded=false}) => {
  const { title, instructor: { full_name } } = response
  return (
    <Card type='course' response={response} expanded={expanded} />
  )
}
```

In this case with the base `<Card />` being passed props, it seems like the approach I've taken in my proposed example will be the one to take (in speed mode semi-psuedocode):

```javascript
const CourseCard = styled(({response, expanded=false, className})
   // stuff
)`
  max-width: ${props => props.expanded ? '500px' : '100'px}
`
```

Thoughts on this more fleshed out example? I'd be happy to update this PR with your version as well! Mostly my big facepalm was missing passing the `className` prop :)

geelen:
Happy for your example to be added actually, since _lots_ of people are missing the fact that `className` is needed, so more examples the merrier. But with `attrs` coming (#365), the example gets even simpler:

```js
const Button = styled.a.attrs({
  className: "some-global-class"
})`
  width: 200px;
`
```

Your other one does, too:

```js
const CourseCard = styled(Card).attrs({
  type: 'course',
  expanded: props => props.expanded || false,
  full_name: props => props.response.instructor.full_name
})`
  max-width: ${props => props.expanded ? '500px' : '100'px}
`
```

tayiorbeii:
@geelen Thanks, that makes it much more clear! You want I should update my PR example with your examples?

mxstbr:
Closing this due to inaction, feel free to submit a followup PR!

PR #490, opened 2017-02-13
[WIP DON'T MERGE] Drop the whitelist

geelen:
This is just an experiment (and a broken one for now) but I wanted somewhere to kick off discussion.

I've been looking for a declarative way to define which properties get passed through to the underlying component in an attempt to get rid of the [global attribute whitelist](https://github.com/styled-components/styled-components/blob/v2/src/utils/test/validAttr.test.js). I had an idea, but it didn't get very far:

```js
const Comp = props => <div {...props}/>

const SC = styled(Comp).attrs({
  primary: false, // meaning don't pass this through
})`
  color: ${ props => props.primary ? 'red' : 'blue' };
`

<SC primary/> // <div className="sc-Comp abcd1234"/>
```

Then, we could define all Styled Components as _not passing attributes by default_ (except for some core React ones) but define all the _DOM aliases_ as having the following:

```js
/* styled-components/index.js */
export styled = (...) => (...)

// DOM Aliases
styled.a = styled('a').attrs({
  href: true,
  target: true,
  rel: true,
  /* ... */
})
styled.button = styled('button').attrs({
  'autofocus disabled form name type value': true // space-separated shorthand maybe?
})
/* etc etc */
```

But I quickly ran into trouble—boolean attributes are already passing `true`, so you end up attaching attributes like `disabled` to _every_ `<button>`, which is no good really. So I'm not sure whether to push forward with this idea or not. Thoughts?


edygar:
What about switch to a blacklist approach? I mean, the styled-component declaration knows all props that it depends on which shouldn't be propagated. By default, any non black-listed `attr` would be propagated, any `undefined` [_.result](https://lodash.com/docs/4.17.4#result) means it shouldn't be propagated. A named export (as `doNotPropagate`) could even be a alias to `undefined`, so it communicates the intention

mxstbr:
Closing, since this is likely not going to be the final API.

Issue #493, opened 2017-02-13
[v2] @font-face breaks styles 


marc-rutkowski:
## Version
2.0.0-3

## Reproduction
https://github.com/marc-rutkowski/styled-components-v2-font-face
(standard create-react-app)

## Steps to reproduce
```
git clone git@github.com:marc-rutkowski/styled-components-v2-font-face.git
yarn
yarn start
```

Dev server starts and the app home page displays a red background and use the Pangolin webfont.
`@font-face` is set by using the `injectGlobal` helper in [App.js](https://github.com/marc-rutkowski/styled-components-v2-font-face/blob/master/src/App.js#L8)

Now, switch to styled-components v2 :
```
yarn remove styled-components
yarn add styled-components@next
yarn start
```

## Expected Behavior
Same rendering than with styled-components v1.x

## Actual Behavior
`@font-face` breaks styling. The red background is not visible anymore and default font is used instead of the custom one.
If the  `@font-face` rule is removed the red background appears has before.

Maybe related to #462 ? 

ps: thanks for this great lib!

thysultan:
Thanks, will push a patch for this.

thysultan:
patched upstream(parser), cc @mxstbr.

k15a:
Awesome ❤️

marc-rutkowski:
@thysultan thank you for the very fast reply and fix! It works fine now.

@k15a maybe this issue should be stay open until deps are updated to stylis@1.1.2+ ?
Let me know if you want a PR for this...

ps: It took me awhile to understand that it was fixed by this [commit](https://github.com/thysultan/stylis.js/commit/b85da030a09bc2fea3edc7a4a6b1bec1eb0b88c9) ;-)

k15a:
@marc-rutkowski Ahh. I thought that we are using version ranges so the patches are covered. OFC it should stay open then^^

PR #496, opened 2017-02-14
[v2] fix @font-face rule support

marc-rutkowski:
This PR :
- add tests for the problem explained into #493
- upgrade stylis to v1.1.3
- do a minor fix on another test that was no longer valid after stylis upgrade

### DIFF
--- a/src/constructors/test/injectGlobal.test.js
+++ b/src/constructors/test/injectGlobal.test.js
@@ -11,10 +11,6 @@ let styled = resetStyled()
 const rule1 = 'width: 100%;'
 const rule2 = 'text-decoration: none;'
 const rule3 = 'color: blue;'
-const rule4 = `@font-face {
-  font-family: 'some font';
-  src: url('fontFile.ttf') format('truetype');
-}`
 
 describe('injectGlobal', () => {
   beforeEach(() => {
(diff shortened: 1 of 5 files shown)


marc-rutkowski:
Maybe the test is superfluous because there is already one into stylis.
I just add one to validate that the problem is here before, and is gone after the upgrade...

src/test/css.test.js
```diff
@@ -29,10 +29,10 @@ describe('css features', () => {
...
       .sc-a {}
       .b {
-        display:-webkit-box;
-        display:-webkit-flex;
-        -ms-flexbox;
-        display:flex;
+        display: -webkit-box;
+        display: -webkit-flex;
+        display:  -ms-flexbox;
+        display: flex;
```
mxstbr: This is great though, please leave it in!

src/constructors/test/injectGlobal.test.js
```diff
@@ -19,6 +23,7 @@ describe('injectGlobal', () => {
 
   it(`should inject rules into the head`, () => {
     injectGlobal`
+      ${rule4}
```
mxstbr: I doubt testing this in styled-components is necessary, mind removing this change?

Issue #471, opened 2017-02-07
Benchmarking


dbertella:
This is not an issue itself, and it's a follow up from #465.

Using a styled-component as the following in RN can lead to some performance problem in your opinion?
```js
const Container = styled.View`
    flex: ${({ flex }) => flex || 1};
    flex-direction: ${({ direction }) => direction || 'column'};
    justify-content: ${({ justify }) => justify || 'flex-start'};
    align-items: ${({ align }) => align || 'stretch'};
    margin: ${({ m }) => m || 0};
    padding: ${({ p }) => p || 0};
`;
```
I know the stylesheet Object is heavily cached, and I think styled-component will use it at some point, am I right?
What about passing props at runtime? What's going to happen?
Because I'm really happy using this component it's really fast to prototype, but if it can be a problem I will find another solution.

I report @jacobp100's answer since it's a good point to start. 
 
> I'm don't think anything has been benchmarked, but it's something I'd like to see. Most of the time spent in styled-components will be parsing, but I don't think we know how much time is spent here.
> 
> In RN, every time your props change in a way that changes the CSS text, it gets re-parsed. This is probably negligible for a single component. However, when you load the app, you'll have to parse CSS for every component that gets mounted. If you have a global theme and change a value, that'll trigger a large re-parse for every component that depends on the theme.
> 
> If you're concerned about perf, you can set the default styles with styled-components, and override the styles in the style property.
> ```js
> const ContainerDefaults = styled.View`
>     flex: 1;
>     flex-direction: column;
>     ...
> `;
> 
> const Container = ({ flex, direction ... }) => (
>     <ContainerDefaults style={{ flex, direction ... }} />
> );
> ```
> This will only require one parse, no matter what styles you pass in.
> 
> If you want to eliminate the parse time, you can look at my [almost identical project](https://github.com/jacobp100/cssta). It's more limited in what you can do, but I think the limitations are a good thing!

jacobp100:
As a first step, we need to gather example projects to see how people use styled-components.

We can benchmark the startup time using react-test-renderer.

mxstbr:
I've been meaning to try to benchmark our stuff for ages, but couldn't figure out a reproducible way to do it. Maybe related to #348?

jacobp100:
See #474

jacobp100:
I've done some benchmarking with `benchmark.js` over on [my project](https://github.com/jacobp100/cssta/blob/master/benchmark/tests.js).

[...]

PR #497, opened 2017-02-14
Add benchmarking

jacobp100:
See #471

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ npm-debug.log.*
 dist
 .DS_Store
 bundle-stats.html
+benchmark/compiled
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ script:
   - yarn run danger run -- --verbose
   - yarn run flow
   - yarn run lint && npm run test
+  - yarn run benchmark
 notifications:
   email:
     on_failure: change
--- a/benchmark/.babelrc
+++ b/benchmark/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": [["latest", { "modules": false }], "react"]
+}
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -1,0 +1,37 @@
+# Benchmark
+
+Benchmarks performance for both browser and native styled-components.
+
+To benchmark,
+
+```bash
+npm install
+npm start
+```
+
+## Adding a Test
+
+Under `tests`, there are files for each platform we test: currently browser and native. Each platform test file has a default export of an object whose keys are the test names and values are test functions.
+
+```js
+export default {
+  'test name': () => {
+    /* Test function body */
+  }
+}
+```
+
+Tests are usually formed by creating a component, and running it through `react-test-renderer`.
+
+```js
+const Component = styled.div`
+  color: ${props => props.color};
+`;
+
+const instance = render(<Component color="black" />);
+instance.update(<Component color="red" />);
+```
+
+All tests are compiled using webpack. React uses production-mode optimisations, and the output is minified.
+
+Tests are benchmarked using Benchmark.js, so are run multiple times. If your test uses a global cache, you should clear the cache before running it.
--- a/benchmark/index.js
+++ b/benchmark/index.js
@@ -1,0 +1,35 @@
+// @flow
+const Benchmark = require('benchmark')
+require('console.table') // eslint-disable-line
+const nativeTests = require('./compiled/native').default // eslint-disable-line
+const browserTests = require('./compiled/browser').default // eslint-disable-line
+
+const runSuite = (suiteName, tests) => {
+  const suite = new Benchmark.Suite()
+
+  const testNames = Object.keys(tests)
+
+  testNames.forEach((testName) => {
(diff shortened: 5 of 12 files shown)


mxstbr:
Oh wow this is looking awesome! /cc @geelen 

Can we run this in CI with consistent results? Maybe run an empty benchmark (i.e. just calling a `noop` function) a bunch of times to get a baseline for the current CI state, then run the actual benchmark a bunch of times and take the diff from the baseline? I think webpack does something like that, /cc @TheLarkInn

jacobp100:
Benchmark.js should handle consistent results! It’s then mostly about the machine and the load on it. Could try it on the CI and see what happens when we run it a few times?

mxstbr:
> Could try it on the CI and see what happens when we run it a few times?

Let's do that, all you should need to do is add it to `.travis.yml`?

mxstbr:
> Error: Cannot find module "react-test-renderer"

Might want to fix that 😉 Also, Travis fails because Flow complains about all the non-flow-typed files.

jacobp100:
The remaining native tests will be fixed when #406 is merged.

mxstbr:
Going to restart CI a bunch of times to see if we get consistent results for the benchmarking. I'll update this comment as the results happen.

**Run 1**

```
Results for native
0                                  1            
---------------------------------  -------------
simple component                   0.503±0.208ms
prop changes                       1.421±0.868ms
prop shorthands                    0.000±0.000ms
prop shorthands with prop changes  0.000±0.000ms

Results for browser
0                                  1            
---------------------------------  -------------
simple component                   0.430±0.224ms
prop changes                       1.118±0.533ms
prop shorthands                    0.364±0.252ms
prop shorthands with prop changes  0.933±0.446ms
```

mxstbr:
I just had a realisation.

We can use Travis [build artifacts](https://docs.travis-ci.com/user/uploading-artifacts/) to store our benchmarking results as JSON, and then compare the current run to the previous run/output the results nicely human-readable from our `Dangerfile.js`!

jacobp100:
I'm not sure about running this on the CI now 😂 The results vary so much, and they seem to vary way more when you start doing more in the tests! Maybe we could try and find a machine with consistent performance.

mxstbr:
I was expecting that, to be honest—it's not like Travis promises to give you the exact same amount of power every time you run a CI process. That's why I said we should add a baseline!

Let's try benchmarking a noop function (`() => { }`) a bunch of times to get a baseline value to compare against, then reporting the diff between running styled-components and the noop function instead of the raw styled-components values.

In theory, by having a baseline we account for changes in CI power which should give us consistent results. What do you think?

jacobp100:
Oh I'll try that!

mxstbr:
Another note, I think by using the `env` key in `.travis.yml` and running the benchmarks based on the `RUN_BENCHMARKS` env variable being set to true we can run a travis instance _just_ for the benchmarks. (see [the travis docs](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Build-Matrix) for more info)

That might be helpful?

mxstbr:
## With baseline results

(all of these are on Node 7)

**Run 1**

```
native results                     Time       
---------------------------------  -----------
simple component                   1.181±0.349
prop changes                       3.764±0.444
prop shorthands                    0.000±0.000
prop shorthands with prop changes  0.000±0.000

browser results                    Time       
---------------------------------  -----------
simple component                   1.292±0.386
prop changes                       2.906±0.494
prop shorthands                    1.254±0.363
prop shorthands with prop changes  3.128±0.547
```

jacobp100:
Hmm. It still seems flaky for a few tests. But this probably the best we'll get. Still interested in adding it to the CI? If so, I'll clean up the branch a bit.

untitled 2

mxstbr:
Running on CI, we're getting much closer than before.

Let's try running the _whole thing_ a bunch of times and taking the average of a handful of runs instead of the diff results?

jacobp100:
Benchmark.js does that for us! I think it's just down to the load of the server at the point you begin.

We could go overkill and run the baseline test between each test, but that would take ages (and Travis might not like that).

This graph might highlight the problem tests a bit better...

bar

mxstbr:
> but that would take ages

Not too worried about that if it runs in its own Travis process, since that means it won't block testing, linting etc. from failing fast. Let's try?

jacobp100:
I think it just got worse 😂

bar-2

mxstbr:
Man that's annoying. Okay, last idea then: what about running master and PR in the same process to make them comparable?

jacobp100:
I think the variation is mostly down to variations in the load over time for Travis. Some of the results vary a huge amount, but some seem pretty okay. Perhaps we could just warn people, but also say that it isn't 100% reliable, and that they can re-run the CI if they want to be sure?

jacobp100:
Maybe we could just always print the results for every CI run and compare it with the last one. We'll just use it as vague advice, and let people decide whether it was performance-changing or not.

philpl:
@jacobp100 Since this is in a separate repo now, let's close it and add it once we have lerna, right? :wink:

jacobp100:
Yea that seems like the best option now!

Issue #494, opened 2017-02-13
Styled components should be "tag agnostic"


cloud-walker:
This is a proposal for an improvement, based on a question I've asked on gitter.

I think that the styled component creation process should be tag agnostic: you shouldn't be forced to specify an element tag prematurely, as you may need to use a different tag semantically, but keep the same presentation aspect, like a button button vs an anchor button.

@xcoderzach suggested the following solution (using `recompose`):

```javascript
const Button = styled(componentFromProp('c'))`
  ...myStyles...
`

<Button c="button" foo="bar" /> // renders <button foo="bar" />
<Button c="a" foo="bar" />  // renders <a foo="bar" />
<Button c={Link} foo="bar" /> // renders react router <Link foo="bar" />
```

And I like it (as I like recompose too), but I still think it should be a direct styled-components feature.

I suggest to default as `div` (the most generic one, or `span`?), and leave as developer choice the possibility to specify a different default tag if needed.

Then leave like a `tag` prop to permit the following form: `<Button tag="a" />` to override the default tag on create element time.

k15a:
You can do this with:
```JS
const e = React.createElement

const Button = styled(({ tag, children, ...props }) => e(tag, props, children))`
  // styles
`
```
which is quite small. IMHO I don't think we should add your proposal to styled-components because it's nothing you need so often and there is a simple workaround.

cloud-walker:
Maybe @k15a is right (compliment for the poro's avatar!) so, what about a docs section about that solution?

k15a:
Yeah it would probably be good to have something like this and https://github.com/styled-components/styled-components/issues/393#issuecomment-274106357 in the docs.

xcoderzach:
I'm on the fence about this one. With both my and @k15a's solutions, you lose the attribute whitelist. You'll end up with a bunch of warnings if you pass in any custom props.  A real workaround is going to be more verbose (blacklisting or whitelisting props 😢).

Buttons that can be an `<a />`, `<input type=submit />`, or `<button />` are pretty common.  But, that's the only example I can think of, so an affordance just for this case may be overkill.

cloud-walker:
Simple improvement with attribute whitelist:

```javascript
const e = React.createElement

const Button = styled(({tag = 'div', children, ...props}) => e(tag, props, children))`
  // styles
`
```

k15a:
You can do this with defaultProps as well:
```JS
const e = React.createElement

[...]

Issue #393, opened 2017-01-20
Props dependant Inheritance


alanzanattadev:
Hi,

I've just discovered styled-components, and it's awesome, great job !
One very cool feature is the composable inheritance like ```const FlexRed = styled(Red)`display: flex` ```.

The thing is, we can set css attributs depending on props, but we can't choose the parent component depending on props.

For example I have an HorizontalLayout and a VerticalLayout, I want a ScrollableLayout component, but I don't want to make HorizontalScrollableLayout and VerticalScrollableLayout, I'd  like to do ```styled(props => props.horizontal ? HorizontalLayout : VerticalLayout)`overflow: auto;`;```

Composition would be so much more powerful with these kind of modularity.

What do you think ? I don't know how styled-components work so I don't know how complicated it is to add the feature

Thank you !

k15a:
You could use the CSS helper.

```JS
import styled, { css } from 'styled-components'

const horizontalLayout = css`
  /* CSS */
`

const verticalLayout = css`
  /* CSS */
`

const ScrollableLayout = styled.div`
  ${props => props.horizontal ? horizontalLayout : verticalLayout}

  overflow: hidden;
`
```

Don't know if this helps you.

alanzanattadev:
It's a workaround, but it doesn't let me reuse the intermediary components as standalone components :/

k15a:
Actually your solution should work if I look twice.

alanzanattadev:
Yes it is indeed :) Thank you !

mxstbr:
That's an amazing solution @k15a!

We should add an example of this to the docs, people are going to love it. What do you think @geelen?

PR #498, opened 2017-02-14
Add 'Component based on props' to docs

k15a:
I've added some examples to show how to choose a component based on props.
Relevant: #494 and #393

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file. If a contri
 ## [Unreleased]
 
 ### Added
+- Add a section in tips and tricks to show how to choose a component based on props. [@k15a](https://github.com/k15a)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


ericdfields:
Can you link to your examples? Thank you.

k15a:
https://github.com/k15a/styled-components/blob/353a7c4f8c4abf19b148857b7c47dad5e7499533/docs/tips-and-tricks.md#choosing-a-component-based-on-props

ericdfields:
@mxstbr you're probably right. I suppose the use case for me is I have a styled component Card that could stand on it's own as a `div` or might be in a list of other Cards. To save a component, I could tell the Card that it should be an `li` instead of a `div`, saving me the trouble of managing styles for `li` (no list style, margin, padding, etc).

There's a good counter argument that the Card's single purpose is to be a card, and the `li`'s purpose is to be the list item.

k15a:
I am open to changes or we could leave this out entirely. I think it's something that could be mentioned in the docs because we had some question if this is possible. Maybe we should add something like this to the docs but with a better solution?

PR #489, opened 2017-02-12
Added react-native-extended-stylesheet

shripadk:
* Supports REM/Percent units
* Supports Media Queries (Fixes #203)

PR #504, opened 2017-02-15
[DISCUSSION] Custom StyleSheet constructors

jacobp100:
Bouncing off #489, what if we allow the user to use what ever stylesheet library they like? To support react-native-extended-stylesheet, you'd create a module,

###### styled-components-extended-stylesheet
```js
import StyleSheet from 'react-native-extended-stylesheet'
StyleSheet.build()
export default styles => StyleSheet.create(styles)
```

Then for the user,

```js
import styled, { setStyleSheetPlugin } from 'styled-components'
import styleSheetPlugin from 'styled-components-extended-stylesheet'

setStyleSheetPlugin(styleSheetPlugin)

const UsesMediaQueries = styled.View`
  color: red;

  @media (min-width: 600px) {
    color: blue;
  }
`
```

Not tested, but if we like this direction, and we get Jest merged, happy to add them!

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "inline-style-prefixer": "^2.0.5",
     "is-function": "^1.0.1",
     "is-plain-object": "^2.0.1",
+    "react-native-extended-stylesheet": "^0.3.1",
     "supports-color": "^3.1.2"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -48,7 +48,6 @@
     "inline-style-prefixer": "^2.0.5",
     "is-function": "^1.0.1",
     "is-plain-object": "^2.0.1",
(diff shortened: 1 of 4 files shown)


shripadk:
I like this idea. I have been tinkering with modifications to react-native-extended-stylesheet as well and have successfully managed to get @media with orientation working (especially useful if you want flexbox based grids in react-native which wouldn't be possible otherwise). It would be great if both these awesome projects could be combined in some way (including, as you suggested, making it available via custom stylesheet constructors).

mxstbr:
I'm not super happy to add random plugins/extension points.

We've so far refrained from any configuration whatsoever, because building a third party component library with some special config means that users might not be able to use it without some other special config—which we want to avoid. I'm curious what @geelen thinks about this.

PR #509, opened 2017-02-16
parse css blocks to objects in attrs

ramitos:
related: https://github.com/styled-components/styled-components/pull/371


geelen:
This is looking good, but I have a quick q (and it might just be because I'm on my phone and missed something)

Does this support passing css-as-inline-styles to anything _other_ than the `style` attr?

ramitos:
> Does this support passing css-as-inline-styles to anything other than the `style` attr?

No, it doesn't 😕

geelen:
Hey @ramitos I think if we can target arbitrary attributes with this it'll be good to go. Any chance you can add that?

ramitos:
@geelen I just pushed a solution that - although not pretty - seems to work. However, I'm not comfortable with Flow enough to create a type that extends Array with a property. Any hints?

geelen:
Closed in favour of #636.

PR #510, opened 2017-02-17
Export attrs constructor

bondz:
Export attrs constructor in #365


mxstbr:
https://github.com/styled-components/styled-components/pull/365#issuecomment-280661599

PR #518, opened 2017-02-19
Explicit extending syntax

geelen:
So, previously there was some logic when you wrote `styled(X)` that checked if `X` was a StyledComponent or not, which I never liked. So here's an attempt at fixing it.

```js
const Parent = styled.div`
  margin: 0.5rem;
  width: 2rem;
  height: 2rem;
  background: papayawhip;
`

/* Old way. This used to merge Parent's rules
 * with A's, but now simply wraps it and passes
 * className to it like it was any other component.
 */
const A = styled(Parent)`
  border: 4px solid palevioletred;
`

/* New way. No extra component is created, which
 * is preferable, but now the behaviour is explicit.
 */
const B = Parent.extend`
  border: 4px solid tomato;
`

/* Now there's an explicit way of changing tags,
 * which is not something to be used all the time
 * but might be useful in some cases.
 */
const C = Parent.extendWith('span')`
  border: 4px solid peru;
`
```

When rendered, you can see that `A` simply wraps `Parent` but the other two don't:

image

We'd need the babel plugin to inject `componentId` and `displayName` correctly in this case, but it should be ok I think. This is a breaking change, because `styled(X)` will no longer inherit static variables like `propTypes` or `defaultProps` from `X` (`X.extend` does though), but given there's a direct upgrade path I'm fairly happy with shipping this as part of v2.


geelen:
> There are library changes, but not tests. That's OK as long as you're refactoring existing code

Danger's not seeing the test changes in this PR, weird.

geelen:
Also what is going on with appveyor? Failing on a whitespace bug. Could be an old transitive dependency (like stylis)? @mxstbr can we make appveyor use yarn instead of npm?

k15a:
I really like this idea 👍 but in my opinion it would be cool if you could extend styled components without the extend.

```JS
const B = Parent`
  border: 4px solid tomato;
`
```

But then you'd have to check if this component is called as a styled-component or React component. Don't know if this is worth the little bit cleaner look.

> Danger's not seeing the test changes in this PR, weird.
That's happening on other PRs as well. Might be related to the move to DangerJS.

geelen:
Yeah that can't be done. Something can't be both a tagged template literal and a constructor function. But also, I think that's pretty confusing so I think `.extend` is good

edygar:
I think there's some weirdness in it. Mainly on `extendWith` case, though I understand the reasoning.
Wouldn't it be better, as a API, `styled-component` which are going to extends another to Always define it's target?

```js
const Parent = styled.div`
  margin: 0.5rem;
  width: 2rem;
  height: 2rem;
  background: papayawhip;
`

const A = styled.div.extends(Parent)`
  border: 4px solid palevioletred;
`

const C = styled(ThirdPartyCmp).extends(Parent)`
  border: 4px solid peru;
`
```

That way we got some consistency and `styled` only deals with its attr's, never giving the illusion to consumer that `tagName` is rewritable in anycase.

GGAlanSmithee:
@geelen @edygar Am I correct in that, if this was to be merged, it will support use-case

```javascript
const Icon = ({className, type}) => (
    <i className={className}>
        {type}
    </i>
)

const StyledIcon = styled(Icon).attrs({
    className: 'material-icons'
})`
    font-size: 2rem;
`

// no attrs :(
const BlueStyledIcon = styled(StyledIcon)`
    color: blue;
`
```
because that would be swell! :)

This is a contrived example, and might not be the most idiomatic way of using `styled-components`, but IMO it's the expected behavior. Implicit or explicit it would be nice to be able to do this.

GGAlanSmithee:
BTW, I think the naming of `extendWIth` might be a bit unfortunate, since the behavior is more like `replaceWIth`. Maybe `replaceWithTag` to make it super clear as to what's going on? It's a bit wordy, but if it's only intended to be used rarely, that might be a good thing (like react's lifecycle methods).

geelen:
Yeah I thought `extendChangingTag` or `extendForComponent` or something might be better.

@GGAlanSmithee not sure what's going on with your `BlueStyledIcon` above, can you clarify. But in general:

```js
styled(X).attrs({/*... attrs ...*/})` /*... styles ...*/ `
```

This _wraps_ `X` and passes any `attrs` including `className` which it generates from the `styles`.

```js
X.extend.attrs({/*... attrs ...*/})` /*... styles ...*/ `
```

This creates a new component that _merges_ `attrs` and `styles` with whatever `X` had previously.

Does that answer your question?

geelen:
@edygar I was initially going to implement that, but decided against it. I think, in the general case, you want to keep whatever underlying tag or component was used previously. With that you'd need to know, unless you allowed `styled.extend(A)` or something, which I didn't like.

edygar:
@geelen, now you got me thinking…
Could you explain the need of `extend`? I mean, `A` went just fine there at your first example.

```js
const A = styled(Parent)`
  border: 4px solid palevioletred;
```

FrancescoA:
Is there a recommended way to use `extend` on a non styled-component? Meaning, we often use styled-components to style the root component of a render function. However, even though the component is using a bunch of styled components, it itself is still a react component. Just stick with `styled(Component)` ?

geelen:
@FrancescoA yep, and pass down `className` to the outermost styled component.

PR #519, opened 2017-02-19
Bringing attrs to Native as well

geelen:
I realised when doing #518 that attrs hadn't been added to StyledNativeComponent. So this is a first pass, but needs someone to pick up who actually uses ReactNative to make sure I haven't misunderstood anything.


philpl:
@geelen Shall we merge this? I can update the native implementation of StyledComponent to match the web one afterwards

philpl:
As discussed I'll redo this on #714

Issue #389, opened 2017-01-19
[1.3.1] breaks react native animation


IljaDaderko:
```
const SContent = styled(View)`
  opacity: ${props => props.fade};
  flex: 1;
  margin-top: 20;
`;

const AContent = Animated.createAnimatedComponent(SContent);

        ...some other code...        

<AContent fade={this.state.contentAnim}>
```

This used to work and update opacity as intended before, right now opacity stays at 0 and never updates for some reason. `this.state.contentAnim` updates as before

---

**EDIT**

if I add callback function to animation that updates state passed to AContent
```
_onAnimateContent = () => {
    Animated.timing(this.state.contentAnim, {
      duration: 200,
      toValue: 1
    }).start(() => this.setState({ contentPointerEvents: 'auto' }));
  }
```

and

```
<AContent fade={this.state.contentAnim}  pointerEvents={this.state.contentPointerEvents}>
```

After that animation finishes and state is updated I can see my component, however it just becomes visible without any animation.

mxstbr:
Are you on the newest version? We _really_ need to pick up #94...

IljaDaderko:
styled components 1.3.1 react v0.40

mxstbr:
And it worked on earlier versions?!

IljaDaderko:
I've posted 2 code snippets below, a bit verbose but might be more informative:
Interesting bit is that Logo animation works fine, it's that AContent bit that is having issues. To double confirm this, just ran whole app on 1.3.0 (deleted node modules with 1.3.1, reinstalled whole thing with 1.3.0) and it works perfectly fine.

layouts/Base.js
---
```
import React, { PropTypes, Component } from 'react';
import styled from 'styled-components/native';
import {
  Image,
  View,
  Animated,
  Easing
} from 'react-native';
import { Match } from 'react-router';
import { connect } from 'react-redux';
import { routerSet } from '../redux/_router';
import Logo from '../components/Logo';
import HelpLink from '../components/HelpLink';
import SignIn from '../scenes/SignIn';
import SignUp from '../scenes/SignUp';
import { padding, dimensions } from '../Styles';
import router from '../emiters/router';

const SImageContainer = styled(Image)`
  flex: 1;
  background-color: transparent;
  align-items: center;
  padding-top: ${padding.baseTop};
`;

const SContent = styled(View)`
  flex: 1;
  margin-top: 20;
  opacity: ${props => props.aopacity};
`;

[...]

PR #528, opened 2017-02-23
Add docs for animations in React Native (assists #389), also allow pa…

jacobp100:
…ssing array of styles to native styled component


jacobp100:
I'm not sure duplicating the text from React Native README is the best thing for maintainability, but I thought it might help the user…

jasan-s:
I tried passing in the following: 

```
{
 transform: [
        {
          rotate: this.state.rotation.interpolate({
            inputRange: [-1, 1],
            outputRange: ['-360deg', '360deg']
          })
        }
      ]
    }
```

I get the Error Transform with key of "rotate" must be a string : {"rotate": "0deg"}

PR #541, opened 2017-03-01
Migrate to Jest (fixes #325)

### LINKED ISSUE #325: PR: Migration to Jest

jacobp100:
By default, Jest uses JSDom. Instead of getting the CSS from `StyleSheet`, we can get it directly from the DOM. As a result, the following tests need fixing or removing,

* styleSheet › reset() › should flush all css
* styleSheet › getCSS() › should list all css in all Components
* styleSheet › getCSS() › should list all css all Components including globally injected styles
* styleSheet › getCSS() › should list all css Components with min: false


philpl:
Mind to get this up to speed, or if I pick this back up? It would really be awesome to have some jest features, and to add this to lint-staged

jacobp100:
There were a couple of failures that it would be good to have someone look at!

geelen:
Ok I've picked this branch up and taken it on for some SSR stuff. So these commits will land, but not via this PR. Thanks for the start, though, @jacobp100!

Issue #536, opened 2017-02-26
domElement enhancement question


didierfranc:
## Version
1.x
2.x

## Question

Is there a reason to use this list over a simple Object.keys(React.DOM) ?
https://github.com/styled-components/styled-components/blob/v2/src/utils/domElements.js

I made it like that in react-sc
👉 https://github.com/didierfranc/react-sc/blob/master/src/index.js#L6

mxstbr:
Hah that's neat! /cc @geelen 

Mind submitting a PR for this against the `v2` branch? I reckon it'd kick down our package size by a lot!

didierfranc:
I'll do it ;)

didierfranc:
This will disappear in React 15.5/16

PR #546, opened 2017-03-01
Remove domElements list, use React.DOM instead

didierfranc:
Everything is here https://github.com/styled-components/styled-components/issues/536


yaycmyk:
This is going away in 15.5, I wouldn't use it: https://github.com/facebook/react/issues/8854

k15a:
Thank you for that information @yaycmyk.

mxstbr:
That's sad, I was really excited for a second. Thanks @yaycmyk!

Issue #452, opened 2017-02-03
AsyncStyledComponent: A right way to manage dynamic styles


theKashey:
So this is not about some bug, and even not about feature.
I come here from #134, look like I have a bit of knowledge how to fix a problem and speedup applying new styles.

So, where problem lies:
 On new component mount it is creating it`s own style, generating className and adds portion of text to current stylesheet.
 This lead to two problems: update of "global" stylesheet will reflow/redraw(in browser terms) all affected components(read - everything). Flush some useful caches and do some useless work. 

On each mount of each component.
Just because you need style to be available **instantly**. 

If you will await styles not right now, but in some point in future, everything can be a bit better.
So, do not apply styles asap, and dont delete them for a some time.
Apply new styles in bulk in new stylesheet, and then, after few seconds of idle, you can move new styles to global stylesheet.

The cons are simply - everything will be rendered in a strange way:
 - first everything will be rendered, and become invisible.
 - next styles will arrive to a browser, and whole tree will gain classNames. And regenerated.
 - then something changes style of component - they will not be changed instantly, but only after few ticks. Ie after flushing new styles to new stylesheet.

As result, it will change behavior how everything works, but may be it worth it.
May be not, cos it can break something.

Almost same logic are used in Yandex Maps API for handling dynamic styles. So it is a bit tested and proven solution. See a little demo: http://jsfiddle.net/gunwL8az/ 
PS: not a true example, cos each marker is SVG->PNG->DATAURI->STYLESHEET.

Sample code:
`
const createStyledComponent = (target: Target, rules: RuleSet, parent?: ReactClass<*>) => {

      generateAndInjectStyles(theme: any, props: any) {
        const executionContext = { ...props, theme }
        // some other implementation, flushes styles on nextTick, returns promise.
        --> return componentStyle._async_generateAndInjectStyles(executionContext)
      }

      componentWillMount() {
        // If there is a theme in the context, subscribe to the event emitter. This
        // is necessary due to pure components blocking context updates, this circumvents
        // that by updating when an event is emitted
        if (this.context[CHANNEL]) {
          const subscribe = this.context[CHANNEL]
          this.unsubscribe = subscribe(nextTheme => {
            // This will be called once immediately
            const { defaultProps } = this.constructor
            // Props should take precedence over ThemeProvider, which should take precedence over
            // defaultProps, but React automatically puts defaultProps on props.
            const isDefaultTheme = defaultProps && this.props.theme === defaultProps.theme
[...]

PR #553, opened 2017-03-03
AsyncStyleSheet and auto removing of unused styles

theKashey:
I come from here https://github.com/styled-components/styled-components/issues/452
This PR not passing all tests(I dont know why), and linting (ok, this is just proof-of-concept)

What is done here:
 1. I replace StyleSheet with very simply AsyncStyleSheet. It just add styles in a bulk, and then compact them in a real huge style.
 2. I add removing of old styles. I was shocked, then I dont find anything about "remove" in Styled. You are very dirty ones. You should clean after you.

What in result
 - there is no need for dirty hacks on stylesheet level to support "millions of rules". You will never have millions of rules. Simultaneously.
 - you can animate your styles. With out penalty.
 - thanks to buck operations everything runs a bit faster
 - (cons) But in a bit different way.

### DIFF
--- a/example/index.html
+++ b/example/index.html
@@ -1,45 +1,86 @@
 <!DOCTYPE html>
 <html>
-  <head>
-    <meta charset="utf-8">
-    <title>Basic Example</title>
-  </head>
-  <body>
-    <h1>Basic Example</h1>
-    <div id="container"></div>
-    <script src="https://unpkg.com/react@15.3.2/dist/react.min.js"></script>
-    <script src="https://unpkg.com/react-dom@15.3.2/dist/react-dom.min.js"></script>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/babel-core/5.8.24/browser.min.js"></script>
-    <script src="/styled-components.js"></script>
-    <script type="text/babel">
-    // Create a <Title> react component that renders an <h1> which is
-    // centered, palevioletred and sized at 1.5em
-    const Title = styled.default.h1`
+<head>
+  <meta charset="utf-8">
+  <title>Basic Example</title>
+</head>
+<body>
+<h1>Basic Example</h1>
+<div id="container"></div>
+<script src="https://unpkg.com/react@15.3.2/dist/react.min.js"></script>
+<script src="https://unpkg.com/react-dom@15.3.2/dist/react-dom.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/babel-core/5.8.24/browser.min.js"></script>
+<script src="/styled-components.js"></script>
+<script type="text/babel">
+  // Create a <Title> react component that renders an <h1> which is
+  // centered, palevioletred and sized at 1.5em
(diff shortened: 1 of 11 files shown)


geelen:
Ok I've had a bit of a read of this PR, and I understand it's just at a proof-of-concept stage, but I'm not convinced the problems described in #452 are in need of something like this. But I want to pick up some discussion from there and move it here, in particular:

```js
The cons are simply - everything will be rendered in a strange way:

first everything will be rendered, and become invisible.
next styles will arrive to a browser, and whole tree will gain classNames. And regenerated.
then something changes style of component - they will not be changed instantly, but only after few ticks. Ie after flushing new styles to new stylesheet.
```

This seems very much the way Aphrodite works, and in particular the "big problem" of Aphrodite is the way the styles are injected after the markup has hit the page: https://github.com/Khan/aphrodite/issues/76

However, the bigger dealbreaker for me is that ordering of CSS rules become unpredictable:

```js
// Default rendering
<Comp/>

// After some user action
<Comp active/>

// Later 
<Comp/>
```

If you remove the css from `Comp` when `active = false`, then add it again later, you've reordered the CSS. Given we support descendant selectors, order is hugely important and informs a lot of the design of Style Components. It's issue #1 after all (worth reading through the linked issues there as well).

As a result, I'm not inclined to merge this or anything that adds a flash-of-unstyled-markup or involves reordering CSS, but I'm happy to continue the discussion, and I do appreciate the work you've put into this already, @theKashey.

theKashey:
@geelen - thanks for reply. I would like to discuss this PR.
So main goals is simply:
 1. Give ability to create new styles. quickly, instantly, for animations, from templates and so on.
To achieve this you have to:
 1. Apply them in bulk.
 2. Remove unused styles. As long some styles during animation are using for a few moments.

Next I just propose the way, we use in Yandex for ages.

What about "problems":
 1. "Lag" before style is required and className set. 
Here I use timeout, just because dont have anything else. Originally it was "nextTick".
In React LifeCycle you can create a TopLevel component with overriden ComponentDidUpdate(and so on) and flush styles on this event. 
Then all children "perform" style update, and you can apply all new styles in a bulk.
In this case you have to return new className in a "old" way, synchronously. Everything will be same, but you will have "bulk". 
So - no problem here.

2. I dont really understand issue. 
Why Horizontal and Vertical get rendered before SVG? 
Rendering must go from top to bottom. I should investigate this problem.
But, from my point of view - it is very very bad idea, to rely on component creation order.
Or, "desired" order can be emulated.
Let me dive deeper inside this problem, I just always avoid situations like this.

mxstbr:
> But, from my point of view - it is very very bad idea, to rely on component creation order.

That's what `styled-components` has to do to work as people expect it to. 🤷‍♂️

philpl:
The new stylesheet that @geelen wrote that is part of the async SSR effort, should deprecate this effort; Furthermore, I think this AsyncStyleSheet doesn't handle the injection order correctly, which leads to specificity and priority issues on the generated CSS

Issue #107, opened 2016-10-17
Internet Explorer support


didierfranc:
<img width="1413" alt="capture d ecran 2016-10-17 a 21 55 19" src="https://cloud.githubusercontent.com/assets/14132464/19453185/753ecfdc-94b4-11e6-8b6c-366ca1df296b.png">

It seems ***styled-component*** doesn't support Internet Explorer 10

👀 https://github.com/styled-components/styled-components/issues/107#issuecomment-254332819

## Version
1.0.5/1.0.6

## Steps to reproduce
Run your app on IE10

## Expected Behavior
Work

## Actual Behavior
Crash

mxstbr:
Thanks for digging in and submitting an issue about this!

While I hate IE as much as the next front-end developer, I understand that sometimes business needs come above technology needs, and that sometimes IE does need to be supported.

It's not super high priority for me to fix this, especially now that we have support for IE11 (see #82), but I'd love if somebody could dig in and add support for lower versions of IE as well!

didierfranc:
I would like to implement ***styled-component***, it will simplify a lot for our applications, unfortunately app from https://github.com/xeonys/ and so Deloitte, needs support for IE10+ 😭

mxstbr:
> unfortunately apps from Xeonys and so Deloitte, needs support for IE10+

That's exactly what I mean!

I (or @geelen) would love to give you some pointers if you want to dig in? It shouldn't be _too_ hard to fix this as far as I understand!

didierfranc:
Yes, it will be great ;)

amsb:
In IE10, the console gives me the error:

> SCRIPT5007: Unable to get property 'appendRule' of undefined or null reference

It looks like `this.insertedRule` is being set to null or undefined in:

https://github.com/styled-components/styled-components/blob/master/src/models/ComponentStyle.js#L25

styleSheet is a global glamour Stylesheet singleton created here:

https://github.com/styled-components/styled-components/blob/master/src/models/StyleSheet.js#L8

which is defined here:

https://github.com/threepointone/glamor/blob/98d4f44ba4335abd0e63bf946edff04ec51d3a41/src/sheet.js#L65

It looks like glamour is being thoughtful about some of the older IE issues (but I'm no expert) so I'm not yet sure why insert is return null at the moment. Is the Stylesheet singleton being created too soon in the document lifecycle on IE10?

didierfranc:
[...]

PR #564, opened 2017-03-07
replaced babel-preset-latest with babel-preset-env

shaunwallace:
With the deprecation of babel-preset-latest and the addition of babel-preset-env per [babel-preset-env](https://github.com/babel/babel-preset-env) the babel.rc file as well as the rollup.config.js have been updated to use the env preset.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,6 @@
 {
   "presets": [
-    "latest",
+    "env",
     "react"
   ],
   "plugins": [
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "babel-plugin-transform-class-properties": "^6.19.0",
     "babel-plugin-transform-flow-strip-types": "^6.21.0",
     "babel-plugin-transform-object-rest-spread": "^6.20.2",
-    "babel-preset-latest": "^6.16.0",
+    "babel-preset-env": "^1.2.1",
     "babel-preset-react": "^6.16.0",
     "chokidar": "^1.6.1",
     "danger": "^0.7.5",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -51,7 +51,7 @@ const plugins = [
   babel({
     babelrc: false,
     presets: [
-      ['latest', { es2015: { modules: false } }],
+      ['env', { modules: false }],
       'react',
     ],
     plugins: [


philpl:
Lgtm :) do we need to add some custom changes to make it transpiler in the same or even in a more minimal way?

shaunwallace:
@philpl Sorry, not sure what you mean by a 

> more minimal way.

Can you elaborate? As for custom changes, the babel docs for babel-preset-env don't mention the option to apply module rules to specific presets like you can with babel-preset-latest. I've reached out to [@left_pad](https://twitter.com/left_pad) to confirm since the rollup config was specifically setting `modules: false` for es2015 but this might already be handled by babel-preset-env.

philpl:
@shaunwallace I mean, we might want to define an env that fits our desired browser support

shaunwallace:
@philpl ya good point, where are you guys at on the browsers that you want to provide support for? Is this documented somewhere?

philpl:
@shaunwallace There's no official list (yet? ;))

It's definitely supporting IE10+ now as part of this issue: #107

So in version 1 the support is dictated by inline-style-prefixer: https://github.com/rofrischmann/inline-style-prefixer#browser-support

In version 2 it will be dictated by stylis though: https://github.com/thysultan/stylis.js

Note, that in version 2 we're thus going as low as IE 8+.

I guess it's very much @mxstbr's call to make what the support should look like. Also, eventually there should be an actual browser support doc in the repo, I guess.

yavorsky:
Wow, it's nice! 😍

Note, that uglify-js supports only es5 and will throw an error if something won't be transformed. So, for minification specify `uglify: true` as a target.

For faster compilation and development process optimization you could make targets more flexible and set latest browsers for dev environment and all the other stuff (like ie 10) for production.

philpl:
@yavorsky @shaunwallace it seems that v2 actually drops down all the way to IE8, which is awesome!

yavorsky:
@philpl It's nice! But actually, it could help only for built-ins. Many of them have ie 10 support [babel-preset-env/data/built-ins.json](https://github.com/babel/babel-preset-env/blob/master/data/built-ins.json), but for es6 transformations will be really effective only ie 11 rejection. 😔 (It supports only block scoped functions and only transform-es2015-block-scoped-functions will be ignored)

existentialism:
> As for custom changes, the babel docs for babel-preset-env don't mention the option to apply module rules to specific presets like you can with babel-preset-latest. I've reached out to @left_pad to confirm since the rollup config was specifically setting modules: false for es2015 but this might already be handled by babel-preset-env.

@shaunwallace just for clarity, the `modules` option on preset-env functions the same as setting the option under `es2015` for preset-latest, they both enable desired module transform (or none at all).

https://github.com/babel/babel-preset-env#modules

shaunwallace:
Ya thnx @existentialism I forgot to bring that back into this thread as @left_pad confirmed this as well. 😄

hzoo:
yep it's the same, https://twitter.com/left_pad/status/842771387622313985 😄

mxstbr:
Just to note, we officially support the same browsers that React does. (anything else doesn't really make sense :wink:) As of today that means IE9+!

Thanks for this PR

PR #567, opened 2017-03-08
Group Built-With Section in Categories

k15a:
I thought some order in the built with section would be good so it's faster to look up for something. Let me know what you think about that change ☺️


philpl:
I think this needs to be updated, but I don't see any reason why we shouldn't do this :)

k15a:
What do you think needs to be updated? :) Other groups, order, naming or something else?

philpl:
I just meant it's out of sync :D the structure itself is perfect

PR #585, opened 2017-03-14
Accept preprocessed CSS from babel plugin

philpl:
- [x] Refactor singleton factories
- [x] Add `css`, `flatten`, and `GlobalStyle` to singleton factories
- [x] Add `stringifyRules` helper to abstract CSS parsing
- [x] Add separate `no-parser` variants of these helpers
- [x] Test no-parser flattener variant

This adds the entrypoint that accepts the preprocessed CSS from the babel plugin, which has an open PR for the preprocessing here: https://github.com/styled-components/babel-plugin-styled-components/pull/26

It needs to add new singleton factories and new arguments to some other factories, since it now passes in `css`, `flatten`, `stringifyRules` (new), and `GlobalStyle` into some things that need them.

Edit: After fixing everything up, there are still 4 failing unit tests left, which are however failing on v2 as well.

### DIFF
--- a/src/models/ComponentStyle.js
+++ b/src/models/ComponentStyle.js
@@ -1,15 +1,14 @@
 // @flow
 import hashStr from 'glamor/lib/hash'
-import stylis from 'stylis'
 
-import type { RuleSet, NameGenerator, Flattener } from '../types'
+import type { RuleSet, NameGenerator, Flattener, Stringifier } from '../types'
 import styleSheet from './StyleSheet'
 
 /*
  ComponentStyle is all the CSS-specific stuff, not
  the React-specific stuff.
  */
-export default (nameGenerator: NameGenerator, flatten: Flattener) => {
+export default (nameGenerator: NameGenerator, flatten: Flattener, stringifyRules: Stringifier) => {
   const inserted = {}
 
   class ComponentStyle {
@@ -41,7 +40,7 @@ export default (nameGenerator: NameGenerator, flatten: Flattener) => {
       if (!inserted[hash]) {
         const selector = nameGenerator(hash)
         inserted[hash] = selector
-        const css = stylis(`.${selector}`, flatCSS, false, false)
+        const css = stringifyRules(flatCSS, selector, false)
         this.insertedRule.appendRule(css)
       }
       return inserted[hash]
--- a/src/models/ComponentStyle.js
+++ b/src/models/ComponentStyle.js
@@ -2,15 +2,14 @@
 import hashStr from 'glamor/lib/hash'
 import stylis from 'stylis'
 
-import type { RuleSet, NameGenerator } from '../types'
-import flatten from '../utils/flatten'
+import type { RuleSet, NameGenerator, Flattener } from '../types'
 import styleSheet from './StyleSheet'
 
 /*
  ComponentStyle is all the CSS-specific stuff, not
  the React-specific stuff.
  */
-export default (nameGenerator: NameGenerator) => {
+export default (nameGenerator: NameGenerator, flatten: Flattener) => {
   const inserted = {}
 
   class ComponentStyle {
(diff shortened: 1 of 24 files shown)


geelen:
OMG this is amazing just 🎉🎉🎉

> Edit: After fixing everything up, there are still 4 failing unit tests left, which are however failing on v2 as well.

Let me jump over there and fix that up for you.

philpl:
@geelen I'm currently testing this (and obviously the babel-plugin) against the styled-components website :)

kentcdodds:
For anyone else interested, here's what it looks like on https://styled-components-xspunjeazi.now.sh/

screen shot 2017-03-15 at 8 04 47 am

Can't get too much more optimized than that (other than SSR I guess). 👏

kentcdodds:
Is there any chance we could tree-shake out the parser? If we're just using the babel plugin and `noParse` then there's no reason to have the parser right? I expect that would reduce the size of styled-components considerably.

philpl:
@kentcdodds The `no-parser` entrypoint shouldn't depend on stylis, thus throwing it out of the bundle, actually ;)

kentcdodds:
Oh, interesting! So the `import` actually changes to the `no-parser` module? Coooool!

kentcdodds:
Will there be an update to the docs/README to show how to use the plugin?

philpl:
@kentcdodds ✨  I want to do the docs soon, in a separate PR. Btw we'd want to keep this stuff as "experimental" for now. This PR adds some notes to the babel-plugin on how to enable it: https://github.com/styled-components/babel-plugin-styled-components/pull/26

Optimally we'd want a doc in this repo that contains a small setup guide and a list of limitations.

kentcdodds:
Also, unless I'm mistaken, this pretty much will _not_ work with [polished](https://polished.js.org/) functions (or even custom functions) which use any form of nesting, need vendor prefixes, etc... The rules of what's possible and what's not with functions for this are complicated. So it should probably be documented that you should not use this plugin with dynamic expressions in your css.

This is a real bummer. I wonder whether there's anything we can do to improve upon this in the future... 🤔

philpl:
@kentcdodds Yep, there's a long list of limitations that I'll add to the docs, I'm planning, on styled-components. The variable thing is the most annoying part. Any variable won't go through stylis, and can cause more issues.

kentcdodds:
Would be nice to have a linter rule to disallow anything that would break when using this plugin. Opportunity for more people to get involved!

rafayepes:
Do we need all.these.chained.calls? Can't it be more flat?

philpl:
@rafayepes if you mean the many factories, then I can only say, that was the one reason why I procrastinated getting this done. But it's very important for code reusability

src/models/ComponentStyle.js
```diff
@@ -36,15 +34,17 @@ export default (nameGenerator: NameGenerator) => {
      * Returns the hash to be injected on render()
      * */
     generateAndInjectStyles(executionContext: Object) {
-      const flatCSS = flatten(this.rules, executionContext).join('')
-        .replace(/^\s*\/\/.*$/gm, '') // replace JS comments
-      const hash = hashStr(this.componentId + flatCSS)
+      const flatCSS = flatten(this.rules, executionContext)
+      const hash = hashStr(this.componentId + flatCSS.join(''))
+
```
theKashey: Here a little error is hidden.
First you check **internal** hash, and next add selector to a **global** css scope. 
What next? You can just duplicate style in two different style tags.

philpl: @theKashey I'm not quite sure what you're referring to? This is no change to the logic before :) we're relying on the fact that the componentid plus the hash are unique

theKashey: I am just pointing to a little bug. "inserted[hash]" must be a global variable for Styled, as long "result" works in global scope. Not sure is this related to this PR, but it is better to change it. Someday.
screen shot 2017-03-18 at 9 55 59 am

PR #597, opened 2017-03-18
withTheme hoist static (fix #596)

### LINKED ISSUE #596: Make withTheme hoist static properties
For reference: https://github.com/react-community/react-navigation/issues/381

brunolemos:
Fix #596.

For reference: https://github.com/react-community/react-navigation/issues/381

--
The big yarn changes happens on master too, I think `yarn.lock` is outdated.


mxstbr:
Sorry, which issues is this fixing? I don't quite understand the problem here.

brunolemos:
@mxstbr 

If you use `react-navigation`, for example, use can configure a scene using a static property:

```js
class MyScene extends React.Component
    static navigationOptions {
        title: 'My page',
    }
}
```

But, if I use `withTheme` in this component, the static property will be "lost" and react-navigation won't have access to it, because it will be only in the underlying component, not in the new component created by withTheme.

This PR makes the static properties be passed to the new component using the same approach redux uses on their `connect` HoC.

k15a:
Is there a disussion on redux why they implemented it? I don't understand why react-navigation isn't using props instead.

philpl:
Hiya, I think it makes more sense to add this to v2, to prevent further divergence of the two branches, and to prepare for this major release.

I've opened a new PR over at #712 that does the same thing. Also cleaned it up a little and added a `displayName` for fun and giggles.

PR #599, opened 2017-03-19
Update the Logo url

k15a:
On NPM the logo was broken because the url was relative.

I've removed the logo from this repository because there is a newer version in `styled-components/brand`.


k15a:
It looks like https://gitcdn.xyz supports that. I will update the PR if I am back home.

philpl:
@k15a I'm pretty sure you just have to replace the commit id with `master` ;)

k15a:
Yeah of course. That makes total sense :)

Issue #574, opened 2017-03-10
[V2] Theming doesn't work with reactstrap 4.2.0


L3ARNING2B:
Hi everyone,

I have read this very recent [article](https://www.smashingmagazine.com/2017/01/styled-components-enforcing-best-practices-component-based-systems/) claiming styled-components is the future so I decided to give a try.

I want to to customize my [reactstrap](https://reactstrap.github.io/) app. _(React equivalent of Bootstrap 4)_ 

I don't wan't to import the bootstrap css stylesheet. I rather make my own styling based on the class used in bootstrap. This is where styled-components come from! I Hoped...

For some reason, the custom theme passed to the `<ThemeProvider theme={theme} />` disappear for the one I've configured in my component `Button.defaultProps` right after a re-rending due to state change.

I've looked into their [__Button__ components](https://github.com/reactstrap/reactstrap/blob/master/src/Button.js) and I still wonder, their components are fairly simple.

I have created a webpackbin to demonstrate the issue.

I have also opened an issue on reactstrap: https://github.com/reactstrap/reactstrap/issues/355

According to @TheSharpieOne

> Yeah, definitely looks like an issue on styled-components side. ... Seems like Styled(*) shouldn't be passing down the props it uses. [...]
IDK, but looking at this line in styled-components it looks like it only subscribes and doesn't set the initial state based on the [default] props.

## Version
styled-components 2.0.0-7
reactstrap 4.2.0

## Reproduction

https://www.webpackbin.com/bins/-KetejqQhRZrV5MyzPpr

## Steps to reproduce

Click on the button within the reproduction and see.

## Expected Behavior

Custom theme to be applied

## Actual Behavior

Default theme is applied

L3ARNING2B:
I've opened an issue on reactstrap as well : https://github.com/reactstrap/reactstrap/issues/355

kopax:
Thanks for reporting. Looks similar to #549

santaclauze:
Yup, working on similar issues. Any leads would be grateful!

santaclauze:
# Version

styled-components 2.0.0-7
reactstrap 4.2.0


# Steps to reproduce

https://www.webpackbin.com/bins/-Kf5Tdg7bfj2oWZzIb1V

Case A - Button: Click on one of the left buttons to reproduce the issue with the component Button.

Case B - ButtonClass: Click on one of the right buttons to reproduce the issue with the component ButtonClass.

[...]

Issue #549, opened 2017-03-02
[V2] Using anchor with react-router break styled <ThemeProvider />


kopax:
I have found a bug in `v2.0.0-5`

When clicking on url that use an anchor, the `<ThemeProvider />` has no more effect.

Reverting to `v1.4.3` solve the issue.

## Version
2.0.0-5

## Reproduction
http://www.webpackbin.com/NJGs7ebqG
https://webpackbin.herokuapp.com/NJGs7ebqG

## Steps to reproduce

Use the anchroScroll method from here in your `react-router` : https://github.com/ReactTraining/react-router/issues/394#issuecomment-220221604

## Expected Behavior

To scroll to the anchor in question

## Actual Behavior

It scroll but It also break the `<ThemeProvider />` and use the theme from `defaultProps` instead.

kopax:
@mxstbr This bug is a bit major. It does not only concern the `react-router`...

I also encounter issue ussing `<Provider />` and `react-redux`. On my LoginForm when I type in the `<input name="username" />`, the store get updated and the `<ThemeProvider />` also lose the theme

Almost all our apps use styled-components v2 and are impacted by this bug.

santaclauze:
Seems to be the same issue : #574

kopax:
Corrected and released in `2.0.0-9`.

PR #605, opened 2017-03-22
Restore PR #450. Fix #445, #574, #549

### LINKED ISSUE #445: StyledComponent uses defaultProps for theme instead of theme provided by ThemeProvider after initial render

kopax:
According to @k15a (see #574)

> Ok this is a bug with v2 only. I think I have already fixed that bug in v1 so I will have to search for the PR.
> The corresponding PR is #450.

- fix issues: #445, #574, #549
- This change to the _CHANGELOG_ is not included in this [commit](https://github.com/styled-components/styled-components/pull/450/commits/7c4f01d44e5d40921de5052d893c7f755ed90b5b#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR13) (irrelevant)


kopax:
> Please include a CHANGELOG entry. You can find it at CHANGELOG.md 

Should I add the changelog ?

kopax:
can we get a new v2 for that please :) ?

PR #616, opened 2017-03-24
Updated README.md

DaveyEdwards:
Adding example for passing props to components

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,32 @@ Here is one input without any content showing the placeholder, and one with some
   </a>
 </div>
 
+
+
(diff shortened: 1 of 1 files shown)


philpl:
Hey, cheers for submitting this! I'm going to write a "Recipes" section in the docs soon. So this might be a better fit for that to prevent the Readme from getting any longer. Are you fine with merging it, once I'm done with it?

DaveyEdwards:
yeah that sounds good

philpl:
This will be clearer on the future docs: https://github.com/styled-components/styled-components-website/blob/master/pages/docs/basics/adapting-based-on-props.js

Issue #583, opened 2017-03-14
Performance with 2.0.0


siddharthkp:
### Disclaimer
This might be something that you guys are already aware of OR it might just be insignificant.

### Version
2.0.0-7 and 1.4.4

### Reproduction

I ran the tests in [styled-components/benchmark](https://github.com/styled-components/benchmark) and 2.0.0-7 was significantly slower.

Processor specs:
2.6 GHz Intel Core i5
8 GB 1600 MHz DDR3

Results:

```

Results for browser (could not get native to work for me)

test                               1.4.4          2.0.0-7        increase
---------------------------------  -------------  -------------  ---------
simple component                   0.085±0.011ms  0.186±0.046ms  2.19x
prop changes                       0.216±0.023ms  0.380±0.075ms  1.76x
prop shorthands                    0.099±0.027ms  0.186±0.066ms  1.88x
prop shorthands with prop changes  0.261±0.071ms  0.359±0.099ms  1.37x

```

Let me know if this is useful

jacobp100:
Interesting! I know that the parser switched from postCSS to stylis—maybe that could be it?

siddharthkp:
Possibly, how can i debug that?

thysultan:
I doubt it is stylis, isolating just the parser it should perform better, will try to compare how postCSS handles this [benchmark](https://raw.githubusercontent.com/thysultan/stylis.js/master/benchmark/index.html) to say by how much.

jacobp100:
Did anything else major change between the two versions?

k15a:
Just the attrs constructor and the component selector. The component selector shouldn't be the bottleneck.

thysultan:
[benchmark](https://github.com/thysultan/stylis.js/tree/master/benchmark) with post-css, stylis, css-tree and stylus rendering the same styles. stylis should improve performance in terms of the parser.

jacobp100:
Strange. Were any other libraries updated, like glamour?

thysultan:
Does 1.4.4 use post-css asynchronously?

thysultan:
ran the bench, the third one is without any parser, replacing stylis to a function that just returns input css.

```
Results for browser v1
                                   styled         cssta        
---------------------------------  -------------  -------------
simple component                   0.124±0.036ms  0.020±0.006ms
prop changes                       0.336±0.085ms  0.140±0.084ms
prop shorthands                    0.198±0.112ms  0.024±0.009ms
[...]

Issue #588, opened 2017-03-17
Split into multiple packages and move to monorepo


jacobp100:
In v2, we're going to be introducing a babel transform to perform some pre-parsing.

From my experience doing this, you'll find that most changes have to take place in both the styled-components and the babel plugin repos. Coordinating PRs between the two will be very difficult.

Not only this, but benchmarking (which is currently in styled-components) will also want to benchmark the results after running through the babel plugin. For this to work without cyclical dependencies, this would introduce yet a third coupled repo.

We should consider moving styled-components, babel-plugin-styled-components, and styled-components-benchmark into a Lerna monorepo for v2.

geelen:
I fully support this. I really want `styled-components` the package to be stand-alone—you shouldn't get the babel plugin unless you choose to opt in. But for maintenance and perf testing it makes total sense for them to be developed together.

jacobp100:
Would be good if we can keep git history. Does anybody know how to do this?

philpl:
I think this is an awesome idea. It's quite easy to add the babel plugin repo as a subtree, to preserve most of its git history, actually.

jacobp100:
Oh neat, [Lerna will do this for us](https://github.com/lerna/lerna#import)! I think the next step is that we'll need to pick a date to do this. We'll want to merge as many pending PRs before this, as it'll likely cause issues afterwards. I'm assuming this is v2 only?

philpl:
@jacobp100 Ooh! I wasn't aware that they've got this built in now :)

philpl:
@jacobp100 let's do this after v2, like prettier?

jacobp100:
I'd say we want it for v2. If there's bugs in the babel transform that require cross project changes, it'll cause a lot of headaches.

mxstbr:
We need to merge v2 into `master` first, before we can do this, but I agree we should do it before the official release.

mxstbr:
Well, that didn't happen before the release, but let's do it now! PRs welcome to make this into a monorepo that contains `css-to-react-native`, the babel plugin, the website and, of course, the core package.

k15a:
@mxstbr How about automatic changelog generation based on PR/Issuer labels with lerna-changelog like create-react-app does it?

mxstbr:
Sure, that sounds awesome! We forget to do changelogs all the time haha

philpl:
[...]

PR #619, opened 2017-03-25
Improve StyledComponent performance

philpl:
#583

- Add second `BaseStyledComponent` class to not repeat methods
- Improve `propsForElement` and `this.attrs` creation
- Stop creating `this.attrs` if `attrs` doesn't exist
- Rewrite `generateAlphabeticName` to be iterative instead of recursive

### Before

```
simple component (construction)     0.019±0.003ms
simple component (first render)     0.132±0.027ms
simple component (repeated render)  0.014±0.001ms
prop changes                        0.309±0.054ms
prop shorthands (first render)      0.175±0.071ms
prop shorthands with prop changes   0.377±0.093ms
```

### After

```
simple component (construction)     0.013±0.002ms
simple component (first render)     0.121±0.026ms
simple component (repeated render)  0.015±0.004ms
prop changes                        0.311±0.101ms
prop shorthands (first render)      0.155±0.052ms
prop shorthands with prop changes   0.344±0.038ms
```

Edit:

### After with preprocessing

```
simple component (construction)     0.011±0.002ms
simple component (first render)     0.090±0.018ms
simple component (repeated render)  0.014±0.001ms
prop changes                        0.195±0.030ms
prop shorthands (first render)      0.095±0.030ms
prop shorthands with prop changes   0.208±0.051ms
```

### DIFF
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -138,7 +138,8 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
           if (
             !isTargetTag || (
               validAttr(propName) &&
-              propName !== 'className'
+              propName !== 'className' &&
+              propName !== 'innerRef'
             )
           ) {
             // eslint-disable-next-line no-param-reassign
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -39,8 +39,11 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
     }
 
     buildExecutionContext(theme: any, props: any) {
-      const { attrs = {} } = this.constructor
+      const { attrs } = this.constructor
       const context = { ...props, theme }
+      if (attrs === undefined) {
+        return context
+      }
 
       this.attrs = Object.keys(attrs).reduce((acc, key) => {
         const attr = attrs[key]
@@ -156,7 +159,7 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
       componentId = generateId(options.displayName || 'sc'),
       ParentComponent = BaseStyledComponent,
       rules: extendingRules,
-      attrs = {},
+      attrs,
     } = options
 
     const warnTooManyClasses = createWarnTooManyClasses(displayName)
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -16,11 +16,14 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
   /* We depend on components having unique IDs */
   const identifiers = {}
   const generateId = (_displayName: string) => {
-    const displayName = _displayName
-      .replace(/[[\].#*$><+~=|^:(),"'`]/g, '-') // Replace all possible CSS selectors
-      .replace(/--+/g, '-') // Replace multiple -- with single -
+    const displayName = typeof _displayName !== 'string' ?
+      'sc' : _displayName
+        .replace(/[[\].#*$><+~=|^:(),"'`]/g, '-') // Replace all possible CSS selectors
+        .replace(/--+/g, '-') // Replace multiple -- with single -
+
     const nr = (identifiers[displayName] || 0) + 1
     identifiers[displayName] = nr
+
     const hash = ComponentStyle.generateName(displayName + nr)
     return `${displayName}-${hash}`
   }
@@ -156,7 +159,7 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
   ) => {
     const {
       displayName = isTag(target) ? `styled.${target}` : `Styled(${target.displayName})`,
-      componentId = generateId(options.displayName || 'sc'),
+      componentId = generateId(options.displayName),
       ParentComponent = BaseStyledComponent,
       rules: extendingRules,
       attrs,
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -25,127 +25,144 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
     return `${displayName}-${hash}`
   }
 
-  const createStyledComponent = (target: Target,
-                                 options: Object,
-                                 rules: RuleSet) => {
+  class BaseStyledComponent extends AbstractStyledComponent {
+    static target: Target
+    static styledComponentId: string
+    static attrs: Object
+    static componentStyle: Object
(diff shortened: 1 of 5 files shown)


src/models/StyledComponent.js
```diff
@@ -16,138 +16,186 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
...
+      return className
+    }
+
+    componentWillMount() {
+      // If there is a theme in the context, subscribe to the event emitter. This
+      // is necessary due to pure components blocking context updates, this circumvents
+      // that by updating when an event is emitted
+      if (this.context[CHANNEL]) {
+        const subscribe = this.context[CHANNEL]
+        this.unsubscribe = subscribe(nextTheme => {
```
wmertens: Am I understanding correctly that if you use a theme, it subscribes to changes? Ouch. This can mean that *all* components are subscribed to theme changes, I'm going to guess that that is not great for resource use.

Since I never change themes in my app while running, I'd like this subscription to be optional.

I have a similar problem with translations, and I resorted to just re-mounting the entire app whenever the language changes (by using the language as a key on the root component). I don't want to pay the listener cost on every piece of displayed text.

How about providing a ThemeListener component that you can put below sCU components to update the tree below with the new context? Or adding a _listen prop on all components that decides whether to add this listener?

philpl: This is not part of this refactor, but anyway, the listener's cost is close to zero, compared to an entire remount. A remount is actually more expensive, and a listener is not doing anything as long as the theme doesn't change.

So for someone who keeps the theme static, the cost is negligible.

wmertens: Hmm, good point, but isn't there overhead in adding and removing tens if not 100s of listeners?

philpl: @wmertens I can write a benchmark for this case soon, but compared to the mounting and unmount in of React components and DOM elements, it should be negligible as well :wink:

src/models/StyledComponent.js
```diff
@@ -16,138 +16,186 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
   /* We depend on components having unique IDs */
   const identifiers = {}
   const generateId = (_displayName: string) => {
-    const displayName = _displayName
-      .replace(/[[\].#*$><+~=|^:(),"'`]/g, '-') // Replace all possible CSS selectors
-      .replace(/--+/g, '-') // Replace multiple -- with single -
+    const displayName = typeof _displayName !== 'string' ?
+      'sc' : _displayName
+        .replace(/[[\].#*$><+~=|^:(),"'`]/g, '-') // Replace all possible CSS selectors
```
evan-scott-zocdoc: if you cache these regexes into variables it should help a little

src/models/StyledComponent.js
```diff
@@ -16,138 +16,186 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
...
+  ) => {
+    const {
+      displayName = isTag(target) ? `styled.${target}` : `Styled(${target.displayName})`,
+      componentId = generateId(options.displayName),
+      ParentComponent = BaseStyledComponent,
+      rules: extendingRules,
+      attrs,
+    } = options
+
+    const warnTooManyClasses = createWarnTooManyClasses(displayName)
```
evan-scott-zocdoc: seems like this should be guarded for dev-mode only - outside the scope of this PR perhaps, but it does introduce a bit of overhead and library size

src/models/StyledComponent.js
philpl: Thanks, that shouldn't change a lot, but let's make every bit count :)

https://jsperf.com/regexp-indexof-perf

src/models/StyledComponent.js
philpl: Nope, this is actually a regression, that I accidentally introduced. To compare with the old creation code:

```js
let warnTooManyClasses		
if (typeof process !== 'undefined' && process.env.NODE_ENV !== 'production') {		
  warnTooManyClasses = createWarnTooManyClasses()		
}
```

src/models/StyledComponent.js
wmertens: So this is what the react-router guys have to say on the subject: https://github.com/ReactTraining/react-router/issues/4629#issuecomment-284202921

src/models/ComponentStyle.js.orig
```diff
@@ -0,0 +1,71 @@
...
+     * */
+    generateAndInjectStyles(executionContext: Object) {
+      const flatCSS = flatten(this.rules, executionContext)
+      const hash = hashStr(this.componentId + flatCSS.join(''))
+
+      if (inserted[hash] === undefined) {
+        const selector = nameGenerator(hash)
+        inserted[hash] = selector
+
+<<<<<<< HEAD
```
wmertens: Ouch  :)

src/models/StyledComponent.js
philpl: You can raise an issue and we can discuss it separately. However I don't think there's a better solution that makes sense for what the theming does

src/models/ComponentStyle.js.orig
philpl: Oops! :D

evan-scott-zocdoc:
Side comment, there is a potential performance benefit from changing the implementation of https://github.com/styled-components/styled-components/blob/d096d6126b606a5c413e353d2d6cedf2309a912d/src/models/AbstractStyledComponent.js#L6 to use `PureComponent` instead

philpl:
@evan-scott-zocdoc That is unfortunately not quite the right way. First of all, it will change the behaviour for the user, and unexpected that is. And also, it doesn't improve the performance of the library itself. It's the responsibility of the user to decide when to turn to reference equality checks for perf optimisations.

The more clever approach is to not trigger state changes, when it's not necessary, but that's not really the scope of this PR. Btw: https://twitter.com/dan_abramov/status/820668074223353858

philpl:
@mxstbr possibly! Next I want to see if there are more obvious things, maybe even related to behaviour as I've said above

evan-scott-zocdoc:
@philpl how are you generating those benchmark stats? would be awesome to just have an npm script we could run when making changes to the library to quickly evaluate perf differences

philpl:
@evan-scott-zocdoc With https://github.com/styled-components/benchmark

I've split up some tests into more granular ones though. Once we do #588, it's going to live in the main repo.

PR #636, opened 2017-04-01
Ramitos attrs with style

geelen:
Updating #509 against current v2.


src/constructors/css.js
```diff
@@ -3,6 +3,8 @@ import interleave from '../utils/interleave'
 import flatten from '../utils/flatten'
 import type { Interpolation, RuleSet } from '../types'
 
-export default (strings: Array<string>, ...interpolations: Array<Interpolation>): RuleSet => (
-  flatten(interleave(strings, interpolations))
-)
+export default (strings: Array<string>, ...interpolations: Array<Interpolation>): RuleSet => {
+  const rules = flatten(interleave(strings, interpolations))
+  rules.isStyledBlock = true
```
geelen: @ramitos this is where the flow error is happening. I'm not 100% sure why you need to do this though?

src/models/StyledComponent.js
```diff
@@ -64,7 +69,27 @@ export default (ComponentStyle: Function) => {
...
-        return componentStyle.generateAndInjectStyles(executionContext)
+        const css = componentStyle.generateAndInjectStyles(executionContext)
+
+        const inlineRules = Object.keys(this.attrs).reduce((styles, name) => {
+          if (!Array.isArray(this.attrs[name])) {
+            return styles
+          }
+
+          if (!this.attrs[name].isStyledBlock) {
+            return styles
```
geelen: @ramitos can you explain why this needs to be handled differently here? Shouldn't it just be able to come through as an array?

src/constructors/css.js
```diff
@@ -3,6 +3,9 @@ import interleave from '../utils/interleave'
 import flatten from '../utils/flatten'
 import type { Interpolation, RuleSet } from '../types'
 
-export default (strings: Array<string>, ...interpolations: Array<Interpolation>): RuleSet => (
-  flatten(interleave(strings, interpolations))
-)
+export default (strings: Array<string>, ...interpolations: Array<Interpolation>): RuleSet => {
+  const rules = flatten(interleave(strings, interpolations))
+  // $FlowFixMe
+  rules.isStyledBlock = true
```
philpl: I don't really like this ☹️ Isn't this also going to break v8 optimisation?

philpl:
I don't really see why this should be this entangled with the "main" API. 😢  Isn't there a way to make this more separate?

mxstbr:
Since there's been no work on this for ages, I'm going to close it out.

PR #638, opened 2017-04-01
fix: prevent leaking the innerRef prop to wrapped components

### LINKED ISSUE #592: innerRef is leaked as a prop to the wrapped component
## Version
1.4.4

## Reproduction
https://www.webpackbin.com/bins/-KfSW7ArY2ALUjC5qswH

## Steps to reproduce
Wrap any component in `styled()` and use the `innerRef` callback to access it.

## Expected Behavior
`innerRef` should not be leaked as a prop to the wrapped child component.

## Actual Behavior
`innerRef` is leaked as a prop to the wrapped child component.

evan-scott-zocdoc:
As you can see, this doesn't break `innerRef` - it just doesn't leak the prop.

Closes #592

### DIFF
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -98,7 +98,7 @@ export default (ComponentStyle: Function) => {
         propsForElement.className = [className, generatedClassName].filter(x => x).join(' ')
         if (innerRef) {
           propsForElement.ref = innerRef
-          if (isTag(target)) delete propsForElement.innerRef
+          delete propsForElement.innerRef
         }
 
         return createElement(target, propsForElement, children)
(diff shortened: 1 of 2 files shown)


philpl:
I don't think we want to change this in `master` and release a new version. It should be sufficient to fix it in `v2`. I've already got my eye on that since there are ~3 issues related to this. 🌦

evan-scott-zocdoc:
@philpl Why not? Without a concrete release date for v2 it could be a while until people are able to upgrade.

mxstbr:
Sidenote: It'll be a month at max, and the upgrade will be pretty painless. I'd also rather merge this into v2. 

Review: I just had a revelation, this should attach a `ref` to the underlying component rather than passing innerRef through?!

evan-scott-zocdoc:
Closed in favor of https://github.com/styled-components/styled-components/pull/639 as requested.

PR #639, opened 2017-04-01
fix: prevent leakage of innerRef to wrapped child components

evan-scott-zocdoc:


k15a:
Isn't the testRef pointing to WrappedStyledComp?

philpl:
I fixed the test so that it checks for the ref to be the inner component, and also discovered that the `className` prop was overwritten in the same way. Btw both bugs are courtesy of me and a lack of sleep. You're welcome 😆 

source

PR #640, opened 2017-04-02
Add `no-parser` entry to the Rollup build

YasserKaddour:
This is my attempt to make @philpl valiant effort on the prepossessing feature benefit the Rollup goodness.
The parsed size shrinked from `32.6kB` to `23.9kB`, and the gziped size from `15.2kB` to `8.4kB` :tada:

However, I have no idea how to make `webpack` aware of the es module, since the `module` field in `package.json` is only for the main entry. I tried finding some online resource with no success, this my first time dealing with npm packages, any guidance on this will be appreciated. 
To make it work I changed `module.exports = require('./lib/no-parser')` to `module.exports = require('./dist/no-parser')` in `/no-parser.js`, but this is not what we want.

Thank you all, for this awesome library.


mxstbr:
Oh wow this looks awesome! 😍  @Rich-Harris @philpl mind giving this a good look?

Also /cc @TheLarkinn, maybe he has clues which entry point we'd use in `package.json` for this.

philpl:
This looks really great! Thanks a bunch!

I'd imagine that for the ES imports entrypoint, we'd have to create a second no-parser.mjs, but that probably doesn't work yet

rollup.config.js
```diff
@@ -65,10 +84,12 @@ const plugins = [
   json(),
 ]
 
-if (prod) plugins.push(uglify(), visualizer({ filename: './bundle-stats.html' }))
+if (prod) {
+  if (noParser) { plugins.push(uglify(), visualizer({ filename: './bundle-stats-no-parser.html' })) } else { plugins.push(uglify(), visualizer({ filename: './bundle-stats.html' })) }
+}
 
 export default {
-  entry: 'src/index.js',
```
philpl: Can you put the entry ternary closer to here or inline? Easier to read i.e. The reader doesn't have to scroll up

rollup.config.js
```diff
@@ -65,10 +84,12 @@ const plugins = [
   json(),
 ]
 
-if (prod) plugins.push(uglify(), visualizer({ filename: './bundle-stats.html' }))
+if (prod) {
+  if (noParser) { plugins.push(uglify(), visualizer({ filename: './bundle-stats-no-parser.html' })) } else { plugins.push(uglify(), visualizer({ filename: './bundle-stats.html' })) }
```
philpl: Can you swap out the if statement for a ternary as the filename value?

rollup.config.js
```diff
@@ -12,18 +12,37 @@ import visualizer from 'rollup-plugin-visualizer'
...
-[
-  { dest: 'dist/styled-components.min.js', format: 'umd' },
-] :
-[
-  { dest: 'dist/styled-components.js', format: 'umd' },
-  { dest: 'dist/styled-components.es.js', format: 'es' },
-]
+const entry = noParser ? 'src/no-parser/index.js' : 'src/index.js'
+
+let targets
```
philpl: This seems a bit convoluted as well.

package.json
```diff
@@ -12,7 +12,7 @@
     "build:flow": "flow-copy-source -v -i '{**/test/*.js,**/*.test.js}' src lib",
     "prebuild:umd": "rm -rf dist/*",
     "prebuild:dist": "rm -rf dist/*",
-    "build:dist": "rollup -c && rollup -c --environment PRODUCTION",
+    "build:dist": "rollup -c && rollup -c --environment PRODUCTION && rollup -c --environment NOPARSER && rollup -c --environment PRODUCTION,NOPARSER",
```
philpl: Maybe we should write a script for this or break it up? @mxstbr

philpl:
@YasserKaddour awesome work! I'd like to change a couple of things keep the config clean. I can do it myself as well, if you're busy. :+1: :tada:

YasserKaddour:
Thank for the feedback :+1: , I didn't know about `.mjs` I will look into it.

Rich-Harris:
LGTM. I'm not sure I understand the question about the entry point?

YasserKaddour:
@Rich-Harris Right now with webpack my app bundles with `/lib/no-parser/index.js` as specified in `/no-parser.js`, not with ES module in `/dist/no-parser.es.js`. 
Whereas, with the main entry point `"main": "lib/index.js"`, webpack is able to find the equivalent ES module because `"module": "dist/styled-components.es.js"` is specified in `package.json`
I am sure I am doing something wrong, I would appreciate your help, Thanks.

philpl:
@YasserKaddour I took the freedom to adjust the script names for the `dist` tasks, but apart from that it's ready to go.

*But* and this is something I've thought about for the last days, what if a library, that builds on styled-components, decides to use the preprocessing? Or what if a library doesn't use it, but a project uses it and includes said library?
If this happens, then we're dead in the water, in the bundling sense, since a project's bundle will then include both of our bundles: no-parser and main.

This is already a problem, and this PR won't change anything i.e. make it worse nor better, but it's something that I think is worth addressing here.

cc/ @mxstbr @geelen

Rich-Harris:
@YasserKaddour you're not doing anything wrong, this is basically an unsolved problem at the moment. Node's proposed solution is to use .mjs files alongside .js files, while an [alternative proposal](https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md#transitioning-libraries) suggests using a `modules.root` key in package.json.

Until one of those solutions wins, the best you can do is document it well.

mxstbr:
That is indeed an issue @philpl, I hadn't thought of that. God dammit.

What do we do in that case?

YasserKaddour:
@Rich-Harris Thank you, I hope they will decide on this soon.

In the meanwhile I come up with a little hacky solution by adding a `module` flag on the babel plugin that changes `require('styled-components/no-parser')` to `require('styled-components/dist/no-parser.es')` 

This doesn't solve @philpl problem, but work fine if you don't use plugins. WDYT?

YasserKaddour:
@philpl These import could be made into a separate bundle / repo like `styled-component-core`:
```
/* Import singletons */
import generateAlphabeticName from '../utils/generateAlphabeticName'
import styleSheet from '../models/StyleSheet'

/* Import singleton constructors */
import _StyledComponent from '../models/StyledComponent'
import _ComponentStyle from '../models/ComponentStyle'
import _GlobalStyle from '../models/GlobalStyle'
import _styled from '../constructors/styled'
import _keyframes from '../constructors/keyframes'
import _injectGlobal from '../constructors/injectGlobal'
import _constructWithOptions from '../constructors/constructWithOptions'

/* Import components */
import ThemeProvider from '../models/ThemeProvider'

/* Import Higher Order Components */
import withTheme from '../hoc/withTheme'
```
Maybe when you will switch to Lerna. It wouldn't fix the problem entirely but it would avoid import duplication.

philpl:
As @geelen had an alternative proposal for addressing this first, I'm going to carefully close this for now :) We can reopen it, if we consider it again, but this might be more of the v3 territory, where we'll integrate preprocessing and static css stuff into the core lib, which will make the split obsolete. This is experimental after all, and I don't think we have to reach some kind of interoperability for now

PR #642, opened 2017-04-03
Add nested `innerRef` for Styled Components

### LINKED ISSUE #629: [v2] inner refs to nested styled components doesn't work
## Version
2.0.0-10

## Reproduction
https://www.webpackbin.com/bins/-KgV_iHd-TqUwqGz5b8J

## Steps to reproduce
Make a styled StyledComponent, and try to get the inner ref. Doesn't return the HTML element.

```js
const StyledComponent = styled.div`
  background: #0c0;
`
const StyledStyledComponent = styled(StyledComponent)`
  background: #c00;
`

// Then in some component
<StyledStyledComponent innerRef={node => this.node = node} />
```

## Expected Behavior
Expecting `this.node` to be a `HTMLDivElement`

## Actual Behavior
`this.node` is a `StyledComponent`

## Potential solution?
I looked through the source and I'm guessing the problem is [here](https://github.com/styled-components/styled-components/blob/master/src/models/StyledComponent.js#L99)?

```diff
- if (innerRef) {
+ if (innerRef && target.name !== 'StyledComponent') {
  propsForElement.ref = innerRef
  if (isTag(target)) delete propsForElement.innerRef
}
```
It looks like the innerRef is not propagated down to nested StyledComponents. I was able to test this locally and it worked, but not PR worthy as `target.name` seems to be unreliable for comparison. Someone with better understanding of the internals might be better suited to take a look at this.

philpl:
As a continuation of #639 
Fix #629 

- Pass the `innerRef` to the nested target if said target is a Styled Component.
- Replace the old `innerRef` tests that weren't actually doing much

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file. If a contri
 - Fix defaultProps used instead of ThemeProvider on first render [@k15a](https://github.com/k15a), restored.
 - Refactor StyledComponent for performance optimization.
 - Prevent leakage of the `innerRef` prop to wrapped child; under the hood it is converted into a normal React `ref`. (see [#592](https://github.com/styled-components/styled-components/issues/592))
+- Pass `innerRef` through to wrapped Styled Components, so that it refers to the actual DOM node. (see [#629](https://github.com/styled-components/styled-components/issues/629))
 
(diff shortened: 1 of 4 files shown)


dmitrika:
Thank you!

PR #648, opened 2017-04-03
Add some files to .npmignore

YasserKaddour:
I would add CONTRIBUTING.md and CODE_OF_CONDUCT.md, but it's your call.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -12,3 +12,5 @@ dangerfile.js
 yarn.lock
 .editorconfig
 .npmignore
+CODE_OF_CONDUCT.md
+CONTRIBUTING.md
--- a/.npmignore
+++ b/.npmignore
@@ -7,3 +7,8 @@ rollup.config.js
 .eslintignore
 .babelrc
 example/
+bundle-stats*
+dangerfile.js
+yarn.lock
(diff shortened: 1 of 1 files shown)


YasserKaddour:
Since  CONTRIBUTING.md and CODE_OF_CONDUCT.md where only added 3 days ago, I will go ahead and add them.

philpl:
Both aren't necessary, so this looks good to me, but do we want to use the explicit "files" list in package.json instead maybe? CC @mxstbr

YasserKaddour:
I think explicit "files" would be better, If that's OK with you I will do that instead.

YasserKaddour:
Also, from what I could see in other packages `docs`, `flow-typed` and `src` folders should get the axe, a let only `dist` and `lib` remain.

geelen:
Keep the src folder! It's such a pain when you're using libraries and you
only have transpired source to read through without jumping to GitHub. Same
goes for docs, next time you're hacking without internet you'll thank me :)

philpl:
Agreed! :smile: Keep all three

YasserKaddour:
@geelen You're right, I will make another PR tomorrow.

PR #660, opened 2017-04-06
Add other solutions section

kentcdodds:
I put a section like this on all my repositories because I think it's pretty handy for people trying to make a decision about what solution fits their use cases and preferences best.


k15a:
A simple question aside. What's the benefit of a review request? Is there a special page where I can see all PR I got requested to review? This would be really handy. But as far as I know it's only another notification isn't it?

kentcdodds:
> as far as I know it's only another notification

That's me as well :man_shrugging:

existentialism:
There's https://github.com/pulls/review-requested

nerdchacha:
Also, here
image

kentcdodds:
ping?

PR #661, opened 2017-04-07
Add dangerfile.js to .npmignore

denisw:
This file doesn't seem useful to publish. Also, it messes up support for Flow, which complains about not finding the `danger` module unless `dangerfile.js` is explicitly added to the `[ignore]` section of `.flowconfig`.


k15a:
There is already a PR which is doing that: https://github.com/styled-components/styled-components/pull/655

PR #662, opened 2017-04-07
Add FlatList, SectionList & VirtualizedList support

Kureev:
As far as 0.43 is out, it would be great to support new list views (ref: https://facebook.github.io/react-native/blog/2017/03/13/better-list-views.html)

P.S. I didn't add TS support due to missing validations in DefinitelyTyped repo


Kureev:
cc @patrick91

Kureev:
Thanks 🙏🏻

PR #663, opened 2017-04-07
Remove glamor from our dependencies

mxstbr:
This'll make our npm installs much faster-we were downloading glamor for
a single file.


philpl:
@mxstbr Changelog entry?

PR #668, opened 2017-04-08
migrate from the deprecated React.propTypes to prop-types package

### LINKED ISSUE #667: React v15.5.0 PropTypes deprication
## Version
1.4.4

## Reproduction
When using react 15.5.0, styled-components cause a deprecation error to come up.
## HOW TO FIX
 This is due to the newest react release from an hour ago [that can be viewed here](https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html). All you need to do is use a separate library called prop-types that works just like React.PropTypes.

YasserKaddour:
Fix #667

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file. If a contri
 
 ### Changed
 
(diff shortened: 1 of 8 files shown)


YasserKaddour:
@philpl done in #670 :v:

Issue #386, opened 2017-01-18
Discuss SSR API and methods


thisguychris:
I just want to kick off a discussion to finalise our SSR API.

So API came from the comments on the #214 

```JS
const { css, html } = StyleSheet.collect(() =>
  ReactDOMServer.renderToString(<App />)
);
```

Please feel free to suggest what the API will look like, (let the bikeshedding begin!)

Also if we're already doing de-duping in SSR level, do we even need reset()? (related issue #378)  Or would that just be a convenience function since we don't need it anymore?

cc @mxstbr / @geelen

acjay:
One thing I would throw out there is that it would be very useful to be able to have separate bundles of CSS. That way, you could target styles to a `style` tag or to a separate stylesheet. When I worked for a publisher, we used this type of optimization to achieve very fast initial renders, with nonessential styles deferred until later.

I can ticket this separately, if that would be helpful. It would require some sort of ability to annotate at the component or CSS level where rulesets would be targeted, and an API for collecting the separate bundles (which might be a `collect` variant).

mxstbr:
This happens either way @acjay. With styled-components, only the styles of components that are actually used on the initial render are injected 👍 

Generating `.css` files for server-side rendered apps would not be complicated to add, but should not live in core. (it's more or less just a `fs.write(css)` away) We're going to work on a babel or webpack plugin to make this possible for non-ssr apps but that's a different issue. (see #59)

dfrankland:
@thisguychris, `reset()` is great for doing testing. Considering that tagged templates have logic in them, that logic may need to be tested.

mxstbr:
That's a great point actually @dfrankland, I hadn't even considered that use case!

nicksrandall:
We will also need to keep track of the classNames that are generated so we can rehydrate them on the client. 

For example, this is how glamor does it... https://github.com/threepointone/glamor/blob/master/docs/server.md

camsjams:
Looks good to me

nicksrandall:
Do we also want to be able to export class names here ( in addition to CSS and HTML) like Aphrodite and Glamor does so that we can consume them on the front end to make sure things stay in sync or are just going to rely on the babel plugin to ensure deterministic class names?

xcoderzach:
[...]

PR #679, opened 2017-04-11
Rework stylesheet for ssr

geelen:
So, this is the all-singing, all-dancing StyleSheet to end all stylesheets. So far I have rehydration working, as well as deterministic local & global stylesheets. The serialised format looks like this:

```html
<style data-styled-components
       data-styled-components-is-local="false">
  /* sc-component-id: sc-global-1455077013 */
 html { font-size: 16px; }
  /* sc-component-id: sc-global-557410406 */
 body { background: papayawhip; }
</style>
<style data-styled-components="gZjIar dzbzmI"
       data-styled-components-is-local="true">
 /* sc-component-id: ONE */
 .ONE {}
 .gZjIar { color: blue; }
 /* sc-component-id: TWO */
 .TWO {}
 .dzbzmI { color: red; }
</style>
```

What this lets us do is figure out that there have been two `injectGlobal` calls and two StyledComponents generated. When the `BrowserStyleSheet` is created, it reads anything in the DOM with a `data-styled-components` attribute and builds up a picture of the existing styles.

In the above case, that means four componentIds `sc-global-1455077013`, `sc-global-557410406`, `ONE` and `TWO`. It also reads the classnames `gZjIar` and `dzbzmI` so it knows not to generate those in future.

**What this means** is that when your app boots, whether or not it's using the `no-parser` stuff, ***it won't inject CSS into the head that it doesn't need to***. But, when a dynamic style comes along, it all still works.

Here's my todo list for this feature (which I'll be updating as I go):

- [x] strip back as close to a single `inject` API call as possible.
- [x] switch all usages of the existing `StyleSheet` across
- [x] handle converting from the initial `<style>` tag with a single text node child to a one-node-per-component so we can use `appendData` to inject new CSS in the right order.
- [x] add a test for keyframes
- [x] make a ServerStyleSheet that generates this exact format
- [x] make the StyleSheetProvider pass the instance through context
- [x] build the SSR api described in #386
- [x] construct the singleton + cloning API on the server
- [x] delete the old code, including Glamor's sheet (now vendored)
- [ ] ~~hoist `@import` statements to the top of style tags (threepointone/glamor#151)~~ Gonna raise this separately
- [x] actually keyframes don't need to be global, make them local
- [x] stop sending down empty classnames for unused components
- [x] dispatch `injectGlobal` and `keyframes` from the singleton to each instance

Lots still to do, but this feels like a really good feature to make SSR really well integrated into Styled Components. Will try to get this finished as soon as I can!


geelen:
**Ready for Review!**

This is a big PR, but if you get a chance to go through it @mxstbr **we now have concurrent SSR with rehydration!!** 🎉 

The API I've gone with is as follows:

```js
const sheet = new ServerStyleSheet()
const html = renderToString(sheet.collectStyles(<Example/>))
const css = sheet.getStyleTags()
```

I think anything shorter is potentially misleading. I'm currently stuck on getting the example to rehydrate properly but I think it has to do with the browser-based babel transform thing, which is annoying. But I think that can come in as a separate PR that someone else could pick up.

To play around, do:

```
yarn dev
open http://localhost:3000/ssr.html
```

I'd love to get this merged as soon as possible so we can cut a release and get some feedback. There's a lot of edge cases around deterministic rehydration so I expect there might be a few bugs to fix, but the more eyes on this the better.

philpl:
Wow! This is amazing news 🙌🙌

I'll go through it this eve

src/constructors/test/css.test.js
```diff
@@ -1 +1,2 @@
 // @flow
+it('fixme', () => {})
```
mxstbr: Rogue old test?

src/constructors/test/element.test.js
```diff
@@ -55,3 +55,5 @@
 //     expect(renderedComp.props.className).toEqual(`${className} ${generatedClassname}`)
 //   })
 // })
+
+it('fixme', () => {})
```
mxstbr: Second rogue old test?

src/models/AbstractStyledComponent.js
```diff
@@ -16,4 +17,5 @@ export default class AbstractStyledComponent extends Component {
 
 AbstractStyledComponent.contextTypes = {
   [CHANNEL]: PropTypes.func,
+  StyleSheet: PropTypes.instanceOf(StyleSheet),
```
mxstbr: Do we want to make this more unique to avoid collisions with other libraries/apps?

src/models/StyleSheet.js
```diff
@@ -1,35 +1,142 @@
 // @flow
+import BrowserStyleSheet from './BrowserStyleSheet'
+import ServerStyleSheet from './ServerStyleSheet'
 
-/* Wraps glamor's stylesheet and exports a singleton for the rest
-*  of the app to use. */
+export const SC_ATTR = 'data-styled-components'
+export const LOCAL_ATTR = 'data-styled-components-is-local'
```
mxstbr: Any reason to make this `data-styled-components` vs `data-s-c`?

src/test/warnTooManyClasses.test.js
```diff
@@ -1,14 +1,13 @@
...
 import { shallow } from 'enzyme'
-import expect from 'expect'
 
 import { resetStyled, expectCSSMatches } from './utils'
 import styleSheet from '../models/StyleSheet'
 
 let styled
 
-describe('warn too many classes', () => {
+describe.skip('warn too many classes', () => {
```
philpl: Is this supposed to be still disabled?

src/utils/extractCompsFromCSS.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+const SC_COMPONENT_ID = /^[^\S\n]*?\/\* sc-component-id:\s+(\S+)\s+\*\//mg
```
philpl: Nice!

src/models/BrowserStyleSheet.js
```diff
@@ -0,0 +1,123 @@
...
+class BrowserTag implements Tag {
+  isLocal: boolean
+  components: Map<string, Object>
+  el: HTMLElement
+  ready: boolean
+
+  constructor(el: HTMLElement, isLocal: boolean, existingSource: string = '') {
+    this.el = el
+    this.isLocal = isLocal
+    this.components = new Map(extractCompsFromCSS(existingSource).map(obj =>
```
philpl: Actually, found this thing here. `Map` is not polyfilled by us, so in environments where the user is not polyfilling it, it might cause issues for a wide variety of old browsers.

src/models/BrowserStyleSheet.js
```diff
@@ -0,0 +1,123 @@
...
+    this.el.parentNode.replaceChild(newEl, this.el)
+    this.el = newEl
+  }
+}
+
+/* Factory function to separate DOM operations from logical ones*/
+export default {
+  create() {
+    const tags = []
+    const names = new Set()
```
philpl: Same for `Set`, I guess

philpl:
@geelen (inb4) Sorry, for messing with your code :wink:

I removed all instances of `Map` and `Set` for now. We can benchmark it against the commit before, if we need to, and evaluate whether drawbacks of a ponyfill/polyfill for them are worth it.

I'd say if you're happy with this branch and @mxstbr's other question around the context is resolved, we can ship this 🛩  :smile:

Another small thing, once this lands SSR will be incompatible with next.js. We should probably raise an issue over on their repo, to discuss our approach.

src/models/StyleSheet.js
philpl: It doesn't really have to be this short, I'd say. This name is pretty unambiguous

src/models/ServerStyleSheet.js
```diff
@@ -67,8 +80,11 @@ export default class ServerStyleSheet {
   }
 
   getStyleTags(): string {
-    clones.delete(this.instance)
-    this.closed = true
+    if (!this.closed) {
+      clones.splice(clones.indexOf(this.instance), 1)
```
geelen: :'( I miss `Set`

geelen:
> Another small thing, once this lands SSR will be incompatible with next.js. We should probably raise an issue over on their repo, to discuss our approach.

Why is that? Just the public API?

geelen:
Thanks for your help! I've tweaked a bunch of things, changed up the Context key, exported StyleSheetManager. I'm gonna merge this and cut a new v2 release.

mxstbr:
Before you cut a new release, this will break a lot of folks. We should add documentation.

geelen:
It... shouldn't break anything except users of the current SSR API, right?

geelen:
@mxstbr @philpl just deployed this to https://frontend.center. I'm using react-snapshot so not actually hitting the SSR api yet but I'm getting server-rendered styles with full rehydration 👌

mxstbr:
> It... shouldn't break anything except users of the current SSR API, right?

Yeah but people are using that and they'll be pissed if we don't at least tell them what the new API is!

philpl:
@geelen Specifically, the API in next.js doesn't let you wrap another provider around the app. Or at least I haven't figured out how to do it in a nice way.

oscar-b:
@geelen Any chance of getting a the styles as a component as well as text? Wouldn't that make it possible to avoid `dangerouslySetInnerHTML`?

oscar-b:
react-helmet for instance has both toString() and toComponent() for similar reasons.

oscar-b:
Proof of concept: https://github.com/oscar-b/styled-components/commit/3ba3a36f3ff5e685a29097a81bf34cf874f610b0

I don't know Flow enough to do it properly, and there are no tests. It works for both global and local styles, by just doing `{sheet.getStyleElement()}` inside `<head></head>`

ramesius:
Has these changes made it into a release on npm yet?

ismay:
@ramesius @mxstbr As per your comment [here](https://github.com/styled-components/styled-components/issues/386#issuecomment-297913181) I thought that this code should be available under the `rc` tag on npm, but it seems that that tag isn't available? Also the 2.0.0-15 release doesn't contain this pr either it seems.

geelen:
Sorry I got stuck on something and haven't pushed it. Gimme a sec!

oscar-b:
@geelen Look at merging #736 as well first?

geelen:
Thanks @oscar-b, I just released one without it, but will take a look at #736 now as well to see if it's ready to go

simenbrekken:
I just ran into an edge-case with the previous API which I couldn't get past even after updating to RC.

The project in question is a Lerna project with common components extracted into a `common` package which is referenced by other packages. All packages have `styled-components` listed as a dependency which in the problem lies:

> Warning: Failed context type: Invalid context `__styled-components-stylesheet__` of type `StyleSheet` supplied to `styled.nav`, expected instance of `StyleSheet`.

This is caused by `lib/models/StyleSheet` exporting a singleton which isn't shared across my packages.

Is there a way to explicitly provide `StyleSheet` via context?

zentuit:
I'm migrating to styled-components v2 for a couple of our sites and I'm getting the same warning as @simenbrekken. We also have common components extracted out into a a couple of packages. Is there any resolution or workaround for this?

etburke:
I'm also experiencing this context type warning issue. Anyone have any further insights on this?

PR #681, opened 2017-04-11
[v2] Minify some parts of the library

philpl:
- Switch over to `babel-preset-env`
- Enable loose mode (learner class declarations)
- Strip propTypes in prod (we didn't have a lot anyway though)
- Remove validAttr util in prod

`31.8kB raw, 12kB gzip` => `25kB raw, 9.1kB gzip`


philpl:
I thought about this again, and I think the only props that we are concerned about are stuff like `width` or `color`, which would actually have some behaviour in HTML. But all these attributes that are actually doing stuff are in our whitelist anyway, so this should be safe.

Another thing is that some props would be stringified, and I'm not sure if we'd incur a performance penalty by passing them through

philpl:
@geelen @mxstbr We might be able to revive the attribute validation optimisation, i.e. removing attr-validation in production. This is because React filters them out anyway. It doesn't even set & sringify them. Quote:

> We are currently warning about “unknown properties” in the DOM.
> For example if you render:
> `<div foo="bar" />`
> you’re going to see a warning from React, and **foo won’t actually turn up in the DOM**.

So I'd say, it's safe in production

https://github.com/facebook/react/issues/9477

gaearon:
To be clear https://github.com/facebook/react/issues/9477 is all about changing this in the 16, and stringifying + outputting them.

philpl:
@gaearon Yea, I'm wondering if passing through props will still be safe in 16. Probably not anymore, I'd imagine?

gaearon:
I don’t know. But facebook/react#9477 is aimed exactly at making it unsafe in 16.

PR #685, opened 2017-04-12
replace filter-forEach combo in favor for single forEach

mephju:
When checking for valid props I think it's more efficient to keep the number of iterations to a minimum. There could potentially be a lot of props passed to a styled component. Therefore I replaced
```javascript
  Object.keys(this.props)
          .filter(propName => !isTag(target) || validAttr(propName))
          .forEach(propName => {
            propsForElement[propName] = this.props[propName]
          })
```

with
```javascript
Object.keys(this.props)
          .forEach(propName => {
            if (isTag(target) || !validAttr(propName)) return
            propsForElement[propName] = this.props[propName]
          })
```

I would have actually liked to use reduce here but your eslint rules forbid assigning props to the accumulator. Anyway, thanks for considering this change :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file. If a contri
 
 ### Changed
 
+- Exchanged filter-forEach combo in favor of single forEach.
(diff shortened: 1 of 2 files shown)


philpl:
Sorry, this has already been corrected in v2 😄 Thanks for the PR anyway! :tada:

https://github.com/styled-components/styled-components/blob/v2/src/models/StyledComponent.js#L148

PR #692, opened 2017-04-13
Add prop-types package as an external for the rollup build

YasserKaddour:


philpl:
This is only against v2, but in general, it's not that pressing for another release, I believe.

YasserKaddour:
@philpl I amended  the commit to follow the React team [recommendation](https://github.com/reactjs/prop-types#how-to-depend-on-this-package); to add the package as a dependency and a peer dependency.

k15a:
I don't think we should move prop-types to peer dependencies. And if I understand it correctly the React team thinks so as well.
>For libraries, we also recommend leaving it in dependencies:
>```JSON
>"dependencies": {
>  "prop-types": "^15.5.7" 
>},
>"peerDependencies": {
>  "react": "^15.5.0" 
>}
>```

YasserKaddour:
@k15a My bad, I rebased and amended the commit. Thanks.

philpl:
I had a think about this, and I believe if we follow the recommendation of it not being a peer dependency, it also shouldn't be an external. Sorry to keep you waiting, @YasserKaddour 😞

YasserKaddour:
@philpl I think there is a misunderstanding here, partly due to my clumsy PR, sorry about that :sweat: .
 
Adding `prop-types` as external in rollup is fine, because rollup will add `import PropType from 'prop-types'` to the flat bundle, and `prop-types` being a dependency any module manager will pick it up.

Bundling 'prop-types' is not doing anybody any favor, most likely it will be already imported by another library like `react-router`, and this will result to `prop-types` being bundled two times.

I hope you understand my reasoning and reconsider, Thank you :v:

k15a:
Hey @YasserKaddour,

we need to create another distribution folder that has all the files compiled by Babel (without the modules transform) but not bundled by Rollup.
We will point Rollup and WebPack to use these files instead of the bundle. This way WebPack and Rollup can take care of any minification on the user site and we don't have to consider that.

After that the bundle is only used when you want a single file with all the dependencies that styled-components is using. For example when you insert styled-components via script tag you don't want to insert the 8 other dependencies as well.

Issue #352, opened 2017-01-04
Consider adding a warnings for incorrect tagged template literal usage in development


cesarandreu:
I was adding styled-components to an app and global styles weren't showing up. There weren't any errors and it wasn't immediately obvious what I was doing wrong. It turned out I was using `injectGlobal` wrong:

```js
// wrong: calling injectGlobal as a function. oops.
injectGlobal(`html { color: red; }`)

// correct: using a tagged template literal
injectGlobal`html { color: red; }`
```

The first argument of tagged template literals includes a [`raw` property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Raw_strings). Adding a check in development to prints out an error when `strings.raw` is missing would help catch this kind of mistake. I'm not sure how frequently this happens, but considering the relative rarity of tagged template literal usage, I figure it wouldn't hurt.

```js
if (process.env.NODE_ENV !== 'production' && !('raw' in strings)) {
  console.error('...')
}
```

vdanchenkov:
I would suggest to check it slightly different way. `injectGlobal` and other helpers expect array in first argument. So, let's check for it.

Something like.
```JS
if (process.env.NODE_ENV !== 'production' && !(Array.isArray(strings))) {
  console.error('...')
}
```

On the other hand, we could support incorrect usage? 
```JS
injectGlobal(`html { color: red; }`) 
// will be equivalent to
injectGlobal([ `html { color: red; }`])
// or 
injectGlobal`html { color: red; }`
```

mxstbr:
This seems reasonable. Whatever we can do to avoid people adding bugs to their apps 👍

iamsolankiamit:
I'll take this one. Just to clarify, we want to allow incorrect usage or show error when incorrectly used?

mxstbr:
Introducing an error now would be a breaking change, so let's just log a warning?

beautyfree:
Is it still open?

chengjianhua:
@maksugr Could you make a PR for your completed commits?

maksugr:
@chengjianhua it is right [here](https://github.com/styled-components/styled-components/pull/1202)

probablyup:
Yup

PR #698, opened 2017-04-15
Add warning messages for incorrect function calls

CompuIves:
For issue: https://github.com/styled-components/styled-components/issues/352

This adds warning messages for these function calls:

```js
styled.div(`
  html {

  }
`)
```

The error message will be:

```
Error: styled.div is called as a function.
It seems like styled.div is called as 'styled.div(`css`)', it is supposed to be called as 'styled.div`css`'
```

I added `sinon` to assert that `console.error` has been called, I'm not sure what your policy on this is. Also, it seems that the `yarn.lock` changed more than just for `sinon`. I tried doing a clean install with the old `yarn.lock`, but it still changed.

Very curious what you think of this approach 😄


k15a:
Thank you for the PR. We might want to wait until we migrated our tests to Jest. AFAIK @geelen is currently working on it.

philpl:
I don't think this is the right approach unfortunately. If we want to enforce our API — which makes sense — then we should put in an invariant that throws. A `console.error` is neither a warning per se, nor an error.

I believe we should just use `invariant` and throw, which is by the way easier to test as well. It's arguably the more popular way to do things, and we'd actually enforce it this way. https://github.com/reactjs/react-redux/blob/4d302257e3b361731f44b1f546e547ed578c8eec/src/components/connectAdvanced.js#L90

Also, I think the function name could be really explicit e.g. `invariantArray(arr)` or `enforceArray(arr)`, instead of just `validate`, which is pretty cryptic

CompuIves:
I agree with this,  we _could_ break applications that unknowingly call the functions the wrong way. In that case we'd need to do a major version bump. 

I'll also update the name, at first I used as a function to do more kinds of validations but this is indeed very cryptic.

philpl:
@CompuIves We could merge this into v2. In that case it would become part of the next major release 😄

CompuIves:
Updated PR based on the feedback!

src/constructors/test/styled.test.js
```diff
@@ -9,4 +10,28 @@ describe('styled', () => {
       expect(styled[domElement]).toExist()
     })
   })
+
+  it('should console.error the correct name for html tags when invalid arguments are passed', () => {
```
philpl: Small nit ;) The test names are still not updated

src/utils/enforce-array-arguments.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+import invariant from 'invariant'
+
+const generateInvalidFunctionError = (func) => (
+`Error: ${func} is called as a function.
```
philpl: I think this first line is not necessary, especially since it will prepend "Invariant Violation:" there, if I'm not mistaken?

```js
error.name = 'Invariant Violation';
```

CompuIves:
Good points, I updated the PR

philpl:
@mxstbr Any idea whether we still want this? If so we probably want it to compile away in prod

probablyup:
This is a better target for a linter, IMO.

PR #707, opened 2017-04-18
add support to jsdom

zvictor:
While trying to implement tests for my styled-components I was constantly facing the following error
> TypeError: Cannot read property 'cssRules' of undefined

I then managed to track its root as being the issue https://github.com/tmpvar/jsdom/issues/992.

Other browsers are also affected by this, not only jsdom, as can be seen at https://github.com/threepointone/glamor/issues/58.

Glamor has already fixed this with the commit https://github.com/threepointone/glamor/commit/98d4f44ba4335abd0e63bf946edff04ec51d3a41, and this PR brings the main change to styled-component.

----

This PR should fix https://github.com/styled-components/jest-styled-components/issues/2 and it could also be related to https://github.com/styled-components/styled-components/issues/124#issuecomment-273914589

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file. If a contri
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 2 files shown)


k15a:
Hey, thank you for the PR. If you could add a changelog entry I am happy to merge this. 🎉

zvictor:
there we go! :)

PR #710, opened 2017-04-19
Add innerRef support to withTheme HOC

philpl:


src/hoc/withTheme.js
```diff
@@ -31,6 +35,22 @@ export default (Component: ReactClass<mixed>) => class extends React.Component {
...
+    }
+
+    return (
+      <Component
+        theme={theme}
+        {...this.props}
+        innerRef={innerRef}
+        ref={ref}
+      />
+    )
```
geelen: Personally, I'd make that render method more like:

```js
const { innerRef } = this.props
const refs = isStyledComponent(Component) ? { innerRef } : { ref: innerRef }

return (
  <Component {...{theme, ...this.props, ...refs}}/>
)
```

philpl: I was trying to avoid the double-merge as a micro-optimisation, but it's definitely cleaner. Should I change it?

philpl: We could also extract a variable, like `_isStyledComponent` and write:

```js
<Component
  theme={theme}
  {...this.props}
  innerRef={_isStyledComponent ? innerRef : undefined}
  ref={_isStyledComponent ? undefined : innerRef}/>
```

PR #712, opened 2017-04-19
Hoist non-static properties on withTheme HOC

### LINKED ISSUE #596: Make withTheme hoist static properties

philpl:
Fix #596 

See for reference:

- https://github.com/reactjs/react-redux/issues/53
- https://github.com/reactjs/react-redux/issues/276

### DIFF
--- a/src/hoc/withTheme.js
+++ b/src/hoc/withTheme.js
@@ -6,7 +6,7 @@ import PropTypes from 'prop-types'
 import hoistStatics from 'hoist-non-react-statics'
 import { CHANNEL } from '../models/ThemeProvider'
 
-const wrapWithTheme = (Component: ReactClass<mixed>) => {
+const wrapWithTheme = (Component: ReactClass<any>) => {
   const componentName = (
     Component.displayName ||
     Component.name ||
--- a/src/hoc/withTheme.js
+++ b/src/hoc/withTheme.js
@@ -3,34 +3,49 @@
 
 import React from 'react'
 import PropTypes from 'prop-types'
+import hoistStatics from 'hoist-non-react-statics'
 import { CHANNEL } from '../models/ThemeProvider'
 
-export default (Component: ReactClass<mixed>) => class extends React.Component {
-  static contextTypes = {
-    [CHANNEL]: PropTypes.func,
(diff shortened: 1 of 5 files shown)


src/hoc/withTheme.js
```diff
@@ -3,34 +3,49 @@
...
+const wrapWithTheme = (Component: ReactClass<any>) => {
+  const componentName = (
+    Component.displayName ||
+    Component.name ||
+    'Component'
+  )
 
-  state: { theme?: ?Object } = {};
-  unsubscribe: () => void;
+  class WithTheme extends React.Component {
```
geelen: Is it possible to do `class WithTheme extends Component`? Would that mean statics are available by simple prototype inheritance?

philpl: That's not a good idea, since we'd be overriding some original logic. I'd stick to the HOC-pattern :wink:

philpl:
@brunolemos I understand you're upset and I apologise for stepping on your toes. Please do understand however that I've thought about this and opened a fresh PR since it's going to be merged against v2 and now intersects/conflicts with #710. I'm currently trying to move quickly towards the v2 release, and I thought it's safe to move your changes over to a new branch.

I'll add an attribution with your name to the changelog, and I hope this doesn't scare you off from contributing to SC in the future. Sorry again! 😅

brunolemos:
@philpl I was just finishing to update this PR with the original commit plus yours, but nevermind, you merged it.

philpl:
@brunolemos Hey, man. I'm sorry I went ahead, but I can't read your mind and guess what you're up to. Yes, this is unfortunate, but at the end of the day it's just code, so please don't be passive aggressive about this, because that doesn't help either of us, since it's not constructive.

I'm happy to buy you a beer and apologise in person, if we end up in the same city one day. We don't have to escalate this further and turn this into something bigger than it is. Sorry again, mate

brunolemos:
@philpl It's constructive in the way that my intention is that you don't do that again with other  contributor or a project member like me. Besides that, no problem, keep moving fast.

PR #717, opened 2017-04-20
[v2] Switch babel over to babel-preset-env

### LINKED ISSUE #684: Array.from is not polyfilled, but IE doesn't support it
## Version
1.0.5

## Reproduction

```
$ npm install styled-components
$ grep -r Array\.from node_modules/styled-components/dist
```

## Question
Array.from is not supported in IE. Without poly-fills, applications importing `styled-components` will have to implement polyfills in order to support IE. Is this intended behaviour, or can we update the `babel-preset-env`?

philpl:
- Switch to babel-preset-env in loose mode
- Remove all propTypes in production bundle
- Remove last occurrence of `Array.from`

Fix #684


kopax:
Hi @philpl, do you think it would be possible to deploy a fix for this when releasing v2?

https://github.com/styled-components/babel-plugin-styled-components/issues/33

I have a library bootstrap styled, made on styled that is almost 670kb. I am about to release but i  I can't get under 500kb and the library produce a warning. I am almost sure this would reduce the size enough for us. Thanks in advance.

philpl:
@kopax I've got a PR for it, I only have to finish it up, so will do 😀

@k15a do you have time to review this and #714? Max and Glen are busy and these two PRs are the only ones left for a potential v2 RC

src/models/BrowserStyleSheet.js
```diff
@@ -113,7 +113,12 @@ export default {
     const names = {}
 
     /* Construct existing state from DOM */
-    Array.from(document.querySelectorAll(`[${SC_ATTR}]`)).forEach(el => {
+    const nodes = document.querySelectorAll(`[${SC_ATTR}]`)
+    const nodesLength = nodes.length
+
+    for (let i = 0; i < nodesLength; i += 1) {
```
k15a: Any drawbacks of using a for of loop here?

philpl: There shouldn't be :) it's just a tiny bit faster and less hackier than converting that into an array

PR #718, opened 2017-04-20
Emoji class names (as a branch)

Un3qual:
Since you got rid of emoji class names, and some people still wanted, maybe you could merge this as a branch so people can have fun class names if they want.


philpl:
I'm not sure whether it was a good idea to open a PR, since:

- It will need to change for v2
- It won't be merged
- We'll need to eventually close it

Maybe you should write a Medium article, or a Gist for it instead? 😄

geelen:
Yeah sorry this can't really happen at the moment. I wanna improve the way we configure Styled Components so that switching this sort of thing in is a lot easier, and will hopefully restore emoji as an option then. I'll come back to this thread and ping you when that happens, @Un3qual

PR #720, opened 2017-04-20
keyframes better support older iOS (8.x) webkit

tyfoo:
When using `styled-components` in one of our projects we noticed that on older iOS devices (8.x, iPhone 6 and 6 Plus) animations didn't work.

After some investigation we've found http://stackoverflow.com/questions/9211261/css3-animation-not-working-in-safari and after some experimentation, the solution suggested in http://stackoverflow.com/a/34013139 worked.

This PR adds changes `keyframes` to generate both `@keyframes {name}` and `@-webkit-keyframes {name}`.


philpl:
LGTM :ship: Please note, that this is is already working in v2, since we've switched to stylis, which does vendor @keyframes for us

jesnil01:
👍  Thanks for this!

This is not only present on IOS devices but older webkit browsers on desktop as well. 

Looking forward for this in a future version!

PR #726, opened 2017-04-21
Use rimraf for Windows compatibility

oscar-b:
Windows doesn't understand `rm -rf`, use package `rimraf` for cross OS compatibility.


package.json
```diff
@@ -8,10 +8,10 @@
   "module": "dist/styled-components.es.js",
   "scripts": {
     "build": "npm run build:lib && npm run build:dist && npm run build:flow",
```
k15a: Is Windows understand `&&`?

oscar-b: Yes, no problem there.

PR #729, opened 2017-04-22
Add native style prop note

griffinmichl:
This was a point of confusion for me when I was getting set up with styled-components in react native.


philpl:
I'll remind myself to add it to the website as well :)

PR #731, opened 2017-04-22
Get rid of whitelists for domElements and props validation

lttb:
This is my suggestion to get rid of whitelists of `domElements` and `validAttr`.

[is-react-prop](https://github.com/lttb/is-react-prop) and [react-dom-elements](https://github.com/lttb/react-dom-elements) provide data from React and ReactDOM instances that allows us not to copy-paste these whitelists, which will reduce a little the final bundle size and make this validations more flexible.


philpl:
It's definitely going to save some bytes, since the code is from React and can thus be reduplicated.

On the other hand, I'm not sure whether we should link to arbitrary files in the React package. This could also already break with React 16

Maybe you could chime in here, @danabramov?

lttb:
@philpl In my opinion it's also ideologically right to reuse validations that React provides.

The important point of this PR is also an attempt to unify these validations and domElements use, because currently a lot of other libraries that implement styled-components interface just copy-paste this logic too.

It would be greate to have a one place where this logic was implemented I think.

Of course there are some questions about capablity with React 16, but eventually we can add version checking in packages above.

gaearon:
This is **entirely unsupported** and will break in React 16 which completely removes `/lib/`.
Please don’t do it.

Additionally, the plan is for React 16 to pass all attributes through. Our eventual goal is to reduce the need for whitelists, and thus reduce the bundle size. We recommend all React-related libraries to do the same. **Any solutions relying on string tag or prop whitelists make apps in React ecosystems a little bit more bloated, and I am concerned about this.** It’s unfortunate, given that React is facing tougher competition from leaner libraries. This makes me more hesitant to recommend `styled-components` to other developers. I understand you are coming from good intentions, but ultimately this direction subverts our effort to make React apps smaller.

IMO `styled-components` should change the API to align with our goal of passing attributes through. I believe I brought this up with @mxstbr many times so I’m sure you folks are aware of our direction. If `styled-components` needs a whitelist, I think it is a design flaw (just like I believe `react-dom` needing a whitelist was a design flaw—which we are fixing). It is suboptimal to ship all those strings (either tag names or attributes) to the user. It also will prevent break React behavior in future versions (e.g. `<div>` and `styled.div` would work differently with regards to custom attributes which is surprising).

lttb:
@gaearon thanks for clarification! 

so what do you think about dom attributes and styles properties separation by the Component interface?
Just for example:
```jsx
<Button type="submit" _color="red" />
```

In this case we allow the user to decide which attributes he wants to pass through, and we can get rid of any whitelists.

gaearon:
I’d avoid any convention that relies on comparing strings (e.g. special property names).
You can solve the problem by clearly separating which props are primary for you: DOM or style.

For example SC API could treat DOM props as first class and put everything else into `look` (never mind the name):

```js
<Button type="submit" look={{ color: 'red' }} />
```

Or it could do the opposite, and treat style props as first class (again, never mind the `attrs` name):

```js
<Button color="red" attrs={{ type: 'submit' }} />
```

The desire to support both at the same level is, in my opinion, the reason the issue has become complicated.

gaearon:
If you ask me, the first approach is the one that makes the most sense intuitively.

It is not super verbose in simple cases:

```js
const Button = styled('div')`
  color: ${variation => variation === primary ? 'red' : 'blue'}
`;

<Button type="submit" variation="primary" />
```

It accommodates for arbitrary cases:

```js
const Button = styled('div')`
  color: ${variation => variation.primary ? 'red' : 'blue'};
  background: ${variation => variation.default ? 'pink' : 'magenta'};
`;

<Button type="submit" variation={{ primary: true, default: false }} />
```

It is friendlier to static typing (`variation` is literally the type of the second argument to `styled`, no dynamic stuff).

lttb:
@gaearon I agree. 
As for me the first approach is more friendly also because it's easier to refactor code from native dom elements to styled.

I just stopped thinking this way only by the fact that it seems less comfortable and natural to use then the same properties level.

gaearon:
I moved my comment here: https://github.com/styled-components/styled-components/issues/439#issuecomment-296399324.

lttb:
Think can close this PR in case of this discussion and #439

kof:
@gaearon 

>The desire to support both at the same level is, in my opinion, the reason the issue has become complicated.

At the same time, it complicates the component interface for the user if we start to separate those things. Flat props is always nicer, especially for small components.

The root of the problem we looking at right now is actually react's decision to move responsibility for passing only supported attributes for the dom components to the user. If react would just ignore unsupported attributes as it was before, we would't have this problem right now. Right now I don't recall the reason why react did it in the first place.

gaearon:
Because:

* it's one of the most commonly and consistently requested features to React (https://github.com/facebook/react/issues/140)
* maintaining a whitelist means growing React size forever since browsers keep adding new attributes
* there is a runtime cost to filtering props

kof:
> it's one of the most commonly and consistently requested features to React (facebook/react#140)

The issue is actually about:  "Allow custom (nonstandard) attributes.". Seems like an API for registering custom attributes is a valid request.

> maintaining a whitelist means growing React size forever since browsers keep adding new attributes

Could be done for dev mode only and by using some database (like caniuse) or at least a separate package which can be well maintained and used across the entire community.

Right now we are at the point where we have a number of libraries which maintain their lists.

> there is a runtime cost to filtering props

It should be very fast, basically checking if a property is in a map. If done in dev mode only doesn't matter much anyways.

kof:
Also this interface:

```jsx
<Button look={{primary}} />
```
sucks a lot compared to:

```jsx
<Button look="primary" />
```

It should be possible to continue using this interface one way or another.

PR #735, opened 2017-04-23
[Proposal] Replace .attrs with .withProps constructor

philpl:


geelen:
Any reason this is against `v2-rc` and not `v2`?

philpl:
@geelen Can't remember. I think because I was just on that branch and didn't think about it :wink:

mxstbr:
Do we want this?

philpl:
@mxstbr Nah, we're going with the namespace solution, so this is just another way of using `.attrs` to solve the problem that namespaces will solve for us.

PR #736, opened 2017-04-23
Get ssr styles as React Element

oscar-b:
Proof of concept, would appreciate comments.

Makes it possible to render `{sheet.getStyleElement()}` directly in <head> using React. Current ssr solution requires that the extracted styles are injected into the html outside of React, for instance via Express string substitution. This isn't suitable in for example GatsbyJS.

This needs tests, and proper Flow thingies.

With this change, I can do something like this in Gatsby:

```jsx
class Html extends React.Component {
  render() {
    const helmet = Helmet.renderStatic();

    // 💅
    sheet.collectStyles(this.props.body);

    return (
      <html lang="sv">
        <head>
          <meta charset="utf-8" />
          <meta name="viewport" content="width=device-width, initial-scale=1" />

          {helmet.title.toComponent()}
          {helmet.meta.toComponent()}
          {helmet.link.toComponent()}
          {helmet.style.toComponent()}

          {sheet.getStyleElement()}

        </head>
        <body>
          <div id="react-mount" dangerouslySetInnerHTML={{ __html: this.props.body }} />

          <script async src={prefixLink(`/bundle.js?t=${BUILD_TIME}`)} />
        </body>
      </html>
    );
  }
}
```

### DIFF
--- a/src/models/ServerStyleSheet.js
+++ b/src/models/ServerStyleSheet.js
@@ -48,6 +48,18 @@ class ServerTag implements Tag {
     return `<style type="text/css" ${namesAttr} ${localAttr}>\n${css}\n</style>`
   }
 
+  toElement() {
+    const attributes = {
+      [SC_ATTR]: this.names.join(' '),
+      [LOCAL_ATTR]: this.isLocal,
+    }
+    const css = Object.keys(this.components)
+      .map(key => this.components[key].css)
+      .join('')
+
+    return <style key={this.names.join('')} type="text/css" {...attributes} dangerouslySetInnerHTML={{ __html: css }} />
+  }
+
   clone() {
     const copy = new ServerTag(this.isLocal)
     copy.names = [].concat(this.names)
@@ -86,6 +98,15 @@ export default class ServerStyleSheet {
     }
 
     return this.instance.toHTML()
+  }
+
+  getStyleElement() {
+    if (!this.closed) {
+      clones.splice(clones.indexOf(this.instance), 1)
+      this.closed = true
+    }
+
+    return this.instance.toElement()
   }
 
   static create() {
--- a/src/models/StyleSheet.js
+++ b/src/models/StyleSheet.js
@@ -1,4 +1,5 @@
 // @flow
+import React from 'react'
 import BrowserStyleSheet from './BrowserStyleSheet'
 import ServerStyleSheet from './ServerStyleSheet'
 
@@ -14,6 +15,7 @@ export interface Tag {
   addComponent(componentId: string): void,
   inject(componentId: string, css: string, name: ?string): void,
   toHTML(): string,
+  toElement(): React.Element<*>,
   clone(): Tag,
 }
 
@@ -103,6 +105,10 @@ export default class StyleSheet {
 
   toHTML() {
     return this.tags.map(tag => tag.toHTML()).join('')
+  }
+
+  toElement() {
+    return this.tags.map(tag => tag.toElement())
   }
 
   getOrCreateTag(componentId: string, isLocal: boolean) {


src/models/StyleSheet.js
```diff
@@ -105,6 +106,10 @@ export default class StyleSheet {
     return this.tags.map(tag => tag.toHTML()).join('')
   }
 
+  toElement() {
+    return this.tags.map(tag => tag.toElement())
```
philpl: The elements here will need to have a `key`. You should be able to just use `cloneElement` and add it here

src/models/ServerStyleSheet.js
```diff
@@ -48,6 +48,18 @@ class ServerTag implements Tag {
     return `<style type="text/css" ${namesAttr} ${localAttr}>\n${css}\n</style>`
   }
 
+  toElement() {
+    const attributes = {
+      [SC_ATTR]: this.names.join(' '),
+      [LOCAL_ATTR]: this.isLocal,
```
philpl: We're searching for `el.getAttribute(LOCAL_ATTR) === 'true'`. I'm not sure whether this works, if React just adds the property shorthand; Can you check please? :)

src/models/StyleSheet.js
```diff
@@ -14,6 +14,7 @@ export interface Tag {
   addComponent(componentId: string): void,
   inject(componentId: string, css: string, name: ?string): void,
   toHTML(): string,
+  toElement(): any,
```
philpl: I'd guess the type here should just be `ReactElement` or sth

src/models/StyleSheet.js
oscar-b: Hm yes but what should I use for keys? `this.names.join('')` in ServerTag.toElement?

src/models/ServerStyleSheet.js
oscar-b: The result written to file is something like this:

```html
<style type="text/css" data-styled-components="biyTF hdZKtG gqDZGS clHZEQ yOoIr fUPLnt inUVIL fSmkBB kZQtlQ PIsWY fPJRNv jgRNmm TsYdo iZXahJ bDCuCY kUYNDb foZtll cqjtay fjMziR gqhtmN cEnRUY dhfSfb NkGiE hFJQDZ cOZxoD jYtjIs jskQpj fIEXtD isATaW btuKsl dReRJd dhLCfp gESlQT hoNHWU hRbfbz bMefOW cHPONr fFcRsM" data-styled-components-is-local="true">
```

You mean it wouldn't work if React omits `="true"`?

src/models/StyleSheet.js
oscar-b: I set the key in `toElement()` instead.

src/models/ServerStyleSheet.js
philpl: I'm not sure, I guess it's possible

src/models/StyleSheet.js
philpl: Just use the index, since this will only be used once and is thus static

oscar-b: Hm fighting with Flow:

```javascript
toElement() {
    return this.tags.map((tag, index) => (
      React.cloneElement(
        tag.toElement(),
        { key: index },
      )
    ))
  }
```

Gives error:
```[flow] property `key` of object literal (Property not found in props of React element `style` See also: call of method `cloneElement`)```

Seems like Flow doesn't like adding props to existing elements: https://github.com/facebook/flow/issues/1606

Any idea how to get around that?

File: src/models/ServerStyleSheet.js (hunk 2 above)
Comment (oscar-b) [inline src/models/ServerStyleSheet.js]: Not sure I follow  :)

**Comment by oscar-b** [comment]
For certain things, output seem to break.

Example:

```css
.jhjnFP:before {
	content: "";
}
```

Is output as:
```css
.jhjnFP:before {
	content: &quot;&quot;;
}
```

Also instead of:
```css
background-image: url(/22d13351d85a31fb85621052174c8493.jpg);
``` 

I get: 
```css
background-image: url(&#x27;/22d13351d85a31fb85621052174c8493.jpg&#x27;);
```

KyleAMathews:
React escapes text automatically. You probably just need to use `dangerouslySetInnerHTML`.

oscar-b:
@KyleAMathews Of course, thanks!

oscar-b:
@philpl Any outstanding issues you want me to look at?

philpl:
@oscar-b No, this is looking good :) @geelen will probably want to take a look at the methods' names though

geelen:
Merged and added a couple of small tweaks, then released as `v2.0.0-17` (`styled-components@rc`). Thanks @oscar-b!

PR #737, opened 2017-04-23
`innerProps` for whitelisting

geelen:


geelen:
See https://github.com/styled-components/styled-components/issues/439#issuecomment-296509149

probablyup:
Given that this is super out of date now and hasn’t been touched in almost a year, gonna close it.

Issue #741, opened 2017-04-24
prop-types warning in enzyme / Jest


philipchmalts:
## Version
1.4.5 / 2.0.0--15

## Reproduction

Running tests with enzyme and jest that use styled components

## Steps to reproduce
![screen shot 2017-04-24 at 5 53 21 pm](https://cloud.githubusercontent.com/assets/14876812/25360417/04b6c772-2917-11e7-8d84-ffec2eab5516.png)

## Expected Behavior
Expected no prop-types warning  when running tests. In browser I do not see the warnings but when in the testing environment I see this. Also noticed that when I use the v2 branch and compile it in my project the warning goes away. Either a new release of the v2 or an update in v1 babel-flow-react-proptypes will fix the issue I think.

## Actual Behavior
  Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.

kopax:
You need to replace in your code `import React, \{ PropTypes \} from 'react';`
with `import React from 'react';\n import PropTypes from 'prop-types';`

You also need to do 

   npm install prop-types --save

philipchmalts:
This seems to be related to #700 and #703. I just checked the package.json 

https://github.com/styled-components/styled-components/blob/v1.4.5/package.json#L63

and `babel-flow-to-proptypes` is still set to 0.18.2 instead of 1.2.0. Can make a PR for this and it will be cleared up.

EnoahNetzach:
The problem was in `babel-flow-to.proptypes`, but now it's been resolved in [**v2.1.3**](https://github.com/brigand/babel-plugin-flow-react-proptypes/pull/81).

In order to adopt the changes, a new publish should be issued here.

kopax:
Could we get a new release of styled v2 with this please ?.

kopax:
I've just tested with the `v2.0.0-17` release and I still have bunch of warning, is it correct ?

philpl:
@kopax I just checked and the package in question seems to be up-to-date there `"babel-plugin-flow-react-proptypes": "^2.1.3",`. Can you post the warnings you're seeing so I can fix it for good? :)

philipchmalts:
@kopax I just updated both my repos that use sc and this looked like it fixed the issue

kopax:
I was too fast, this is the warning I have left, I thought they were gonna disappear with styled v2.0.0-17 also. @philipchmalts do you have them?

    console.error node_modules/fbjs/lib/warning.js:36
[...]

PR #746, opened 2017-04-26
Resolve remaining prop-types warnings

EnoahNetzach:
`babel-plugin-flow-react-proptypes` updated to include https://github.com/brigand/babel-plugin-flow-react-proptypes/pull/81.

Solves #741.


philpl:
@k15a Can you confirm that this fixes all the warnings like it does on v2? I can release a patch, I suppose, if it does

k15a:
I don't know why it shouldn't but haven't tried it as well ;)

PR #764, opened 2017-05-03
[v2] Restore setNativeProps

### LINKED ISSUE #763: [v2] setNativeProps is missing
## Version
2.0.0-17

## Reproduction

```javascript
import React from 'react';
import styled from 'styled-components/native';
import { TouchableHighlight, View } from 'react-native';

const Text = styled.Text``;

export default class App extends React.Component {
  render() {
    return (
      <View>
        <TouchableHighlight>
          <Text>Test</Text>
        </TouchableHighlight>
      </View>
    );
  }
}
```

## Expected Behavior

Should not throw any errors.

## Actual Behavior

Throw the following error : 

```
invariant.js:44 Uncaught Error: Touchable child must either be native or forward setNativeProps to a native component
    at invariant (invariant.js:44)
    at ensureComponentIsNative (ensureComponentIsNative.js:20)
    at Object.componentDidMount (TouchableHighlight.js:124)
    at Object.chainedFunction [as componentDidMount] (factory.js:531)
    at ReactCompositeComponent.js:265
    at measureLifeCyclePerf (ReactCompositeComponent.js:75)
    at ReactCompositeComponent.js:264
    at CallbackQueue.notifyAll (CallbackQueue.js:76)
    at ReactReconcileTransaction.close (ReactReconcileTransaction.js:80)
    at ReactReconcileTransaction.closeAll (Transaction.js:206)
```

## Reason

```setNativeProps``` is missing in ```StyledNativeComponent``` in v2 branch.

### PR DESCRIPTION
**Pull request by MatthieuLemoine**
In ```StyledNativeComponent```, ```setNativeProps``` has been deleted between v1 and v2.

This PR restores it to allow forwarding of native props to the inner component.

Fix #763.

### DIFF
--- a/src/models/StyledNativeComponent.js
+++ b/src/models/StyledNativeComponent.js
@@ -24,6 +24,7 @@ export default (constructWithOptions: Function) => {
       theme: null,
       generatedStyles: undefined,
     }
+    root: Object
 
     buildExecutionContext(theme: any, props: any) {
       const { attrs } = this.constructor
@@ -96,12 +97,22 @@ export default (constructWithOptions: Function) => {
     }
 
     setNativeProps(nativeProps: Object) {
+      this.root.setNativeProps(nativeProps)
+    }
+
+    generateRef() {
       const { innerRef } = this.props
-      innerRef.setNativeProps(nativeProps)
+      return (component: any) => {
+        this.root = component
+
+        if (innerRef) {
+          innerRef(component)
+        }
+      }
     }
 
     render() {
-      const { children, style, innerRef } = this.props
+      const { children, style } = this.props
       const { generatedStyles } = this.state
       const { target } = this.constructor
 
@@ -112,7 +123,7 @@ export default (constructWithOptions: Function) => {
       }
 
       if (!isStyledComponent(target)) {
-        propsForElement.ref = innerRef
+        propsForElement.ref = this.generateRef()
         delete propsForElement.innerRef
       }
 
--- a/src/models/StyledNativeComponent.js
+++ b/src/models/StyledNativeComponent.js
@@ -100,14 +100,12 @@ export default (constructWithOptions: Function) => {
       this.root.setNativeProps(nativeProps)
     }
 
-    generateRef() {
+    getRef(node: any) {
       const { innerRef } = this.props
-      return (component: any) => {
-        this.root = component
+      this.root = node
 
-        if (innerRef) {
-          innerRef(component)
-        }
+      if (typeof innerRef === 'function') {
+        innerRef(node)
       }
     }
 
@@ -123,7 +121,7 @@ export default (constructWithOptions: Function) => {
       }
 
       if (!isStyledComponent(target)) {
-        propsForElement.ref = this.generateRef()
+        propsForElement.ref = this.getRef
         delete propsForElement.innerRef
       }
 
--- a/src/models/StyledNativeComponent.js
+++ b/src/models/StyledNativeComponent.js
@@ -100,13 +100,17 @@ export default (constructWithOptions: Function) => {
       this.root.setNativeProps(nativeProps)
     }
 
-    getRef(node: any) {
+    onRef(node: any) {
       const { innerRef } = this.props
       this.root = node
 
       if (typeof innerRef === 'function') {
         innerRef(node)
       }
+    }
+
+    onStyledRef(node: any) {
+      this.root = node
     }
 
     render() {
@@ -121,8 +125,10 @@ export default (constructWithOptions: Function) => {
       }
 
       if (!isStyledComponent(target)) {
-        propsForElement.ref = this.getRef
+        propsForElement.ref = this.onRef
         delete propsForElement.innerRef
+      } else {
+        propsForElement.ref = this.onStyledRef
       }
 
       return createElement(target, propsForElement, children)
--- a/src/models/StyledNativeComponent.js
+++ b/src/models/StyledNativeComponent.js
@@ -18,13 +18,13 @@ export default (constructWithOptions: Function) => {
     static styledComponentId: string
     static attrs: Object
     static inlineStyle: Object
+    root: Object
 
     attrs = {}
     state = {
       theme: null,
       generatedStyles: undefined,
     }
-    root: Object
 
     buildExecutionContext(theme: any, props: any) {
       const { attrs } = this.constructor
--- a/src/models/StyledNativeComponent.js
+++ b/src/models/StyledNativeComponent.js
@@ -100,7 +100,7 @@ export default (constructWithOptions: Function) => {
       this.root.setNativeProps(nativeProps)
     }
 
-    onRef(node: any) {
+    onRef = (node: any) => {
       const { innerRef } = this.props
       this.root = node
 
@@ -121,10 +121,10 @@ export default (constructWithOptions: Function) => {
       }
 
       if (!isStyledComponent(target)) {
-        propsForElement.ref = this.onRef.bind(this)
+        propsForElement.ref = this.onRef
         delete propsForElement.innerRef
       } else {
-        propsForElement.innerRef = this.onRef.bind(this)
+        propsForElement.innerRef = this.onRef
       }
 
       return createElement(target, propsForElement, children)
--- a/src/models/StyledNativeComponent.js
+++ b/src/models/StyledNativeComponent.js
@@ -109,10 +109,6 @@ export default (constructWithOptions: Function) => {
       }
     }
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by philpl** [comment]
Hm, I wasn't aware of this functionality.

It looks like you're calling `setNativeProps` on the `innerRef` _prop_ though.
You'll probably want to call that on an actual ref right?

So we'll need to always pass down a `ref` or `innerRef` prop, store the result, and then pipe the result to `innerRef` as well, if it exists.

**Comment by MatthieuLemoine** [comment]
@philpl Is ```innerRef``` a ref to the wrapped component i.e ```Text``` in ```styled.Text``` ?

**Comment by philpl** [comment]
@matthieulemoine it indeed is :) it's an optional prop. Best thing is probably to add a new method that sits between the passed innerref prop and the ref/innerRef logic in render

**Comment by MatthieuLemoine** [comment]
In v1, there was a ```generateRef``` function to keep a reference to the wrapped component whether or not ```innerRef``` was passed : 

```javascript
generateRef() {
  const { innerRef } = this.props
  const innerRefType = typeof innerRef
  
  if (innerRefType === 'string') {
    this.root = innerRef
    return innerRef
  }
  
  return (component: any) => {
    this.root = component
    
    if (innerRefType === 'function') {
      innerRef(component)
    }
  }
}
```

And in render : 

```javascript
propsForElement.ref = this.generateRef()
```

```this.root``` was then used ```setNativeProps```:

```javascript
setNativeProps(nativeProps: Object) {
  const root = typeof this.root === 'string' ? this.refs[this.root] : this.root 
  root.setNativeProps(nativeProps)
}
```

Should I restore this function too ?

philpl:
It'll need some modifications, and I think we should use a method instead to keep the function instance static.

Apart from that:

- We don't support passing strings as refs anymore
- `innerRef` is not passed as `ref` (i.e. not "unwrapped") if the underlying component is a StyledComponent (this should already be in place though)

philpl:
@MatthieuLemoine Are you making progress on this and do you still have time for this? :)

It's quite a far reaching issue and we should push for fixing it soon-ish, I guess

MatthieuLemoine:
@philpl Yes if you have time to review what I did it would be nice !

To be able to forward ```setNativeProps```, we need a ref to the underlying component. It's only possible if the component is not a stateless function. 

We can't have both **setNativeProps forwarding** and **stateless component usage**.

What we can do is to add a ref to the child component if it isn't stateless that will enable **setNa
tiveProps forwarding** and do nothing otherwise.

With this solution we will able to allow stateless component to be wrapped by ```styled-components``` and to support having a styled component as first child of a ```Touchable```. 
We will just have to add a note in the README to inform the users that they can't use a wrapped stateless component as ```Touchable```'s first child but should use class component instead for this very case.

src/models/StyledNativeComponent.js
```diff
@@ -95,8 +96,23 @@ export default (constructWithOptions: Function) => {
       }
     }
 
+    setNativeProps(nativeProps: Object) {
+      this.root.setNativeProps(nativeProps)
+    }
+
+    generateRef() {
```
philpl: You don't need to generate a new method on every call, but can instead reuse the same method for the callback over and over again :wink: 🎐

src/models/StyledNativeComponent.js
```diff
@@ -95,8 +96,23 @@ export default (constructWithOptions: Function) => {
...
+      const { innerRef } = this.props
+      return (component: any) => {
```
philpl: When you modify this to be a method, it'd be great to call this `instance` or `node`, or sth, since it's not a component :)

src/models/StyledNativeComponent.js
```diff
@@ -95,8 +96,23 @@ export default (constructWithOptions: Function) => {
...
+        this.root = component
+
+        if (innerRef) {
```
philpl: This is hurting perf on a micro level and since we're having efforts to improve perf at the same time it'd be great to fix this and just swap it out with a simple `typeof innerRef === 'function'` please

src/models/StyledNativeComponent.js
```diff
@@ -24,6 +24,7 @@ export default (constructWithOptions: Function) => {
       theme: null,
       generatedStyles: undefined,
     }
+    root: Object
```
philpl: Should probably be grouped with the static flow properties (types) above, instead of the class instances (values)

src/models/StyledNativeComponent.js
```diff
@@ -107,7 +123,7 @@ export default (constructWithOptions: Function) => {
       }
 
       if (!isStyledComponent(target)) {
-        propsForElement.ref = innerRef
+        propsForElement.ref = this.generateRef()
```
philpl: The `setNativeProps` method needs to work for nested Styled Components too, so we need to add a case for passing `innerRef` as well.

philpl:
Hey, sorry for dropping the ball here for quite some time. I hope you can still finish this? If not that's totally fine as well :)

I added some remarks, most of them minor, except two.

philpl:
@MatthieuLemoine Sorry for pinging you again. It looks like this is the last issue that separates us from the v2 release unfortunately. If you don't mind I can take this PR over and do the changes myself, if you lack the time. Sorry for being a bit pushy

src/models/StyledNativeComponent.js
MatthieuLemoine: Can you be more specific ? I've never used Flow actually.

philpl: Just out it below the static definitions above this block :)

philpl:
@matthieulemoine thanks for picking this up so quickly again! You are the best, mate!

For the functional compinents, maybe we should just wrap every one of them for now in a normal component? :/

MatthieuLemoine:
@philpl I have created two methods : 
* ```onRef``` : to set ```this.root``` for ```setNativeProps``` and forward innerNode to ```innerRef```
* ```onStyledRef``` : to set ```this.root``` for nested Styled Components

MatthieuLemoine:
There is still a remaining issue with stateless component as we can't define a ref to a statelesss component

mxstbr:
Let's just go with however it works right now? I don't want us to wrap every component in another one realistically :cry:

src/models/StyledNativeComponent.js
MatthieuLemoine: Oh yes of course !

src/models/StyledNativeComponent.js
```diff
@@ -107,8 +125,10 @@ export default (constructWithOptions: Function) => {
       }
 
       if (!isStyledComponent(target)) {
-        propsForElement.ref = innerRef
+        propsForElement.ref = this.onRef
         delete propsForElement.innerRef
+      } else {
+        propsForElement.ref = this.onStyledRef
```
philpl: this will only have to be:

```js
propsForElement.innerRef = this.onRef
```

philpl:
@MatthieuLemoine there's only the remaining last change to make this work with nested Styled Components. The code to make it work is fortunately very simple :) Let's merge it afterwards

src/models/StyledNativeComponent.js
MatthieuLemoine: Oh yeah smart !

MatthieuLemoine:
@philpl Actually one test is failing because the ```innerRef``` of a nested Styled Component is no longer equal to the innerRef of the outer component.

philpl:
@MatthieuLemoine I guess you can remove that test then :) also a small nit (that we don't have to fix now) `.bind(this)` creates a new reference, so it negates the purpose of making the method static instead of generating a new function all the time.

you could just do:

```js
onRef = node => {...}
```

On the class, but we don't have to do this really. Let's get the test removed and this code merged :)

MatthieuLemoine:
@philpl Done

philpl:
@MatthieuLemoine thank a bunch for being so persistent and helpful! 💟 

I'll get this merged once the tests pass :smile:

src/models/StyledNativeComponent.js
```diff
@@ -95,8 +96,21 @@ export default (constructWithOptions: Function) => {
       }
     }
 
+    setNativeProps(nativeProps: Object) {
```
philpl: One last thing—Sorry!—this can now be deleted I think

MatthieuLemoine: We need ```setNativeProps``` to forward nativeProps to the underlying native component. The reason of this PR is that it's missing from v2.

philpl: I was being stupid. Sorry

PR #768, opened 2017-05-05
Add custom classname

ahmed3mar:
Add custom `classname` over the `generated name`.

I know that generated name is unique but sometimes i just need to specify my custom name of the class

for example,
```JS
import styled from 'styled-compontents';

const Section = styled('section.class-name')`
  color: red;
`;
```

it will generate
```CSS
.class-name {
  color: red;
}

### DIFF
--- a/src/constructors/styled.js
+++ b/src/constructors/styled.js
@@ -3,10 +3,22 @@ import css from './css'
 import type { Interpolation, Target } from '../types'
 import domElements from '../utils/domElements'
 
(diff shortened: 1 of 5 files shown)


philpl:
Unfortunately this is already supported by the attrs method in v2. There's some more considerations that we're discussing on whitelisting props however, if you're interested

ahmed3mar:
I did not see it before :\

i'll check it and i'll try to help.

PR #771, opened 2017-05-06
Remove unnecessary listing `flow-typed` in [libs]

luftywiranda13:
It's already included by default

https://flow.org/en/docs/config/libs/


philpl:
@styled-components/typers

PR #783, opened 2017-05-11
Added createAnimatedComponent to native documentation

### LINKED ISSUE #782: Animations with react-native not working
Think you found a bug?
======================
To enable Animation in react-native you need to use `Animated.[View]` to enable style manipulation. 
In the readme you just bind a regular view: https://github.com/styled-components/styled-components/blob/master/docs/react-native.md
This results in `Attempted to assign to readonly property.

So in current version animations doesn't seem to work for react-native.

Example code: 
```javascript
const StyledImage = styled.Image`
    justifyContent: center;
    zIndex: 100;
`

<StyledImage
            source={require('images/icon.png')}
            style={{
              width: this.state.icon_size,
              height: this.state.icon_size,}}
          />
```

nenti:
This pull request resolves #782

### DIFF
--- a/docs/react-native.md
+++ b/docs/react-native.md
@@ -40,14 +40,15 @@ const RotatedBox = styled.View`
 
 ## Animations
 
-To get React Native animations working, you'll want to define all your non-changing styles in the usual way, and then pass your `Animated.Value`s in as style props.
+To get React Native animations working, you'll want to define all your non-changing styles in the usual way, and then bind animation with `createAnimatedComponent` and pass your `Animated.Value`s in as style props.
 
 ```js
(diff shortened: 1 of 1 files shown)


docs/react-native.md
```diff
@@ -68,7 +69,7 @@ class AnimateOpacity extends Component {
   render() {
     const { opacity } = this.state;
     // Pass in your animated values here!
-    return <BaseStyles style={{ opacity }} />
+    return <StyledView style={{ opacity }} />
```
k15a: This should be `AnimatedStyledView`then?

nenti:
Good point @k15a. Commited fix.

Issue #774, opened 2017-05-08
Parent selectors resolve to different rules due to hash classnames


drFabio:
I think I found a possible missbehaviour when using props and siblings . (Or probably just the className placeholder in general).

## Version
1.0.5

## Reproduction
https://www.webpackbin.com/bins/-Kjc_PkTewh0QoJ0kJNn

## Steps to reproduce

Create a component with the sibling syntax that use some props for it's rules.

## Expected Behavior

The sibling rules would be applied to all props class variants

## Actual Behavior

The component get's styled with the sibling rules on the prop specific class.

diondiondion:
Interesting. This wouldn't happen if the `&` selector used the `sc-`-prefixed className that doesn't change based on props, but not sure if this would have other side effects.

drFabio:
Will take a look on this prefix and let you know. It may be at least a workaround for this specific case.
If this is indeed a missbehaviour I would be glad to make a PR with tests in the near future

philpl:
This is not misbehaviour but intended.

Imagine a case where you'd write this:

```js
const Button = styled.button`
  color: black;

  ${p => p.addHover && css`
    &:hover {
      color: violet;
    }
  `}
`

<Button />
<Button addHover />
```

What you're referring to with the `sc-*` classes are classes based on the component id. The button has a single component id, and its `sc-x` class refers to all its variations.

In the above example, if we'd target this `sc-x` class instead for the parental selector (i.e. ampersand selector), then both buttons would suddenly have the above pseudo class.

I'd suggest you to just avoid using this selector altogether while altering the classes. Don't forget that two components, with different props, rendering different CSS, will also be decoupled from each other. By this thinking, this behaviour does make sense ;)

drFabio:
I get the rationale that each component with a prop would be it's own thing.
I didn't quite got this "sc-" part. Can I leverage it to get this behaviour of the same rule for all siblings even if they use different props?

> I'd suggest you to just avoid using this selector altogether while altering the classes.

By that quote You are meaning sibling selectors?

diondiondion:
> This is not misbehaviour but intended.

Let's say it's intended misbehaviour then. ;)

I've always thought that generating an entirely new CSS class for each variation was kind of inelegant.

Ideally styled-components would (somehow) smartly split all the CSS rules for a given component into dynamic & static bits and assign each bit its own hashed className (and add all of these classes to the rendered component). Similar to how you'd write classes and modifier classes in a methodology like BEM.
[...]

PR #784, opened 2017-05-12
Parent selectors hash resolution

drFabio:
This fix aims to solve the problem described [here](https://github.com/styled-components/styled-components/issues/774) upon which sibling selectors is unexpectedly applied when using props on a component.

This is not complete being submited for the time just for input and guidance.

### DIFF
--- a/src/test/props.test.js
+++ b/src/test/props.test.js
@@ -39,20 +39,16 @@ describe('props', () => {
       }
       `
     })
-    it('Should give the same root class to the prop children and the propless one', ()=> {
+    it('Should give the same class to the prop children and the propless one', ()=> {
       const wrapper = mount(<div><Comp /><Comp red /></div>)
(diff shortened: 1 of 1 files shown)


geelen:
Hmm, I think there's a few more cases you need to consider...

```js
const Comp = styled.div`
  color: ${props => props.red ? 'red' : 'white'}
  & + & {
    color: 'black'
  }
`

const Comp = styled.div`
  ${props => props.red && css`
    & + & {
      color: 'black'
    }
  `}
`
```

Does the `&` have different meanings in the two cases? Should the `& + &` example actually be `& + ${this}` to explicitly refer to the base class in the second one? In both?

drFabio:
Yeah will update the cases and implement that then.

mxstbr:
Any news @drFabio? :blush:

probablyup:
Superceded by #2071

Issue #791, opened 2017-05-15
Add mediaQuery helper method to manage media queries


JohnAlbin:
Motivation: I really like the media query suggestions in tips-and-tricks.md, but I want something simpler to use.

In tips-and-tricks.md, we have:

```JS
// style-utils.js
import { css } from 'styled-components'

export const media = {
  handheld: (...args) => css`@media (max-width: 420px) { ${ css(...args) } }`,
  tablet: (...args) => css`@media (min-width: 421px) { ${ css(...args) } }`,
}
```

```JS
import { media } from '../style-utils';

const Box = styled.div`
  font-size: 16px;
  ${media.handheld`
    font-size: 14px;
  `}
`;
```

The best part of this tip is that it allows the developer to pick whatever name they want when defining the static object that holds all the media query helpers (e.g. `media`, `respondTo`, `atMedia`, …). And, since it is a static object, IDEs will offer auto-complete suggestions when you type the name of the object. I.e. typing `media.` in your IDE will offer `media.handheld`, `media.tablet` and all other media queries defined in your object.

The only problem with the tip is that there is a lot of boilerplate. ``(...args) => css`@media (query goes here) { ${ css(...args) } }`,`` and it is not possible (for mere mortals) to memorize that.

At first I wrote a simple helper function that setup the boilerplate and could be used like this:
```JS
const media = {
  tablet: mediaHelper('(min-width: 420px)'),
}
```

But I realized that it would be nice if we could use a tagged template literal for the media query string too. Like this:

```JS
const media = {
  tablet: mediaQuery`(min-width: 420px)`,
}
```

This had the bonus effect that it removed one set of parenthesis and made the code easier to read.

NOTE: if you are wondering why the mediaQuery method can't handle the parenthesis around the query string so that you could just type this: `` mediaQuery`min-width: 420px` ``, consider the following query: `@media tv, (max-width: 520px) {}`, which is not wrapped entirely by a single set of parenthesis.

After a bit of playing with styled-component's css() method, I came up with this:

```JS
const mediaQuery = (...query) => (...rules) => css`
  @media ${css(...query)} {
    ${css(...rules)}
  }
`
```

I love how this function looks. Thank you, css() method! You rock.

Now that we have that method, we can do stuff like this:

```js
// style-utils.js
import { mediaQuery } from 'styled-components' // assuming this feature is added!

const sizes = {
  tablet: 768,
  desktop: 992,
  giant: 1170,
}

[...]

PR #792, opened 2017-05-15
Add mediaQuery helper method to manage media queries

JohnAlbin:
This fixes https://github.com/styled-components/styled-components/issues/791 There's a much longer explanation there.

While my original code was just this:

```JS
const mediaQuery = (...query) => (...rules) => css`
  @media ${css(...query)} {
    ${css(...rules)}
  }
`
```

I had to complicate it a bit to add Flow typing on the params and return types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file. If a contri
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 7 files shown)


philpl:
Copying over my comment from the issue:

> Just a note: Please beware that css(...args) and similar usages of it will break preprocessing. I'm looking out for this kind of code snippets and in the future we won't recommend anything that will break preprocessing in the future 😄

> Edit: Btw as a closing note. This is a CSS helper and those do belong into polished or a separate pkg, since the API doesn't contain any helpers like these and shouldn't be bloaty in the sense of delivering anything that is not immediately needed.

PR #804, opened 2017-05-21
add more false values

shashkovdanil:


philpl:
This is unfortunately not what you'd expect, when using Styled Components. :wink:
`0` and `-0` are totally valid values, that should be inlined in to the resulting CSS code.

psixdev:
What about `NaN`?

kitten:
@psixdev if you find any notes on what you'd like to see improved please open a new issue ✌️ it's more likely to see changes being kicked off by new issues rather than PRs/issues that have been closed >1yr ago since the status quo of the repo has changed by that time.

Issue #246, opened 2016-11-23
Building reusable Text component


steida:
I am confused. I like the idea of the coupling element with its styles, but now I am not sure how to reuse basic text styles. Here is my text component: 

```jsx
/* @flow */
import React from 'react';
import styled from 'styled-components';

type Props = {
  bold?: bool,
  inverted?: bool,
};

const Text = styled.span`
  color: ${props => props.inverted ? props.theme.colors.white : props.theme.colors.black};
  font-family: ${props => props.theme.fontFamily};
  font-size: ${props => props.theme.fontSize.text};
  font-weight: ${props => props.bold ? props.theme.fontWeight : 'normal'};
  line-height: ${props => props.theme.lineHeight };
`;

export default (props: Props) => <Text {...props} />;

```

And now, I would like to create Link component.

```jsx
/* @flow */
import React from 'react';
import styled from 'styled-components';
import { Link as RouterLink } from 'react-router';

type Props = {
  bold?: bool,
  exactly?: bool,
  inverted?: bool,
  to: string,
};

const AnchorOrRouterLink = ({ bold, exactly, inverted, to, ...props }: Props) => {
  const isExternalLink = to.includes('://');
  return isExternalLink ?
    <a
      {...props}
      href={to}
    />
  :
    <RouterLink
      {...props}
      activeOnlyWhenExact={exactly}
      activeStyle={{ textDecoration: 'underline' }}
      to={to}
    />
};

const Link = styled(AnchorOrRouterLink)`
  color: ${props => props.inverted ? props.theme.colors.white : props.theme.colors.primary};
  text-decoration: none;
  
  font-family: ${props => props.theme.fontFamily};
  font-size: ${props => props.theme.fontSize.text};
  font-weight: ${props => props.bold ? props.theme.fontWeight : 'normal'};
  line-height: ${props => props.theme.lineHeight };

  &:hover {
    text-decoration: underline;
  }
  &:active {
    text-decoration: underline;
  }
`;

export default (props: Props) => <Link {...props} />;
```

As we can see, Link component is duplicating font-family, font-size, font-weight, line-height from Text, which is IMHO not nice. As far I know, we have these options:

Remove text styles from Link entirely.

```jsx
<Text><Link to="/">text</Link></Text>
```

Not sure whether this is OK. I would rather have Text API on Link component.

```jsx
<Link bold to="/">text</Link>
```

But how to implement `bold` on top of the Link?

[...]

PR #814, opened 2017-05-23
Changed extendWith to withComponent

geelen:
From #246. I think this will make it easier for people to understand what `withComponent` does.

Also, the old `extendWith` behaviour is available like so:

```js
/* old */
const X = Y.extendWith('z')` ... `

/* new */
const X = Y.withComponent('z').extend` ... `
```

My feel is you won't often want to do both at the same time, but it does make this pattern quite nice:

```js
const Heading = styled.h1`
  ...
`

Heading.h1 = Heading
Heading.h2 = Heading.withComponent('h2')
Heading.h3 = Heading.withComponent('h3')
Heading.h4 = Heading.withComponent('h4')
Heading.h5 = Heading.withComponent('h5')
Heading.h6 = Heading.withComponent('h6')
Heading.Link = Heading.withComponent(Link)

export default Header
```

```js
<Header.h2>This will be a h2 tag!</Header.h2>
```


Nudelsieb:
Doesn't work for me:
const Div = Paragraph.withComponent('div')`
  font-size: 16px;
`;

TextField.jsx:43 Uncaught TypeError: _Paragraph2.default.withComponent is not a function

Is it not in the trunk yet? I found it in the documentation and wanted to use it? If that's so the documentation is further than the stable release.

igl:
it's
```
const Div = Paragraph.withComponent('div').extend`
  font-size: 16px;
`;
```

Issue #775, opened 2017-05-09
[v2] @supports query always results to true, if there´s a nested @media query


oennen:
## Version
2.0.0-17

## Reproduction

https://www.webpackbin.com/bins/-Kji-ewR2MGAKe_s1IPJ

## Steps to reproduce

see example above

## Expected Behavior

`@supports not (display: block)` should result to false and content should be ignored.
Title color should be palevioletred.

## Actual Behavior

`@supports not (display: block)` results to true and content is being included.
Title color is green.

philpl:
@ThySultan This seems to be a problem with Stylis :) https://www.webpackbin.com/bins/-Kji2D_tAEbrLoEipsCh

thysultan:
Landed a patch in v2.0.5.

PR #817, opened 2017-05-23
updated package.json to allow patches of stylis when installing deps

silverlight513:
There have been multiple patch updates to stylis with fixes for multiple issues. Styled components currently installs 2.0.3. The latest stylis is 2.0.12. 

Examples of a few issues:
https://github.com/thysultan/stylis.js/issues/16
https://github.com/styled-components/styled-components/issues/775

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "is-function": "^1.0.1",
     "is-plain-object": "^2.0.1",
     "prop-types": "^15.5.4",
-    "stylis": "^2.0.0",
+    "stylis": "~2.0.0",
     "supports-color": "^3.2.3"
   },
   "devDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -5645,9 +5645,9 @@ strip-json-comments@~2.0.1:
   version "2.0.1"
   resolved "https://registry.yarnpkg.com/strip-json-comments/-/strip-json-comments-2.0.1.tgz#3c531942e908c2697c0ec344858c286c7ca0a60a"
 
-stylis@^2.0.0:
-  version "2.0.3"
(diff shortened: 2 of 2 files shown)


k15a:
Actually, the `^` already includes patches. The bigger problem is that we bundle stylis with styled-components which locks the version for webpack users. We should probably update our build process before v2 final release.

philpl:
@k15a We should probably add stylis as an external for the `.es.js` build

silverlight513:
Ah okay, fair enough. Would an updated patch version of stylis in the yarn.lock not help webpack users at all?

philpl:
@silverlight513 no, webpack 2 and rollup will point to the ES bundle

k15a:
I think we should create another distribution folder which is compiled without modules.

So webpack and rollup can use this folder and bundle it alone and everyone who wants to include styled-components via script tag can use unpkg.com and use the bundles.

silverlight513:
Oh, I didn't realise that. Shall we close this PR then as it's pointless to the project. Other than discussion about the issue.

philpl:
@k15a don't think that's a good idea. We can just change the `.es.js` module alone, since that's not the UMD bundle that you'd use on script tags, etc.

k15a:
Good point. Is there a way to automatically exclude all dependencies from the `.es.js` bundle?

philpl:
@k15a yea, adding all keys of `dependencies` to `externals`, but that's not a good idea either ^^ It'd probably introduce some quirks over time. I'll open a PR that modifies the ES bundle in a few minutes.

k15a:
Could you explain why this isn't a good idea? I mean webpack and rollup can resolve the dependencies on their own when they are not in the bundle. What's the downside of that?

philpl:
@k15a It just adds a layer of unexpectedness imho. I'd change as little as possible for now, since I don't want any weird surprises after the v2 release ^^

PR #818, opened 2017-05-23
Fix Type for Broadcast.subscribe's Callback

jameskraus:
Broadcast.subscribe takes a parameter, so it currently throws a flow type error because the type says it takes no parameters:
```
  node_modules/styled-components/src/utils/create-broadcast.js:20
   20:       listeners.forEach(listener => listener(currentValue))
                                                    ^^^^^^^^^^^^ unused function argument
     10:   subscribe: (listener: () => void) => () => void
                                 ^^^^^^^^^^ function type expects no arguments
```

This small change should fix this problem.


jameskraus:
Looks like a fix for this was merged a a few days after I sent this PR

PR #819, opened 2017-05-23
Allow updates to stylis from being used in ES bundle

philpl:
Before, even when stylis had a new patch version, Webpack 2 and Rollup would use the ES Bundle, which includes stylis. This PR changes the ES Bundle to import stylis instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "build:lib": "babel --out-dir lib src",
     "build:flow": "flow-copy-source -v -i '{**/test/*.js,**/*.test.js}' src lib",
     "prebuild:dist": "rimraf dist/*",
-    "build:dist": "rollup -c && rollup -c --environment PRODUCTION",
+    "build:dist": "rollup -c && rollup -c --environment ESBUNDLE&& rollup -c --environment PRODUCTION",
     "build:watch": "npm run build:lib -- --watch",
     "test": "npm run test:web && npm run test:native",
     "test:web": "jest",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
     "build:lib": "babel --out-dir lib src",
     "build:flow": "flow-copy-source -v -i '{**/test/*.js,**/*.test.js}' src lib",
     "prebuild:dist": "rimraf dist/*",
-    "build:dist": "rollup -c && rollup -c --environment PRODUCTION",
+    "build:dist": "rollup -c && rollup -c --environment ESBUNDLE&& rollup -c --environment PRODUCTION",
```
k15a: little nitpick: space is missing before &&
I don't know if some shells have a problem with it.

philpl: Ah oops

silverlight513:
Hey, I just tested this work out on one of my projects and there's a slight hiccup. If you use npm, everything works fine, but if you use yarn, the yarn.lock file in this project is forcing the use of old packages. I.e. stylis at version 2.0.3 in the lock file which doesn't have some lovely fixes I was looking to use.

It's an easy switch to use npm for all users but forcing some users not to use yarn isn't great. Would it be okay if I provide a PR with an updated yarn.lock file?

mxstbr:
Oh wow that's a good catch @silverlight513, please submit a PR with the latest 2.x version!

silverlight513:
I'll get on it once I get home from work :)

PR #821, opened 2017-05-24
Combine name with id if both present

griffinmichl:
Issue: https://github.com/styled-components/babel-plugin-styled-components/issues/44

This will allow users to use the babel plugin with ssr (which sets the componentId at compile time) and displayName by combining them at runtime if both are present.


philpl:
I'll see later what these kinds of ids look like :)

src/test/expanded-api.test.js
```diff
@@ -46,13 +46,22 @@ describe('expanded api', () => {
...
+      const Comp2 = styled.div.withConfig({ componentId: 'OMGLOL' })``
       expect(Comp.styledComponentId).toBe('LOLOMG')
       expect(shallow(<Comp />).prop('className')).toMatch(/LOLOMG/)
       expect(Comp2.styledComponentId).toBe('OMGLOL')
       expect(shallow(<Comp2 />).prop('className')).toMatch(/OMGLOL/)
     })
+
+    it('should be combined with displayName if both passed in', () => {
+      const Comp = styled.div.withConfig({ displayName: 'Comp', componentId: 'LOLOMG' })``
+      const Comp2 = styled.div.withConfig({ displayName: 'Comp2', componentId: 'OMGLOL' })``
```
mxstbr: This is an ace test case 💯  very on brand for this repo 😆

src/models/StyledComponent.js
```diff
@@ -181,14 +181,17 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
       attrs,
     } = options
 
+    const styledComponentId = (options.displayName && options.componentId) ?
+      `${options.displayName}-${options.componentId}` : componentId
```
mxstbr: Do we know for sure here that `componentId` will always be a thing, even if the babel plugin isn't used?

philpl: It defaults to an autogenerated one :) `componentId = generateId(options.displayName),`

@griffinmichl can you put in a rule that doesn't change the `styledComponentId` to this concatenated version if `options.displayName` is present, but `componentId` is not, since it'll default to sth that is similar to the `displayName`?

griffinmichl: @philpl Not sure I follow. I believe that, because I'm using `options.displayName` and `options.componentId`, this behavior will be achieved because options isn't mutated by the default restructuring value. If I am thinking about this correctly, it's covered by the `should be generated from displayName + hash` test case.  
  
I might be missing something totally obvious. Sorry!

philpl: yea, you're right! sorry.

mxstbr:
Awesome, thanks @griffinmichl, this is a lot better 👌

PR #829, opened 2017-05-25
Upgrade to Stylis 3

philpl:


philpl:
@thysultan I'm seeing some weird ` { ; }` artefacts. Most of them disappeared when I switched on the `compress` option, but not all of them.

philpl:
@thysultan some more bugs in `ssr.test`:

```diff
     /* sc-component-id: sc-keyframes-keyframe_0 */
    -@-webkit-keyframes keyframe_0{ 0%{ opacity:0; }
    +@-webkit-keyframes keyframe_0{ 0%{ opacity:0}
     }
    -@keyframes keyframe_0{ 0%{ opacity:0; }
    +@keyframes keyframe_0{ 0%{ opacity:0}
     }
     </style>
```

thysultan:
> I'm seeing some weird { ; } artefacts.

V3 adds semicolons where omitted to support no-semicolons,

> some duplication in ssr.test:

This looks like a bug, whats the full input?

thysultan:
@philpl patched some of theses in 3.0.1, BTW the cascade option should be true(default), false turns of cascading.

thysultan:
Managed to patch it...

Where does `.sc-a { }` come from?

mxstbr:
>Where do these empty blocks .sc-a { } come from?

Those are on purpose, they are per-component rather than the others, which are per-instance-of-css-fragment :blush:

thysultan:
I was trying to track down where and how they are passed to stylis, since stylis removes empty blocks its a bit odd that they would appear if they are getting passed through it.

thysultan:
Most failed tests now are whitespace related... 

Is there a way to have the tests break up on a failed test to mitigate having to view all the failed tests at once.

philpl:
Hiya, I'll finish this PR this evening :) yea, you can use jest's file filter with the P key

philpl:
@thysultan found another bug... I think:

```diff
-  .sc-a{ } .c{ color: blue; } .c > h1{ font-size: 4rem; } .sc-b{ } .d{ color: blue; } .d > h1{ font-size: 4rem; } .d{ color: red; }
+ .sc-a{ } .c{ color: blue; } .c > h1{ font-size: 4rem; } .sc-b{ } .d{ color: blue; color: red; } .d > h1{ font-size: 4rem; }
```

Source:

```js
    const Parent = styled.div`color: blue;`
    const Child = Parent.extend`color: red;`
```

It seems to be combining the two `color` properties here, but I'm unsure whether this changes the behaviour in terms of specificity, due to it being the same, but in a different order? (A bit confused)

geelen:
Hmm, is the red in the above expected or what v3 is putting out. Because I think it's more correct.

geelen:
Depends on what exactly is in that test case. Could you post a link to the source?

philpl:
@geelen it's merging the two rules that have the same selectors in the correct order. So (see Edit2) i think it's correct.

philpl:
@geelen the updated test case is here: https://github.com/styled-components/styled-components/pull/829/commits/7c7a0a8ea6ccb541bde52e50f3f56828929f95f0

Wanna review this? :)

thysultan:
It's by design, it now behaves like it would in [sass/postcss](https://www.sassmeister.com/) we where talking about this before here [#286](https://github.com/styled-components/styled-components/pull/286#issuecomment-273436290) decided it's better to have parity with scss to avoid unexpected results, i.e nested `@media`block issues.

BTW, is there any reason for this [stringifyRules.js#L23-L25](https://github.com/styled-components/styled-components/blob/79a7552cb4d89f22eddb3b04acdca5bbdaa1f9ef/src/utils/stringifyRules.js#L23-L25)

```js
  const cssStr = (selector && prefix) ?
    `${prefix} ${selector} { ${flatCSS} }` :
    flatCSS

// and
.replace(/^\s*\/\/.*$/gm, '') // replace JS comments
```

philpl:
@thysultan yea, these lines are for several cases that need to line up with the preprocessing:

On the normal side: https://github.com/styled-components/styled-components/blob/master/src/constructors/keyframes.js#L19

On the preprocessing side we then use it like this: https://github.com/styled-components/styled-components/blob/master/src/no-parser/stringifyRules.js#L13

It's all stuff that will be cleaned up with v3, since the entire preprocessing logic and the way we'll use stylis will change then :) (as discussed prior)

brunolemos:
@philpl don't you get this error? Maybe it's because I'm using `react-native-web`:

`ReferenceError: escade is not defined #30`

https://github.com/thysultan/stylis.js/issues/30

philpl:
@brunolemos Oh, I didn't get that error in any test. Weird. Seems to be fixed now and we'll just need to update the yarn lockfile on release

PR #837, opened 2017-05-26
Add missing TS typings for V2

patrick91:
Hi everyone, this is still in progress, I've started creating the missing typings for V2.

Here's a recap of what's missing/done:

- [x] SSR
- [x] extend
- [x] withComponent
- [x] attrs
- [x] function themes

Probably I've missed something (I'm about to catch a plane :) ) let me know what's missing and I'll add here :)


Igorbek:
Sorry for my silence, I was in a long trip. Nice work! Let me take a look more closely today-tomorrow.

typings/styled-components.d.ts
```diff
@@ -24,9 +24,100 @@ export interface InterpolationFunction<P> {
   (props: P): Interpolation<P>;
 }
 
+export interface StyledComponentClass<P, T> extends ComponentClass<P> {
+  extend: ThemedStyledFunction<P, T>;
+
+  withComponent(
```
Igorbek: Regarding [your question](https://github.com/styled-components/styled-components/pull/837#issuecomment-304453938), I think we can very efficiently use mapped types here.
Say, we define all the tags and corresponding props as following:
```ts
interface HtmlTags {
  a: HTMLAnchorElement;
  area: HTMLAreaElement;
  // ..., svg
}
```
then define like this:
```ts
type WithComponentOverloads<T> = {
  [K in keyof HtmlTags]: (tag: K) => StyledComponentClass<HtmlTags[K], T>;
};
interface StyledComponent<T> {
  withComponent: WithComponentOverloads<T>[keyof HtmlTags];
  //...
}
```
same `HtmlTags` can also be used in declaration of `ThemedStyledInterface`.

I'm not really sure how well it's gonna work. I can try to experiment with that.

typings/styled-components.d.ts
```diff
@@ -24,9 +24,100 @@ export interface InterpolationFunction<P> {
   (props: P): Interpolation<P>;
 }
 
+export interface StyledComponentClass<P, T> extends ComponentClass<P> {
```
patrick91: done thanks!

patrick91:
Hey @Igorbek thanks for the suggestions, now it looks definitively better. Still have to finish it tho.

Can you check why this [line](https://github.com/patrick91/styled-components/blob/bda9b483beaa7ba2dfbcd22c1680188308eeaf1d/typings/tests/main-test.tsx#L57) is not getting the correct type anymore? I might have messed up the definitions.

http://s.patrick.wtf/Txa8tKpfg3

Igorbek:
@patrick91 it was very exhaustive investigation... ok, so it seems that `@types/react@=0.14.57` has actually screwed it up. It is working well on 0.14.56 and 15.0.25. I think we should be safe to update to the latest `@types/react`

patrick91:
@Igorbek that's better! Thanks :)

Almost ready to go :)

typings/styled-components.d.ts
```diff
@@ -24,9 +26,21 @@ export interface InterpolationFunction<P> {
...
+  withComponent<K extends keyof HTMLTags>(tag: K): WithComponentOverloads<HTMLTags, T>[K];
+  withComponent<K extends keyof SVGTags>(tag: K): WithComponentOverloads<SVGTags, T>[K];
+  withComponent(element: ComponentClass<P>): StyledComponentClass<ComponentClass<P>, T>;
+}
+
 export interface ThemedStyledFunction<P, T> {
-  (strings: TemplateStringsArray, ...interpolations: Interpolation<ThemedStyledProps<P, T>>[]): ComponentClass<ThemedOuterStyledProps<P, T>>;
-  <U>(strings: TemplateStringsArray, ...interpolations: Interpolation<ThemedStyledProps<P & U, T>>[]): ComponentClass<ThemedOuterStyledProps<P & U, T>>;
+  (strings: TemplateStringsArray, ...interpolations: Interpolation<ThemedStyledProps<P, T>>[]): StyledComponentClass<P, T>;
+  <U>(strings: TemplateStringsArray, ...interpolations: Interpolation<ThemedStyledProps<P & U, T>>[]): StyledComponentClass<P, T>;
```
Igorbek: should return `StyledComponentClass<P & U, T>` here

typings/styled-components.d.ts
```diff
@@ -36,148 +50,17 @@ export type HtmlStyledFunction<E> = ThemedHtmlStyledFunction<E, any>;
 export type ThemedSvgStyledFunction<E extends SVGElement, T> = ThemedStyledFunction<React.SVGAttributes<E>, T>;
 export type SvgStyledFunction<E extends SVGElement> = ThemedSvgStyledFunction<E, any>;
 
-export interface ThemedBaseStyledInterface<T> {
+type Base<T> = {
+    [K in keyof HTMLTags]: ThemedHtmlStyledFunction<HTMLTags[K], T>;
```
Igorbek: SVG tags are missing

typings/styled-components.d.ts
```diff
@@ -36,148 +50,17 @@ export type HtmlStyledFunction<E> = ThemedHtmlStyledFunction<E, any>;
 export type ThemedSvgStyledFunction<E extends SVGElement, T> = ThemedStyledFunction<React.SVGAttributes<E>, T>;
 export type SvgStyledFunction<E extends SVGElement> = ThemedSvgStyledFunction<E, any>;
 
-export interface ThemedBaseStyledInterface<T> {
+type Base<T> = {
```
Igorbek: better name: `ThemedStyledComponentFactories`, or something similar?

typings/styled-components.d.ts
```diff
@@ -207,9 +90,12 @@ export function injectGlobal(strings: TemplateStringsArray, ...interpolations: S
 
 export const ThemeProvider: ThemeProviderComponent<any>;
 
+export class StyleSheetManager extends React.Component<{ sheet: ServerStyleSheet }, any> { }
+
 export class ServerStyleSheet {
-  collectStyles(children: ReactElement<any>): ReactElement<any>
-  getStyleTags(): string
+  collectStyles(children: ReactElement<any>): StyleSheetManager
+  collectStyles(tree: any): StyleSheetManager;
```
Igorbek: is anything better than `any`? Maybe `React.ReactNode`? Is that what we need?

typings/styled-components.d.ts
patrick91: Done!

typings/styled-components.d.ts
patrick91: Done, I had to use an hack to make it work with both, but seems fine :)

typings/styled-components.d.ts
patrick91: ReactNode is defined as `type React.ReactNode = string | number | boolean | {} | React.ReactElement<any> | (string | number | boolean | any[] | React.ReactElement<any>)[]
`
seems to accept anything anyway

Igorbek: It's not really an `any`, so I would argue :) and `ReactNode` would leave just a single simple signature: `collectStyles(tree: React.ReactNode): StyleSheetManager`.
We should be very careful with `any`, it's always a source of many hidden bugs.
My personal opinion is to start with more restrictive types and extend them if needed, other direction is always a breaking change.

typings/styled-components.d.ts
```diff
@@ -24,9 +26,21 @@ export interface InterpolationFunction<P> {
...
+type WithComponentOverloads<Tags, T> = {
+  [K in keyof Tags]: StyledComponentClass<Tags[K], T>;
+};
+
+export interface StyledComponentClass<P, T> extends ComponentClass<ThemedOuterStyledProps<P, T>> {
+  extend: ThemedStyledFunction<P, T>;
+
+  withComponent<K extends keyof HTMLTags>(tag: K): WithComponentOverloads<HTMLTags, T>[K];
+  withComponent<K extends keyof SVGTags>(tag: K): WithComponentOverloads<SVGTags, T>[K];
+  withComponent(element: ComponentClass<P>): StyledComponentClass<ComponentClass<P>, T>;
```
Igorbek: ~`withComponent(element: ComponentClass<P>): StyledComponentClass<ComponentClass<P>, T>;`~
`withComponent(element: ComponentClass<P>): StyledComponentClass<P, T>;`

typings/styled-components.d.ts
```diff
@@ -207,9 +96,12 @@ export function injectGlobal(strings: TemplateStringsArray, ...interpolations: S
 
 export const ThemeProvider: ThemeProviderComponent<any>;
 
+export class StyleSheetManager extends React.Component<{ sheet: ServerStyleSheet }, any> { }
+
 export class ServerStyleSheet {
-  collectStyles(children: ReactElement<any>): ReactElement<any>
-  getStyleTags(): string
+  collectStyles(children: ReactElement<any>): StyleSheetManager
```
Igorbek: By looking [the docs](https://www.styled-components.com/docs/advanced#server-side-rendering) I figure `collectStyles` doesn't really return a `StyleSheetManager` instance, it returns _a react element_. So maybe:
```ts
collectStyles(tree: React.ReactNode): ReactElement<{ sheet: ServerStyleSheet; }>;
```

And `{ sheet: ServerStyleSheet; }` can be defined in its own interface since it's reused.

Igorbek:
BTW, @patrick91 do you need any help with coding that?

philpl:
Just a quick FYI here. I'd like to publish a patch or minor version soon, with stylis 3, but I'm waiting for this to be merged :)

patrick91:
@philpl I should be done by the weekend :)

@Igorbek you're already doing a lot! thanks. I might need some help with attrs, but I'm not sure yet, still have to start working on it :)

patrick91:
@Igorbek can you have a look at the attrs definition? A part from the duplication (that can be easily fixed) I'm not sure if that's the best way to do it :)

typings/styled-components.d.ts
```diff
@@ -24,10 +26,29 @@ export interface InterpolationFunction<P> {
...
+  // [K in keyof P]: (props: ThemedOuterStyledProps<P, T>) => any;
+}
+
+export interface StyledComponentClass<P, T> extends ComponentClass<ThemedOuterStyledProps<P, T>> {
+  extend: ThemedStyledFunction<P, T>;
+
+  withComponent<K extends keyof HTMLTags>(tag: K): WithComponentOverloads<HTMLTags, T>[K];
+  withComponent<K extends keyof SVGTags>(tag: K): WithComponentOverloads<SVGTags, T>[K];
+  withComponent(element: ComponentClass<P>): StyledComponentClass<P, T>;
+  attrs(attrs: P | PropsToFunctions<ThemedOuterStyledProps<P, T>, T>): ThemedStyledFunction<P, T>;
```
Igorbek: - `attrs` does not belong to the styled component class, it should be within `ThemedStyledFunction`
- the argument for `attrs` is an object that mixes: values for **declared** props, values for **fresh** props, and functions that accept existing inner props and return values for existing or fresh props
- `attrs` should return a themed styled function that carries information about provided and introduced props (styled function can create a styled component without specifying required props, and with using introduced props)
It's most likely impossible to define this properly in typescript type system. It currently lacks proper higher order functions support. I've played a bit with `attrs` definition yesterday and could only come up with the following:
```ts
type Attrs<P, A extends Partial<P>, T> = {
  [K in keyof A]: ((props: ThemedStyledProps<P, T>) => A[K]) | A[K];
};

export interface ThemedStyledFunction<P, T, O = P> {
  (strings: TemplateStringsArray, ...interpolations: Interpolation<ThemedStyledProps<P, T>>[]): StyledComponentClass<O, T>;
  <U>(strings: TemplateStringsArray, ...interpolations: Interpolation<ThemedStyledProps<P & U, T>>[]): StyledComponentClass<O & U, T>;
  attrs<U, A  extends Partial<P> = {}>(attrs: Attrs<P & U, A, T>): ThemedStyledFunction<P & A & U, T, O & U>;
}
```
- I had to introduce a new generic type parameter `O` as a type for _outer props_ of future style components
- I discovered that functions inside are not contextually typed, and filed a bug in TS https://github.com/Microsoft/TypeScript/issues/16209
- specifying `U` explicitly in `attrs` breaks inference for `A` because of https://github.com/Microsoft/TypeScript/issues/14400

typings/styled-components.d.ts
```diff
@@ -24,10 +26,29 @@ export interface InterpolationFunction<P> {
...
+}
+
 export interface ThemedStyledFunction<P, T> {
-  (strings: TemplateStringsArray, ...interpolations: Interpolation<ThemedStyledProps<P, T>>[]): ComponentClass<ThemedOuterStyledProps<P, T>>;
-  <U>(strings: TemplateStringsArray, ...interpolations: Interpolation<ThemedStyledProps<P & U, T>>[]): ComponentClass<ThemedOuterStyledProps<P & U, T>>;
+  (strings: TemplateStringsArray, ...interpolations: Interpolation<ThemedStyledProps<P, T>>[]): StyledComponentClass<P, T>;
+  <U>(strings: TemplateStringsArray, ...interpolations: Interpolation<ThemedStyledProps<P & U, T>>[]): StyledComponentClass<P, T>;
+  attrs(attrs: P | PropsToFunctions<ThemedOuterStyledProps<P, T>, T>): ThemedStyledFunction<P, T>;
```
Igorbek: You actually have `attrs` here too, sorry, didn't notice.

typings/styled-components.d.ts
patrick91: Thanks, that's definitively better!

typings/styled-components.d.ts
patrick91: Yeah, I had to add them in both places other ways this wouldn't have worked:

```ts
const Button = styled.button`
  color: palevioletred;
  font-size: 1em;
  margin: 1em;
  padding: 0.25em 1em;
  border: 2px solid palevioletred;
  border-radius: 3px;
`;

const TomatoButton = Button.attrs({
  title: (props) => "title"
})`
  color: red;
`;

```

Igorbek: Hm, I didn't know this syntax was supported too.

typings/styled-components.d.ts
```diff
@@ -1,10 +1,15 @@
...
+import { HTMLTags, SVGTags } from "./tags";
+
 type Component<P> = ComponentClass<P> | StatelessComponent<P>;
 
+type ThemeFunction<T> = (theme: any | T) => any | T;
+type Theme<T> = any | ThemeFunction<T>;
+
 export interface ThemeProps<T> {
-  theme: T;
+  theme: Theme<T>;
```
Igorbek: I think this is _inner_ props, so they should be strictly `T`, type of the theme. However, outer props can have optional `theme` as `T | (theme: T) => T`.

typings/styled-components.d.ts
patrick91: Had to change

```ts
  attrs<U, A extends Partial<P> = {}>(attrs: Attrs<P & U, A, T>): ThemedStyledFunction<P & A & U, T, O & U>;
```

to 

```ts
  attrs<U, A extends Partial<P> = {}>(attrs: Attrs<P | U, A, T>): ThemedStyledFunction<P & A & U, T, O & U>;
```

typings/styled-components.d.ts
```diff
@@ -36,148 +60,23 @@ export type HtmlStyledFunction<E> = ThemedHtmlStyledFunction<E, any>;
 export type ThemedSvgStyledFunction<E extends SVGElement, T> = ThemedStyledFunction<React.SVGAttributes<E>, T>;
 export type SvgStyledFunction<E extends SVGElement> = ThemedSvgStyledFunction<E, any>;
 
-export interface ThemedBaseStyledInterface<T> {
+type ThemedStyledComponentFactoriesHTML<T> = {
```
Igorbek: Minor: would Pascal-case `Html`, `Svg` be better for consistency (`ThemedShtmlStyledFunction`)

typings/styled-components.d.ts
```diff
@@ -24,10 +29,29 @@ export interface InterpolationFunction<P> {
...
+};
+
+type Attrs<P, A extends Partial<P>, T> = {
+  [K in keyof A]: ((props: ThemedStyledProps<P, T>) => A[K]) | A[K];
+};
+
+export interface StyledComponentClass<P, T> extends ComponentClass<ThemedOuterStyledProps<P, T>> {
+  extend: ThemedStyledFunction<P, T>;
+
+  withComponent<K extends keyof HTMLTags>(tag: K): WithComponentOverloads<HTMLTags, T>[K];
```
Igorbek: simpler:
`withComponent<K extends keyof HtmlTags>(tag: K): StyledComponentClass<HtmlTags[K], T>;`

typings/styled-components.d.ts
```diff
@@ -24,10 +29,29 @@ export interface InterpolationFunction<P> {
...
+  [K in keyof A]: ((props: ThemedStyledProps<P, T>) => A[K]) | A[K];
+};
+
+export interface StyledComponentClass<P, T, O = P> extends ComponentClass<ThemedOuterStyledProps<P, T>> {
+  extend: ThemedStyledFunction<P, T>;
+
+  withComponent<K extends keyof HTMLTags>(tag: K): WithComponentOverloads<HTMLTags, T>[K];
+  withComponent<K extends keyof SVGTags>(tag: K): WithComponentOverloads<SVGTags, T>[K];
+  withComponent(element: ComponentClass<P>): StyledComponentClass<P, T>;
+  attrs<U, A  extends Partial<P> = {}>(attrs: Attrs<P | U, A, T>): ThemedStyledFunction<P & A & U, T, O & U>;
```
Igorbek: No, it has to be exactly that `Attrs<P & U, A, T>`
`U` represents a newly introduced fresh props. They will be added to props accessible in the attr's functions.

typings/styled-components.d.ts
Igorbek: and maybe `ThemedStyledComponentHtmlFactories`?

typings/styled-components.d.ts
Igorbek: why's that?

typings/styled-components.d.ts
patrick91: Then I get this error: http://s.patrick.wtf/tyx78QkcOF

typings/styled-components.d.ts
patrick91: it is actually not supported! sorry :)

typings/styled-components.d.ts
patrick91: Well TypeScript definition are all uppercase: http://s.patrick.wtf/0ygEY1jRl4
Maybe I can replace Html with HTML everywhere :)

Igorbek:
Hey @patrick91 I can back to reviewing Sunday evening only. I'm in a trip.

patrick91:
@Igorbek no worries :)

typings/styled-components.d.ts
Igorbek: this's gonna work:
```ts
attrs<U, A extends Partial<P & U> = {}>(attrs: Attrs<P & U, A, T>): ThemedStyledFunction<P & A & U, T, O & U>;
```

Igorbek:
Hey @patrick91 I made a few changes based on your branch and my own feedback to avoid miscommunication. Take a look comparison here: https://github.com/patrick91/styled-components/compare/feature/typings-2...Igorbek:feature/typings-2
I can either create a new PR here or create a PR to your repo, if you like. I'm not sure if my changes would seem acceptable to you. One note I need to mention, is that I remove `StyleSheetManager.create()` since it returns `StyleSheet` which needs its own definition and currently is being recognized as a DOM's StyleSheet, which is incorrect, obviously.

patrick91:
Wow that's great! Thanks

patrick91:
@Igorbek I made a small change, instead of `any` we have now `object` for the theme prop. So that only objects (and functions) can be passed to the theme.

If that's ok, we are ready to merge :)

patrick91:
done!

@philpl shall we merge? :)

philpl:
Looking great! Awesome work 🌈

mxstbr:
Awesome!! :tada:

codepunkt:
This is amazing. Just removed 2.0.0 and installed from GitHub. Thanks guys!

cesalberca:
This is just what I needed 😊! When is this expected to be released? Also, do we have to change the way we re-export styled-components as the [documentation](https://www.styled-components.com/docs/api#typescript) says?

philpl:
@cesalberca We'll release a 3.0.1 patch today :)

patrick91:
@cesalberca no need to re-export if you don't need theming or auto complete support for it :)

cesalberca:
Great, I'll be among the first downloads ;) Also, I don't know if this has to do with the current typings or if I should create an issue for this, but I can't refer to other components. My use case is exactly the same as the one described in the [documentation](https://www.styled-components.com/docs/faqs#can-i-refer-to-other-components).

When in my code I try to do the following:

```javascript
import * as React from 'react'
import styled from './../constants/styledComponents'

interface Props {
  title: string
  onClick?: React.EventHandler<React.MouseEvent<HTMLDivElement>>
}

const Card: React.StatelessComponent<Props> = props => {
  const { title, onClick } = props
  return (
    <StyledCard onClick={onClick}>
      <Container>
        <Title>{title}</Title>
      </Container>
      <Backdrop>{title[0]}</Backdrop>
    </StyledCard>
  )
}

const StyledCard = styled.div`
  box-shadow: 0 0 30px var(--background-medium);
  background-color: white;
  border-radius: var(--border-radius);
  transition: 0.3s ease all;
  position: relative;
  overflow: hidden;
  width: 200px;
  height: 100px;

  &:hover {
    cursor: pointer;
    transform: translateY(-5px);
  }

  &:active {
    transform: translateY(7px);
  }
`

const Container = styled.div`
  padding: calc(var(--spacing-unit) / 2) var(--spacing-unit);
  margin: calc(var(--spacing-unit) / 2);
  display: flex;
  position: relative;
  justify-content: center;
  z-index: 2;
`

const Title = styled.h3`
  margin-top: 0;
  color: var(--background-medium);
`

const Backdrop = styled.span`
  position: absolute;
  right: 0;
  z-index: 1;
  top: 0;
  bottom: 0;
  font-weight: bold;
  color: var(--background-light);
  transform: translateX(25%);
  font-size: 15rem;
  transition: 0.3s ease all;

  ${StyledCard}:hover & {
    color: var(--primary);
  }
`

export default Card
```

I get the following Typescript error:

```
Argument of type 'ComponentClass<ThemedOuterStyledProps<HTMLProps<HTMLDivElement>, ITheme>>' is not assignable to parameter of type 'Interpolation<ThemedStyledProps<HTMLProps<HTMLSpanElement>, ITheme>>'.
  Type 'ComponentClass<ThemedOuterStyledProps<HTMLProps<HTMLDivElement>, ITheme>>' is not assignable to type 'ReadonlyArray<string | number | InterpolationFunction<ThemedStyledProps<HTMLProps<HTMLSpanElement...'.
    Property 'includes' is missing in type 'ComponentClass<ThemedOuterStyledProps<HTMLProps<HTMLDivElement>, ITheme>>'.
```

I've followed the [Typescript section](https://www.styled-components.com/docs/api#typescript) in the documentation and I was wondering if this error will go away with the new typings, if not, any ideas why is this happening? Thanks!

Igorbek:
Ah, we totally forgot about this feature.

patrick91:
@Igorbek this should be fine

```ts
export interface ThemedStyledFunction<P, T, O = P> {
  (strings: TemplateStringsArray, ...interpolations: (StyledComponentClass<P, T> | Interpolation<ThemedStyledProps<P, T>>)[]): StyledComponentClass<P, T, O>;
  <U>(strings: TemplateStringsArray, ...interpolations: (StyledComponentClass<P, T> | Interpolation<ThemedStyledProps<P & U, T>>)[]): StyledComponentClass<P & U, T, O & U>;
  attrs<U, A extends Partial<P & U> = {}>(attrs: Attrs<P & U, A, T>): ThemedStyledFunction<P & A & U, T, O & U>;
}
```

haven't tested it properly yet. I'm wondering if this work in css`xxx` as well

philpl:
@patrick91 @Igorbek Is this sth that will need to be fixed before the patch release?

Igorbek:
I'd say yes, otherwise we'd need another patch release later. It will take a couple hours for me to create a PR.

cesalberca:
I feel like I've opened Pandora's box 😅. Good luck!

Igorbek:
@cesalberca it's been done in #882

arnettha:
Are these typings can already be used? I tried to use extend and attrs (with v2.0.1) but still got "not a function error". Any help is appreciated! 😃 

> TypeError: __WEBPACK_IMPORTED_MODULE_6_styled_components__.default.input.attrs is not a function

philpl:
@arnettha that looks like a webpack issue, not a TS one. Are you sure you're using v2?

arnettha:
@philpl by js error do you mean syntax error? Or something else? Here is how I used the attrs
```
const Radio = styled.input.attrs({type: 'radio'})`
...
`;
```
Yes, I'm sure I'm using v2. I checked it by `npm list styled-components` in my project directory and it showed something like this:

PR #859, opened 2017-05-31
Copy nested rules to multiple child

brunoqueiros:
With this PR you can extend a component that is extending another component, in current version you only get the rules from the direct parent.

```js
const Parent = styled.div`color: red;`
const Child = Parent.extendWith('span')`font-size: 10px;`
const ChildChild = Child.extendWith('span')`background-color: pink;`
```

ChildChild will have both Parent and Child rules.


philpl:
I think this wasn't up-to-date with v2, since it still contains `extendsWith`, but @brunolemos fixed it here: https://github.com/styled-components/styled-components/pull/871

Thanks a lot for the PR anyway! :) :+1:

Issue #843, opened 2017-05-28
Flow-typed libdef for v2


relekang:
Just opening this task if someone is looking for v2 libdef so they know it is in progress. I am working on it at https://github.com/relekang/flow-typed/tree/fix/styled-components-v2. 

cc @styled-components/typers

philpl:
Hiya, do you have any update on this? :) Unfortunately my Flow skills are not up to my TS skills, so I'd rather no offer my help. Sorry

relekang:
Sorry no, my oss time has been spent on other stuff. I can take another try at it today and se how far I get.

I was wondering tho if we should take it back into the repo, but keep it as a libdef? Maybe put it in `typings/flow/index.flow.js` and `typings/flow/native.flow.js`. The workflow for adding it to the project would then be to install styled-components and add `node_modules/styled-components/typings/flow` under `[libs]` section of `.flowconfig`.

cc @ryyppy

philpl:
@relekang It'd definitely be nicer, and I'm not quite clear on why we moved it over to flow-typed in the first place. Shouldn't it even add these libs automatically? I'm having some issues on v2 that it's using the included ones by default until I add the ignore statement to the config

relekang:
> I'm not quite clear on why we moved it over to flow-typed in the first place.

This [tweet(-storm)](https://twitter.com/lbljeffmo/status/813434277018955776) 🙃 

> Shouldn't it even add these libs automatically?

The generate functionality in flow is currently broken as far as i know, have not had the time to test with 0.47 yet.

> I'm having some issues on v2 that it's using the included ones by default until I add the ignore statement to the config

Yeah, that should be fixed by my recent PR. One of the main problems here is that it surfaces relations with styled-components dependencies in the project using styled-components, which gives a bunch of weird errors. 

Moving to libdef was a good call, but I think it might make sense to keep them here. We could always PR them to flow-typed as well if people want to get them from there.

IvanCoronado:
I am facing this problem just now. Waiting for that PR.
Thanks you @relekang

relekang:
@philpl I opened the PR on flow-typed, it is not complete, but should be a good start. I will look into moving the libdef to this repo when I have more time. Need to figure out to test it without the flow-typed cli.

relekang:
The PR was just merged to flow-typed. Please test it and report any issues here 🙏 

I did not get the time to add `attrs`, will look into that next week if nobody else want to fix it.

[...]

PR #868, opened 2017-06-02
Remove the flow copy logic from build process

relekang:
These should not be used anymore as we have written in the docs lately
libdef is now the easiest way to use flow with styled-components. Using
the sources as we did earlier generates a lot of errors that is hard to
avoid.

cc @styled-components/typers


philpl:
I'm pretty sure I was actually running into some issues due to this on Friday, so thanks for taking a look :)

Andarist:
@styled-components/typers @relekang @philpl could I ask how do you maintain public typings now? it seems really problematic as they live in the other global `flow-typed` repo?

relekang:
There is an ongoing discussion/effort in #843 on moving it back in to this repo, but still as a libdef not sources like this pr removes. I don't know when I will be able to work on that, probably not for a few weeks. There is a bunch of trade-offs with both tho 😅 Mainly flow-typed makes it harder to keep up to date, however, putting it in this repo makes it harder to support several incompatible versions of flow.

To answer your question on how it is maintained it is mostly a catch-up game where a few of us try to keep it up to date with changes in the library whenever we have time to spare.

PR #869, opened 2017-06-02
Added reference to styled-components-theme lib

erikras:
I find it very useful. But then I'm biased. 😆


k15a:
Could you please move this to the end of the helpers list to be fair to others?

erikras:
I was going by alphabetical order.

erikras:
Also, my lib predates all the other "Helpers". But okay...

k15a:
We always handled it this way so it's just fair :)

k15a:
There is also https://github.com/romellogood/awesome-styled-components if you want to submit have your library on another list as well.

PR #871, opened 2017-06-05
Fix extend for 3 or more inheritances (Fix #780)

### LINKED ISSUE #780: [V2] Double extend not working in v2.0.0-17
Hi, extending an already extended styled-component is not working in v2.0.0-17.

```
const a = styled.div`...`
const b = a.extend`...` ---> works
const c = b.extend`...` ---> not working
```

I hope this explains the issue enough, otherwise I can create an example gist

brunolemos:
Fix #780

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file. If a contri
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


brunolemos:
No problem!

Another thing to add in the refactoring roadmap is to avoid all the code duplication between web and native code, it's too easy to make a change on one and forget to do in the other. 

There are some Design Patterns that would fit well in the project, specially the [Strategy](https://sourcemaking.com/design_patterns/strategy) one.

philpl:
@brunolemos Do you want to open a discussion with some thoughts on that maybe? The native/web code duplication is a big problem as well, and it'd be nice to finally solve it :)

PR #873, opened 2017-06-06
Add additional proxy methods for NativeMethodsMixin

kitten:
*Making SC Native future proof (TM)*

This will make Styled Components future-proof for
Yoga Nodes, which depends on a couple of additional
methods to exist. It's unclear whether this will actually
be enabled by default, but I'm sure it doesn't hurt to support
Yoga Nodes for those who need it.

The methods that were added: https://facebook.github.io/react-native/docs/nativemethodsmixin.html

### DIFF
--- a/src/models/StyledNativeComponent.js
+++ b/src/models/StyledNativeComponent.js
@@ -147,7 +147,6 @@ export default (constructWithOptions: Function) => {
     // $FlowFixMe
     BaseStyledNativeComponent.prototype[methodKey] = function nativeProxy(...args) {
       if (this.root !== undefined) {
-        // $FlowFixMe
         this.root[methodKey](...args)
       } else {
         const { displayName } = this.constructor
--- a/src/models/StyledNativeComponent.js
+++ b/src/models/StyledNativeComponent.js
@@ -12,6 +12,15 @@ import { CHANNEL } from './ThemeProvider'
 import InlineStyle from './InlineStyle'
 import AbstractStyledComponent from './AbstractStyledComponent'
 
+const nativeMethodKeys = [
+  'setNativeProps',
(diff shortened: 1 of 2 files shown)


philpl:
Or maybe @k15a ? :)

k15a:
Unfortunately I not using React Native very much so I don't know if this is used or not ;) Mabye we should wait until someone actually requests this feature?

philpl:
@k15a I'd rather add it now, since it doesn't really hurt anyone and is making the proxying more complete

philpl:
Ok, let's keep this around for the next patch release :)

src/models/StyledNativeComponent.js
```diff
@@ -150,6 +143,24 @@ export default (constructWithOptions: Function) => {
...
+      if (this.root !== undefined) {
+        this.root[methodKey](...args)
+      } else {
+        const { displayName } = this.constructor
+
+        // eslint-disable-next-line no-console
+        console.warn(
+          `${methodKey} was called on a Styled Component wrapping a stateless functional component. ` +
+          'In this case no ref will be stored, and instead an innerRef prop will be passed on.\n' +
+          `Check whether the stateless functional component is passing on innerRef as a ref in ${displayName}.`,
```
geelen: Is this the right message? Is it really to do with ref and innerRef?

philpl: Yes, we can assume that when the this.root isn't set, that we couldn't set it because the underlying component is a SFC, and that the user hasn't passed on innerRef :)

mxstbr:
Can we ship this once conflicts are released?

probablyup:
@kitten do we still want to ship this?

PR #880, opened 2017-06-07
Update inline-style-prefixer to the latest version

rofrischmann:
Updating the prefixer to the latest version basically improves prefixing performance by 4-5x
Definitely something you should consider here!

Drawback:
We updated the browser support versions which might change the prefixing behavior. You can create a custom prefixer instance to solve that - let me know and I update the PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "css-to-react-native": "^2.0.3",
     "fbjs": "^0.8.9",
     "hoist-non-react-statics": "^1.2.0",
-    "inline-style-prefixer": "^2.0.5",
+    "inline-style-prefixer": "^3.0.5",
     "is-function": "^1.0.1",
     "is-plain-object": "^2.0.1",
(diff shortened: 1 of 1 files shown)


k15a:
We don't use inline-style-prefixer anymore because stylis is automatically prefixing for us. Don't know why we still have inline-style-prefixer in the package.json. @philpl any idea?

philpl:
Oh, sorry! @k15a is correct, we're not using it anymore, and there's some dead code left over from it.

rofrischmann:
Ah I see. Was wondering why it was still in the package.json and included in the autoprefix util.

philpl:
Opened a PR to fix this: https://github.com/styled-components/styled-components/pull/881

@rofrischmann We had some dead code while we migrated to v2, and some of it wasn't deleted it seems :/ I'm not really sure why we didn't pick up earlier on that

PR #882, opened 2017-06-07
Add TypeScript support of using other styled components in selectors

Igorbek:
A following up PR to #837

- adds support of using other components in interpolations
  - tests
- removes a few unnecessary `any`s
- update `CHANGELOG`

### DIFF
--- a/typings/styled-components.d.ts
+++ b/typings/styled-components.d.ts
@@ -20,7 +20,7 @@ export type OuterStyledProps<P> = ThemedOuterStyledProps<P, any>;
 
 export type Interpolation<P> = FlattenInterpolation<P> | ReadonlyArray<FlattenInterpolation<P> | ReadonlyArray<FlattenInterpolation<P>>>;
 export type FlattenInterpolation<P> = InterpolationValue | InterpolationFunction<P>;
-export type InterpolationValue = string | number;
+export type InterpolationValue = string | number | StyledComponentClass<any, any>;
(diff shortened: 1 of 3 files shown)


Igorbek:
@philpl @patrick91 @brunolemos please review.
BTW is it ok that I change the `CHANGELOG`?

typings/styled-components.d.ts
```diff
@@ -20,7 +20,7 @@ export type OuterStyledProps<P> = ThemedOuterStyledProps<P, any>;
 
 export type Interpolation<P> = FlattenInterpolation<P> | ReadonlyArray<FlattenInterpolation<P> | ReadonlyArray<FlattenInterpolation<P>>>;
 export type FlattenInterpolation<P> = InterpolationValue | InterpolationFunction<P>;
-export type InterpolationValue = string | number;
+export type InterpolationValue = string | number | StyledComponentClass<any, any>;
```
brunolemos: Not sure an interpolation can be a StyledComponentClass?

Igorbek: It can. So `StyledComponentClass` produced by a styled function:
```ts
const Link: StyledComponentClass<any, any> = styled.a` color: blue; `;
```
then it can be used in both functional and constant interpolations:
```ts
const Section = styled.section`
  color: black;
  ${Link} {
    color: red;
  }
  ${p => p.theme.linkClass || Link} {
    color: red;
  }
`;
```

brunolemos: Oh right, that's new from v2

brunolemos:
@Igorbek I'll let for people who use TypeScript to review your code

PR #904, opened 2017-06-11
add react-primitives target

mathieudutour:
@mxstbr Here it is :)

Couple of things:
- I add some abstraction in `InlineStyle.js` and `StyledNativeComponent.js` to pass the Stylesheet instead of requiring it from `react-native`. Probably need to change the name `StyledNativeComponent` but I was short on idea.
- The tests are failing on my machine for RN with `Cannot find module 'ReactElementType' from 'ReactRef.js'`. I'm opening this to get the result from the CI 🤞

### DIFF
--- a/primitives.js
+++ b/primitives.js
@@ -1,2 +1,0 @@
-/* eslint-disable flowtype/require-valid-file-annotation */
-module.exports = require('./lib/primitives')
--- a/primitives.js
+++ b/primitives.js
@@ -1,0 +1,2 @@
+/* eslint-disable flowtype/require-valid-file-annotation */
+module.exports = require('./lib/primitives')
--- a/src/models/InlineStyle.js
+++ b/src/models/InlineStyle.js
@@ -3,7 +3,7 @@
 import transformDeclPairs from 'css-to-react-native'
 
 import hashStr from '../vendor/glamor/hash'
-import type { RuleSet } from '../types'
+import type { RuleSet, TStyleSheet } from '../types'
 import flatten from '../utils/flatten'
 import parse from '../vendor/postcss-safe-parser/parse'
 
@@ -18,9 +18,9 @@ export const resetStyleCache = () => {
  */
 export default class InlineStyle {
   rules: RuleSet
-  StyleSheet: { create: Function }
+  StyleSheet: TStyleSheet
 
-  constructor(StyleSheet: { create: Function }, rules: RuleSet) {
+  constructor(StyleSheet: TStyleSheet, rules: RuleSet) {
     this.StyleSheet = StyleSheet
     this.rules = rules
   }
@@ -52,6 +52,7 @@ export default class InlineStyle {
       const styles = this.StyleSheet.create({
         generated: styleObject,
       })
+      console.log(styles)
       generated[hash] = styles.generated
     }
     return generated[hash]
--- a/src/models/InlineStyle.js
+++ b/src/models/InlineStyle.js
@@ -3,7 +3,7 @@
 import transformDeclPairs from 'css-to-react-native'
 
 import hashStr from '../vendor/glamor/hash'
-import type { RuleSet, TStyleSheet } from '../types'
+import type { RuleSet, StyleSheet } from '../types'
 import flatten from '../utils/flatten'
 import parse from '../vendor/postcss-safe-parser/parse'
 
@@ -18,14 +18,12 @@ export const resetStyleCache = () => {
  */
 export default class InlineStyle {
   rules: RuleSet
-  StyleSheet: TStyleSheet
 
-  constructor(StyleSheet: TStyleSheet, rules: RuleSet) {
-    this.StyleSheet = StyleSheet
+  constructor(rules: RuleSet) {
     this.rules = rules
   }
 
-  generateStyleObject(executionContext: Object) {
+  generateStyleObject(executionContext: Object, styleSheet: StyleSheet) {
     const flatCSS = flatten(this.rules, executionContext).join('')
     const hash = hashStr(flatCSS)
     if (!generated[hash]) {
@@ -49,10 +47,9 @@ export default class InlineStyle {
         'borderColor',
         'borderStyle',
       ])
-      const styles = this.StyleSheet.create({
+      const styles = styleSheet.create({
         generated: styleObject,
       })
-      console.log(styles)
       generated[hash] = styles.generated
     }
     return generated[hash]
--- a/src/models/InlineStyle.js
+++ b/src/models/InlineStyle.js
@@ -16,42 +16,46 @@ export const resetStyleCache = () => {
 /*
  InlineStyle takes arbitrary CSS and generates a flat object
  */
-export default class InlineStyle {
-  rules: RuleSet
+export default (styleSheet: StyleSheet) => {
+  class InlineStyle {
+    rules: RuleSet
 
-  constructor(rules: RuleSet) {
-    this.rules = rules
-  }
-
-  generateStyleObject(executionContext: Object, styleSheet: StyleSheet) {
-    const flatCSS = flatten(this.rules, executionContext).join('')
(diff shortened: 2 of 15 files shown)


philpl:
Sorry, I don't know of any prior discussions and haven't read the code yet ^^

Why do we need this when with `react-native-web` the user can theoretically just alias `react-native` to `react-native-web` in babel or webpack?

mathieudutour:
For use with the other platforms that react-primitives support (currently only Sketch using react-sketchapp) but as react-as-a-platform grows, I'm expecting more to appear and hopefully we will converge toward a "standard" set of primitives on each platform.

src/primitives/index.js
```diff
@@ -0,0 +1,37 @@
...
+const constructWithOptions = _constructWithOptions(css)
+const StyledNativeComponent = _StyledNativeComponent(
+  constructWithOptions,
+  reactPrimitives.StyleSheet,
+)
+const styled = (tag: Target) => constructWithOptions(StyledNativeComponent, tag)
+
+/* React native lazy-requires each of these modules for some reason, so let's
+*  assume it's for a good reason and not eagerly load them all */
+const aliases = 'Animated Image Text Touchable View '
```
philpl: `Animated` isn't a target (assuming it's the same as in React Native), but it does have properties for it's target e.g. `Animated.View`, so this can't be added as an alias.

src/primitives/test/primitives.test.js
```diff
@@ -0,0 +1,273 @@
+import 'react-primitives'
+import React from 'react'
+
+import styled from '../index'
+import { shallow, mount } from 'enzyme'
+
+// NOTE: These tests are like the ones for Web but a "light-version" of them
```
philpl: Can you update this comment to reflect that it's copied from the native tests?

src/types.js
```diff
@@ -22,3 +22,7 @@ export type Stringifier = (
   selector: ?string,
   prefix: ?string
 ) => string
+
+export type TStyleSheet = {
```
philpl: The name `TStyleSheet` is kind of clunky. Can it just be `StyleSheet`?

src/models/StyledNativeComponent.js
```diff
@@ -163,6 +163,7 @@ export default (constructWithOptions: Function) => {
     } = options
 
     const inlineStyle = new InlineStyle(
+      StyleSheet,
```
philpl: Instead of doing this, the `InlineStyle` model should become a factory function that accepts an RN/RP StyleSheet.

have look on how some other models do it :) This is the closest example obviously: https://github.com/mathieudutour/styled-components/blob/c65a1cb54d81eddbac48fa845627e48fb007d3ef/src/models/ComponentStyle.js#L11

So instead of passing your StyleSheet in, we can pass the generated InlineStyle in

src/models/InlineStyle.js
```diff
@@ -48,9 +49,10 @@ export default class InlineStyle {
         'borderColor',
         'borderStyle',
       ])
-      const styles = StyleSheet.create({
+      const styles = this.StyleSheet.create({
         generated: styleObject,
       })
+      console.log(styles)
```
philpl: console.log to be deleted

primitives.js
```diff
@@ -0,0 +1,2 @@
+/* eslint-disable flowtype/require-valid-file-annotation */
+module.exports = require('./lib/primitives')
```
philpl: We can probably delete this file and `/native.js`, since we have a `native/` and `primitives/` folder now

cc @patrick91 @mxstbr

philpl:
Awesome! So I think the last outstanding issue is this one here:

> Instead of doing this, the InlineStyle model should become a factory function that accepts an RN/RP StyleSheet.
> have look on how some other models do it :) This is the closest example obviously: https://github.com/mathieudutour/styled-components/blob/c65a1cb54d81eddbac48fa845627e48fb007d3ef/src/models/ComponentStyle.js#L11
> So instead of passing your StyleSheet in, we can pass the generated InlineStyle in

mathieudutour:
Hum I did change it. Isn't it you meant? https://github.com/mathieudutour/styled-components/blob/d21ef5ef28a07bf6700002f35e723524d62f8f0d/src/models/InlineStyle.js#L26

mxstbr:
I'm so excited, I've been wanting to make this happen for a while now! Can't wait to use this :blush:

philpl:
@mathieudutour It's not quite what I meant :smile: so basically in `ComponentStyle` we're doing this:

```js
export default (nameGenerator: NameGenerator, flatten: Flattener, stringifyRules: Stringifier) => {
```

So the signature of that export is `(...args) => ComponentStyle`.

For `InlineStyle` we can do the same:

```js
export default (StyleSheet: StyleSheet) => {
  class InlineStyle {
    // ...
  }

  return InlineStyle
}
```

It's the pattern that we've adopted throughout the repo :wink:

mathieudutour:
I think I get it. Is that right?

philpl:
@mathieudutour That's perfect! :heart:

@mxstbr Since you're around, do you have time to review this as well :smile:

jongold:
Great work @mathieudutour - excited to play with it.

@philpl echoing Mathieu's answer - react-primitives actually exposes react-native-web as the renderer for web — aliasing is the simple solution but I'm super super excited about the promise of react-primitives as a universal abstraction for every platform. Also I really want react-sketchapp & styled-components to be BFFs.

mxstbr:
> Also I really want react-sketchapp & styled-components to be BFFs.

:+1:

If @geelen veto's this to be in core for whatever reason I'll personally copy and paste this to make `styled-components-primitives` with the hackiest imports ever 😜

jongold:
This is so dope
image

geelen:
Sorry friends, was on the road all last week and basically not checking the issue tracker at all. I'm super keen for SC to support react sketchapp (that screenshot got me :shaq_wiggle_gif:) but I'm not clear on how this diff makes that possible, or what the downsides would be.

I just watched Leland's talk and I think I understand the concept, but it looks like you're deleting the `styled-components/native` endpoint in this PR? I certainly don't have a problem with introducing a `styled-components/primitives` endpoint (or even a new package down the line) to get this out there and see how it feels, but I wouldn't want it to be the only interface to Native yet. I'm also not sure about how the platform-specific extensions would work across a library boundary like this...

Personally, I'm big on the ideas of unifying UI development workflow but I'm also firmly in the camp that the web has very specific problems and I'm not willing to compromise on the API for more generality. It's why things like nesting are in SC, despite being specific to real web CSS.

That said, I'm extremely interested to see where this leads so assuming that there are no breaking changes I think we should push this out under a `styled-components/primitives` entry point and get some real-world feedback.

mathieudutour:
There is no breaking change, I removed the native.js file because there is already a folder called “native” with an index.js inside. I did it following the feedback of @philpl, shall I roll back?

mxstbr:
> I removed the native.js file because there is already a folder called “native” with an index.js inside

I'm unsure if that'll do what we want it to, let's revert that change for now to make 100% sure nothing breaks and ship this?

mathieudutour:
done

primitives.js
geelen: Ah yes sorry I hadn't noticed those being added. I vaguely prefer `/native.js` rather than `/native/index.js` since they're just aliases into `/lib`, but I don't know where the typescript definitions would live then (`/typings` perhaps?).

Anyway, we can figure that out later. For now, better to only have one place for everything, so you can delete these two files.

geelen:
Sorry this was my mistake. You can indeed remove those extra entry points.

mathieudutour:
No worries. Done

philpl:
:tada:

mxstbr:
Going to write a blogpost about this release, this is **a big deal**!

sebastienfi:
This is dope

PR #956, opened 2017-06-29
Add CodeSandbox to 'Built with' section

CompuIves:


philpl:
Omg how did we not add it sooner?!  🤣

CompuIves:
😅

I was thinking we should maybe rather link to https://codesandbox.io/s/new since the homepage is not built with styled-components, just plain css and html. Waddaya think?

mxstbr:
That makes sense to me!

PR #961, opened 2017-07-01
Remove dependencies from es bundle

### LINKED ISSUE #438: Module uses bundled code for building with webpack
## Version
2.0

## Steps to reproduce
* Install SC, use webpack to bundle

## Expected Behavior
* Webpack uses `dependencies` for bundling, including some deps that may occur elsewhere in the project

## Actual Behavior
* Webpack includes the rollup .es.js file which bundles everything

In the package.json, `main` is set to `lib/index.js` which is the ES5 modular transpilation. `jsnext:main` is `dist/styled-components.es.js` which is a pre-bundled version that only excludes React. There is also a `module` key set to the same bundle.

I think that `jsnext:main` and `module` should also point to `lib/index.js`, or else the es bundle should exclude regular `dependencies` as well as `peerDependencies`.

k15a:
Fixes #438

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "react-native": "^0.39.2",
     "react-primitives": "^0.4.2",
     "rimraf": "^2.6.1",
-    "rollup": "^0.37.0",
+    "rollup": "0.43.0",
     "rollup-plugin-babel": "^2.7.1",
     "rollup-plugin-commonjs": "^6.0.0",
     "rollup-plugin-flow": "^1.1.1",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -75,7 +75,7 @@ if (prod) plugins.push(uglify(), visualizer({ filename: './bundle-stats.html' })
 export default {
   entry: 'src/index.js',
   moduleName: 'styled',
-  external: esbundle ? Object.keys(pkg.dependencies) : ['react'],
+  external: ['react'].concat(esbundle ? Object.keys(pkg.dependencies) : []),
   exports: 'named',
   targets,
   plugins,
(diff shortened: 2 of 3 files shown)


mxstbr:
Hmm, CI already fails?

k15a:
Yes, I forgot that react is installed as a development dependency.

mxstbr:
Can you post a diff of the `.es.js` bundle?

k15a:
https://www.diffchecker.com/1H5rkQGt

mxstbr:
Okay, that looks pretty good! Does this work if `npm linked` into a project that uses the `.es.js` bundle?

k15a:
It's loading the modules from node_modules so it's working.

screen shot 2017-07-01 at 19 27 11

mxstbr:
Nice, this looks good to me then! Let's wait for @Rich-Harris to (hopefully) chime in, otherwise I say we try and ship a pre-release with this? /cc @geelen

Rich-Harris:
Yep, makes total sense if it's likely that those dependencies will be shared by other packages — bundling external deps makes the most sense when you're creating a UMD build for the browser.

FYI in more recent versions of Rollup you can import package.json:

```diff
-// We need to use require here because import creates a parse error
-const pkg = require('./package.json')
+import pkg from './package.json'
```

k15a:
Do we still need rollup-plugin-json if rollup can handle JSON files now? Or is this only for the config?

Rich-Harris:
> Or is this only for the config?

yep, only for the config

k15a:
Should we update Rollup and refactor the import statement before that?

philpl:
@k15a it'd be nice to upgrade rollup as well, if it's not too much trouble :) otherwise we can do that in another PR. Up to you 👍

PR #964, opened 2017-07-03
change prepublish script to prepare. fixes #963

### LINKED ISSUE #963: Not installable via GitHub with new NPM
Hello everyone!

The frequency of styled-components releases to npm is not excessively high. Every small bugfix might potentially be helpful for someone and warrant a patch version release.

Another alternative for those needing a recent commit to master in their working version is the installation via GitHub, e.g. `npm install styled-components/styled-components`. At the moment, this doesn't work because the `dist` and `lib` folders are not committed to the repository 👍 but also not created automatically on installation 👎!

The `prepublish` npm script that is used to build both `dist` and `lib` used to be executed on both publishing a package and installing it. Starting with npm@5 this is no longer the case and `prepublish` is run only when publishing the package.

See https://docs.npmjs.com/misc/scripts#deprecation-note for more information on that topic.

codepunkt:
This fixes #963 and runs the build process of `dist` and `lib` when either publishing or installing the package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,154 +1,152 @@
 {
-  "name": "styled-components",
-  "version": "2.1.0",
-  "description": "Visual primitives for the component age. Use the best bits of ES6 and CSS to style your apps without stress 💅",
-  "main": "lib/index.js",
-  "typings": "typings/styled-components.d.ts",
-  "jsnext:main": "dist/styled-components.es.js",
-  "module": "dist/styled-components.es.js",
-  "scripts": {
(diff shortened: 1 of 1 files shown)


mxstbr:
That means everybody who types `npm install styled-components` builds the code locally even though it's already built, right?

philpl:
I think we discussed this thoroughly and explained why it won't work for us :/ Unfortunately we forgot to close this PR

https://github.com/styled-components/styled-components/issues/963#issuecomment-312687522

PR #997, opened 2017-07-15
Add links to Rebass and styled-system in README.md

jxnblk:
Thought it might make sense to link to a couple more styled-components-based libraries. Happy to change the order and placement however you prefer. 😎 

Also note: I'm seeing failing tests locally, which I'm assuming are not related to the changes here – they appear to be related to vendor prefix differences

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,6 @@ We could use your help to get syntax highlighting support to other editors! If y
 ## Built with `styled-components`
 
 ### Libraries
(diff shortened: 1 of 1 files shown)


k15a:
Hey, would you mind putting the on the bottom of the section? We want to be fair to people who where earlier :)

jxnblk:
Of course, that's why I asked – should be updated now

PR #999, opened 2017-07-16
Added link to video on creating a style guide using NPM

notrab:
As suggested by [Max](https://twitter.com/mxstbr/status/886378768301924354) I thought I'd submit a PR. Hopefully this will help anyone else thinking of doing the same thing as I did.


k15a:
You mean dash-case or whatever this is called :) Good catch (:

PR #1004, opened 2017-07-18
Replace `$FlowInvalidInputTest` with `$FlowFixMe`.

wtgtybhertgeghgtwtg:
Just as the title says.  `$FlowInvalidInputTest` doesn't silence Flow errors in projects depending on `styled-components`, since it's in this package's `.flowconfig`.  And you already use `$FlowFixMe`.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -17,5 +17,4 @@
 
 [options]
 suppress_comment=.*\\$FlowFixMe
-suppress_comment=.*\\$FlowInvalidInputTest
 unsafe.enable_getters_and_setters=true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file. If a contri
 ## Unreleased
 
 - Upgrade stylis to 2.3 and use constructor to fix bugs with multiple libs using stylis simultaneously (see [#962](https://github.com/styled-components/styled-components/pull/962))
+- Replace `$FlowInvalidInputTest` with `$FlowFixMe`.  [#1004](https://github.com/styled-components/styled-components/pull/1004)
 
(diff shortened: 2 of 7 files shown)


.flowconfig
```diff
@@ -17,5 +17,4 @@
 
 [options]
 suppress_comment=.*\\$FlowFixMe
```
wtgtybhertgeghgtwtg: You could also get rid of this line, since `$FlowFixMe` is suppressed by default, but that's not really related to this PR.

philpl:
cc @styled-components/typers

relekang:
I think we should try to communicate that this is expected errors as `FlowFixMe` kind of communicates that it should be fixed. Maybe we can do `// $FlowFixMe expected error on test of invalid data`?

wtgtybhertgeghgtwtg:
Shouldn't that apply to existing instances of `$FlowFixMe` in tests?

philpl:
@relekang So this is not a desirable change, am I correct in understanding this?

@wtgtybhertgeghgtwtg I don't really see that it changes anything substantially, and is just swapping out one suppression comment for another that's more generic

relekang:
> So this is not a desirable change, am I correct in understanding this?

Not in it's current form as it removes the context. It would be fine tho if the context is added as a part of the message as mentioned above.

> Shouldn't that apply to existing instances of `$FlowFixMe` in tests?

Depends on the case. Some of them might not be related to testing unexpected data.

philpl:
Then I'd go for closing this tbh. Sorry, @wtgtybhertgeghgtwtg ! But I think a separate FlowFixMe name is better than writing:

```js
// $FlowFixMe: Invalid input for testing
```

Sounds about right, @relekang? :)

wtgtybhertgeghgtwtg:
As you wish.

PR #1007, opened 2017-07-18
Add styled-shortcut-components to readme

donavon:
>- [styled-shortcut-components](https://github.com/donavon/styled-shortcut-components): A convenience package that wraps `styled-components` with `styled-shortcuts`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,7 @@ We could use your help to get syntax highlighting support to other editors! If y
 - [styled-map](https://github.com/scf4/styled-map): Super simple lib to map props to styles with `styled-components`
 - [styled-system](https://github.com/jxnblk/styled-system): Design system utilities for styled-components – used in grid-styled and Rebass ([demo](http://jxnblk.com/styled-system/))
 - [styled-shortcuts](https://github.com/donavon/styled-shortcuts): Use simple string notation to access props. No functions necessary! Example: `font-size: ${'theme.fontSize:px'};` ([demo](https://codesandbox.io/s/jRE0XxR9v))
+- [styled-shortcut-components](https://github.com/donavon/styled-shortcut-components): A convenience package that wraps `styled-components` with `styled-shortcuts`.
 
 ### Boilerplates
 - [react-redux-styled-hot-universal](https://github.com/krasevych/react-redux-styled-hot-universal) (SSR, Universal Webpack, Redux, React-router, Webpack 2, Babel, Styled Components and more...)


philpl:
@donavon 

> This functionality should be in the base styled-components, IMO

We have a very narrow definition of what styled-components does: Embracing styling's domain specific language CSS. This obviously through tagged template literals and its interpolations. Now these interpolations are pretty simple and intuitive in what they do, and are flexible enough to allow something like `styled-shortcuts` to be built on top. But we're not fans of introducing extra bulk on top of the existing API for the sake of having it in the base library. :wink:

Btw explanations for why different feature requests weren't pursued can be found in various closed GH issues. It sounds harsh, but we'd rather keep extra-stuff out of the library and make the API lean, simple, and easy to learn, rather than to introduce one too many features that can also be external.

>  It would obviously be a breaking change

This expands on the stuff before; (Correct me if I'm wrong. I don't know styled-shortcuts in detail) Why introduce a breaking change that removes string-interpolation for their normal, obvious use, for a feature that only a fraction of the userbase might prefer?

> not until 2.0. But if/until then, I'm here

v2 is out :smile:

Hope that wasn't too harsh, and gives you an insight on our general position on feature requests. If you're still interested in discussing this further, you can open an issue, and expand on what this would mean for the library, i.e. new APIs, breaking changes, etc

donavon:
@philpl,

> a feature that only a fraction of the userbase might prefer?

I disagree. If the functionality of `styled-shortcuts` were baked into `styled-components`, I think you would find that it would be used a LOT.

Here's the "Theming" example from your own website (I added border-radius to further drive home my point):
```
border-radius: ${props => props.theme.borderRadius}px;
color: ${props => props.theme.main};
border: 2px solid ${props => props.theme.main};
```
That's a lot of crazy boilerplate code! And here it is using `styled-shortcuts`:
```
border-radius: ${'theme.borderRadius:px'};
color: ${'theme.main'};
border: 2px solid ${'theme.main'};
```
You don't think that shown this example, that more than "a fraction of the userbase" would use it? (We don't have to agree here.) 

Passing `props` to a styled component seems to be what you are showing people as  best practice (and one that I support). I'm making that process more less tedious. 

> features that can also be external

And while you bring it up. Did you see what hoops I had to jump through to make this work? It would be nice to have a "hook" into SC for add-ins. That would not be a breaking change, and a nice additions to the product. And without adding too much overhead.

> v2 is out 😄

Yeah... Obviously I meant 3.0. 😜

Oh, and thx for contributing such a wonderful addition to the React community!

PR #1008, opened 2017-07-19
Add warning when template literal is passed incorrectly to injectGlobal.

iamsolankiamit:
```
injectGlobal(`
  body {
    font-family: sans-serif;
  }
`);
```
:x: This raises a warning to not use parentheses.
```
injectGlobal`
  body {
    font-family: sans-serif;
  }
`;
```
 :+1:  Correct Way


philpl:
We have another PR that was adding some checks there, but as of now we're unsure whether to merge it. Sorry! https://github.com/styled-components/styled-components/pull/698

PR #1010, opened 2017-07-19
Add styled-media-query to libraries

morajabi:
I think it's useful for others using styled-components.


morajabi:
@k15a I thought it is ordered in stars count. 
I'm moving it to the last but I have a suggestion: Isn't it better to order them in an order for example alphabetic order or whatever like others do? I don't think the order of addition is a good thing for a plugin list. But it is just a suggestion.

philpl:
@morajabi eventually we should probably get rid of this list inside the readme and have something more organised on the website. But it's not our goal to promote all the best projects in order, but just to show some to express to the reader that we have a healthy community of third-party projects

morajabi:
If you wish I can make a separate page or something you say for them. @philpl

philpl:
@morajabi It's still unclear where several pages for the website will go. Maybe we should start a discussion on Spectrum and talk about it, as I haven't designed or thought about pages like the "Wall of Love", "Companies", "Third party projects", and so on yet.

morajabi:
@philpl OK whenever that happened, I'd like to help on that.

philpl:
@morajabi https://spectrum.chat/thread/25aad1ea-4c6d-4558-97a6-3214f47d3287

PR #1016, opened 2017-07-20
Add Griz to the list of grid systems

josephrexme:


josephrexme:
@k15a I thought it's alphabetical? I'd definitely do that

PR #1022, opened 2017-07-22
feat: __webpack_nonce__ support in SSR + Browser contexts

### LINKED ISSUE #887: Add __webpack__nonce to style tags
I already had a brief talk with @mxstbr about this a couple of weeks ago: https://twitter.com/mxstbr/status/824626370760364032

When trying to enforce strict rules about which content is allowed on a specific page, a Content Security Policy (CSP) is the way to go.

Styled-Components currently put all of their Styles as an inline style to the DOM. Therefore, a CSP would need to contain: style-src `unsafe-inline`. As the name already puts: This is unsafe and not the recommended way of securing your app.

Luckily, there is the concept of `nonce` attributes. Style/Script tags with a whitelisted `nonce` can be specified in the CSP. If Styled-Components would set this `nonce`, we could get rid of the `unsafe-inline` setting.

Luckily, WebPack is already supporting the concept of a global `__webpack_nonce__` variable. Once set, all dynamically injected code created by WebPack will have a nonce attribute with the correct value.

What needs Styled-Components to do?

If the __webpack_nonce__ variable is set, pass the value to every generated style tag inside the `nonce` attribute of the `style` tag.

Unluckily, the `__webpack_nonce__` is not well documented and kind of a `myth`. See Google https://www.google.at/search?q=webpack_nonce__&oq=webpack_nonce__&aqs=chrome..69i57.4663j0j7&sourceid=chrome&ie=UTF-8#q=__webpack_nonce__ and GitHub: https://github.com/webpack/webpack/search?utf8=%E2%9C%93&q=__webpack_nonce__&type=

Further readings:

- https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
- https://developers.google.com/web/fundamentals/security/csp/
- https://csp.withgoogle.com/docs/index.html

probablyup:
Closes #887

### DIFF
--- a/src/models/ServerStyleSheet.js
+++ b/src/models/ServerStyleSheet.js
@@ -1,4 +1,5 @@
 // @flow
+/* eslint-disable no-underscore-dangle */
 import React from 'react'
 import type { Tag } from './StyleSheet'
 import StyleSheet, { SC_ATTR, LOCAL_ATTR, clones } from './StyleSheet'
@@ -27,6 +28,10 @@ class ServerTag implements Tag {
     this.size += 1
   }
 
+  concatenateCSS() {
+    return Object.keys(this.components).reduce((styles, k) => (styles + this.components[k].css), '')
+  }
+
   inject(componentId: string, css: string, name: ?string) {
     const comp = this.components[componentId]
 
@@ -39,28 +44,33 @@ class ServerTag implements Tag {
   }
 
   toHTML() {
-    const namesAttr = `${SC_ATTR}="${this.names.join(' ')}"`
-    const localAttr = `${LOCAL_ATTR}="${this.isLocal ? 'true' : 'false'}"`
-    const css = Object.keys(this.components)
-      .map(key => this.components[key].css)
-      .join('')
+    const attrs = [
+      'type="text/css"',
+      `${SC_ATTR}="${this.names.join(' ')}"`,
+      `${LOCAL_ATTR}="${this.isLocal ? 'true' : 'false'}"`,
+    ]
 
-    return `<style type="text/css" ${namesAttr} ${localAttr}>\n${css}\n</style>`
+    if (typeof global !== 'undefined' && global.__webpack_nonce__) {
+      attrs.push(`nonce="${global.__webpack_nonce__}"`)
+    }
+
+    return `<style ${attrs.join(' ')}>${this.concatenateCSS()}</style>`
   }
 
   toReactElement(key: string) {
-    const attributes = {
+    const attrs: Object = {
       [SC_ATTR]: this.names.join(' '),
       [LOCAL_ATTR]: this.isLocal.toString(),
     }
-    const css = Object.keys(this.components)
-      .map(k => this.components[k].css)
-      .join('')
+
+    if (typeof global !== 'undefined' && global.__webpack_nonce__) {
+      attrs.nonce = global.__webpack_nonce__
+    }
 
     return (
       <style
-        key={key} type="text/css" {...attributes}
-        dangerouslySetInnerHTML={{ __html: css }}
+        key={key} type="text/css" {...attrs}
+        dangerouslySetInnerHTML={{ __html: this.concatenateCSS() }}
       />
     )
   }
--- a/src/test/ssr.test.js
+++ b/src/test/ssr.test.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import { renderToString } from  'react-dom/server'
 import ServerStyleSheet from '../models/ServerStyleSheet'
-import { resetStyled, stripWhitespace, nameGenerator } from './utils'
+import { resetStyled, stripWhitespace } from './utils'
 import _injectGlobal from '../constructors/injectGlobal'
 import _keyframes from '../constructors/keyframes'
 import stringifyRules from '../utils/stringifyRules'
@@ -21,6 +21,9 @@ const format = css => stripWhitespace(css)
 describe('ssr', () => {
   beforeEach(() => {
     styled = resetStyled(true)
+
+    // eslint-disable-next-line no-underscore-dangle
+    global.__webpack_nonce__ = undefined
   })
 
   it('should extract the CSS in a simple case', () => {
@@ -63,6 +66,37 @@ describe('ssr', () => {
 
       </style>
       <style type="text/css" data-styled-components="b" data-styled-components-is-local="true">
+      /* sc-component-id: sc-a */
+      .sc-a {}
+      .b { color: red; }
+
+      </style>
+    `))
+  })
+
+  it('should add a nonce to the stylesheet if webpack nonce is detected in the global scope', () => {
+    injectGlobal`
+      body { background: papayawhip; }
+    `
+    const Heading = styled.h1`
+      color: red;
+    `
+
+    // eslint-disable-next-line no-underscore-dangle
+    global.__webpack_nonce__ = 'foo'
+
+    const sheet = new ServerStyleSheet()
+    const html = renderToString(sheet.collectStyles(<Heading>Hello SSR!</Heading>))
+    const css = format(sheet.getStyleTags())
+
+    expect(html).toEqual('<h1 class="sc-a b" data-reactroot="" data-reactid="1" data-react-checksum="197727696">Hello SSR!</h1>')
+    expect(css).toEqual(format(`
+      <style type="text/css" data-styled-components="" data-styled-components-is-local="false" nonce="foo">
+      /* sc-component-id: sc-global-2303210225 */
+      body { background: papayawhip; }
+
+      </style>
+      <style type="text/css" data-styled-components="b" data-styled-components-is-local="true" nonce="foo">
       /* sc-component-id: sc-a */
       .sc-a {}
       .b { color: red; }
--- a/src/test/ssr.test.js
+++ b/src/test/ssr.test.js
@@ -273,4 +273,43 @@ describe('ssr', () => {
       </style>
     `))
   })
+
+  it('should return a generated React style element', () => {
+    injectGlobal`
+      body { background: papayawhip; }
+    `
+    const Heading = styled.h1`
+      color: red;
+    `
+
+    const sheet = new ServerStyleSheet()
+    const html = renderToString(sheet.collectStyles(<Heading>Hello SSR!</Heading>))
+    const elements = sheet.getStyleElement()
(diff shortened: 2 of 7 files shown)


src/models/ServerStyleSheet.js
```diff
@@ -39,13 +39,30 @@ class ServerTag implements Tag {
...
+      html += ` nonce="${global.__webpack_nonce__}"`
+    }
+    /* eslint-enable */
+
+    html += ` ${SC_ATTR}="${this.names.join(' ')}"`
+    html += ` ${LOCAL_ATTR}="${this.isLocal ? 'true' : 'false'}"`
+    html += '>'
+
+    /* eslint-disable */
+    const css = Object.keys(this.components).reduce((styles, key) => {
```
probablyup: was able to remove an array iteration here

src/test/ssr.test.js
```diff
@@ -71,6 +74,37 @@ describe('ssr', () => {
...
+
+    // eslint-disable-next-line no-underscore-dangle
+    global.__webpack_nonce__ = 'foo'
+
+    const sheet = new ServerStyleSheet()
+    const html = renderToString(sheet.collectStyles(<Heading>Hello SSR!</Heading>))
+    const css = format(sheet.getStyleTags())
+
+    expect(html).toEqual('<h1 class="sc-a b" data-reactroot="" data-reactid="1" data-react-checksum="197727696">Hello SSR!</h1>')
+    expect(css).toEqual(format(`
```
probablyup: @mxstbr would you guys be open to switching these to snapshot tests? I think it might simplify these a bit.

probablyup:
Not really sure why rollup is failing. The error seems to indicate the JSX isn't being transformed?

src/models/ServerStyleSheet.js
```diff
@@ -39,13 +39,25 @@ class ServerTag implements Tag {
   }
 
   toHTML() {
-    const namesAttr = `${SC_ATTR}="${this.names.join(' ')}"`
-    const localAttr = `${LOCAL_ATTR}="${this.isLocal ? 'true' : 'false'}"`
-    const css = Object.keys(this.components)
-      .map(key => this.components[key].css)
-      .join('')
+    const attrs = [
+      'type="text/css"',
```
probablyup: I think this can actually be removed.

src/test/ssr.test.js
mxstbr: Let's open an issue/another PR for that?

src/models/ServerStyleSheet.js
```diff
@@ -39,13 +39,25 @@ class ServerTag implements Tag {
...
+    const css = Object.keys(this.components).reduce((styles, key) => {
+      return (styles += this.components[key].css), styles
+    }, '')
+    /* eslint-enable */
 
-    return `<style type="text/css" ${namesAttr} ${localAttr}>\n${css}\n</style>`
+    return `<style ${attrs.join(' ')}>${css}</style>`
   }
 
   toReactElement(key: string) {
```
philpl: The same props will need to be added to this method down here

src/models/ServerStyleSheet.js
```diff
@@ -39,13 +39,25 @@ class ServerTag implements Tag {
...
-    const css = Object.keys(this.components)
-      .map(key => this.components[key].css)
-      .join('')
+    const attrs = [
+      'type="text/css"',
+      `${SC_ATTR}="${this.names.join(' ')}"`,
+      `${LOCAL_ATTR}="${this.isLocal ? 'true' : 'false'}"`,
+    ]
+
+    /* eslint-disable no-underscore-dangle */
```
philpl: Can you instead just disable this globally for this file. Shouldn't be an issue.

Also what does the disabled eslint down below prevent from throwing an error? Seems like a bad idea to just disable everything without a comment.

src/models/ServerStyleSheet.js
```diff
@@ -39,13 +39,25 @@ class ServerTag implements Tag {
...
+    if (global.__webpack_nonce__) {
+      attrs.push(`nonce="${global.__webpack_nonce__}"`)
+    }
+    /* eslint-enable */
+
+    /* eslint-disable */
+    const css = Object.keys(this.components).reduce((styles, key) => {
+      return (styles += this.components[key].css), styles
```
philpl: you can just return `styles + this.components[key].css` and even turn it into an arrow functions. Gets rid of the mutation and is easier to read.

src/models/ServerStyleSheet.js
philpl: Let's still keep it for completeness sake. We don't control what kind of HTML type this renders in. A bit redundant in 2017, but better than having people nitpick it.

src/models/ServerStyleSheet.js
probablyup: It was complaining about using a comma and I guess there's an eslint rule that all arrow functions have to be one-liners without curlies?

philpl: Ok, in that case, just apply the suggestion below and it should become unnecessary ;)

src/models/ServerStyleSheet.js
probablyup: I actually ended up making this into a reusable function since the same thing happens in `toReactElement`

probablyup:
Looking into the rollup thing - might need to add this plugin: https://www.npmjs.com/package/rollup-plugin-node-globals

probablyup:
Okay I found a way around it, using this seems to work without modifying rollup:

```js
if (typeof __webpack_nonce__ !== 'undefined') {
      attrs.push(`nonce="${__webpack_nonce__}"`)
    }
```

Basically just letting it implicitly refer to global scope

src/test/ssr.test.js
probablyup: Will do. I added some new tests for previously uncovered code which use snaps but I'll do a separate PR to convert existing tests where it makes sense.

probablyup:
@philpl addressed your comments

probablyup:
Nvm the lack of global keyword seems to only work in node 8 - looking for other solutions...

src/test/ssr.test.js
```diff
@@ -239,4 +273,43 @@ describe('ssr', () => {
...
+    // eslint-disable-next-line no-underscore-dangle
+    global.__webpack_nonce__ = 'foo'
+
+    const sheet = new ServerStyleSheet()
+    const html = renderToString(sheet.collectStyles(<Heading>Hello SSR!</Heading>))
+    const elements = sheet.getStyleElement()
+
+    expect(elements).toHaveLength(2);
+
+    expect(elements[0].props).toMatchSnapshot()
```
philpl: I have yet another small nit here, sorry :wink:

This test is basically the same as the above one, but we care about the nonce being present and set to
the global above. So we should probably just check for this directly instead of duplicating the above snapshots and having to check manually :smile:

philpl:
@probablyup Awesome work! I've only got one small nit left (see above comment)

probablyup:
This might just be me, but I always test the base state and modifications
separately. Easier to track individual regressions.

philpl:
@probablyup that's a good default testing strategy, but in this case we don't care about anything but the nonce there. We're only testing:

- Premise: A nonce is set as a global
- Assumption: A nonce attr/prop is present

Nothing about the rest of the output. The two concerns don't overlap :)

probablyup:
@philpl done

johannesnagl:
@mxstbr sorry for Not giving any feedback. But my github handle is unfortunately Not @jollife.

Will investigate this Next week!

mxstbr:
Oh no worries, please try it out and let us know if it works!

johannesnagl:
Hi guys,

I tested this together with @stnwk and we found out the following:

1. Implementation is working
2. Nevertheless, the implementation is not using the correct `__webpack_nonce__`. 

screen shot 2017-07-24 at 13 29 42


Clarification:

You're looking for `window.__webpack_nonce__` or `global.__webpack_nonce__`. In our tests, when setting these values, the code is working fine. Although, WebPack itself is using the magic variable `__webpack_nonce__`, which is not a `window` or `global` variable. 

See here: 
https://github.com/webpack/webpack/blob/ab2270263e9af5b0dd83e454c20ab3aa1797424a/lib/APIPlugin.js#L26

It's getting replaced as a simple string.

So, the correct implementation would be the following:

```
    if (typeof __webpack_nonce__ !== 'undefined' && __webpack_nonce__) {
      this.el.setAttribute('nonce', __webpack_nonce__)
    }
```

We still have a problem with `injectGlobal`, where it's not being correctly set, but we haven't yet fully understand if it's a problem on our side or within styled-components.

stnwk:
For the `injectGlobal` issue I think the following might be the reason:

In [injectGlobal.js#L14](https://github.com/styled-components/styled-components/blob/master/src/constructors/injectGlobal.js#L14) we are calling the function 
`inject(componentId, false, stringifyRules(rules))` on [StyleSheet.js#L84](https://github.com/styled-components/styled-components/blob/master/src/models/StyleSheet.js#L84) - which actually takes a forth and a fifth parameter (`hash` and `name`). But we are not passing either one of them.

In the case of being on the client-side, e.g. using `BrowserStyleSheet` this means the following function is called: [BrowserStyleSheet.js#L61](https://github.com/styled-components/styled-components/blob/master/src/models/BrowserStyleSheet.js#L61).

In this function [we check](https://github.com/styled-components/styled-components/blob/master/src/models/BrowserStyleSheet.js#L69) wether the `name` parameter was passed before setting the nonce attribute, which isn't set when calling `injectGlobal`, because we are not passing it.

I'm not sure wether the check for the `name` parameter is on purpose there, 
but that's the culprit I think.

Would also suggest to add a test for this specific case using `injectGlobal`.

Cheers :)

philpl:
@probablyup are you open to creating a new PR with the above changes? :) 

Fixing injectGlobal should be corrected by moving the nonce setAttribute logic out of the surrounding if-name clause, right?

probablyup:
Ironically I had it that way for a while haha. I'm not sure how we are
going to unit test a magic variable though since it's not actually in any
scope.

stnwk:
Maybe taking a look at [how webpack tests magic variables ](https://github.com/webpack/webpack/blob/master/test/cases/nonce/set-nonce/index.js)helps? @probablyup

probablyup:
Yeah or making a util that returns it and mocking the util. That's probably
the most straightforward.

PR #1023, opened 2017-07-22
test: jest 20, switch ssr tests to snapshots and use node environment

probablyup:
See commits


philpl:
Nice! @geelen should probably take a look, but he's been kind of busy. But since this is not actually changing any logic or test semantics it should be fine.

cc @mxstbr any additional thoughts?

philpl:
@mxstbr I don't see a reason why we shouldn't switch to this :smile:

@probablyup Can you add a changelog entry so we can merge this?

probablyup:
@philpl yup, probably tonight or tomorrow

probablyup:
@philpl done

philpl:
cheers! :)

PR #1040, opened 2017-07-27
refactor use charsLength instead of chars.length

wongterrencew:
Trivial change, but figured maybe this was accidentally let through :p

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ All notable changes to this project will be documented in this file. If a contri
 
 - Add react-native `ImageBackground` alias (see [#1028](https://github.com/styled-components/styled-components/pull/1028))
 
+- Refactor variable in generateAlphabeticName.js (see [#1040](https://github.com/styled-components/styled-components/pull/1040))
+
 ## [v2.1.0] - 2017-06-15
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ All notable changes to this project will be documented in this file. If a contri
 
 - Add react-native `ImageBackground` alias (see [#1028](https://github.com/styled-components/styled-components/pull/1028))
 
+- Refactor variable in generateAlphabeticName.js
```
philpl: Can you add the `(...)` note with a link to this PR please? :)

k15a:
I think we did this to cache the property access because it's done pretty often. Don't know if the engines do this optimization from default :)

philpl:
@k15a Yes, I've optimised it and overlooked the residual call there :wink: The engines don't do this optimisation automatically, since the `.length` i.e. the array might be changed by side effects at any time.

k15a:
My bad, I thought this PR changes this back to chars.length but that's not the case.

CHANGELOG.md
wongterrencew: 👌  done

PR #1041, opened 2017-07-27
Ignore comments in InlineStyle

ckknight:


philpl:
Hiya, sorry for not following up on this earlier.

Can you please add an entry to the Changelog? This way I also won't have to pull and test this manually 😅 

Can you also please add a test for comments to the native tests so that we won't have another regression? https://github.com/styled-components/styled-components/blob/master/src/native/test/native.test.js

ckknight:
@philpl Ball's back in your court. 🏀

philpl:
Seems to be working as intended! 👍

philpl:
Cheers, sorry that it took so long :)

PR #1044, opened 2017-07-28
Add .extend and .withComponent deterministic ID generation

### LINKED ISSUE #1031: Add .extend and .withComponent deterministic ID generation
For server-side rendering it is often crucial to generate deterministic and stable IDs during transpilation time, so that the import-order of modules—and thus the order in which StyledComponents are instantiated—doesn't influence the `componentId`. A failure to do so leads to a checksum mismatch between the SSR result and React on the client side.

To generate these IDs we have the `ssr` option in our babel-plugin: https://github.com/styled-components/babel-plugin-styled-components

> By adding a unique identifier to every styled component this plugin avoids checksum mismatches due to different class generation on the client and on the server. [...]

But due to the way it works, it adds `withConfig` to any call that creates a new Styled Component. This is tough to implement correctly as of now, and the way we communicate the component ids in the transpiled code will change completely in styled-components v3, when we introduce stable preprocessing.

So what we can do now is add some dynamic `componentId` generation to the runtime instead.

https://github.com/styled-components/styled-components/blob/master/src/models/StyledComponent.js#L219

In `extend` on this line, instead of filtering out `componentId`, we'll want to append a hash of the new CSS to it. This can probably just use `ComponentStyle.generateName` together with all the concatenated string templates, which is a "unique-enough" source (as a temporary heuristic here).

https://github.com/styled-components/styled-components/blob/master/src/models/StyledComponent.js#L211

In `withComponent` we can just concatenate the tag string ([or the `displayName` if it's a component](https://github.com/styled-components/styled-components/blob/master/src/models/StyledComponent.js#L177)).

Apart from the sources of the suffix, the implementation will mostly look like this:

```js
const newComponentId = `${componentId}-${extraId}`
```

neoziro:
Will fix #1031

I added a test with the new behaviour. This should fix server-side rendering issues with `.extend` and `.withComponent`.

### DIFF
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -225,7 +225,6 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
 
       static get extend() {
         const {
-          componentId: previousComponentId,
           rules: rulesFromOptions,
           ...optionsToCopy
         } = options
@@ -235,20 +234,22 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
             ? rules
             : rulesFromOptions.concat(rules)
 
-        const newComponentId =
-          previousComponentId &&
-          `${previousComponentId}-${ComponentStyle.generateName(
-            newRules.join(''),
-          )}`
-
         const newOptions = {
           ...optionsToCopy,
-          componentId: newComponentId,
           rules: newRules,
           ParentComponent: StyledComponent,
         }
 
-        return constructWithOptions(createStyledComponent, target, newOptions)
+        const createWithHash = (tag, opts, css) => {
+          const newOpts = opts.componentId ? {
+            ...opts,
+            componentId: `${opts.componentId}-${ComponentStyle.generateName(css.join(''))}`,
+          } : opts
+
+          return createStyledComponent(tag, newOpts, css)
+        }
+
+        return constructWithOptions(createWithHash, target, newOptions)
       }
     }
 
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -22,7 +22,7 @@ const multiDashRegex = /--+/g
 export default (ComponentStyle: Function, constructWithOptions: Function) => {
   /* We depend on components having unique IDs */
   const identifiers = {}
-  const generateId = (_displayName: string) => {
+  const generateId = (_displayName: string, parentComponentId: string) => {
     const displayName = typeof _displayName !== 'string' ?
       'sc' : _displayName
         .replace(escapeRegex, '-') // Replace all possible CSS selectors
@@ -32,7 +32,8 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
     identifiers[displayName] = nr
 
     const hash = ComponentStyle.generateName(displayName + nr)
-    return `${displayName}-${hash}`
+    const componentId = `${displayName}-${hash}`
+    return parentComponentId ? `${parentComponentId}-${componentId}` : componentId
   }
 
   class BaseStyledComponent extends AbstractStyledComponent {
@@ -175,7 +176,7 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
   ) => {
     const {
       displayName = isTag(target) ? `styled.${target}` : `Styled(${getComponentName(target)})`,
-      componentId = generateId(options.displayName),
+      componentId = generateId(options.displayName, options.parentComponentId),
       ParentComponent = BaseStyledComponent,
       rules: extendingRules,
       attrs,
@@ -226,6 +227,7 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
       static get extend() {
         const {
           rules: rulesFromOptions,
+          componentId: parentComponentId,
           ...optionsToCopy
         } = options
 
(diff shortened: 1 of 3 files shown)


src/test/expanded-api.test.js
```diff
@@ -62,6 +62,19 @@ describe('expanded api', () => {
       expect(Comp2.styledComponentId).toBe('Comp2-OMGLOL')
       expect(shallow(<Comp2 />).prop('className')).toMatch(/Comp2-OMGLOL/)
     })
+
+    it('should work with `.extend` and `.withComponent`', () => {
```
philpl: Can we separate the `.extend` from the `.withComponent` tests? That should make these tests a bit shorter :)

src/models/StyledComponent.js
```diff
@@ -208,25 +208,42 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
...
-          : rulesFromOptions.concat(rules)
+        const newRules =
+          rulesFromOptions === undefined
+            ? rules
+            : rulesFromOptions.concat(rules)
+
+        const newComponentId =
+          previousComponentId &&
+          `${previousComponentId}-${ComponentStyle.generateName(
+            newRules.join(''),
```
philpl: These rules might contain some interpolations. I'm not sure whether it's a good idea to blindly join them 🤔  cc @geelen

neoziro: @philpl what do you suggest?

neoziro:
@philpl I modified the `.extend`, I wasn't taking extended rules. Making a `.join('')` is not a problem, you can see that it work in test. But I can also skip functions or replace it by a string "FUNC" if you prefer.

philpl:
@neoziro That is looking fine. The only thing I'm concerned about is performance in this case :wink: I'll take a look later and see whether it's sufficient

src/models/StyledComponent.js
```diff
@@ -235,20 +234,22 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
...
-
         const newOptions = {
           ...optionsToCopy,
-          componentId: newComponentId,
           rules: newRules,
           ParentComponent: StyledComponent,
         }
 
-        return constructWithOptions(createStyledComponent, target, newOptions)
+        const createWithHash = (tag, opts, css) => {
```
philpl: This is not quite looking optimal, in terms of V8 preoptimisation. It will probably deopt this function, since it's created on every call. Any reason why you were unhappy with your previous implementation?

neoziro: @philpl previously the hash was not built with the rules of `.extend`. It was built with the rule of the original component. That is not what we want. This is the only way to do it. Also about performances, I think we can do this only if we have a component id.

philpl: @neoziro In this case we might want to think about a new option instead. Something like `parentId` which just passes down the old `componentId`. Then we let `componentId` be `undefined` and inside the StyledComponent factory we pass the `parentId` into `generateId` and prepend it to the new id.

cc @geelen

neoziro: @philpl I don't know the entire codebase in mind, I let you decide what is the best! Should I introduce a `parentId` concept or I let you do it?

philpl: @neoziro Since you're already fully at it and getting your hands dirty, I can guide you to what I mean :smile:

Here we could just pass in the `componentId` as a `parentComponentId`: https://github.com/styled-components/styled-components/blob/master/src/models/StyledComponent.js#L229

Then it'd eventually arrive back in the factory here: https://github.com/styled-components/styled-components/blob/master/src/models/StyledComponent.js#L178

There we can add a new argument to `generateId`:

```js
    const {
      componentId = generateId(options.displayName, options.parentComponentId)
    } = options
```

Then inside `generateId` you can concatenate this with `displayName` to make the new component id "more unique" i.e. different from the parent's one: https://github.com/styled-components/styled-components/blob/master/src/models/StyledComponent.js#L26

One thing we also need to do (**probably not in this PR**) is passing in the runtime inferred (default) `displayName` into `generateId` as well instead of just `options.displayName`: https://github.com/styled-components/styled-components/blob/master/src/models/StyledComponent.js#L177

neoziro: @philpl I understood everything, will do it soon

neoziro:
@philpl done

src/models/StyledComponent.js
```diff
@@ -32,7 +32,8 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
     identifiers[displayName] = nr
 
     const hash = ComponentStyle.generateName(displayName + nr)
-    return `${displayName}-${hash}`
+    const componentId = `${displayName}-${hash}`
+    return parentComponentId ? `${parentComponentId}-${componentId}` : componentId
```
philpl: Maybe `parentComponentId !== undefined`? Sorry for the nit, but we have an ongoing perf pr and it'd be good to keep things optimised 😅

src/models/StyledComponent.js
```diff
@@ -208,26 +209,37 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
       static target = target
 
       static withComponent(tag) {
-        const { displayName: _, componentId: __, ...optionsToCopy } = options
-        const newOptions = { ...optionsToCopy, ParentComponent: StyledComponent }
+        const { componentId: previousComponentId, ...optionsToCopy } = options
+
+        const newComponentId =
+          previousComponentId &&
+          `${previousComponentId}-${isTag(tag) ? tag : getComponentName(tag)}`
```
philpl: Maybe we should extract this logic into a small function at the top of this file, now that it's used twice?

philpl:
@neoziro This is looking perfect! :tada: I added two small nits. I'll test this and see if it fixes the SSR issues, but I don't see why it wouldn't :smile:

src/models/StyledComponent.js
neoziro: This is not exactly the same, so I prefer to keep it separated.

src/models/StyledComponent.js
neoziro: Done

neoziro:
@philpl ready to be merged

philpl:
Awesome! :tada:

mxstbr:
Yay! :tada:

PR #1046, opened 2017-07-28
Add missing instance to ServerStyleSheet in TypeScript definition

mohsen1:


philpl:
So first of all, the instance is a regular StyleSheet class, to get that out of the way. And there's no definition for it since it isn't publicly accessible / supposed to be used directly.

While we do write about a workaround in the docs, it's going to be absolute, and is just that: just a workaround. So it'll be removed and not recommended in the future.

It's good to know about this implementation detail right now, however we shouldn't expose this, since it limits what we can safely deprecate and change in a minor release. So if we wanted to change the SSR implementation, we couldn't if the instance property is documented and guaranteed to work.

cc @geelen @styled-components/typers

Issue #931, opened 2017-06-19
createBroadcast implementation causes performance issues


IIIristraM:
this lines
```
return () => {
    listeners = listeners.filter(item => item !== listener)
 }
```
produce a lot of new objects which leads to a frequent garbage collections, which take sometimes up to 250 ms.

this can be probably avoided by changing `listeners` from array to a simple hash object like {  [id]: listner }, and the lines above to

```
return () => {
    delete listeners[id]
 }
```

mxstbr:
Great catch, PR welcome!

philpl:
We should probably just switch to iamstarkov/theming soon.

But it depends on `brcast` which uses the same filtering logic unfortunately: https://github.com/vesparny/brcast/blob/master/index.js#L16

vesparny:
I suspect what a bout using splice?

```js
listeners.splice(listeners.indexOf(listener), 1)
```

vesparny:
I'll be happy to apply the patch to `brcast`.
Just wondering what's the best way to gain performance.

mxstbr:
Wrote a very simple microbench (likely doing things wrong, please double check my code): https://esbench.com/bench/5950cbe599634800a034871c

On my machine the `delete listeners[key]` version is ~3x faster that `.splice` and ~20x faster than `.filter`!

vesparny:
I have published a new version of `brcast` that addresses this issue.
Feel free to steal the code from there while you wait to switch to `theming`

xcoderzach:
@schwers Can you make a pull request from that fork?  It would be very useful for that to be in the official release 😄

vesparny:
Note I've released a new version of brcast fixing another issue.
It should also be a little bit more perdormant.

Do you want me to submit a PR that replaces SC implementation with brcast?

xcoderzach:
@vesparny that would be awesome!  I'd use it immediately (I have a page that takes 2 seconds to unmount in production 😬 )

vesparny:
Didn't you have plans for switching to theming?

xcoderzach:
Looks like it fell by the wayside.  A pull request for that would be welcome too, but may be more difficult.

mxstbr:
Yeah let's switch to theming, assuming all our existing tests pass, PR welcome!

mxstbr:
This was fixed in #1048!

PR #1051, opened 2017-07-31
Use theming library in place of custom theming components.

xcoderzach:
Stuff to note:

- I had to bump the bundle size to 14.2Kb. :-(
- I had to wrap `theming`'s `withTheme` implementation in another component to handle `innerRef` translation.  :-(
- I just ignored `theming`'s lack of flow annotations with a `$FlowFixMe` :-(
- I left the existing theme tests mostly as is, even though we're now just running our own tests against `theming`.  IMO This is probably a good idea, since we're publicly exporting parts of the library.


xcoderzach:
This is related to discussions in #1048, #931.

philpl:
Thanks a lot! 🙏 This comes right when we needed it

@schwers do you mind benchmarking this against your patch so we can weigh the pros and cons?

schwers:
@philpl sure! taking a look now

schwers:
I benchmarked this by locally running my app's virtualized-scroller back and forth for about 12 seconds, and looked at the first 11s. This view in our app is very information dense, with a high number of rows on screen.

image


With the low-allocation strategy in https://github.com/styled-components/styled-components/pull/1048  I'm seeing:

image

|Patch|JSHeap Max|Typical Animation Frame Sizes **|Time % spent in GC|
|------|------------|---------------------------------|--------------------|
| `theming`| 186MB  (note: after the 11s sample, memory ends up getting released)| 25-42ms| 4.7%|
| https://github.com/styled-components/styled-components/pull/1048 | 90MB | 22-39 | 3.9% |

** this is very anecdotal. We have a optimizations to work out in our scrolling. so this is ballparking 'typical' animation frames were only a few rows are added and there's no forced layouts, etc

I'd say they're both an improvement over the current implementation of `createBroadcast`. However the lower allocation version does _feel_ smoother with less visual stutter

probablyup:
Hmm the sawtoothing isn't great. Trading major GC pauses for minor ones (still an improvement over the current code though.) I'd keep looking for opportunities to reduce allocations where possibl.

schwers:
Yup! totally agree. fwiw I think part of this is the application I'm working in, itself.

On the subject of GC opportunities, one thing I've been thinking of is [`buildExecutionContext`](https://github.com/styled-components/styled-components/blob/master/src/models/StyledComponent.js#L54). 

e.g. If the api for dynamic styles changes from `(props) => ...` to `(theme, props) => ...` then we wouldn't have to build execution contexts. (though, this would be a breaking change with `attrs`, not sure how much that api is used)

probablyup:
Could it be (props, theme) instead? Then it wouldn't be breaking I think.

schwers:
For what I had in mind, I think it would be a breaking change regardless of ordering. I'm pretty open to having `props` come first, but I find myself mostly using `theme`, so having shorter signatures could be nice. Plus it reminds me a little of the `reselect` api.

To elaborate more, I was thinking that `props` could be all props of the styled-component, `theme`. Adding `theme` to the `props` argument would necessitate making a new `props` object that's only used for styles (vs passing props along directly).

e.g. your styles would go from

```js
const before = styled.div`
   color: ${ props => props.isPrimary ? props.theme.primary : props.theme.secondary };
};

const after = styled.div`
   color :${ (theme, props) => props.isPrimary ? theme.primary : theme.secondary };
`
```

(aside -- maybe this is feasible to add code-mod for transitioning?)

probablyup:
Oh I see! I actually like this API a lot more heh but you're right, it'd be
a hard breaking change that could be difficult to migrate from.

mxstbr:
@geelen and I discussed about this specific API _a lot_ when we first created styled-components. I think this was the most contentious point of the whole API, we talked about it over the span of a couple of days to figure out what we like best.

The reason we didn't go with two separate arguments (in either direction, `(theme, props) =>` or `(props, theme) =>`) is that it's annoying to have to assign an unused variable if you don't need the first argument and there is no clear "This argument is used 99% more often than the other without the other" thing here where that'd only happen one out of a million times. (quite on the contrary)

```JS
// Ugh this sucks
const After = styled.div`
   color :${ (_, theme) => theme.primary};
`

// This sucks too
const After = styled.div`
   color :${ (_, props) => props.color};
`
```

The option we explored was to leverage destructuring to get a "named arguments"-kinda thing going in JavaScript to avoid the unused variable problem. That would look something like this:

```JS
// This looks great from a consumers perspective 😍😍😍
const After = styled.div`
   color :${ ({ theme, props }) => props.isPrimary ? theme.primary : theme.secondary };
`
```

From an API perspective that would allow you to pick just what you want, while also looking really nice and clean. (though it probably wouldn't help with perf in this case)

The reason we didn't go with that is it's a) confusing for newcomers who might not understand destructuring and it's one more thing to learn and b) it's unclear what the argument to the function is when it's not destructured. (`input`? ugh)

Merging `theme` into `props` also allows for some cool interplay with standard React APIs like e.g. `defaultProps` to set the default theme:

```JS
const After = styled.div`
   color :${ ({ theme, props }) => props.isPrimary ? theme.primary : theme.secondary };
`

// If this component is not wrapped in a `ThemeProvider` set the default theme
After.defaultProps = {
  theme: {
    primary: '#000',
    secondary: '#999',
  }
}
```

All of this together should explain why we went with `props.theme` over multiple arguments or destructuring. I hope that makes sense, I really don't want to change that API as either options from a consumer perspective are worse.

That being sad, if passing two arguments means styled-components will be 100x less memory intensive/more performant we might talk about it, but let's open an issue instead of doing such a massive change in this PR.

schwers:
@mxstbr thanks for all the context! Really interesting to know the thinking behind the current api. 

> "This argument is used 99% more often than the other without the other" thing here where that'd only happen one out of a million times. (quite on the contrary)

very true ;)

> Merging theme into props also allows for some cool interplay with standard React APIs like e.g. defaultProps to set the default theme:

Yeah this is a nice feature :)

> but let's open an issue instead of doing such a massive change in this PR.

Absolutely! I'd like to explore this a bit more in dev before going further. I think it just came up in reference to looking for further GC wins. My apologies for any confusion here.

mxstbr:
No worries, I'm glad I got to document the thinking behind the decision since I don't think it's been written down anywhere yet. 😊 Happy to discuss in a separate issue if you think this'll help a lot!

philpl:
It's also important to note that Glamorous recently switched to having theme being merged into the props object as well. While they had a separate argument for it for quite a while, the usage / DX benefits of it outweighed the performance impact 😬

src/hoc/withTheme.js
```diff
@@ -1,66 +1,18 @@
...
-          ref={isStyledComponent ? undefined : innerRef}
-        />
-      )
-    }
-  }
-
+  const WithTheme = withTheme(({ innerRef, ...props }) => (
+    <Component
+      {...props}
+      innerRef={isStyledComponent ? innerRef : undefined}
```
philpl: Instead of adding another pure functional component here, as you mention in the description, can we try to use the theme listener, and build our own `withTheme` helper? This way we avoid adding two levels of components here, which would help perf a lot

xcoderzach: Done, it added even more to the bundle size tho :-(

xcoderzach:
The current CI failures are the bundlesize, because now we're packaging in two `withTheme` implementations.  :-S

src/models/test/ThemeProvider.test.js
```diff
@@ -89,14 +87,13 @@ describe('ThemeProvider', () => {
...
       componentWillMount() {
-        this.context[CHANNEL](theme => {
-          // eslint-disable-next-line react/prop-types
-          expect(theme).toEqual(themes[this.props.shouldHaveTheme])
-          childRendered++ // eslint-disable-line no-plusplus
-          if (childRendered === Object.keys(themes).length) {
-            done()
-          }
-        })
+        const theme = this.context[CHANNEL].getState()
```
schwers: I'm not familiar with `theming`, does this imply there would be a similar context type change as discussed in https://github.com/styled-components/styled-components/pull/1048#discussion_r130274738?

xcoderzach: yep, it does

xcoderzach:
I wrote a benchmark.js script for brcast unsubscribe vs unsubscribing based on id:

id unsubscribe 1,184 ops/sec ±0.96% (94 runs sampled)
fn unsubscribe 728 ops/sec ±0.81% (91 runs sampled)

https://github.com/xcoderzach/broadcast-benchmark

philpl:
@xcoderzach: So @vesparny is currently doing these optimisations as well. https://github.com/vesparny/brcast/pull/6

I'm more concerned about the increased bundle size. @iamstarkov Have you looked into how to decrease the size of theming already, by any chance? :)

> The current CI failures are the bundlesize, because now we're packaging in two withTheme implementations. :-S

Any idea why the `withTheme` export is not tree shaken out of our bundle?

xcoderzach:
It's not getting tree shook because `withTheme` comes from `createWithTheme` which is a factory that gets called with the custom channel when you call `createTheming`.

xcoderzach:
basically calling `createTheming` calls `createWithTheme`, which means it doesn't removed.

vesparny:
@philpl `brcast@3.0.0` has been released

I've already open PRs to `glamorous` and `theming`

mxstbr:
Since this is pretty hopelessly out of date I'm going to close it, but a follow-up PR implementing this on the new code @schwers wrote would be much appreciated!

PR #1067, opened 2017-08-06
feat(typescript): type domTag (HTML only) component factories

luke-john:
**What**:

Adding typing for styled function to accept dom (html only) node tags (with optional props).

**Why**:

This allows you to type props inside the tagged template string, and also use wittAttrs with dom tags.

**How**:

Extended the ThemedBaseStyledInterface interface with a new interace which has typing for all the HTML dom factories.

inspired by the glamorous typings https://github.com/paypal/glamorous/blob/master/typings/built-in-component-factories.d.ts

if this is okay, I'll proceed with adding the svg typings.

**Checklist**:
- [ ] Documentation N/A
- [x] Tests
- [ ] Ready to be merged

### DIFF
--- a/typings/built-in-component-factories.d.ts
+++ b/typings/built-in-component-factories.d.ts
@@ -1,0 +1,126 @@
+import {
+  WithOptionalTheme,
+  ThemedStyledFunction,
+} from './styled-components'
+
+type HTMLStyledComponentFactory<Theme, Props, Element> = ThemedStyledFunction<
+  Props,
+  Theme,
+  WithOptionalTheme<Props, Theme>
+>;
+
+export interface HTMLComponentFactory<Theme, Props extends { theme?: Theme; }> {
+  a: HTMLStyledComponentFactory<Theme, Props, HTMLAnchorElement>
+  abbr: HTMLStyledComponentFactory<Theme, Props, HTMLElement>
+  address: HTMLStyledComponentFactory<Theme, Props, HTMLElement>
+  area: HTMLStyledComponentFactory<Theme, Props, HTMLAreaElement>
+  article: HTMLStyledComponentFactory<Theme, Props, HTMLElement>
+  aside: HTMLStyledComponentFactory<Theme, Props, HTMLElement>
+  audio: HTMLStyledComponentFactory<Theme, Props, HTMLAudioElement>
+  b: HTMLStyledComponentFactory<Theme, Props, HTMLElement>
+  base: HTMLStyledComponentFactory<Theme, Props, HTMLBaseElement>
+  bdi: HTMLStyledComponentFactory<Theme, Props, HTMLElement>
+  bdo: HTMLStyledComponentFactory<Theme, Props, HTMLElement>
+  big: HTMLStyledComponentFactory<Theme, Props, HTMLElement>
+  blockquote: HTMLStyledComponentFactory<Theme, Props, HTMLElement>
+  body: HTMLStyledComponentFactory<Theme, Props, HTMLBodyElement>
+  br: HTMLStyledComponentFactory<Theme, Props, HTMLBRElement>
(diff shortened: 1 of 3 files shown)


devrelm:
I think that we actually have another issue that this would PR further engrain.

The [latest react types](DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts) have moved towards having smaller, more detailed types for each element to reduce the number of unsupported props on a given intrinsic element's type.

For instance, right now, using the latest react and styled-components typings in TS 2.4, I cannot do this:

```ts
type ResponsiveSize = 'small' | 'large';
type ButtonProps = {size: ResponsiveSize};

const Button = styled.button`
    width: ${({size}: ButtonProps) => size === 'small' ? '30px' : '50px'};
`;

const MyComponent = () =>
    <Button size='small' />;
            ^^^^^^^^^^^^ -- TS error -- Type '"small"' is not assignable to type '(number & "small") | (number & "large")'.
```

What we _should_ do instead is to have all of styled-components' code refer to `JSX.IntrinsicElements` instead of maintaining our own list of tags. I may put in a PR for that later. It would require a handful of other changes as well, and would allow us to get rid of some special-case  handling of svg types vs html types.

philpl:
cc @styled-components/typers

typings/built-in-component-factories.d.ts
```diff
@@ -0,0 +1,126 @@
...
+  ThemedStyledFunction,
+} from './styled-components'
+
+type HTMLStyledComponentFactory<Theme, Props, Element> = ThemedStyledFunction<
+  Props,
+  Theme,
+  WithOptionalTheme<Props, Theme>
+>;
+
+export interface HTMLComponentFactory<Theme, Props extends { theme?: Theme; }> {
```
Igorbek: I don't see how is the type used anywhere in the tests and in the main typings? the file has not been imported anywhere.

typings/styled-components.d.ts
```diff
@@ -65,6 +65,15 @@ type ThemedStyledComponentFactories<T> = ThemedStyledComponentFactoriesHTML<T> &
 export interface ThemedBaseStyledInterface<T> extends ThemedStyledComponentFactories<T> {
   <P extends { theme?: T; }>(component: Component<P>): ThemedStyledFunction<P, T, WithOptionalTheme<P, T>>;
   <P>(component: Component<P>): ThemedStyledFunction<P, T>;
+
+  // Html styled component factories
+  <Props extends { theme?: T; }>(
```
Igorbek: to support proper html tags here, we need to take the tag as a unit type and use information from `ThemedStyledComponentFactories<T>` to retrieve proper factory type. Something like this (simplified):
```ts
<K extends keyof HTMLTags>(htmlTag: K): ThemedStyledComponentFactoriesHTML<T>[K];
<K extends keyof SVGTags>(svgTag: K): ThemedStyledComponentFactoriesSVG<T>[K];
```

goodmind:
Is there any progress?

Igorbek:
@luke-john are you willing to keep working on this?

luke-john:
Hey @Igorbek, apologies, I lost track of this PR.

I'll try and update this in the next day.

typings/styled-components.d.ts
luke-john: just playing about with this and `K` is still a union of all keys of HTMLTags with this style.

I'm not actually sure if this can be achieved without using a giant overload along the following lines.

```ts
interface DomTagStyledFunction {
  (tag: 'div'):  ThemedStyledComponentFactoriesHTML<T>['div']
```

I'm going to open up a pr with something in glamorous (which has this same issue) and will can redo a pr with whatever style ends up landing there.

mxstbr:
Any news here? :blush:

luke-john:
Have redone this based on the feedback here in the following new pr.

https://github.com/styled-components/styled-components/pull/1138

PR #1088, opened 2017-08-14
fix: include __webpack_nonce__ for global styles in Browser context

mzmiric5:
Since `injectGlobal` doesn't pass the name prop to the `BrowserStyleSheet`, style tag for the global CSS would not include the nonce attribute.

This PR just moves the nonce code outside of the name check logic.


src/test/styles.test.js
```diff
@@ -191,4 +195,29 @@ describe('with styles', () => {
...
     Array.from(document.querySelectorAll('style')).forEach(el => expect(el.getAttribute('nonce')).toBe('foo'))
   })
+
+  it('should add a webpack nonce to the global style tags if one is available in the global scope', () => {
+    // eslint-disable-next-line no-underscore-dangle
+    window.__webpack_nonce__ = 'foo'
+
+    injectGlobal`
+        background: red;
+      `
```
mxstbr: Can we make this a proper, valid example? E.g.

```JS
injectGlobal`
  html {
    background: red;
  }
`
```

mzmiric5: whoops, pushing the fix now

PR #1102, opened 2017-08-23
Allow specifying the target element for appending style tags

paul-veevers:
Allows specifying any element to append stye tags to, rather than `document.head`. In combination with `StyleSheetManager` it allows you to specify where all children CSS should be appended.

This is especially useful for apps that use iframes or shadow dom. See https://github.com/styled-components/styled-components/issues/659

For example:
```js
import React from "react";
import ReactDOM from "react-dom";
import PropTypes from "prop-types";
import { StyleSheetManager, StyleSheet, tagConstructorWithTarget } from "styled-components";

export default class StyleManager extends React.Component {
  static propTypes = {
    children: PropTypes.element.isRequired
  };

  constructor(props) {
    super(props);
    const target = document.querySelector('#iframe').contentDocument.head;
    this.stylesheetInstance = new StyleSheet(tagConstructorWithTarget(target));
  }

  render() {
    return (
      <StyleSheetManager sheet={this.stylesheetInstance}>
        {this.props.children}
      </StyleSheetManager>
    );
  }
}
```

I've tried to keep the changes as minimal as possible, but I recognise exporting `StyleSheet` and `tagConstructorWithTarget` isn't ideal since it increases API surface area.

Let me know if there's anything I can do to get this merged.

### DIFF
--- a/src/models/test/StyleSheetManager.test.js
+++ b/src/models/test/StyleSheetManager.test.js
@@ -1,0 +1,76 @@
+// @flow
+/* eslint-disable react/no-multi-comp */
+import React from 'react'
+import { renderToString } from 'react-dom/server'
+import { shallow, mount } from 'enzyme'
+import styled, { ServerStyleSheet, StyleSheetManager } from '../../index'
+
+describe('StyleSheetManager', () => {
+  it('should render its child', () => {
+    const target = document.head
+    const Title = styled.h1`color: palevioletred;`
+    const child = <Title />
+    const renderedComp = shallow(
+      <StyleSheetManager target={target}>
+        { child }
+      </StyleSheetManager>
+    )
+    expect(renderedComp.contains(child)).toEqual(true)
+  })
+
+  // it('should use given stylesheet instance', (done) => {
+  //   const sheet = new ServerStyleSheet()
+  //   const Title = styled.h1`color: palevioletred;`
+  //   renderToString(
+  //     <StyleSheetManager sheet={sheet}>
+  //       <Title />
+  //     </StyleSheetManager>
+  //   )
+  //   expect(sheet.getStyleTags().includes(`palevioletred`)).toEqual(true)
+  // })
+
+  it('should append style to given target', (done) => {
+    const target = document.body
+    const Title = styled.h1`color: palevioletred;`
+    class Child extends React.Component {
+      componentDidMount() {
+        // $FlowFixMe
+        const styles = target.querySelector('style').textContent
+        expect(styles.includes(`palevioletred`)).toEqual(true)
+        done();
+      }
+      render() { return <Title /> }
+    }
+    mount(
+      <StyleSheetManager target={target}>
+        <Child />
+      </StyleSheetManager>
+    )
+  })
+
+  it('should append style to given target in iframe', (done) => {
+    const iframe = document.createElement('iframe')
+    const app = document.createElement('div')
+    // $FlowFixMe
+    document.body.appendChild(iframe)
+    // $FlowFixMe
+    iframe.contentDocument.body.appendChild(app)
+    const target = iframe.contentDocument.head
+    const Title = styled.h1`color: palevioletred;`
+    class Child extends React.Component {
+      componentDidMount() {
+        // $FlowFixMe
+        const styles = target.querySelector('style').textContent
+        expect(styles.includes(`palevioletred`)).toEqual(true)
+        done();
+      }
+      render() { return <Title /> }
+    }
+    mount(
+      <StyleSheetManager target={target}>
+        <Child />
+      </StyleSheetManager>,
+      { attachTo: app }
+    )
+  })
+})
--- a/src/models/test/StyleSheetManager.test.js
+++ b/src/models/test/StyleSheetManager.test.js
@@ -6,6 +6,18 @@ import { shallow, mount } from 'enzyme'
 import styled, { ServerStyleSheet, StyleSheetManager } from '../../index'
 
 describe('StyleSheetManager', () => {
+
+  it('should use given stylesheet instance', () => {
+    const sheet = new ServerStyleSheet()
+    const Title = styled.h1`color: palevioletred;`
+    renderToString(
+      <StyleSheetManager sheet={sheet.instance}>
+        <Title />
+      </StyleSheetManager>
+    )
+    expect(sheet.getStyleTags().includes(`palevioletred`)).toEqual(true)
+  })
+
   it('should render its child', () => {
     const target = document.head
     const Title = styled.h1`color: palevioletred;`
@@ -18,18 +30,7 @@ describe('StyleSheetManager', () => {
     expect(renderedComp.contains(child)).toEqual(true)
   })
 
-  // it('should use given stylesheet instance', (done) => {
-  //   const sheet = new ServerStyleSheet()
-  //   const Title = styled.h1`color: palevioletred;`
-  //   renderToString(
-  //     <StyleSheetManager sheet={sheet}>
-  //       <Title />
-  //     </StyleSheetManager>
-  //   )
-  //   expect(sheet.getStyleTags().includes(`palevioletred`)).toEqual(true)
-  // })
-
-  it('should append style to given target', (done) => {
+  it('should append style to given target', () => {
     const target = document.body
     const Title = styled.h1`color: palevioletred;`
     class Child extends React.Component {
@@ -37,7 +38,6 @@ describe('StyleSheetManager', () => {
         // $FlowFixMe
         const styles = target.querySelector('style').textContent
         expect(styles.includes(`palevioletred`)).toEqual(true)
-        done();
       }
       render() { return <Title /> }
     }
@@ -48,7 +48,7 @@ describe('StyleSheetManager', () => {
     )
   })
 
-  it('should append style to given target in iframe', (done) => {
+  it('should append style to given target in iframe', () => {
     const iframe = document.createElement('iframe')
     const app = document.createElement('div')
     // $FlowFixMe
@@ -62,7 +62,6 @@ describe('StyleSheetManager', () => {
         // $FlowFixMe
         const styles = target.querySelector('style').textContent
         expect(styles.includes(`palevioletred`)).toEqual(true)
-        done();
       }
       render() { return <Title /> }
     }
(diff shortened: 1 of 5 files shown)


philpl:
I don't think the API here is optimal but it's definitely a great idea! cc @geelen: Can you take a look please? :wink: Looks like this might be able to resolve all iframe / shadow DOM problems?

paul-veevers:
Interesting. I think that will solve some related problems I've seen in the issues, but won't solve  iframes/shadow-dom use cases?

For example, here's what I want to write:

```js
<StyleManager >
  <StyledCmpntA />
  <Frame>
    <StyleManager>
      <StyledCmpntB />
    </StyleManager>
  </Frame>
  <Frame>
    <StyleManager>
      <StyledCmpntC />
    </StyleManager>
  </Frame>
</StyleManager>
```

Where `Frame` is an iframe, e.g. https://github.com/ryanseddon/react-frame-component

In this case, the style tag for `StyledCmpntA` will be in the main window `head`, the styles for `StyledCmpntB` will be in the `head` of its parent `frame`, and the styles for `StyledCmpntC` will be in the `head` of its parent `frame`.

That's all achievable with the changes I've made, but I don't think it is with a `<!-- sc-insertion-point -->` api?

It's also completely understandable if you don't want to support my crazy use case :)

mxstbr:
Hmm, yeah that's true... /cc @geelen let's see what he thinks about this change.

mxstbr:
What do you think about that API @paul-veevers?

mxstbr:
I've talked to @geelen, he approves of that API! Want to go ahead and implement it and also add some tests inserting the CSS into shadow DOM?

paul-veevers:
Awesome! I will do it in the next 48 hours or so.

Is there anything I should be aware of?

mxstbr:
> Is there anything I should be aware of?

I don't think so, tests would be great though! Can't wait 👌

paul-veevers:
Some things to note:
- Updating the `target` prop won’t have any effect. Even if I implemented `componentWillReceiveProps` the context won't update? I think it's an ok limitation for now since I don't see the use case for updating the target.
- There doesn't appear to be a way to clean up the stylesheet on unmount?
- A potential future improvement is to auto resolve the target via ref. So it would simply look like `<StyleSheetManager><Children/></StyleSheetManager>`. But, getting a ref seems to be tricky (if not impossible?) without wrapping the children in an element.
- jsdom lacks a shadow dom implementation so there's no test for it
- I haven't used Flow before so let me know if you see anything wrong. I used `$FlowFixMe` in the test file to silence some `Method cannot be called on possibly null value` errors.

src/models/test/StyleSheetManager.test.js
```diff
@@ -0,0 +1,76 @@
...
+    const child = <Title />
+    const renderedComp = shallow(
+      <StyleSheetManager target={target}>
+        { child }
+      </StyleSheetManager>
+    )
+    expect(renderedComp.contains(child)).toEqual(true)
+  })
+
+  // it('should use given stylesheet instance', (done) => {
```
mxstbr: Why was this test commented out? It should work, right?

paul-veevers: Ah yes, I forgot to fix it.

It's now fixed, but just a note that I had to make it the first test in the file. I think it's because `StyleSheet.instance` is a singleton so it was using a `BrowserStyleSheet` if it was after one of the other tests? Not sure, but something to keep in mind.

AllenWilliamson:
I've been testing this out and it looks promising, but what about the functionality of `injectGlobal`? I was expecting the global styles to be added to the custom target element instead of the document `head`.

Kauhsa:
I've been testing this branch as well (since the functionality is very very useful for a project I'm working on). I use this with `react-frame-component`, and while this otherwise seems to work fine, in some cases the following seem to happen. Let's say we roughly the following components:

```
const A = styled.div`
  color: red;
`

const B = styled(A)`
 color: blue;
`
```

...and upon rendering B, the color is still `red`, because the style for A is inserted to `<style>` after the style for B. This is likely not a working test case but an oversimplified explanation what seems to happen for me, and it will be a while until I have time to produce one. :(

In my case, I can avoid this by replacing `styled(A)` with `A.extend`, and everything is fine.

philpl:
@Kauhsa we’ve been trying to understand this bug for a while, but haven’t found a reliable reproduction yet. It seems to be unrelated to this branch however. Can you post a reproduction on this issue if you find one please? :) https://github.com/styled-components/styled-components/issues/1164

Kauhsa:
@philpl ah, sure thing. I doesn't happen though when I use the same components outside of `<StyleSheetManager>` in a different part of code, but many other are also different there so it might not be the cause :)

Kauhsa:
Ping @philpl @paul-veevers:

Well, well. Managed to make an example, on the first try, even! In this case it's broken by `<StyleSheetManager>` because it works correctly without it, but maybe it's related to #1164 anyway? :)

https://github.com/Kauhsa/styled-components-1102-issue

mxstbr:
That definitely looks like an issue with the `StyleSheetManager` if it doesn't happen if you don't use it! @paul-veevers mind taking a look at that repo, and maybe adding it as a test case so we can make sure to fix it? Ordering problems are the worst kind of problems :blush:

Kauhsa:
Ok, I tinkered even more, and this is shows the issue as well (I realize I'm probably doing things I shouldn't here, and I don't know how reliable `jsdom` is, but... yeah). With this code, it's reproducible with `v2.2.0` as well and doesn't require the pull request.

```
import React from 'react';
import ReactDOM from 'react-dom';
import styled, { StyleSheetManager } from 'styled-components';
import StyleSheet from 'styled-components/lib/models/StyleSheet';
import { JSDOM } from 'jsdom'

const dom = new JSDOM(`
  <!DOCTYPE html>
  <html>
    <head>
    </head>
    <body>
      <div id="root" />
    </body>
  </html>
`)

global.document = dom.window.document
global.window = dom.window

const Thing = styled.div`
  color: red;
`

const AnotherThing = styled(Thing)`
  color: blue;
`

// explicitly not server stylesheet, though since `document` exists
// it should be false by default anyway
const sheet = StyleSheet.create(false)

const App = () => 
  <StyleSheetManager sheet={sheet}>
    <AnotherThing>Why am I red</AnotherThing>
  </StyleSheetManager>

ReactDOM.render(<App />, document.getElementById('root'))

// shows the style for Thing added after AnotherThing
console.log(dom.window.document.documentElement.outerHTML)
```

I'm not sure if this implies I'm hitting the same issue #1164 is talking about, since reproducing doesn't require the usage of the new `target` feature implemented in this PR. @mxstbr, @philpl, your two cents?

mxstbr:
Can you try submitting this as a test against master? If that test fails, it means you've managed to reproduce #1164 (which would be amazing!), if it passes it means you found a hole in this PR. (which would also be amazing!)

Sooo... PR pl0x? :blush:

Kauhsa:
Yeah, sure :smile: I need to leave my computer now, but I'll try to get around it as soon as I have time.

Also I'm wondering whether creating an instance of `BrowserStyleSheet` manually and passing it as parameter to `StyleSheetManager` is kosher in the first place, since `StyleSheet` isn't part of the public API, right?

Fer0x:
I agreed with @AllenWilliamson, changing API of `injectGlobal` is also important thing for working with frames.
The API can be look like this:
```js
injectGlobal(document.head)`
     body: {
     }
`;
```

AllenWilliamson:
@Fer0x I solved this issue by creating a style component that uses stylis to process the css (styled-components uses stylis under the hood)

https://gist.github.com/AllenWilliamson/e7ae4c11c8fe0b91f7969d60a508e9b7

You can use it like so:
```javascript
<Style>
    :host{
        border: 1px solid red;
    }
</Style>
```

Modifying injectGlobal like that probably wouldn't work well. It would be difficult to specify a react dom node unless you used it within componentDidMount.

I think the API @paul-veevers has added is great. I've been using it along with that gist to style shadow dom inside web components with no major issues. Some feedback on using stylis this way would be appreciated though. I think if you have a need to specify a node for components, you will eventually need to inject some globals in there as well.

mxstbr:
I feel like having a `<GlobalStyle>` component that moves `injectGlobal` to a component-context makes sense, especially since that also means you'd be able to use your theme in the global styles, right?

/cc @philpl

philpl:
> I feel like having a <GlobalStyle> component that moves injectGlobal to a component-context makes sense, especially since that also means you'd be able to use your theme in the global styles, right?

That would introduce more problems, like the `keyframes` helper. Generally it shouldn't matter where global styles like those go as long as they're there. Maybe @geelen has a plan on how to handle this.

Generally as long as we can spot the stylesheet **somewhere** we should be fine. https://github.com/paul-veevers/styled-components/blob/f599f472fa3233382caa76bbe4714863485fa33f/src/models/BrowserStyleSheet.js#L139

Opening those two helpers up for more context is potentially a problem for their previous usage unfortunately. So we have to tread carefully here :smile:

ivanakimov:
@mxstbr @philpl any chance this might get fixed this month? :sweat_smile::cold_sweat:

mxstbr:
@ivanakimov depends on @paul-veevers, not us! Ping him to see if he has any time to push this over the finish line :blush:

daveaspinall:
We unfortunately had to remove styled-components from two big client projects because of this issue (we don’t have control of document.head) 🙁 Would love to see this feature implemented!

Fer0x:
@mxstbr Is problem only in commented test?

mxstbr:
I don't know, the branch needs to be rebased, conflicts resolved, the CI failures fixed and then we gotta figure out if that test has anything to do with these changes. If somebody wants to go ahead and dig into this that'd be awesome!

bteng22:
hey @mxstbr this PR would be super helpful for my team and I. We pulled down @paul-veevers PR and rebased the branch with master and fixed up the conflicts. The tests are all green except for the `native.test.js`, which we're assuming should fail since it's for mobile (let us know if this isn't the case). 

Should we open a new PR with the branch that we've forked? Let us know 😄 !

alexandrarprice:
Hey, following up with @bteng22 this PR is working really well for us locally! Let us know how we can help get it to prod!

mxstbr:
Yes please, can you open up a new PR with those changes? That'd be awesome!

rpellerin:
Any updates on this?

mxstbr:
Yes, #1324

kitten:
Closing since this was addressed and picked up by #1491 (merged)

jackrobertscott:
See here for progress: https://github.com/styled-components/styled-components/pull/1919

probablyup:
This is a different concern

Issue #1109, opened 2017-08-24
[docs suggestion] Import order of global styles


geelen:
Just hit something demo-ing SC to a mate. If you do this:

```js
import { injectGlobal } from 'styled-components'
import App from './App'

injectGlobal`
  * { all: initial; }
`

ReactDOM.render(<App />, document.getElementById('root'))
```

Because this file depends upon its imports, `App.js` is _always_ evaluated before this file. Which means if it declares any styled components they will be inserted before the `injectGlobal` call here. It means something like a reset is running in the wrong order. You just need to extract another import:

```js
import './global-styles'
import App from './App'

ReactDOM.render(<App />, document.getElementById('root'))
```

```js
/* global-styles.js */
import { injectGlobal } from 'styled-components'

injectGlobal`
  * { all: initial; }
`
```

Then, order is preserved. This should probably be somewhere in the docs. Maybe it is already? Who knows, not me!

k15a:
I think this is pretty much related to https://github.com/styled-components/styled-components/issues/1108 and we should.change the behaviour :)

mxstbr:
Superseded by #1333, which will solve this once and for all.

tysonmatanich:
Just wasted several hours because of this. I understand a fix is coming, however adding it to the documentation until then is the right thing to do.

probablyup:
@tysonmatanich what are you seeing? The behavior has changed a lot since this issue. Please open a new one if you have a specific bug to report.

tysonmatanich:
@probablyup Not a bug, just think it would be good to call out in the [documentation](https://www.styled-components.com/docs/api#injectglobal). The issue I encountered seems exactly the same as @geelen reported. The workaround of needing the global styles import before the App is imported works great. I'm using SSR if that impacts anything. Tried on both v3.2.5 and v3.2.6.

I thought `injectGlobal` just had to be called before `<App />`  was rendered but that is not the case. The styles import has to occur before the App import. 

**Works**
```
import '../styles';
import App from '../app';
```

PR #1117, opened 2017-08-27
fix: coalesce tags with similar locality prior to injection

### LINKED ISSUE #1108: Alternating local/nonlocal components create many separate style tags
## Version
2.1.0

## Reproduction

Types of injections out of order create multiple of each style block type (https://codesandbox.io/s/7403wnx2mx):
screen shot 2017-08-23 at 8 49 33 pm

Each type in order produces only one of each style block (https://codesandbox.io/s/y7vrq9853x):
screen shot 2017-08-23 at 8 49 38 pm

## Expected Behavior
A unified style tag is created for non-local and for local styles. The non-local styles should probably be injected before the local styles.

## Actual Behavior
Multiple style tags of each type are created if local (`styled.x`) vs nonlocal (`injectGlobal`) injections are detected heterogeneously.

This bug is an issue because it can cause styles to be injected out of order, which causes specificity issues.

probablyup:
Prior to an injection run being completed, it's safe to concatenate non-local styles together into a single style element and likewise with local component styles. This helps reduce the overall number
of injected style tags in a highly diverse component tree.

Fixes #1108

### DIFF
--- a/src/models/StyleSheet.js
+++ b/src/models/StyleSheet.js
@@ -117,18 +117,32 @@ export default class StyleSheet {
       return existingTag
     }
 
-    const lastTag = this.tags[this.tags.length - 1]
-    const componentTag = (!lastTag || lastTag.isFull() || lastTag.isLocal !== isLocal)
-      ? this.createNewTag(isLocal)
-      : lastTag
-    this.componentTags[componentId] = componentTag
-    componentTag.addComponent(componentId)
-    return componentTag
+    let tag = this.tags.find(t => t.isLocal === isLocal && !t.isFull())
+    if (!tag) {
+      tag = this.createNewTag(isLocal)
+    }
+
+    this.componentTags[componentId] = tag
+    tag.addComponent(componentId)
+    return tag
   }
 
   createNewTag(isLocal: boolean) {
     const newTag = this.tagConstructor(isLocal)
-    this.tags.push(newTag)
+
+    /**
+     * The first non-local tag should go in position zero.
+     *
+     * Subsequent non-local tags should be appended as normal because
+     * calling this func again means the prior one was "closed" due to a
+     * finished injection.
+     */
+    if (!isLocal && !this.tags.some(t => !t.isLocal)) {
+      this.tags.unshift(newTag)
+    } else {
+      this.tags.push(newTag)
+    }
+
     return newTag
   }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file. If a contri
 
 - Improve component mount and unmount performance with changes to `createBroadcast`. Deprecates usage of `CHANNEL` as a function, will be update to `CHANNEL_NEXT`'s propType in a future version. (see [#1048](https://github.com/styled-components/styled-components/pull/1048))
 - Fixed comments in react-native (see [#1041](https://github.com/styled-components/styled-components/pull/1041))
+- Reduce the number of injected style tags when using many types of injection (component style, `injectGlobal`, etc) interleaved in a single component tree (see [#1117](https://github.com/styled-components/styled-components/pull/1117))
 
 ## [v2.1.2] - 2017-08-04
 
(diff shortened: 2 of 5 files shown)


philpl:
lgtm cc @geelen

mxstbr:
This might clash with the `batch-updates` branch that @geelen has been working on?

philpl:
@mxstbr yea, it most likely will in some ways

probablyup:
My change is only pre-injection, though. It wouldn't affect updates.

philpl:
@probablyup The stylesheets and tags have been refactored by @geelen on that branch. In fact, there's now an in-memory tag that defers the "clean up" / flushing of these initial browser tags. Then when the first, new CSS is injected, it cleans up after itself. I think this or similar is how it works, but I haven't read through the entire code / all the changes yet.

mxstbr:
> FYI it seems that batch-updates won’t be sufficient to bring us up to the speedy-optimisation’s perf

Yeah that's what I'm seeing too, @geelen hasn't had time to look at that yet though.

geelen:
Ok caught up on this. This would be a pretty big change, since it's not "safe" to reorder arbitrary CSS at all. Even though component styles and global css would _usually_ not conflict, they definitely can:

```js
injectGlobal`
  .link { 
    text-decoration: none;
  }
  .link:hover {
    color: inherit;
    text-decoration: underline;
  }
`

const Link = styled.a.attrs({
  className: 'link',
}`
  text-decoration: underline;
  &:hover {
    color: red;
  }
`
```

If those two statements are reversed when injected into the DOM, since they have the same specificity, the `Link` styles won't override the `.link` ones. So we _guarantee_ that whatever order you write your CSS in, be it in an `injectGlobal` call or a `styled.x` one, that's the order the CSS will be injected in. The behavior in #1108 isn't actually a bug—it's that your JS bundle defines local styles before injecting global ones. See the example in #1109 for a fix, you need to extract a file you import before you import your components.

It means if you interleave global and local CSS you'll get lots of style tags, yep. And they have to be separate tags in case you use a `@font-face` declaration somewhere #400.

I admit this behaviour can be surprising and should probably be changed, but the implementation in this PR isn't the right fix. I can see two decent alternatives:

* Redefine `injectGlobal` to _always_ inject above _all_ local CSS.
* Add a special case to the current implementation. When `injectGlobal` is called (i.e. global CSS is injected) _before_ any of the currently-defined SCs are _rendered_ (i.e. local CSS is injected), inject it into a style tag _above_ the first local style tag. If, for some reason, `injectGlobal` is called dynamically down the track, it would interleave, creating extra style tags.

Thoughts? I'm leaning towards the first one, but it'd be a breaking change...

probablyup:
The second bullet is exactly what this PR does.

src/models/StyleSheet.js
```diff
@@ -117,18 +117,32 @@ export default class StyleSheet {
...
     }
 
-    const lastTag = this.tags[this.tags.length - 1]
-    const componentTag = (!lastTag || lastTag.isFull() || lastTag.isLocal !== isLocal)
-      ? this.createNewTag(isLocal)
-      : lastTag
-    this.componentTags[componentId] = componentTag
-    componentTag.addComponent(componentId)
-    return componentTag
+    let tag = this.tags.find(t => t.isLocal === isLocal && !t.isFull())
```
geelen: Won't this line find previous local tags that aren't full and append them there? Effectively maintaining two "last tags", one for local, and one for global styles? Then they both get appended until they're full and more get created?

probablyup:
Nope. Once injection happens, the old tags are "closed" and new tags are
created as normal.

geelen:
Hmm, I don't think so. This line asserts only 2 tags are created, there should probably be 4: https://github.com/styled-components/styled-components/pull/1117/files#diff-ff4f58ec4bfe127b48d620e3f45853e5R230

Imo any kind of inconsistent behaviour (even if with good reason) is undesirable—I'd be happier just saying that all global styles precede all local ones. If it had been that way from the start I don't think anyone would have complained...

probablyup:
My changes just enforce that all global styles go first prior to the first
flush. Once that's completed and the tags have been emitted, they are
considered closed and new tags will be created and flushed as normal
afterward.

It's really just a first-run optimization and as you can see from the
tests, shouldn't break anything.

philpl:
@probablyup Hm, the previous notes from @geelen make me a bit wary of this change. My problem is, will SSR and non-SSR still yield consistent results?

Also, this test seems to test for some dangerous behaviour: https://github.com/styled-components/styled-components/pull/1117/files#diff-ff4f58ec4bfe127b48d620e3f45853e5R207

If `Heading` contains a selector rule, and the `injectGlobal` after it targets the same selector, then won't we change which rule applies due to the swapped order?

I believe this is a problem we need to discuss first, but it might not be that significant, since for it to become a problem `injectGlobal` would have to use the class that the StyledComponent uses. Thus a conflict of specificity by order might be unlikely, if I'm not mistaken?

probablyup:
Irrelevant with the stylesheet refactor.

PR #1122, opened 2017-08-29
chore: Update Danger to 2.x

orta:
Reduces your dependency tree ;)


orta:
Alright, this is something I found in testing with the sandboxing + regenerator-runtime async behavior but couldn't re-pro in tests, gonna close this, you don't need a hanging PR - will give it another shot once I am happy that I've figured that out.

PR #1123, opened 2017-08-29
Support style objects in ts definitions

nbostrom:
The utility functions in [styled-system](https://github.com/jxnblk/styled-system) creates style objects to use together with css-in-js libs. I wanted to create ts definitions for it and use it together with styled-components, but the definitions don't support styles as objects yet.

This PR only supports the most basic kind of usage and the properties are not type checked, like in glamor or typestyle.

Note that I'm fairly new to typescript so you should probably be very thorough in your review! This is a way for me to learn.


k15a:
@styled-components/typers

nbostrom:
Just noticed an error myself. This does not support nested objects. Will fix and update later.

nbostrom:
I have a question regarding tests. Would you like the tests for this thing to live in its own file (ie style-objects-tests.tsx) or is it central enough to be thrown into the kitchen sink style tests in main-test.tsx? @Igorbek @patrick91

Igorbek:
To me, a separate test file is much preferable. However, if your feature spreads across all the use cases, you can add it in existing files. I have no strong opinion, just slightly concerned of growing main-test.tsx.

nbostrom:
Looking at it some more, this feature feels kinda similar to the "nested" feature in that it's a special kind of thing that goes inside the template literal. The "nested" feature has its own file, so I'll try to follow that example while covering as many cases as possible.

nbostrom:
Alright, this should now be ready for review

typings/styled-components.d.ts
```diff
@@ -18,8 +18,11 @@ export type OuterStyledProps<P> = ThemedOuterStyledProps<P, any>;
 
 export type Interpolation<P> = FlattenInterpolation<P> | ReadonlyArray<FlattenInterpolation<P> | ReadonlyArray<FlattenInterpolation<P>>>;
 export type FlattenInterpolation<P> = InterpolationValue | InterpolationFunction<P>;
-export type InterpolationValue = string | number | StyledComponentClass<any, any>;
+export type InterpolationValue = string | number | Styles | StyledComponentClass<any, any>;
 export type SimpleInterpolation = InterpolationValue | ReadonlyArray<InterpolationValue | ReadonlyArray<InterpolationValue>>;
+export interface Styles {
+  [ruleOrSelector: string]: string | Styles;
```
Igorbek: Just curious, does the style only allowed to be a string. Are any other types accepted, like numbers, undefined, nulls, and so on?

nbostrom: It's not allowed to be a number, because sc does not allow unit less styles, but it might allow null or undefined. I'll have to check when I'm at my computer. Nice catch 👍

nbostrom: I have investigated this some more, and while it "works" to give it anything, SC will just stringify anything that isn't an object and the result will most likely be an invalid property value. If we allow numbers I think that would make people believe that the numbers will be automatically converted, like they would within the style prop of react elements.

sveinburne:
Any updates ? Would love this defs to be published.

typings/styled-components.d.ts
philpl: @nbostrom Let's get this change in; However, we'll need `number` to be added here. `InterpolationValue` contains numbers, so the same must be accepted for object values. :wink:

cc @Igorbek

Igorbek: I'm ok with either way: with or without numbers. If numbers are really disallowed (or, going to be deprecated), than let's not include them.

philpl: @Igorbek they're neither disallowed nor deprecated. You can use them in your non-object interpolations as well, which is why it makes sense to add them to the types here as well

Igorbek: sure, let's add them. @nbostrom could you please add `number` here.

philpl:
@nbostrom can you add a quick entry to our changelog please? :wink: We tend to keep losing track of some TS changes because of some left out changelog entries, since Danger ignores those for this folder 😞

nbostrom:
@philpl sure! Should I put it in the "unreleased" section?

sveinburne:
Thanks for the work folks. Looks good to maintainers ?

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file. If a contri
 - Refactor out theme logic in StyledComponent's componentWillMount & componentWillReceiveProps (see [#1130](https://github.com/styled-components/styled-components/issues/1130))
 - Add onReset to valid react props list (see [#1234](https://github.com/styled-components/styled-components/pull/1234))
 - Add support for ServerStyleSheet PropType in both StyleSheetManager and StyledComponent (see [#1245](https://github.com/styled-components/styled-components/pull/1245))
+- Added basic support for style objects in the Typescript definitions (see [#1123](https://github.com/styled-components/styled-components/pull/1123))
```
Igorbek: Just very minor comment, that for consistency with other entries in the change log, use imperative tense here: _"Add~~ed~~ basic..."_

mxstbr:
This has been sitting around for way too long, thanks @nbostrom! Going to ship this :100:

madou:
Any chance of backporting this to v1?

mxstbr:
Probably not @madou. Given that v2 is a drop-in replacement, what's holding you back from upgrading?

madou:
[Atlaskit](https://atlaskit.atlassian.com/) has a dependency on v1, unsure when they're going to upgrade :).

v1 can still parse style objects yeah? might try having a go at tweaking the type defs.

mxstbr:
Damn, @JedWatson @thejameskyle  y'all should upgrade styled-components in AtlasKit! v2 is like half the bundle size and much faster, and it shouldn't break anything.

JedWatson:
@mxstbr we did a bunch of performance work on Atlaskit late last year and upgrading to styled-components v2 caused some meaningful performance problems, which was a real surprise. So while it's compatible, the runtime cost is way too high at the moment (outweighs the benefits of a smaller bundle size, certainly)

Not sure why our results are different to others, like you said v2 is supposed to be faster. I _suspect_ the fact that we're seeing different performance characteristics is related to the way we're using it - lots of dynamic styles, theming, etc.

Until we get to the bottom of it, though, we need to stick with v1 😕

(sorry @madou!)

Issue #216, opened 2016-11-19
Add tests for preact-compat and inferno-compat


busypeoples:
Not sure If this has been discussed already, but I think it would be great to see styled-components work with Preact, Inferno and other Virtual Dom implementations. Are there any thoughts regarding making styled-components JSX and especially React agnostic? 
I can help with this, if there is an interest in making this happen and a general idea on how to approach the topic.

mxstbr:
Yes.

Preact should already work (untested), and in general the only React-specific part of `styled-components` is in two or three files. (`StyledComponent.js` and `ThemeProvider.js`) We're looking towards adding support for web components (ref: #128) and I know @trueadm is looking into adding inferno support! I've also experimented with `choo` support, haven't looked too deeply into it but adding that shouldn't be hard.

Since `styled-components` returns a component (in whatever framework you use), making it fully agnostic isn't possible I'm pretty sure, but we're going to continuously add support for more and more libraries and frameworks! The ultimate goal is: `styled-components` everywhere 😎

whmountains:
I'm going to see about implementing inferno support.

wenwuwu:
Me too. Looking for compatible version for inferno and preact. Thx.

philpl:
Styled-components does work with preact-compat perfectly

k15a:
Is there a way we can run our existing tests with `preact-compat` and `inferno-compat` as well?

SunHuawei:
@whmountains and @wenwuwu , any progress?

philpl:
@k15a We should add some additional tests but making the tests 3x longer is a bit overkill ;)

whmountains:
@SunHuawei Nope.  I got busy with other stuff, unfortunately.

alvinsight:
Hey guys, 
Great job on v2! Since it is out now, any chance this issue could be talked about again ? 
Thanks!
Alvin

bkazi:
Has there been any progress on this? Is it still something that needs working on?

Maybe I'm over-simplifying it but, would it be possible to run the existing tests aliasing `react` with `preact-compat` and `inferno-compat` using something like [jest-alias-module-loader](https://www.npmjs.com/package/jest-alias-module-loader)?

mxstbr:
No progress so far, PRs welcome!

philpl:
[...]

PR #1126, opened 2017-08-29
Test compat libs

bkazi:
Aim to implement #216 

Alias React using [`moduleNameMapper`](https://facebook.github.io/jest/docs/en/configuration.html#modulenamemapper-object-string-string)
Tried porting [native/primitive tests](https://github.com/styled-components/styled-components/issues/216#issuecomment-325725186) with a simple find and replace of `View` to `div` as a start. Results in failing 14 tests and passing 2 tests
Running the web tests with aliased `preact-compat` results in 3 test suites failing
- basic > jsdom
- theming
- ssr

(currently only working on using preact-compat)


mxstbr:
The web styles are probably the way to go since the native styles expect the `style` prop to be set which won't be set in Preact.

bkazi:
Ahh that makes more sense now.
Are there specific test suites I should target or just aim to pass all the tests in the test folder?
(sorry it's my first contribution so I'm a bit confused)

mxstbr:
Everything apart from SSR should pass I think? I dunno if theming works in Preact.

bkazi:
The reason for `basic > jsdom` tests failing is that the innerRef function never gets called
and `theming` fails because the props are not updated when trying to update the theme

mxstbr:
For one, maybe `ref`s aren't a thing with Preact? Same with context for theming? I've never used Preact so I'm unsure where the differences with React are, those might be it?

philpl:
@mxstbr `ref`s were added to preact-compat in v4 (?) I believe, and the context is supported as well. The only notable difference is that `children` are sometimes handled quite differently, which we probably won't have to deal with in our tests, I'd imagine

bkazi:
Yeah preact-compat should have feature parity with React

The issue was with Jest not aliasing React and React-DOM all the way in Enzyme
Used a transformer that uses [babel-plugin-module-alias](https://www.npmjs.com/package/babel-plugin-module-alias) as an option to [babel-jest](https://www.npmjs.com/package/babel-jest)
Results in all tests passing for preact-compat, I'll start work on inferno-compat soon.

bkazi:
Actually no
I messed up big time
I wanted to know for sure if the aliasing was working and turns out it only passes because of jest caching and isn't really being aliased
So currently this is a false positive 

I am however working on a fix using [preact-compat-enzyme](https://github.com/windyGex/preact-compat-enzyme) which seems to be the recommended way to go until enzyme v3 which will allow for easier Preact support (reference: developit/preact-compat#82)

mxstbr:
Good catch!

mxstbr:
Any news here? enzyme v3 was released yesterday... :wink:

bkazi:
Yess testing preact with v2 and the above stated libraries proved to be really difficult since there's no shallow renderer support
There is an ongoing adapter implementation for enzyme v3 and preact [preact-enzyme-adapter](https://github.com/aweary/preact-enzyme-adapater), so I'm waiting on progress there
In the meantime it might be worth upgrading the project to enzyme v3

mxstbr:
Is this ever going to be able to be done @bkazi? :cry: Or did the enzyme v3 update just totally fuck us over?

kitten:
Closing due to inactivity, sorry 😖 I also anticipate that a couple of things will change in canary and our hooks implementation will likely not be compatible anymore with a lot of React-like libraries

PR #1138, opened 2017-09-03
feat(typescript): type domTag component factories

luke-john:
Based on feedback on https://github.com/styled-components/styled-components/pull/1067

Apologies on the delays in following up the feedback on that pr.

**What**:

Adding typing for styled function to accept dom tags (with optional props).

**Why**:

This allows you to type props inside the tagged template string.

**How**:

Extended the ThemedBaseStyledInterface interface with a new interface `ThemedBaseStyledInterface` which takes the theme and then extends with interfaces for each dom tag.

Follows the strategy currently used by glamorous https://github.com/paypal/glamorous/pull/310.

**Checklist**:
- [ ] Documentation
- [x] Tests
- [ ] Ready to be merged


typings/dom-tag-component-factories.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export interface DomTagComponentFactory<DomTag extends keyof JSX.IntrinsicElements, Theme> {
+    <Props>(domTag: DomTag): ThemedStyledFunction<Props & JSX.IntrinsicElements[DomTag], Theme, WithOptionalTheme<Props & JSX.IntrinsicElements[DomTag], Theme>>;
+    <Props>(domTag: DomTag): ThemedStyledFunction<Props & JSX.IntrinsicElements[DomTag], Theme>;
+};
+
+
+export interface DomTagComponentFactories<Theme> extends
+  DomTagComponentFactory<"a", Theme>,
+  DomTagComponentFactory<"a", Theme>,
```
devrelm: You list `"a"` twice here.

Also, is there no way to do this with the `JSX.IntrinsicElements` list using [Mapped Types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#mapped-types), or the `keyof` keyword in general? I would have thought that something like `interface DomTagComponentFactories<Theme> extends DomTagComponentFactory<keyof JSX.IntrinsicElement>`, though I'm pretty sure that that exact syntax wouldn't work...

luke-john: I tried a few things, but I couldn't find a way to limit the key to a single key in the union or to get something like mapped types to work with function overloads.

Would love if someone finds a way though :).

and will fix the double a shortly

typings/dom-tag-component-factories.ts
```diff
@@ -0,0 +1,141 @@
...
+  DomTagComponentFactory<"path", Theme>,
+  DomTagComponentFactory<"pattern", Theme>,
+  DomTagComponentFactory<"polygon", Theme>,
+  DomTagComponentFactory<"polyline", Theme>,
+  DomTagComponentFactory<"radialGradient", Theme>,
+  DomTagComponentFactory<"rect", Theme>,
+  DomTagComponentFactory<"stop", Theme>,
+  DomTagComponentFactory<"svg", Theme>,
+  DomTagComponentFactory<"text", Theme>,
+  DomTagComponentFactory<"tspan", Theme> {}
```
Igorbek: Sorry, that is really something wrong with that approach if you had to list all the tags in that way.
You probably need to use mapped types along with string union types.

luke-john: I'd love if someone could work out a way to achieve it with mapped types and string union types.

Afaict mapped types don't work with function arguments.

mxstbr:
Closed in favor of #1139, thanks so much for taking the time to submit a PR though!

Issue #1080, opened 2017-08-10
[TypeScript] styled argument do not accept string for tagnames


renatorib:
It does not work:

![image](https://user-images.githubusercontent.com/3277185/29194229-0ec54c6a-7dff-11e7-8e96-1a07ecf01954.png)

It works:

![image](https://user-images.githubusercontent.com/3277185/29194326-6752f684-7dff-11e7-93e5-e6577ee8ecb4.png)

Following `styled typings` I found that first argument receive `Component<P>`:
```tsx
type Component<P> = ComponentClass<P> | StatelessComponent<P>;
```

If I add `| string;` at Component type, the first example works successfully.

What I can do in this case? And why Component argument do not accept string?

philpl:
This does look like an oversight to me. Rest assured that you can still use the shorthands. Please check out our docs for some examples of them: https://www.styled-components.com/docs

cc @styled-components/typers

renatorib:
Sorry, I guess I did not express the problem very well.

I want to create styled components with prop types and using string for html tags

```tsx
const Comp = styled<Props>('div')``;
//                 ^ [ts] Argument of type "'div'" is not assignable to parameter of type 'Component<Props>'
```

But typescript does not let me do that, because of this type: 
```tsx 
type Component<P> = ComponentClass<P> | StatelessComponent<P>;
```

renatorib:
Ops, I've closed issue by missclick. Sorry.

Igorbek:
I'm sure #1067 should be addressing this feature.

renatorib:
@Igorbek exactly that issue.

For now, I've created a wrapper to handle this:
``` tsx
import React, { StatelessComponent, ComponentClass } from 'react';
import styled, { Interpolation, ThemedStyledProps } from 'styled-components';

function styledTyped<P>(
  Component: StatelessComponent | ComponentClass | string,
) {
  return (
    strings: TemplateStringsArray,
    ...interpolations: Interpolation<ThemedStyledProps<P, any>>[],
  ) => {
    return styled<P>(() => <Component />)(strings, ...interpolations);
  };
}

// after

type LineProps = {
  background: string;
};

const Line = styledTyped<LineProps>('div')`
  width: 100%;
  height: 1px;
  background: ${p => p.background};
`;
```

philpl:
@renatorib is there any reason why you're using this workaround instead of the shorthands?

renatorib:
@philpl because I can't pass types using shorthands.

[...]

PR #1139, opened 2017-09-03
[TypeScript] Allow dom tag strings to be passed to the `styled` method

devrelm:
Proposed fix to #1080

Alternative to #1138

This is a quick fix for allowing tag strings to be passed to the `styled` factory method.

Note that this implementation is not perfect: if you want to pass a props type param, you must also pass the tag-name as a type param. It requires the users to repeating themselves a little bit, but requires much less overhead on our end in terms of defining and maintaining the types.


devrelm:
Bump?

philpl:
cc @styled-components/typers

Igorbek:
Very nice!

philpl:
@Igorbek @patrick91 Shall we merge this and close #1138?

patrick91:
Yup :)

Issue #719, opened 2017-04-20
Consider toggling vendor prefixing on/off (stylis v3)


gajus:
This is regarding `2.0.0-15`.

Using flex boxes results in:

<img width="338" alt="screen shot 2017-04-20 at 13 36 48" src="https://cloud.githubusercontent.com/assets/973543/25231208/f091f8d4-25ce-11e7-8639-7333ca3f3c9a.png">

There is absolutely no reason to add these... and it is likely a performance hit (processing unknown CSS properties is just another cycle).

There is a boolean way to check whether a property is supported in the browser. See Verou's article http://lea.verou.me/2009/02/check-if-a-css-property-is-supported/.

Possible performance issues aside, it is a nuisance from the DX perspective (as visible in the screenshot).

philpl:
@ThySultan Is there any upside or downside to checking which CSS properties are supported in the runtime?

I don't think we had a discussion about this before, but please link it if I'm wrong :smile:

philpl:
So I've benchmarked this and it turns out it's fortunately not slow, but still, since it's O(n), we wouldn't want it in prod. Also the parsing of CSS outside of Stylis with a middleware would make this slower.

```
property in document.body x 4,233,144 ops/sec ±3.69% (53 runs sampled)
```

The big problem is: If we'd add this to development only, then we risk inconsistencies with prod.

gajus:
> The big problem is: If we'd add this to development only, then we risk inconsistencies with prod.

Whats the reason you would even think of adding this in dev only?

`styled-components` is runtime. It can check which styles to apply at the time of constructing the styles object.

philpl:
Adding this to the runtime will make things slower. It's a question of how much slower, but at the same time we're trying to make v2 faster, so that's a conflict.

As mentioned, checking for a property runs with: `4,233,144 ops/sec ±3.69%`

We'd want to cache this to get rid of unnecessary DOM access and speed things up. Without thinking about setting a property, getting one out of 10,000 runs with: `9,443,351.89 ± 0.74% op/s`

This is most certainly not going to be faster, than the v8 optimised pattern matching and vendoring in stylis. Thus one argument is definitely performance.

The second part is that there's literally no reason to add this to production. The only argument is: It's a nuisance to some. So, that means in production this argument doesn't matter, since you wouldn't look and edit the styles in prod that often. It's really something you can live with there.

[...]

PR #1143, opened 2017-09-04
Proposal: Top-level configuration

geelen:
First up, a bit of backstory. Styled-components has always been a zero-config library, pushing a "chosen" combination of features for simplicity and unification. If you're using SC, there's only one way to do things, which makes it quicker to get going with, makes all documentation relevant, etc.

I'm still committed to that design principle, but there are a few areas were global configuration would be really useful: turning off autoprefixing #719, and enabling "speedy" mode. In both cases, there are legit reasons why you'd want it (writing an electron app, writing an app with 10k rules), but they don't make sense to be the default (shitty browser support, shitty devtools experience). The point being neither of these config options change your _usage_ of the library, just its runtime behaviour. So it wouldn't fragment things the way, say, a configurable PostCSS pipeline would have.

As far as I can tell, there are really only 2 options for configuration in JS—factories and mutation. Factories are better from a software engineering point of view, but the way JS modules work make them suck. As soon as you want to change one config option you have to replace `import styled from 'styled-components'` with `import styled from '../../shared/custom-styled'` throughout your whole codebase. A nicer syntax, though it's much more limited and introduces other problems, would be using context:

```js
import { configure } from 'styled-components'
import App from './App'

export configure({
  speedy: true.
})(App)
```

It looks nice, but I don't think it's viable. Since `styled.div` runs _outside_ of the render tree, anything relying on context has to wait until render. So it limits the kinds of things we can make configurable, since they have to be evaluated in the context of a real React render cycle. But it also kinda raises the question of having config be defined at any point in the render tree, which confuses things a fair bit. I don't think it makes any sense for half the render tree to be using `insertRule` and the rest to not. It certainly would complicate the implementation to try to support that. Much better to be one or the other.

So, if mutation is the other option, there's obviously a bad way to do it:

```js
// ...
styled.SPEEDY = true
```

Again, the problem here is consistency. If in between those two lines you construct all your components, what happens when a config setting changes? What happens if it changes while your app is running?

The way I was trying to do this back before the release of v2 was effectively:

```js
import styled, { configure } from 'styled-components'

// Until styled is called, configure can be called.
configure({ prefix_css: false })

// Now the config is "locked"
styled.div` ... `

// This causes an explosion since SC is already configured.
configure({ speedy: true }) // 💥
```

That's tricky, but possible—we have to find every potentially config-sensitive function inside SC and "lock" the config when it's called. We can do that with a proxy object (not necessarily an ES6 proxy just something that intercepts function calls) but then we're kinda injecting this layer in between us and the users (most of whom don't need to configure anything) which feels wrong.

So, by a process of elimination, we get this:

### Glen's Least Terrible Way of Doing Config in JS That He Can Think Of

This is the usage:

```js
import { configure } from 'styled-components/config'

configure({
  prefix_css: false,
  speedy: true
})
```

This references the file `config.js` which is the only file with any mutability, and it exports two methods `configure` and `getConfiguration`. It's also an ES5 CJS file so it can be consumed as a top-level import above.

The key is that `src/config.js`, which is internal to SC, uses `require('../config.js')` to force Rollup not to inline the module. It calls `config.getConfiguration()` which locks the config. That gets imported by files that need it like `stringifyRules.js`, which is imported from `index.js`, so the final result is that **once you've imported 'styled-components' the config is locked**.

This is a bit dank, I know. Using internal module resolution to change something before some files are loaded doesn't seem on the right side of the clever/stupid divide. But the more I think about it, the more I think it's offers actually a pretty good DX:

* If you don't need to configure anything, literally nothing changes. The objects you import when you use styled components are all exactly as they were—not proxies, not factories, etc.
* When you want to change some config, do this:
  * Create a new file `configuration.js` (can use an existing file as long as it doesn't depend on SC)
  * `import { configure } from 'styled-components/config'` and call `configure` with whatever settings you want to change
  * Import your `configuration.js` _in index.js before importing App.js or anything else that depends on SC` ([example](https://github.com/geelen/sc-webpack2-test/blob/master/src/index.js#L3))

Because we're relying on import order, this is going to be pretty unexpected. But it does match the semantics of modules—they're executed in the order they're imported. And it does let us completely isolate the configuration step from the rest of the app.

I whipped up a test using CRA for both [webpack 1](https://github.com/geelen/sc-webpack1-test) and [2](https://github.com/geelen/sc-webpack2-test) to make sure things were still working even with our es bundle exports, and they are. You can try it out using this PR branch and `yarn link` if you like. Only `prefix_css` is so far implemented, speedy is gonna take a fair bit more work.

So, what do you think? I know it's not great, but I'm fairly confident the alternatives are worse 😉

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file. If a contri
 
 *The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/).*
 
+## Experimental
+
+- Configuration endpoint & ability to disable autoprefixing
+
 ## Unreleased
 
 - Fixed nonce missing from global styles (see [#1088](https://github.com/styled-components/styled-components/pull/1088))
--- a/config.js
+++ b/config.js
@@ -1,2 +1,24 @@
-/* eslint-disable */
-module.exports = require('./lib/config')
+/* Styled Components top level default configuration.
+ *
+ * Allows for some settings to be overridden before SC is
+ * loaded. I.e. you must import 'styled-components/config'
+ * BEFORE importing 'styled-components'
+ *
+ * For more information, see [DOCUMENTATION URL]
+ */
+
+var config = {
+  prefix_css: true,
+}
+
+var config_already_read = false
+exports.configure = function configure(overrides) {
+  if (config_already_read) throw new Error("Can only configure Styled Components before you've used it!")
+  var new_configuration = typeof overrides === 'function' ? overrides(config) : overrides
+  Object.keys(new_configuration).forEach(key => config[key] = new_configuration[key])
+}
+
+exports.getConfiguration = function getConfiguration() {
+  config_already_read = true
+  return config
+}
--- a/config.js
+++ b/config.js
@@ -1,0 +1,2 @@
+/* eslint-disable */
+module.exports = require('./lib/config')
--- a/config.js
+++ b/config.js
@@ -1,3 +1,4 @@
+/* eslint-disable */
 /* Styled Components top level default configuration.
  *
  * Allows for some settings to be overridden before SC is
@@ -7,15 +8,16 @@
  * For more information, see [DOCUMENTATION URL]
  */
 
-var config = {
+const config = {
   prefix_css: true,
 }
 
-var config_already_read = false
+let config_already_read = false
 exports.configure = function configure(overrides) {
   if (config_already_read) throw new Error("Can only configure Styled Components before you've used it!")
-  var new_configuration = typeof overrides === 'function' ? overrides(config) : overrides
-  Object.keys(new_configuration).forEach(key => config[key] = new_configuration[key])
+  Object.keys(overrides).forEach((key) => {
+    config[key] = overrides[key]
+  })
 }
 
 exports.getConfiguration = function getConfiguration() {
--- a/package.json
+++ b/package.json
@@ -10,6 +10,9 @@
     "build": "npm run build:lib && npm run build:dist",
     "prebuild:lib": "rimraf lib/*",
     "build:lib": "babel --out-dir lib --ignore \"*.test.js\" src",
+    "prebuild:es6": "rimraf dist/*",
+    "build:es6": "rollup -c --environment ESBUNDLE",
+    "build:es6:watch": "npm run build:es6 -- -w",
     "prebuild:dist": "rimraf dist/*",
     "build:dist": "rollup -c && rollup -c --environment ESBUNDLE && rollup -c --environment PRODUCTION",
     "build:watch": "npm run build:lib -- --watch",
@@ -122,6 +125,7 @@
     "rollup-plugin-replace": "^1.1.1",
     "rollup-plugin-uglify": "^1.0.1",
     "rollup-plugin-visualizer": "^0.1.5",
+    "rollup-watch": "^4.3.1",
     "tslint": "^4.3.1",
     "typescript": "^2.4.1"
   },
--- a/src/config.js
+++ b/src/config.js
@@ -1,19 +1,9 @@
-console.log('CONFIG FILE BEING LOADED')
+/* This loads the configuration (possibly after it has been changed)
+ * and exports the current values.
+ *
+ * It needs to be calling 'require' otherwise Rollup would inline it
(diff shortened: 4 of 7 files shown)


mxstbr:
**I'm against this change. Adding configuration introduces a lot more complexity to the library and the codebase to cater to tiny subsets of users.**

First, there've been some interesting arguments for turning off autoprefixing in #719, but really most of it boils down to "It's a very slightly nicer development experience because when I use Chromium I need to press two checkboxes and not one". That alone to me is not a good reason for introducing the complexity of configuration, on top of that it's a minor (if vocal) subset of users who complain about this, but it'll never be the breaking point of somebody deciding to not use `styled-components`. (if somebody goes "Oh I'm not going to use this because it autoprefixes in development" it's a strawman argument and they weren't going to use the library anyway)

Second, and more importantly, **we should enable speedy mode in production**, full stop. The only reason we haven't done that historically is because it disables editing rules in Chrome DevTools. As seen in all the tweets and Medium rantposts we've gotten, **a much much bigger subset of users cares about good performance** (in this case meaning using a library that's high up in microbenchmarks, even if those are stupid) than about editing production styles in DevTools. (I've literally not seen a single person say "Thank you" for that)

The performance in stupid microbenchmarks is the main reason why folks are deciding against`styled-components`; will not being able to edit rules in Chrome DevTools make anybody decide to not use `styled-components`? I don't think so. On top of that, with the per-component-classnames we introduced in v2 you can just add a new ruleset a component with a single button click in the DevTools, even in speedy mode, so that whole argument is pretty moot.

Overall I don't see any good reasons for introducing configuration. This is especially true because there's literally (as you said) only a "Least Terrible Way" of doing it that'll likely be quite confusing and we'll get a lot of questions about.

**I vote for not introducing configuration and just doing speedy mode in production.**

philpl:
I suspected I’d become a “tie breaker” here, so let me reflect my thoughts and opinion here, as carefully as possible.

styled-components might be the most popular css-in-is library and have a vibrant community of supporters. However, there’s a lot of growth we still have yet to see.

Libraries like CRA have gained a lot of support based on a simple principle: Sane and developer friendly defaults, weighed against all “nice-to-haves”, without hardly any configuration. I don’t want to go too much into detail, but we should continue to strive for this ideal. And I believe it very much applies to us.

Now, to back up a little again, we’ve gotten close to becoming the defacto go-to solution, but we need more support to actually get there. If we go against our original ideas, we risk alienating our existing supporters without gaining any ground. This is one point to keep in mind, and mostly how I think about these changes.

Introducing speedy mode is the primary point here around which revolves the discussion. But it’s a question that has been answered by all other libraries already. Today AFAIK there’s no other library in our position that has yet to implement it. Most add it without it being configurable or optional. Overall it’s a sane default and if we’d add configurability t should be on by default.

The same goes for autoprefixing, but I’m not going to go into detail for it. SC is predictable and has a single class for each component with styling on it. Why do you need devtools then?

In the past months the most common problem with css-in-is that has been brought up is slow run times. When people bring up a problem with styled-components it’s its slowness. Well, we know that’s not because of our runtime-cost, but the css injection time. But people either don’t know that, don’t care, or want it fixed anyway. Therefore, it’s an extremely important improvement. People who were proposing it to be the default in CRA for instance, ask themselves whether it’s fast enough. And that’s unfortunately a public debate that holds us back.

I don’t think more options need to be added anytime soon. And I don’t think it contributes to the flexibility of the library, but only forgettable improvements for DX.

Everything else about the implementation and about our principles have been said already, so I won’t repeat it.

**TLDR and finally: Overall I’d vote against configurability in styled-components**

geelen:
Yeah cool, I see your points and I'm happy to consider myself outvoted, and simply work towards getting speedy mode shipped in production for all.

It's going to be a little tricky, since SSR will be needing to consuming inlined style tags and converting them into insertRule statements, but poentially we can change the SSR output to make that easier (so we don't have to ship a damn CSS parser just to rehydrate). But it should be doable regardless.

The only thing I will say Max is that "nobody thanking us for not breaking dev tools" doesn't mean that nobody is going to care when we do! The absence of evidence is not the evidence of absence and all that. But I guess we'll see how big a deal it truly is when we launch it.

Anyway, you're both right. Config sucks. 😜

mxstbr:
Yeah that is very true, not a great argument I put out there :thinking: Anyways, can't wait to see speedy mode!

PR #1148, opened 2017-09-05
fix(1127) - Excessive warnings while running tests

### LINKED ISSUE #1127: Excessive warnings while running tests
Running the tests locally results in a lot of warnings since React has recently moved the test utils and renderers to different libraries. This add unnecessary noise while debugging the tests.

Can be fixed as referenced [here](https://github.com/airbnb/enzyme/issues/968#issuecomment-310996577), by removing `react-addons-test-utils` as a a dependency and adding `react-test-renderer` instead

SaraVieira:
closes #1127


geelen:
Wow I have been putting up with these for a long time I wish I'd realised how small the fix was! Thanks for this @SaraVieira 😃

Issue #1110, opened 2017-08-24
css does not filter falsy values


wmertens:
I have a component (Box) that I use _a lot_ in my app, and as such I'd like it to be more optimized.

I replaced the css string with a css object so that I could calculate all values with one function call. 

```js
const Box = styled.div`
	${p => ({
		position: p.relative ? 'relative' : p.absolute ? 'absolute' : false,

		display: p.inline ? 'inline-flex' : 'flex',

		// Size
		width: !(p.fitChildren || p.inline) && `100%`,
        …etc
    })}
```

However, if I don't specify a position, the resulting css has `position: false`. Since the only valid falsy css value is `0`, it would be nice if https://github.com/styled-components/styled-components/blob/master/src/utils/flatten.js#L10 filtered out any falsy values that are not `0`.

Would you accept a PR that does this?

philpl:
Hm, I mean we have the rules for it right below:

https://github.com/styled-components/styled-components/blob/master/src/utils/flatten.js#L20

So it is possible to fix. But I'm kind of stunned by the example. It is pretty contrived, since you could just default to an empty string.

I'm sure anyway that it'd be good practice to fix.

cc @mxstbr

geelen:
Wait, returning `false` should result in an empty string, so you _should_ be getting `position: ;`. Are you saying you get `position: false;` instead? Because that may be an issue with `objToCss` as you mention.

To clarify, the string `false` should never be output, it should be replaced with an empty string. Any other behaviour is a bug.

mxstbr:
That should be a filter, right? We should be doing a `Object.keys(obj).filter(key => key === 0 || Boolean(key))` before the map so we don't even output `position: ;` but skip that rule entirely?

geelen:
Yeah..... I think so! There's... no way an empty value could ever be valid or meaningful, right? Not even if its a media query or nesting or something, if the value is falsey and not 0 you can delete the key. If someone could PR with a test case to cover that would be swell!

reznord:
I'd be happy to take this if it's still free!

philpl:
@reznord Awesome! Feel free to implement the change as suggested by @mxstbr. It should however probably align with the filter logic for false values that flatten is applying. You can probably extract a common function for it and reuse it to filter the object. For the tests, there should already be a test (on my phone& forgot where it is) that tests the object interpolations. (It might just be in the basic test file).

wmertens:
@reznord I implemented it…

[...]

PR #1149, opened 2017-09-06
flatten: ignore falsy non-0 css values

wmertens:
This implements #1110.

Note that the output still includes spurious `;`, I have no idea where they come from, I even searched the code for all `;` :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file. If a contri
 
 ## Unreleased
 
-- N/A
+- Don't emit empty rules like `color: ;` when using objects for CSS. Note that this still happens with template interpolations (see [#1149](https://github.com/styled-components/styled-components/pull/1149))
 
 ## [v2.2.0] - 2017-09-27
 
--- a/src/utils/flatten.js
+++ b/src/utils/flatten.js
@@ -5,13 +5,23 @@ import isPlainObject from 'is-plain-object'
 import type { Interpolation } from '../types'
 
 export const objToCss = (obj: Object, prevKey?: string): string => {
-  const css = Object.keys(obj).map(key => {
-    if (isPlainObject(obj[key])) return objToCss(obj[key], key)
-    return `${hyphenate(key)}: ${obj[key]};`
-  }).join(' ')
-  return prevKey ? `${prevKey} {
+  const css = Object.keys(obj)
+    .filter(key => {
+      const chunk = obj[key]
+      return (
+        chunk !== undefined && chunk !== null && chunk !== false && chunk !== ''
+      )
+    })
+    .map(key => {
+      if (isPlainObject(obj[key])) return objToCss(obj[key], key)
+      return `${hyphenate(key)}: ${obj[key]};`
+    })
+    .join(' ')
+  return prevKey
+    ? `${prevKey} {
   ${css}
-}` : css
+}`
+    : css
 }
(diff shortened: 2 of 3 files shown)


src/utils/flatten.js
```diff
@@ -5,13 +5,18 @@ import isPlainObject from 'is-plain-object'
 import type { Interpolation } from '../types'
 
 export const objToCss = (obj: Object, prevKey?: string): string => {
-  const css = Object.keys(obj).map(key => {
-    if (isPlainObject(obj[key])) return objToCss(obj[key], key)
-    return `${hyphenate(key)}: ${obj[key]};`
-  }).join(' ')
-  return prevKey ? `${prevKey} {
+  const css = Object.keys(obj)
+    .filter(key => obj[key] === 0 || obj[key])
```
philpl: Can you use the same check as the one we’re using inside flatten? It’d be nice to share it even if it shouldn’t make a difference here :)

philpl: @wmertens Bump :wink:

wmertens:
Oops missed that :)

I gave it a quick shot, first factoring out the test to `isInvalidCssValue`
and using it in both places, but then I couldn't commit because the flow
eslint didn't see that the function is filtering out falsy chunks, so now I
just copied the test verbatim. I must say I prefer my `obj[key] === 0 ||
obj[key]` but ok.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file. If a contri
 - Fixed nonce missing from global styles (see [#1088](https://github.com/styled-components/styled-components/pull/1088))
 - Improve component mount and unmount performance with changes to `createBroadcast`. Deprecates usage of `CHANNEL` as a function, will be update to `CHANNEL_NEXT`'s propType in a future version. (see [#1048](https://github.com/styled-components/styled-components/pull/1048))
 - Fixed comments in react-native (see [#1041](https://github.com/styled-components/styled-components/pull/1041))
+- Don't emit empty rules like `color: ;` when using objects for CSS. Note that this still happens with template interpolations.
```
philpl: Can you refer to this PR in the changelog here please? :)

src/utils/flatten.js
```diff
@@ -5,13 +5,26 @@ import isPlainObject from 'is-plain-object'
...
 export const objToCss = (obj: Object, prevKey?: string): string => {
-  const css = Object.keys(obj).map(key => {
-    if (isPlainObject(obj[key])) return objToCss(obj[key], key)
-    return `${hyphenate(key)}: ${obj[key]};`
-  }).join(' ')
-  return prevKey ? `${prevKey} {
+  const css = Object.keys(obj)
+    .filter(key => {
+      const chunk = obj[key]
+      return !(
```
philpl: Can you instead just negate the reference equality? It's a lot easier to miss a negation up here.

wmertens: Good point! Done.

mxstbr:
Ah, I wanted to release this with `2.2.0`, my bad! Mind resolving the CHANGELOG conflict @wmertens? I'll get this out as 2.2.1!

philpl:
@mxstbr We can land a patch together with https://github.com/styled-components/styled-components/pull/1069 potentially?

wmertens:
Rebased :)

mxstbr:
Awesome, thank you Wout!

PR #1150, opened 2017-09-06
Thorough minification

### LINKED ISSUE #1135: Some issues preventing thorough minification
1. There are some instances of `typeof process !== 'undefined'` in the codebase which will defeat DCE in some build tools like webpack. By default, webpack only replaces `process.env.NODE_ENV` in production mode (`-p`) and would not touch a naked use of `process`.

2. `process.env.NODE_ENV` is being replaced by rollup for non-minified builds (e.g. `dist/styled-components.es.js`) which then does not allow webpack to replace it downstream properly and for uglify to DCE it due to the resulting invalid comparison (`"production" !== "production"`)

https://github.com/styled-components/styled-components/blob/master/rollup.config.js#L50-L52

An example of the output in `dist/styled-components.es.js`:

```
var warnTooManyClasses = void 0;
    if (typeof process !== 'undefined' && "development" !== 'production') {
      warnTooManyClasses = createWarnTooManyClasses(displayName);
    }
```

Recommendations:

1. Remove usage of `typeof process !== 'undefined'`
1. Only run the replace plugin on the minified build

bkazi:
Aims to fix #1135 
I found one usage of `typeof process` (should there be more?)
Changed the rollup config to only replace and shim process in production

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file. If a contri
 
 ## Unreleased
 
+- Fixed downstream minification issue with replacing `process` (see [#1150](https://github.com/styled-components/styled-components/pull/1150))
 - Fixed nonce missing from global styles (see [#1088](https://github.com/styled-components/styled-components/pull/1088))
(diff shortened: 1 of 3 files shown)


probablyup:
Yup, looks great!

probablyup:
@bkazi just need to add a CHANGELOG entry :)

bkazi:
Added the changelog entry!

Issue #1164, opened 2017-09-13
Classname reordering when extending styles is inconsistent.


Aleksion:
I'm reopening this since #1083 got closed without the issue being resolve. (or this might be another issue.

 I'm experiencing this 50% of the time, when I deploy my code to production. (On Netlify)

It seems to happen whenever I've used .extend or styled(COMPONENT), but then again, only 50% of the time. This is before and after a refresh:

<img width="214" alt="screen shot 2017-09-06 at 11 36 33" src="https://user-images.githubusercontent.com/3367641/30401319-d8c69a42-98a7-11e7-807c-d7dadc9506e7.png">
<img width="217" alt="screen shot 2017-09-06 at 11 36 49" src="https://user-images.githubusercontent.com/3367641/30401318-d8c33352-98a7-11e7-83ce-b7a6c26fd32d.png">

I've been trying different versions of styled-components and it seems to be an issue all the way back to 2.0.0

philpl:
That’s true! The discussion diverged a lot from the original issue and wasn’t resolved.

We’ll need a stable reproduction first, before tackling this issue apart from a screenshot unfortunately. Can you provide a CodeSandbox or sth similar to reproduce this?

cc @geelen: I think we pinned down the issue to styled(COMPONENT) any idea why it might take higher priority?

Aleksion:
What makes it so hard, is that I haven't been able to replicate it locally. I've been trying in code sandbox, without "luck"

One thing to note is that it doesn't do it on all styles throughout the page. It seems it becomes more of an issue on larger projects.

And I NEVER get it when I run it locally. I only get it when I serve the build online (through netlify's cdn). I don't think how it's served should affect it, but yeah, it only happens in production builds.

It happens for BOTH styled(COMPONENT) and STYLED_COMPONENT.extend

I initially changed all .extend to styled() because I thought it was only an issue with .extend

Aleksion:
Oh, and it's happening on a gatsby project. Might be relevant.

philpl:
@Aleksion:

> It happens for BOTH styled(COMPONENT) and STYLED_COMPONENT.extend

`.extend` combines the parent's rules with the new ones and creates a **single class** that's totally separate from the parent's. Thus it's impossible for it to trigger this bug. It must somehow be triggered by using `styled(StyledComponent)` where the deferred rule injection fails to work.

[...]

PR #1194, opened 2017-09-30
Test for BrowserStyleSheet style rendering order

Kauhsa:
As discussed here: https://github.com/styled-components/styled-components/pull/1102#issuecomment-333295857

The added test fails, which catches a bug that happens when #1102 is used, perhaps related to #1164 as well.

CC: @mxstbr, @philpl, @paul-veevers

### DIFF
--- a/src/models/test/BrowserStyleSheet.test.js
+++ b/src/models/test/BrowserStyleSheet.test.js
@@ -1,0 +1,45 @@
+// @flow
+
+import React from 'react'
+import { mount } from 'enzyme'
+
+import styled from '../../index'
+import BrowserStyleSheet from '../BrowserStyleSheet'
+import StyleSheetManager from '../StyleSheetManager'
+import { expectCSSMatches, resetStyled } from '../../test/utils'
+
+describe('BrowserStyleSheet', () => {
+  it('should render styles in correct order when styled(StyledComponent) and StyleSheetManager is used', () => {
+    const Red = styled.div`
+      color: red;
+    `
+
+    const RedChangedToBlue = styled(Red)`
+      color: blue;
+    `
+
+    const sheet = BrowserStyleSheet.create()
+
+    const App = () =>
+      <StyleSheetManager sheet={sheet}>
+        <RedChangedToBlue>I should be blue</RedChangedToBlue>
+      </StyleSheetManager>
+
+    // $FlowFixMe
+    const attachPoint = document.body.appendChild(document.createElement('div'))
+    mount(<App />, { attachTo: attachPoint })
+
+    // window.getComputedStyles would be perfect, but it seems that JSDOM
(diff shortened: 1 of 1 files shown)


src/models/test/BrowserStyleSheet.test.js
```diff
@@ -0,0 +1,45 @@
...
+    // it.
+    // $FlowFixMe
+    const source = document.documentElement.outerHTML
+
+    // regex in case test is run against minified CSS in the future
+    const indexOfRedStyle = source.search('color: red')
+    const indexOfBlueStyle = source.search('color: blue')
+    expect(indexOfRedStyle).toBeGreaterThanOrEqual(0)
+    expect(indexOfBlueStyle).toBeGreaterThanOrEqual(0)
+    expect(indexOfBlueStyle).toBeGreaterThan(indexOfRedStyle)
```
Kauhsa: Yeah, I really wanted to use `window.getComputedStyled`, but it didn't seem to work. While this is a hacky way to achieve the test, it should determine correctly whether the issue happens or not.

src/models/test/BrowserStyleSheet.test.js
```diff
@@ -0,0 +1,45 @@
...
+    `
+
+    const RedChangedToBlue = styled(Red)`
+      color: blue;
+    `
+
+    const sheet = BrowserStyleSheet.create()
+
+    const App = () =>
+      <StyleSheetManager sheet={sheet}>
```
Kauhsa: So: I'm manually passing `BrowserStyleSheet` to `StyleSheetManager`. This is probably valid way to catch a bug in #1102, since that behaviour is identical to what that pull request essentially does, but I'm unsure whether using `StyleSheetManager` in this manner says anything about #1164.

philpl:
Hiya, thanks a lot for this. I’m not sure why it’s now failing this consistently! Let’s see!

Can you thought try to match the test here which is already testing the same? I hope you can reuse a similar helper for testing that the styles match the Stylesheet: https://github.com/Kauhsa/styled-components/blob/d1d954812031043e99fa555d3c2c9d2c8da26d3b/src/test/basic.test.js#L172

Kauhsa:
Heh, yeah, I saw the helper in the other test but had some issues with it (maybe because of `StyleSheetManager`? idk), but I didn't tinker around for very long. I'll try simplifying it down a bit.

mxstbr:
Hey @Kauhsa, any chance you have some time to fix this? :blush:

Fer0x:
What's the purpose of using `styled.extend(Red)`? Replace with `Red.extend` and this test will pass.

probablyup:
@Kauhsa want to get this rebased against the develop branch (v4) and we'll try to get it in?

probablyup:
Merged the rebased version, so going to close this. Thanks for your efforts!

PR #1201, opened 2017-10-02
Regression: Typings for withComponent don't support extended attributes

beheh:
This is a bug report reported via a failing test, as per the issue template.

Since styled-components 2.2.0 I'm no longer able to use attributes which are only present in the child component after using `withComponent`. Specifically, when extending a `span` to an `a`, I'm not able to use the `href` attribute on the `a` element since Typescript gives me a typings error:

```
typings/tests/with-component-test.tsx(47,27): error TS2339: Property 'href' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<ThemedOuterStyledProps<DetailedHTMLProps...'.
```

This worked fine in 2.1.2, so it must be a regression introduced somewhere between then and 2.2.0.

### DIFF
--- a/typings/tests/with-component-test.tsx
+++ b/typings/tests/with-component-test.tsx
@@ -37,7 +37,13 @@ class Random extends React.Component<any, any> {
 }
 
 const H2 = H1.withComponent("h2");
-const a = H1.withComponent("a");
+const AnchorElement = H1.withComponent("a");
 const abbr = H1.withComponent("abbr");
 
 const RandomHeading = H1.withComponent(Random);
+
+class LinkedHeading extends React.Component<any, any> {
+  render() {
+    return <AnchorElement href="https://example.com">Hello World</AnchorElement>;
+  }
+}


Igorbek:
Thank you for the repro case. I'll try to find where it was introduced. Could you also give what TS version you're using.
If you're not willing to fix it in this pr, I think it should be a new issue, not a pr.

beheh:
Sure! I'm using Typescript 2.5.3. I also just bisected and found the culprit to be 39271809fb927c79e18b00a4d6c5d4c6ac2a53d5.

I might be able to fix it, but I just filed this here because the issue template stated "Think you found a bug? The best bug report is a failing test in the repository as a pull request." and I assumed a contributor would just push ontop of it.

devrelm:
Good catch. We might want to look at what was done in #1139 to support tags passed to the `styled` factory. It shouldn't be much different. I can see what I can do, but it might be good to have someone else take a stab at it as I'm pretty swamped at the moment.

Igorbek:
Hm, here's a dilemma where I am not really sure what's the best.
The issue with `withComponent` that it substitutes the underlying component and breaks all the previous assumptions we had about its props.

Let's assume you have
- component `A extends Component<{a: number; className?: string; }>` and
- component `B extends Component<{b: number; className?: string; }>`

These components are not interchangeable - they have different contracts.
So now if we start styling `A`:
```ts
const StyledA = styled(A)`
  color: ${(props: { color: string; }) => props.color}
`;
```
we made `StyledA` require an additional prop. So that now it accepts `{a: number; className?: string; color: string; }` as props.
Next, we can do this:
```ts
const FirstStyledA = styled(StyledA).attrs({
  a: 1 // we provided explicit value for a, so it is now excluded from props
})``;
```
now, `FirstStyledA` takes `{ className?: string; color: string; }` as props.
So now, all the sudden, we apply `withComponent`:
```ts
const FirstStyledB = FirstStyledA.withComponent(B);
```
So the question is what is the type of `FirstStyledB` (type of the props)?
The correct answer would be `{ b: number; className?: string; color: string; }`.
But we have to infer this type from type of `FirstStyledA` which would require reapplying all the operations we made on top of its first `A` component. Currently TypeScript is not that powerful enough to make it happen. So from practical perspective, we need to choose one of these options:
- drop all the knowledge of original component and reset props to props of the new component
- assume that original component has the same signature with the new one and use collected knowledge

In the example above the options are:
- `{ b: number; className?: string; }`
- `{ className?: string; color: string; }`

Neither is correct. You are allowed either to omit `b` or to omit `color`.

My personal preference here is assuming that original components were of the same contract and leave props unchanged. That is more frequent case from my experience.

What do you think?

devrelm:
@Igorbek I'm seeing the same thing. The basic fix for this PR's issue seems fairly simple, but it definitely turns into a game of whack-a-mole. That's not necessarily a bad thing as long as we're writing tests along the way to define the expected behavior and prevent regressions.

That said, my opinion is that a full rewrite is in order — breaking changes and all — to better map out how props are changed by each method. The existing types have served us well, but they reached their current state through slow iteration and I think that that painted us into a corner.

I can try to get a new one out, working off of the existing tests to start, but it would be better to first have an open issue to discuss all the corner cases and interactions. You already brought up a lot of good points, and it would be good to have a dedicated place to continue discussion.

In the meantime I can open up a PR that at least gets @beheh's test running.

devrelm:
@Igorbek So I had forgotten about that second to last paragraph between reading it last night and responding this morning. You're probably right that the regression is better than the fix. I'm going to open the other PR anyway, and we can discuss it further there.

probablyup:
Can this be closed?

PR #1202, opened 2017-10-02
Fixes #352 Add warning for incorrect using injectGlobal

### LINKED ISSUE #352: Consider adding a warnings for incorrect tagged template literal usage in development

maksugr:
Add warning for incorrect using injectGlobal. More: https://github.com/styled-components/styled-components/issues/352 
Ready for suggestions.

### DIFF
--- a/src/constructors/injectGlobal.js
+++ b/src/constructors/injectGlobal.js
@@ -5,6 +5,10 @@ import type { Interpolation, Stringifier } from '../types'
 
 export default (stringifyRules: Stringifier, css: Function) => {
   const injectGlobal = (strings: Array<string>, ...interpolations: Array<Interpolation>) => {
+    if (process.env.NODE_ENV !== 'production' && !(Array.isArray(strings))) {
+      // eslint-disable-next-line no-console
+      console.warn('Incorrect using injectGlobal as a function. Instead use it as a tagged template literal, see: https://www.styled-components.com/docs/api')
+    }
     const rules = css(strings, ...interpolations)
     const hash = hashStr(JSON.stringify(rules))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ All notable changes to this project will be documented in this file. If a contri
 ## Unreleased
 
 - Don't emit empty rules like `color: ;` when using objects for CSS. Note that this still happens with template interpolations (see [#1149](https://github.com/styled-components/styled-components/pull/1149))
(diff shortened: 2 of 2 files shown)


maksugr:
Excuse me, some problems with this PR? Nobody watched it:)

TrySound:
Just spent some time to debug why this doesn't work
```js
injectGlobal(`
  body {
    font-family: sans-serif;
    margin: 0;
  }
`);
```
This kind of errors will improve DX

src/constructors/injectGlobal.js
```diff
@@ -5,6 +5,10 @@ import type { Interpolation, Stringifier } from '../types'
 
 export default (stringifyRules: Stringifier, css: Function) => {
   const injectGlobal = (strings: Array<string>, ...interpolations: Array<Interpolation>) => {
+    if (process.env.NODE_ENV !== 'production' && !(Array.isArray(strings))) {
+      // eslint-disable-next-line no-console
+      console.warn('Incorrect using injectGlobal as a function. Instead use it as a tagged template literal, see: https://www.styled-components.com/docs/api')
```
TrySound: I think using `console.error` is more correct, since it's not deprecation or something like deoptimization.

TrySound:
@maksugr You can test this case with temporary patching `console.error` before using injectGlobal. Don't forget to reverse

probablyup:
I’d rather see this land in an eslint plugin.

kitten:
@probablyup I agree with you.

@maksugr sorry, this is mostly the reason why I haven’t pushed for this PR or a couple of past ones.

It’s not realistic to add warnings for every single misuse unfortunately. We have already made an effort to add a lot of new warnings that are very explicit but almost none of that refer to misuse of the API :/

Thanks for your effort either way :+1:

PR #1203, opened 2017-10-02
Adding link to awesome-styled-components

romellogood:
Fix for https://github.com/styled-components/awesome-styled-components/issues/13


philpl:
cc @mxstbr

romellogood:
@philpl I agree that your message is a lot more thoughtful and friendlier. I added that in and reverted back to the Built with styled-components header

philpl:
Super! Thanks a lot for taking the initiative here 🙌

Let’s see if Max has some more suggestions

Issue #1204, opened 2017-10-03
Ref warning using withTheme HOC and stateless function components


MatthieuLemoine:
## Version
2.2.0

## Reproduction

Using `styled-components/native`, trying to style a functional component that forwards props and renders a `withTheme` wrapped  functional component will result in a warning about ref.

```javascript
import { withTheme, ThemeProvider } from 'styled-components';
import styled from 'styled-components/native';

const Component = props =>
  <div {...props} style={{ color: props.theme.colors.BLUE }}>Hi</div>;

const ThemedComponent = withTheme(Component);

const WrapperComponent = props => <ThemedComponent {...props} />;

const StyledComponent = styled(WrapperComponent)`
  background-color : blue;
`;

const App = () =>
  <ThemeProvider theme={theme}>
      <StyledComponent />
  </ThemeProvider>;
```

## Expected Behavior

Should not display a warning.

## Actual Behavior

Following warning : 
```
Warning: Stateless function components cannot be given refs. Attempts to access this ref will fail.
```

## Reason

When styling a functional component, an` innerRef` will be passed to this component. When an `innerRef` is forwarded to a `withTheme` wrapped component, this innerRef will be set as a ref on the underlying component.

See : https://github.com/styled-components/styled-components/blob/05c29f504430c9cb0661f3d1a2d699d46aa2dd02/src/hoc/withTheme.js#L54-L68

## Fix

Only set ref in `withTheme` if `Component` is not a stateless function component.

PR #1205, opened 2017-10-03
Fix ref warning when using withTheme HOC and stateless function components

MatthieuLemoine:
See #1204 for context.

I added a check like in `models/StyledNativeComponent.js` for stateless function components and a test case.


MatthieuLemoine:
@mxstbr I added the test case but I can only indirectly test that a ref has not been set by testing for an innerRef instead because no warning is raised when using ref on stateless function components in enzyme.

MatthieuLemoine:
Rebased.

Issue #1130, opened 2017-08-30
withTheme should not break if a default theme is provided


kutyel:
We are migrating a quite big React UI-kit to styled components 🎉 🎉 🎉  but found the following problem, every time we want to do some theming, the code for our components looks like this:

```js
import * as theme from '../Theme'; // this is our fallback theme

const Title = styled.span`
    font-weight: bold;
    color: ${props => props.theme.color || theme.color};
    font-size: {props => props.theme.fontSizeMedium || theme.fontSizeMedium}
`;
```

And the code gets even uglier when you have to adjust the styling to conditional prop values... you can imagine. 

Besides that, when using `withTheme` HOC in components and you don't wrap **everything** in a `ThemeProvider` the component itself breaks.

Is there a better/nicer way to provide a **default ThemeProvider** without having to do all that boilerplate code?

Thanks in advance and awesome work btw! ❤️

philpl:
You can use the `defaultProps` to define fallbacks.

```js
const Title = styled.span`
    font-weight: bold;
    color: ${props => props.theme.color};
    font-size: {props => props.theme.fontSizeMedium}
`;

Title.defaultProps = {
  theme: defaultTheme
}
```

> Besides that, when using withTheme HOC in components and you don't wrap everything in a ThemeProvider the component itself breaks.

It shouldn't actually break but just output a `console.error`:

https://github.com/styled-components/styled-components/blob/master/src/hoc/withTheme.js#L37

mxstbr:
I would recommend you export your own `ThemeProvider` that sets a default theme in case the users don't provide a custom one. Don't even need `.defaultProps` everywhere in that case, but of course if you prefer users are aware of you using `styled-components` using `defaultProps` will work perfectly too!

See this draft documentation we never finished: https://github.com/styled-components/styled-components-experimentation/blob/master/component-libraries/shared-component-libraries.md

kutyel:
I was trying to accomplish that without having to provide an extra `ThemePromider`, anyway, is there a wat to provide such with already a theme obj inside? That way I dont have to export both the first `defaultTheme` and the `ThemeProvider`as well...

mxstbr:
> is there a wat to provide such with already a theme obj inside? 

Sorry I don't quite understand the question?

kutyel:
If I do what the mentioned doc purposes, I end up with something like this:

```jsx
import { ThemeProvider, theme, Button, Link, Title } from 'my-custom-library'

[...]

PR #1214, opened 2017-10-07
Extract theme logic from StyledComponent lifecycle methods

lukebelliveau:
This logic was being repeated in two places, and it looks like we want to place similar logic into the withTheme HOC (see [#1130](https://github.com/styled-components/styled-components/issues/1130).

I'd like some feedback on this before I further generalize it for withTheme.

### DIFF
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -10,6 +10,7 @@ import validAttr from '../utils/validAttr'
 import isTag from '../utils/isTag'
 import isStyledComponent from '../utils/isStyledComponent'
 import getComponentName from '../utils/getComponentName'
+import determineTheme from '../utils/determineTheme'
 import type { RuleSet, Target } from '../types'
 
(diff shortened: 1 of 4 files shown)


src/models/StyledComponent.js
```diff
@@ -96,6 +96,19 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
       }
     }
 
+    getThemeAndGeneratedClassName(props: any, fallbackTheme: any) {
+    // Props should take precedence over ThemeProvider, which should take precedence over
```
philpl: As per your comment in the other thread:

Let’s make this a helper and reuse it across the Native component and withTheme, right?

philpl:
cc @mxstbr

src/utils/determineTheme.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+
+export default (props: any, fallbackTheme: any, defaultProps: any) => {
```
mxstbr: Let's type this a small bit better?

```JS
type Props = {
  theme?: any,
} 

export default (props: Props, fallbackTheme?: any, defaultProps?: Props) => 
```

PR #1215, opened 2017-10-07
Minor change in readme

pravdomil:
npm now saves dependencies by default so --save is no more needed


README.md
```diff
@@ -6,7 +6,7 @@
 Visual primitives for the component age. Use the best bits of ES6 and CSS to style your apps without stress 💅
 
 ```
-npm install --save styled-components
+npm i styled-components
```
philpl: While we’re at it, let’s add sth like

// or
yarn add

PR #1220, opened 2017-10-09
 Prevent withTheme from breaking when a theme is provided via defaultProps

### LINKED ISSUE #1130: withTheme should not break if a default theme is provided

kutyel:
Here is my promised PR 🎉  

Fixes #1130


mxstbr:
Test failure: 

screen shot 2017-10-09 at 11 49 32 am

kutyel:
Ok, everything breaks because `styledContext` is `undefined`, I didn't see that one coming 😅, @philpl recommended workaround?

src/hoc/withTheme.js
```diff
@@ -32,15 +33,14 @@ const wrapWithTheme = (Component: ReactClass<any>) => {
 
     componentWillMount() {
       const styledContext = this.context[CHANNEL_NEXT]
-      if (styledContext === undefined) {
+      if (!styledContext && !this.constructor.defaultProps) {
         // eslint-disable-next-line no-console
-        console.error('[withTheme] Please use ThemeProvider to be able to use withTheme')
-        return
+        console.warn('[withTheme] You are not using a ThemeProvider nor passing a theme in defaultProps')
       }
```
mxstbr: Needs a case here for when `if (!styledContext && this.contructor.defaultProps)` to use the defaultProps or props I think.

kutyel: What should that case do then? Just a `setState` with the theme from `defaultProps`?

mxstbr: Should probably still use `determineTheme` with the props, with undefined as the context theme?

kutyel: Ok, thank you! 👌

mxstbr:
Tests pass! :tada: Now the last thing is to adapt the flow types to make FLow happy 👌

src/hoc/withTheme.js
```diff
@@ -32,17 +33,19 @@ const wrapWithTheme = (Component: ReactClass<any>) => {
...
+      if (!styledContext && !this.constructor.defaultProps) {
         // eslint-disable-next-line no-console
-        console.error('[withTheme] Please use ThemeProvider to be able to use withTheme')
-        return
-      }
-
-
-      const { subscribe } = styledContext
-      this.unsubscribeId = subscribe(theme => {
+        console.warn('[withTheme] You are not using a ThemeProvider nor passing a theme in defaultProps')
```
philpl: This warning has incorrect conditions (defaultProps is set) and an inaccurate message now.

Can you change it to show up when the styles context is not present and the result of the util is undefined?

The message will also have to change to something that indicates that no themeprovider was used, no theme prop was passed, and no default theme prop is present

This will also only be necessary in development mode as a warning now since it’s pretty predictable and deterministic behaviour

src/hoc/withTheme.js
```diff
@@ -32,17 +33,19 @@ const wrapWithTheme = (Component: ReactClass<any>) => {
...
+      } else if (!styledContext && this.contructor.defaultProps) {
+        const theme = determineTheme(this.props, undefined, this.constructor.defaultProps)
```
philpl: There’s no such logic in componentWillReceiveProps which will allow the theme prop to update

kutyel:
Thanks for the feedback @philpl, should look better now 😉👍

src/hoc/withTheme.js
```diff
@@ -31,18 +32,21 @@ const wrapWithTheme = (Component: ReactClass<any>) => {
     unsubscribeId: number = -1
 
     componentWillMount() {
+      const { defaultProps } = this.constructor
       const styledContext = this.context[CHANNEL_NEXT]
-      if (styledContext === undefined) {
+      const defaultTheme = determineTheme(this.props, undefined, defaultProps)
```
philpl: nit: the variable name here is confusing as it should be more like `themeProp`

src/hoc/withTheme.js
```diff
@@ -31,18 +32,21 @@ const wrapWithTheme = (Component: ReactClass<any>) => {
...
+      if (!styledContext && !defaultTheme) {
```
philpl: Can we add `process.env.NODE_ENV !== "production"` here?

cc @mxstbr I don't think that such a message would be valuable outside of prod. This behaviour is so deterministic, that someone not passing anything might be an actual, real use-case sometimes.

src/hoc/withTheme.js
```diff
@@ -31,18 +32,21 @@ const wrapWithTheme = (Component: ReactClass<any>) => {
...
+          this.setState({ theme })
+        })
       }
-
-
-      const { subscribe } = styledContext
-      this.unsubscribeId = subscribe(theme => {
-        this.setState({ theme })
-      })
     }
```
philpl: componentWillReceiveProps hook is still missing :wink:

src/hoc/withTheme.js
```diff
@@ -31,18 +32,21 @@ const wrapWithTheme = (Component: ReactClass<any>) => {
...
+      const { defaultProps } = this.constructor
       const styledContext = this.context[CHANNEL_NEXT]
-      if (styledContext === undefined) {
+      const defaultTheme = determineTheme(this.props, undefined, defaultProps)
+      if (!styledContext && !defaultTheme) {
         // eslint-disable-next-line no-console
-        console.error('[withTheme] Please use ThemeProvider to be able to use withTheme')
-        return
+        console.warn('[withTheme] You are not using a ThemeProvider nor passing a theme prop or a theme in defaultProps')
+      } else if (!styledContext && defaultProps) {
```
philpl: For this `else if` statement and the `if` statement above:

Can you please change it to explicit `!==` and `===` `undefined` please? This would save us some time when optimising this for v8 perf.

kutyel:
Done, @philpl is it OK the `componentWillReceiveProps` now or there is more logic that should go inside? 🤔

kutyel:
@mxstbr @philpl how about this? 🤓

kutyel:
@mxstbr should be OK now, I also took the liberty to add some tests to the `determineTheme` function, to increase some code coverage 😉 👍

kutyel:
@mxstbr @philpl I think this is ready for merge 👍

Issue #1218, opened 2017-10-08
position: sticky not prefixed for OS X Safari


NSLS:
## Version
2.2.1

babel-plugin-styled-components version: 1.2.1

## Reproduction

https://www.webpackbin.com/bins/-KvxLUPUpjAScnEh25lc

## Steps to reproduce

1. Add a position: sticky and top: 0 to your styles
2. Open your projects on safari on OS X
3. See that it doesn't work, because there is no prefix =)

## Expected Behavior

The following css should be generated for position: sticky
```
position: sticky;
position: -webkit-sticky;
```
## Actual Behavior

Only
```
position: sticky;
```
gets generated

thysultan:
Might be an issue of the plugin using/installing an older version of [stylis](https://stylis.js.org/) which added support for `sticky` in [3.2.11](https://github.com/thysultan/stylis.js/blob/master/CHANGELOG.md#3211-august-24-2017)

philpl:
@thysultan stylis only runs in the Babel-plugin with preprocessing enabled, which the reproduction shouldn’t have enabled :)

I’m on my phone so I can’t look into this further

NSLS:
@thysultan npm list stylis shows me the version of 3.3.0, so it shouldn't be the case

NSLS:
> needs more info

What info do you need exactly?

I have provided a reproduction demo and all the environment information.

If there is anything else I should provide you with - please tell me

mxstbr:
Both the main package and the Babel plugin have `"stylis": "^3.2.1"` in their package.jsons. A PR to change both of those to `"stylis": "3.x"` would be appreciated so that you always have the newest version. (/cc @thysultan please don't break our users :wink: )

jaydenseric:
@mxstbr not sure what you mean by that, because the only difference between `^3.2.1` and `3.x` is that `3.x` is also able to install versions older than `3.2.1`. [Both syntaxes install the latest minor release](https://docs.npmjs.com/getting-started/semantic-versioning#semver-for-consumers). You can see it in action [here](https://semver.npmjs.com).

PR #1246, opened 2017-10-16
Update stylis dependency to 3.x

theshortcut:
Updates to stylis 3.3 and should help address issues like https://github.com/styled-components/styled-components/issues/1218 in the future.


src/test/styles.test.js
```diff
@@ -91,7 +91,7 @@ describe('with styles', () => {
       ${rule1}
     `
     shallow(<Comp />)
-    expectCSSMatches('.sc-a {} .b { background-color: blue; } .b:hover { -webkit-text-decoration: underline; text-decoration: underline; }')
```
mxstbr: Any ideas why this doesn't get prefixed anymore? /cc @thysultan

src/test/props.test.js
```diff
@@ -34,6 +34,6 @@ describe('props', () => {
       })};
     `
     shallow(<Comp fg="red" />)
-    expectCSSMatches('.sc-a {} .b { border-width: 0; ; ; }')
```
mxstbr: @wmertens will be very happy to see this fixed! 💯

src/test/styles.test.js
thysultan: Vendor prefixing `text-decoration` was a bug.

mxstbr: Perfect, thank you for confirming.

Issue #1200, opened 2017-10-02
Create a Babel macro


mxstbr:
We should create a [Babel macro](https://github.com/kentcdodds/babel-macros) and have a `macro.js` file at the root of the project. This would let users do the following:

```JS
import styled from 'styled-components/macro';

const Bla = styled.div`
  color: blue;
`
```

The big difference is that the macro version would do all the stuff [our Babel plugin](https://github.com/styled-components/babel-plugin-styled-components) does via `babel-macros`. (I'm unsure if all of it is possible, but it's worth a try)

This would be awesome to get the Babel plugin into `create-react-app`: https://github.com/facebookincubator/create-react-app/issues/2730#issuecomment-329577078

/cc @gaearon @kentcdodds @threepointone

PiereDome:
I am currently working on this. There are a few things that I have discovered. 

- We won't be able to modify the import statement, so we cannot append the /no-parser that happens with the preprocess setting. 
- Also it looks like we would need to import styled-components as well as the macro because the macro is "consumed" during the transpilation of babel-macros

```
import styled from 'styled-components'
import macro from 'styled-components/macro'

const macro.div`
  background: red;
`
```

becomes (with all settings set to true)

```
import styled from 'styled-components';


styled.div.withConfig({
  displayName: 'src',
  componentId: 's5geyt2-0'
})(['\n  background: red;\n']);
```

is this still valuable to proceed working on?

kentcdodds:
> We won't be able to modify the import statement, so we cannot append the /no-parser that happens with the preprocess setting.
> Also it looks like we would need to import styled-components as well as the macro because the macro is "consumed" during the transpilation of babel-macros

You may not be able to modify the import statement (because it'll be removed), but you _can_ add a new import statement by getting the `Program` node and unshifting a new import onto the body of the program :)

So no, people wont have to import both, you can make it work without that :)

PiereDome:
Sweet. Thanks for the tip. I will get that implemented then.

kentcdodds:
I created a macro for glamorous that does what I suggested above. Here it is for your reference 😉
https://github.com/bernard-lin/babel-plugin-glamorous-displayname/blob/master/src/macro.js

probablyup:
Closed via https://github.com/styled-components/styled-components/pull/2032

PR #1256, opened 2017-10-20
babel-plugin-macros #1200

PiereDome:
Prototype for create babel Macro issue #1200 

I added a babel-macro in the /src/macro folder and used the files from the babel-plugin-styled-components library that I needed. Be aware that there is a /* eslint-disable */ on all of the files. I wanted some feedback on this to see if there is anything glaring that I'm missing as I work to get the flow-types in and the linting fixes for the code brought in.


PiereDome:
Ah, yes. That is a great idea. Will get that implemented and should save me from fixing the linting and specifying the flow types

PiereDome:
Just added the changes and have linting and flow passing

PiereDome:
Yeah, I'll work on that this evening

PiereDome:
Built out the tests and started to build out fixtures for each option for the plugin and realized I might of started to get carried away. So only built out ssr and minify as different from default. Let me know if you want more or less tests. Oh and let me know if you'd like me to rebase as well.

PiereDome:
@mxstbr I have built out a simple test that the plugin is doing what it should. Do you want to look and see if there is more you want from the tests

philpl:
@mxstbr I haven’t picked up a review right from the start. Mind having another look?

mxstbr:
This looks good code-wise, just waiting for news from @gaearon in https://github.com/facebookincubator/create-react-app/issues/2730

philpl:
@mxstbr I have a project that’s set up with Babel-macros so I can give it a go later 😬

While we’re here, would it be worth moving the Babel plugin into the same repo eventually? That’d be helping me out a lot for v3 actually, and could ease migration in the future.

mxstbr:
Yeah didn't we plan on using lerna? #588 or something

philpl:
@mxstbr we did originally, but in general I think we could just add “styled-components/Babel” and keep the plugin src in this repo too. I’m not too sure about this but it might be helpful soon

mxstbr:
Hmm, I guess so. Not sure I like that we keep adding `/primitives`, `/native`, `/no-parser` etc. :confused:

package.json
```diff
@@ -78,9 +79,11 @@
     "babel-core": "^6.17.0",
     "babel-eslint": "^7.1.1",
     "babel-loader": "^6.2.10",
+    "babel-macros": "^1.1.1",
```
kentcdodds: Note: `babel-macros` has had a name change to improve compatibility with babel v7. It's now `babel-plugin-macros`. That's the only breaking change. It's currently on `2.0.0`.

probablyup:
Is this still a thing?

probablyup:
Going to close this for now since there hasn't been any movement in a long time.

mxstbr:
Definitely still a thing, `create-react-app` will have built-in support for Babel macros as of `v2` so we should definitely ship this.

TrySound:
How's it going? Can I help? I definitely need this feature right now.

mxstbr:
Anybody want to pick this up? create-react-app v2 is happening soon, so we should add a macro so folks using cra can also use the styled-components Babel plugin!

mareksuscak:
Can you share more details on what remains to be done here? I might take the plunge if there's no one else.

probablyup:
#2032 picks this up and finishes it, so closing

Issue #613, opened 2017-03-23
Global namespace


dapids:
Hi guys!!

First of all thank you very much for `styled-components`. We're massively using it in our company and we're superhappy with it! Great job!!

Lately we were forced to start using a global namespace for our css and we noticed that `styled-components` does not really play well with it.

Here you can find a bin with a practical example: https://www.webpackbin.com/bins/-KfvQIyzL57v9tfNRauJ .

**Actual behaviour:**
The background is `green`.

**Expected behaviour:**
The background is `papayawhip`.

So far we're handling it by adding to each styled component a class `self` in order to style them in this way:
```
&.self {
  background: papayawhip;
}
```

But of course we're not satisfied with this solution.

We thought it would be useful to pass to the `ThemeProvider` a global namespace to use for increasing the `styled-components` selectors specificity.

Do you have any other suggestion/solution for such a case?

thysultan:
maybe `:global()` or increment specificity `&&& {}` is what you're looking for?

dapids:
Hi @thysultan, thanks a lot for your reply. :)

Increasing the specificity with `&&...` is a nice idea to avoid using a class for that purpose. But it still doesn't look neat to us to be forced to wrap each component's style in it.

Regarding `:global()`, I don't quite understand how it could help us. Would you mind providing a small example?

thysultan:
I don't think i understand the problem yet. Is it a specificity issue? or selector issue? 

> Regarding :global(), I don't quite understand how it could help us. Would you mind providing a small example?

since `${A}` and `:global()` together `:global(${A})` could be used to select arbitrary components from outside of themselves i thought that might be it but maybe it's actually a specificity issue.

dapids:
> I don't think i understand the problem yet. Is it a specificity issue? or selector issue?

It is a specificity issue. :)

thysultan:
Does changing the point at which you are loading the external global css help?

dapids:
No. The loading order does not influence the outcome.

Eventually `.namespace div` overrides `.h4sH`. :/

geelen:
Yeah, if you've got specificity problems, the only way is more specificity, sadly. So `&&` is your way out, either that or use `!important` or inline styles 😒 

If you don't want to do one of those, you have to refactor your CSS to not have `.namespace div` rules setting things that the styled components is trying to override.

dapids:
@geelen, thanks a lot for your reply.

[...]

PR #1260, opened 2017-10-23
Add option for include a css namespace

nvanselow:
Fixes: https://github.com/styled-components/styled-components/issues/613

The specificity of the classes applied to styled components are not strong enough to override existing styles when styled components are used in an existing page that might also include styles from other sources (like a legacy stylesheet). This PR works with (PR to babel-plugin) to allow for additional selectors to increase the specificity.

Using the babel-plugin is a convenient way to turn this option on.

``` json
{
  "plugins": [
    ["babel-plugin-styled-components", {
      "namespaceClasses": "moreSpecific"
    }]
  ]
}
```
adds the namespace to all styled-components
``` css
.moreSpecific .mx12n49 {
  background-color: yellow;
}
```

or for a list of classes
``` json
{
  "plugins": [
    ["babel-plugin-styled-components", {
      "namespaceClasses": ["moreSpecific", "verySpecific"]
    }]
  ]
}
```
``` css
.moreSpecific .verySpecific .mx12n49 {
  background-color: yellow;
}
```

### DIFF
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -43,162 +43,174 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
       : componentId
   }
 
-  class BaseStyledComponent extends Component {
-    static target: Target
-    static styledComponentId: string
-    static attrs: Object
-    static componentStyle: Object
-    static warnTooManyClasses: Function
+  const createBaseStyledComponent = (options: Object) => {
+    class BaseStyledComponent extends Component {
+      static target: Target
+      static styledComponentId: string
+      static attrs: Object
+      static componentStyle: Object
+      static warnTooManyClasses: Function
 
-    attrs = {}
-    state = {
-      theme: null,
-      generatedClassName: '',
-    }
-    unsubscribeId: number = -1
-
-    unsubscribeFromContext() {
-      if (this.unsubscribeId !== -1) {
-        this.context[CHANNEL_NEXT].unsubscribe(this.unsubscribeId)
+      attrs = {}
+      state = {
+        theme: null,
+        generatedClassName: '',
       }
-    }
+      unsubscribeId: number = -1
 
-    buildExecutionContext(theme: any, props: any) {
-      const { attrs } = this.constructor
-      const context = { ...props, theme }
-      if (attrs === undefined) {
-        return context
+      unsubscribeFromContext() {
+        if (this.unsubscribeId !== -1) {
+          this.context[CHANNEL_NEXT].unsubscribe(this.unsubscribeId)
+        }
       }
 
-      this.attrs = Object.keys(attrs).reduce((acc, key) => {
-        const attr = attrs[key]
-        // eslint-disable-next-line no-param-reassign
-        acc[key] = typeof attr === 'function' ? attr(context) : attr
-        return acc
-      }, {})
+      buildExecutionContext(theme: any, props: any) {
+        const { attrs } = this.constructor
+        const context = { ...props, theme }
+        if (attrs === undefined) {
+          return context
+        }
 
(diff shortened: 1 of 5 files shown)


src/models/StyledComponent.js
```diff
@@ -43,162 +43,174 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
...
-    state = {
-      theme: null,
-      generatedClassName: '',
-    }
-    unsubscribeId: number = -1
-
-    unsubscribeFromContext() {
-      if (this.unsubscribeId !== -1) {
-        this.context[CHANNEL_NEXT].unsubscribe(this.unsubscribeId)
+  const createBaseStyledComponent = (options: Object) => {
```
nvanselow: The primary change here is wrapping the BaseStyledComponent in a function so that the options can be passed through to `generateAndInjectStyles`. For a clearer diff, I suggest looking at this one with `?w=1`.

mxstbr:
Thanks for the PRs!

As a heads up for transparency, I have no idea if we want to merge a feature like this, so this'll require a discussion and is not something we'll just merge.

Overall I'm not against it, I see the usefulness, though I really wish it didn't require any changes in styled-components itself :confused: How do you feel about that @philpl @geelen?

nvanselow:
Thanks for taking a look @mxstbr! I understand that this one might take some discussion due to the nature of the changes. It was a solution we needed and thought others may find useful as well. I look forward to the input and discussion.

philpl:
@mxstbr I don’t really see the point of adding this as a workaround to an “existing problem” in the users code base.

To clarify, I see how it solves the problem, but I don’t think it’s elegant or the right solution. It doesn’t only increase the “scope” of our code base with another feature, but it introduces something that should just work: correct specificity.

So we can’t increase the specificity of all our classes indefinitely with each more extreme rule that comes from a library. That would require making every class that we generate more specific than *any* of the external rules that are being used. (Correct me if I’m mistaken of course)

So an equivalent change would be to predetermine where our style tag will be injected to make it appear after the external libraries rules, so that rules with the same specificity from SC take precedence.

So that would still require the user to override rules when there’s a highly specific target being used in the external CSS, but since we don’t want to attach unlimited classes for these cases anyway to increase our specificity that is a problem even with this solution.

Also, I’m not comfortable with introducing configuration through the Babel plugin that is “just a runtime thing” technically.

This will be something that might make more sense in v3 when we’ll have a CSS pipeline, but right now it feels out of place and unlike the other Babel plugin features. So in short, it breaks the zero configuration aspect majorly.

nvanselow:
Thanks for the comments @philpl 

I agree with your comment that this solves a problem that is in the original CSS and not due to styled-components. Unfortunately, in many cases, we don't have the benefit of completely refactoring the surrounding CSS when making a gradual upgrade of the code. Allowing for a feature like this helps people who need to gradually introduce or replace pieces of the UI in a legacy codebase. It could allow for easier adoption of the styled-components framework (which is an awesome framework!)

I like what you said in your update. I will see if I can get this working using a separate babel plugin instead of adding the option directly to the existing one. Great idea!

nvanselow:
Thanks for the suggestion about using a separate babel-plugin @philpl. I was able to get it to work that way. It will be much simpler and can be brought in my people having this problem rather than affecting the whole library. I'll make that solution public as soon as I can get everything setup in npm and CI.

rpellerin:
Any updates on this babel plugin @nvanselow ?

nvanselow:
@rpellerin - Everything is ready to go and the plugin has helped us gradually transition our legacy web app over to styled components. Unfortunately, this is our company's first time open-sourcing so we've had to jump through some hurdles to get it out. We have some additional meetings the second week of January which we hope will be the last step.

PR #1280, opened 2017-11-01
Disable static styled in case of HMR

### LINKED ISSUE #1225: [HMR] Styles not updating with v2.2.1
## Version
2.2.1
babel-plugin-styled-components version: N/A

## Reproduction

HMR is broken for styled component styles with upgrade from 2.2.0 to 2.2.1. I believe that the culprit may be #1069. Perhaps we could do a check if the code is running in dev environment and not cache the statics then.

## Steps to reproduce
Run webpack-dev-server with HMR using the react-hot-loader plugin

## Expected Behavior
When I change a style attribute in my styledComponent I would expect the style to change in the browser without a full page refresh. I did test that this is working in 2.2.0.

## Actual Behavior
The style change does not happen until I do a full page refresh

theKashey:
Closes #1225 
Just disables static styles at all for HRM(RHL) and dev mode.
There is no other way :(


src/models/ComponentStyle.js
```diff
@@ -32,6 +32,8 @@ const isStaticRules = (rules: RuleSet, attrs?: Object): boolean => {
   return true
 }
 
+const isHRMEnabled = () => typeof module !== 'undefined' && module.hot && process.env.NODE_ENV !== 'production'
```
philpl: This doesn't have to be a function at all, since these are all static checks

theKashey:
Any updates on this?

philpl:
@schwers are there any negative ramifications by this change? Or actually, do you think there’ll be more edge cases with statically styled components?

mhaagens:
Would appreciate a merge of this PR.

schwers:
@philpl This should completely disable the static caching in dev.  

The only question I have, is should we do the production check at all? It seems like the [rollup config](https://github.com/styled-components/styled-components/blob/master/rollup.config.js#L51) is hardcoding `production`, so I think a developer would have to `npm link` for this to work as is?

theKashey:
@schwers - look like production build will produce only .min file. Any other - is not prop.
As long package.json contains
```js
  "main": "lib/index.js",
  "typings": "typings/styled-components.d.ts",
  "jsnext:main": "dist/styled-components.es.js", <--- Hey, not a prod for webpack!
  "module": "dist/styled-components.es.js",
```
Everything might be ok.

schwers:
@theKashey oh neat, okay, this sounds good

PR #1281, opened 2017-11-01
Fix for strict flag in TypeScript

jbw:
## Reproduction

Dependencies: 
```
 "dependencies": {
    "react": "^16.0.0",
    "react-dom": "^16.0.0",
    "styled-components": "^2.2.1"
  }
}

```
tsconfig.json with strict flag set to true. 
```
{
    "compilerOptions": {
        "outDir": "dist",
        "target": "es6",
        "module": "es2015",
        "declaration": true,
        "sourceMap": true,
        "noImplicitAny": true,
        "strictNullChecks": true,
        "noUnusedLocals": true,
        "removeComments": true,
        "jsx": "react",
        "moduleResolution": "node",
        "strict": true,
        "baseUrl": "."
    },
    "include": [
        "src/**/*"
    ],
    "exclude": [
        "node_modules"
    ]
}
```

Example:
```typescript
import * as React from 'react';

export interface MyCompProps  { 
    className: string;
}

export class MyComp extends React.Component<MyCompProps> {

   constructor(props: MyCompProps) { 
       super(props);
    }

    public render() {
        return (
            <div className={this.props.className}>
               Hello
            </div>
        );
    }
}

const MyCompStyled = styled(MyComp) `
    background: red;
`;

export default MyCompStyled;
```

## Error 

```sh
ERROR in [at-loader] ./src/components/core/heading.tsx:61:37
    TS2345: Argument of type 'typeof MyComp' is not assignable to parameter of type 'StatelessComponent<{}>'.
  Type 'typeof MyComp' provides no match for the signature '(props: { children?: ReactNode; }, context?: any): ReactElement<any> | null'.
```
Same error occurs if I provide the props type e.g: `styled<MyCompProps>(MyComp)`.. 

## Workaround 

The workaround is to cast the component type to an `any`:

```
const MyCompStyled = styled(MyComp as any) `
    background: red;
`;
```
The problem is that the shape of `MyComp` is different to `React.Component` so it does not pass the `strict` guard. This won't occur if you leave out the constructor (implicitly any) or set your constructor props to `any`. 

**Edit: fixed example**

### DIFF
--- a/typings/styled-components.d.ts
+++ b/typings/styled-components.d.ts
@@ -51,7 +51,7 @@ export interface ThemedBaseStyledInterface<T> extends ThemedStyledComponentFacto
   <P, TTag extends keyof JSX.IntrinsicElements>(tag: TTag): ThemedStyledFunction<P, T, P & JSX.IntrinsicElements[TTag]>;
   <P, O>(component: StyledComponentClass<P, T, O>): ThemedStyledFunction<P, T, O>;
   <P extends { theme: T; }>(component: React.ComponentClass<P>): ThemedStyledFunction<P, T, WithOptionalTheme<P, T>>;
-  <P>(component: React.ComponentClass<P>): ThemedStyledFunction<P, T>;
+  <P>(component: React.ComponentClass<P> | React.ComponentClass<T>): ThemedStyledFunction<P, T>;
   <P extends { [prop: string]: any; theme?: T; }>(component: React.StatelessComponent<P>): ThemedStyledFunction<P, T, WithOptionalTheme<P, T>>;
 }
 export type BaseStyledInterface = ThemedBaseStyledInterface<any>;


philpl:
cc @styled-components/typers

typings/styled-components.d.ts
```diff
@@ -51,7 +51,7 @@ export interface ThemedBaseStyledInterface<T> extends ThemedStyledComponentFacto
   <P, TTag extends keyof JSX.IntrinsicElements>(tag: TTag): ThemedStyledFunction<P, T, P & JSX.IntrinsicElements[TTag]>;
   <P, O>(component: StyledComponentClass<P, T, O>): ThemedStyledFunction<P, T, O>;
   <P extends { theme: T; }>(component: React.ComponentClass<P>): ThemedStyledFunction<P, T, WithOptionalTheme<P, T>>;
-  <P>(component: React.ComponentClass<P>): ThemedStyledFunction<P, T>;
+  <P>(component: React.ComponentClass<P> | React.ComponentClass<T>): ThemedStyledFunction<P, T>;
```
Igorbek: hmm... it doesn't seem to be correct. This definition says that we can pass a component that takes props in shape of theme data. It doesn't make sense to me.

jbw: Yeah I agree. I initially went with `any` but following along the interfaces it looked like T props are what the compiler was complaining about.  It looks like React.ComponentClass<P> should be enough for this to match the shape but it doesn't. 
It seems there are some props on T that the compiler is expecting which this union resolves...  

I'm not very familiar with the internals of styled-components but are the styles added as props causing this misshape?

Igorbek: let me look closer to the issue you're trying to solve, I may come up with another pr.

Igorbek:
hm, I could not reproduce your error on the latest build.
Packages used: `typescript@2.6.1`, `@types/react-dom@16.0.2`, `@types/react@16.0.19`, `@types/node@8.0.47`
Strict mode was used.
Please make sure you're using latest `@types/*`

jbw:
@Igorbek thanks, updating to 2.2.3 has fixed this!

Igorbek:
@jbw I tried again on a fresh project, and did work.
See exact files I used: https://gist.github.com/Igorbek/5230426eca53232b0742e9b6bbcc40c0
Changes comparing to your sample:
- `tsconfig.json`: changed `declaration` to `false`, since TS cannot export private names
- `styled-components`: used version 2.2.3

I included `yarn.lock` so you can see exact versions of other packages that were used.
Let me know if it helps you or not.

Igorbek:
Also, make sure your `awesome-typescript-loader` uses the same `typescript` version.

jbw:
You need to have a constructor with props and will fail.

Igorbek:
@jbw I don't understand. I took the example you posted in the beginning. Can you please give exact code that is failing, a reproduction code?

Igorbek:
And BTW, I did comment out these lines and it still worked correctly.

jbw:
You should be able to modify the gist you sent to include a constructor in your component.

```typescript
import * as React from 'react';

export interface MyCompProps  { 
    className: string;
}

export class MyComp extends React.Component<MyCompProps> {
   
   constructor(props:MyCompProps) { 
       super(props);
    }

    public render() {
        return (
            <div className={this.props.className}>
               Hello
            </div>
        );
    }
}

const MyCompStyled = styled(MyComp) `
    background: red;
`;

export default MyCompStyled;
```

Igorbek:
Ok, now I see it. You first posted it with the constructor.
I just explained why that happens here: https://github.com/styled-components/styled-components/issues/1287#issuecomment-341997404
So basically, the issue in `@types/react`.

jbw:
*without. Yes, you're right and I can see your confusion now :)

PR #1283, opened 2017-11-01
Typings for withComponent don't update props type (Closes #1201)

devrelm:
This is a fix for #1201 which is a bug report issue, despite being opened as a PR (for the record, I appreciate this approach.)

**Note:** This does _itself_ contain a regression, as best described by @Igorbek [here](https://github.com/styled-components/styled-components/pull/1201#issuecomment-333936999).

### DIFF
--- a/typings/styled-components.d.ts
+++ b/typings/styled-components.d.ts
@@ -31,7 +31,7 @@ type Attrs<P, A extends Partial<P>, T> = {
 export interface StyledComponentClass<P, T, O = P> extends ComponentClass<ThemedOuterStyledProps<O, T>> {
   extend: ThemedStyledFunction<P, T, O>;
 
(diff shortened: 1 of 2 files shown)


Igorbek:
Hey @devrelm I would say we need first add tests that fix the current behavior when we really don't want to hot-swap all the props and evolve them instead. Something similar to my example from #1201.
I really doubt that `withComponent` should reset all the previous collected knowledge, but let's see how it would play with both cases.

probablyup:
Hey @devrelm, would you have time to implement @Igorbek's suggestion?

devrelm:
@probablyup Unfortunately I can't commit to this. I just got back from paternity leave and am getting back into the swing of things with work and a few projects.

I wish that I had paid more attention, though. I would've loved to have done a rewrite of the type defs before the major-version bump.

PR #1292, opened 2017-11-04
Update to Danger 2.0

orta:
🎉 🌮 

So, this lets Danger host the flowtypes instead of having it inline. And it's Danger 2.wow.


mxstbr:
Thaaaanks!

PR #1296, opened 2017-11-06
deps: upgrade stylis

### LINKED ISSUE #1218: position: sticky not prefixed for OS X Safari

### LINKED ISSUE #1247: Is there any reason to automatically bump up Stylis versions?
Apologies if this has come up before, but I couldn't find anything in old issues.

Currently the Stylis dependency version is set to `"3.x"` in the `package.json`. Is there any reason this isn't a standard caret version number?

We have snapshot tests running in our CI pipeline, and because the CI installs the dependencies from scratch every time we often get a new minor release of Stylis without meaning to, which has caused our snapshots to be mismatched with the local versions quite a few times now. The most recent case was a sudden rogue `-webkit-clip-path` (caused by an inadvertent upgrade to [Stylis 3.3.0](https://github.com/thysultan/stylis.js/releases/tag/v3.3.0)).

Is there any reason we would want minor releases (with potentially breaking changes) to happen automatically?

evan-scott-zocdoc:
Closes #1247 
Fixes #1218 

I think there was something about the "3.x" syntax that was causing this subdependency from being automatically updated externally.

In my own projects, when running `npm update` I have not been getting automatic stylis updates as they are released.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file. If a contri
 
 ## Unreleased
 
-- n/a
+- Bump minimum `stylis` version to 3.4.0, adjust the semver caret target (see [#1296](https://github.com/styled-components/styled-components/pull/1296))
 
 ## [v2.2.3] - 2017-10-29
 
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "is-function": "^1.0.1",
     "is-plain-object": "^2.0.1",
     "prop-types": "^15.5.4",
-    "stylis": "3.x",
+    "stylis": "^3.4.0",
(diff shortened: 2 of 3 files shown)


mxstbr:
Dunno, sure

PR #1313, opened 2017-11-16
Es fix escaping

evan-scott-zocdoc:


src/utils/escape.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+const escapeRegex = /[[\].#*$><+~=|^:(),"'`]/g
```
philpl: Since we’re already here, let’s add a dash to this chat group and add a plus to the end, so we can get rid of the multi dash regex below

philpl:
@evan-scott-zocdoc thanks for picking this up after our quick convo in the other thread :)

src/utils/escape.js
evan-scott-zocdoc: I don't think we can since the dash replacement itself happens as a result of that regex

philpl: We can 😉

.replace(/[-()]+/g, '-')

At least I’m 99% sure... left some things out above since I’m on my phone

evan-scott-zocdoc: hmm ok I'll give it a try

evan-scott-zocdoc:
@philpl should be good now, and I managed to avoid having to bump the bundle size by finding a few things that shouldn't have been in the minified bundle 😄

src/models/StyledComponent.js
```diff
@@ -250,7 +246,7 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
 
         const newComponentId =
           previousComponentId &&
-          `${previousComponentId}-${isTag(tag) ? tag : getComponentName(tag)}`
```
philpl: The tag here doesn’t need escaping :)

evan-scott-zocdoc: Same question - why does this one not need to be escaped? Seems like whenever we are dealing with `displayName`, we should always escape to be safe?

evan-scott-zocdoc:
Looks like `test-results.json` has to be checked in for danger to work I guess... seems prone to failure. I dropped the commit though.

evan-scott-zocdoc:
Side note, updating rollup will net a small bundle size reduction

src/models/StyledComponent.js
philpl: Well the component name here will need escaping but if isTag is true then it’s just an element name

src/utils/escape.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+declare var CSS: Object
+
+const escapeRegex = /[[\].#*$><+~=|^:(),"'`-]+/g
+const dashesAtEnds = /(^-|-$)/g
+
+const CSS_ESCAPE_AVAILABLE = typeof CSS !== 'undefined' && 'escape' in CSS
```
philpl: I’m a bit anxious about adding this now without tests :/ can you add some or remove this for now please?

I’m sure we’ll need it more in the future but we’ll need to polyfill it; so right now this switch could lead to some unexpected consequences

evan-scott-zocdoc: @philpl Would you be ok with me testing by loading this polyfill? https://github.com/mathiasbynens/CSS.escape

evan-scott-zocdoc: Or I can just drop that commit. It'll be a while until browser support is good enough that we'd be able to drop the regex anyways

philpl: Let’s drop it for now :) but please leave a comment with a note there, saying that it’s a good place to try replacing that piece of code with the polyfill in the future

PR #1353, opened 2017-12-15
Update test-results.json

existentialism:
While updating the prettier PR, I noticed test-results needed updating.


philpl:
I'm pretty sure that this file will always keep changing since it contains some time stamps:

> `"startTime":1513352499561`

existentialism:
Ahh... makes sense!

Issue #1355, opened 2017-12-15
Non minified distribution is broken


maciej-ka:
## Version
styled-components: 2.2.4
babel-plugin-styled-components: 1.3.0

## Reproduction

https://www.webpackbin.com/bins/-KeeZCr0xKfutOfOujxN
https://codesandbox.io/s/rkmNRByE4

## Steps to reproduce

1. `cd /tmp`
1. `mkdir project1`
1. `cd project1/`
1. `/tmp/project1`
1. `yarn init -y`
1. `yarn add styled-components`
1. open node_modules/styled-components/dist/styled-components.js

## Expected Behavior
Distribution will not contain `process.env.NODE_ENV`

## Actual Behavior
Several `process.env.NODE_ENV` statements

mxstbr:
That is a UMD bundle that's meant to be used in a `<script>` tag, so it should definitely not contain any `process.env.X` code! (contrary to the `.es.js` bundle)

maciej-ka:
The problem is also present, when I build project:

1. `git clone https://github.com/styled-components/styled-components`
1. `cd styled-components`
1. `yarn && yarn build && yarn dev`
1. visit `localhost:3000`

Actual: browser has an error
<img width="337" alt="screen shot 2017-12-15 at 20 30 30" src="https://user-images.githubusercontent.com/5403694/34057345-d8fa198a-e1d6-11e7-9c7a-c434ff16bdcf.png">

maciej-ka:
rollup.config.js:52 should replace process.env.NODE_ENV
```js
  prod && replace({
    'process.env.NODE_ENV': JSON.stringify(prod ? 'production' : 'development'),
  }),
```

Invocation of replace looks ok to me, and condition `prod` is true when running `node_modules/.bin/rollup -c --environment PRODUCTION`.

I found out that we could have newer version in package.json:124
```js
    "rollup-plugin-replace": "^1.1.1",
```
however when using version 2.0.0 of this plugin problem is still present.

mxstbr:
Ohh it's because the `styled-components.js` bundle is generated with `rollup -c --environment DEVELOPMENT` I think??

We should have that replace always in there, unless it's the `.es.js` bundle!

maciej-ka:
Right, yes, how could I missed that, PR coming in a minute, thanks!

PR #1356, opened 2017-12-15
fix: allow process.env only in es bundle

maciej-ka:
Closes: https://github.com/styled-components/styled-components/issues/1355


maciej-ka:
ready

maciej-ka:
🎉 totally ready now 😄

PR #1362, opened 2017-12-17
Flat bundles for each possible entry

### LINKED ISSUE #1359: Ship flat bundles instead of a lib folder
Based on [this blogpost by the React core team](https://reactjs.org/blog/2017/12/15/improving-the-repository-infrastructure.html#compiling-flat-bundles) we should ship flat bundles instead of the `lib` folder we have right now.  Essentially, rather than shipping tons of files we'd only ship two bundles that contain the CommonJS code.

This will make it possible for us to experiment with Google Closure Compiler, ref #1360.

Andarist:
fixes #1359

- [x] fix tests (dunno why some tests are failing atm)


mxstbr:
screen shot 2017-12-17 at 15 43 52

mxstbr:
Danger works again :tada: Tests are still failing though :cry:

Andarist:
Yeah, seems like some problem with jest not hooking into require for whatever reason. Need to dig into this later.

rollup.config.js
```diff
@@ -47,37 +86,30 @@ const plugins = [
...
+    }),
+  babel(),
 ].filter(Boolean)
 
-if (prod) plugins.push(uglify(), visualizer({ filename: './bundle-stats.html' }))
+if (PRODUCTION) {
+  plugins.push(uglify(), visualizer({ filename: './bundle-stats.html' }))
+}
 
 export default {
```
philpl: Just a quick note, you can use an array here to avoid bloating the package scripts 😬

Not sure if you explicitly decided against it

Andarist: just went with the approach already used in the repo, can refactor to array later

Andarist:
Ok, I think I'v fixed the tests, let's see what CI has to say about that 😉

Andarist:
@philpl @mxstbr the PR is ready for code review 🎉

rollup.config.js
```diff
@@ -47,37 +86,30 @@ const plugins = [
...
-  exports: 'named',
-  targets,
+  input,
+  external: ['react', 'react-native', 'react-primitives'].concat(
+    !UMD ? Object.keys(pkg.dependencies) : [],
+  ),
+  output,
   plugins,
   globals: { react: 'React' },
+  outro: BABEL_ENV === 'cjs' ? "module.exports = exports['default']" : '',
```
mxstbr: Isn't this technically a breaking change? Did people using the CJS bundles have to do `const styled = require('styled-components').default` before but not now?

Andarist: Both flavours are supported now:
- `const styled = require('styled-components').default`
- `const styled = require('styled-components')`

Adding the support for latter is not a breaking change, but rather an 'enhancement'.

But anyway `babel-plugin-add-module-exports` was already used on master on cjs target - see [here](https://github.com/styled-components/styled-components/blob/470be50c1ac0138c1ff60a27aa1021db6c5ca56b/.babelrc#L8). So it would be a breaking change to remove that.

It is somewhat a breaking change to remove `lib` directory. People could import things from 'internal' `lib` directory before this PR, probably not many did this, but that's a question for you if you consider removing this a breaking change. IMHO it shouldnt be treated as such, lib structure is in my eyes an implementation detail and reaching to it should be considered a bad practice.

maciej-ka:
`yarn dev` is failing

rollup.config.js
mxstbr: Ohh I see, that makes sense! Thanks for the explanation.

Yeah definitely a bad practice, I would definitely not call that a breaking change.

package.json
```diff
@@ -2,17 +2,22 @@
...
+    "build": "run-p build:**",
+    "prebuild": "rimraf dist",
+    "build:es": "cross-env ESBUNDLE=true rollup -c",
+    "build:cjs": "cross-env BABEL_ENV=cjs CJSBUNDLE=true rollup -c",
+    "build:umd:prod": "cross-env UMD=true PRODUCTION=true rollup -c",
+    "build:umd:dev": "cross-env UMD=true rollup -c",
+    "build:native": "cross-env BABEL_ENV=cjs NATIVEBUNDLE=true rollup -c",
+    "build:primitives": "cross-env BABEL_ENV=cjs PRIMITIVESBUNDLE=true rollup -c",
+    "build:no-parser:es": "cross-env NOPARSERBUNDLE=true rollup -c",
+    "build:no-parser:cjs": "cross-env BABEL_ENV=cjs NOPARSERBUNDLE=true rollup -c",
```
maciej-ka: Maybe let's move details to a new `scripts/build.js` and here provide only one task that we could call: 
`yarn build` to build all
`yarn build native` to build selected (passed by args)

philpl: That's a great idea! it'd enable us to do both a general build, or a granular one 👍

no-parser/package.json
```diff
@@ -0,0 +1,7 @@
+{
```
philpl: We could replace this with `no-parser/index.js` similarly to `native` as it's intended to mirror the same structure

package.json
Andarist: are granular builds really that useful and needed here? personally i liked the proposal of stuffing everything into a single rollup config (arrayed etc) - i guess though that with that in place it shouldnt be hard to introduce a possibility of granular builds by using some env variable like `BUILD_ONLY=granular_type`

maciej-ka: My suggestion is about moving implementation details out of `package.json` to keep it readable.

no-parser/package.json
Andarist: actually we cannot do this, i've discussed this briefly yesterday with @mxstbr - the problem is that `no-parser` entry is used in combination with babel plugin, but when used with module aware bundler (webpack2+, rollup) the import rewriting which is done by the babel plugin points to this `no-parser` which is in commonjs format, so we've traded es modules for cjs even if using module aware bundlers

this way `styled-components/no-parser` will point correctly to appropriate module format and will be resolved with 'regular rules' based on what type of bundler you use

when I think about it now, `styled-components/primitives` should get similar `package.json` instead of proxying file as it is supposed to be runnable in the browser too, right? 🤔

Andarist:
@maciej-ka good catch, gonna fix it when refactoring the scripts

no-parser/package.json
mxstbr: Yep true, that'd be great!

Andarist:
Ready for another round of code review. I've decided not to introduce possibility of granular builds because it would complex the code quite a bit for imho questionable gains. Let me know what do you think about it.

rollup.config.js
```diff
@@ -47,37 +40,94 @@ const plugins = [
...
+  output: {
+    ...umdConfig.output,
+    file: 'dist/styled-components.min.js',
+  },
+  plugins: umdConfig.plugins.concat(
+    replace({
+      'process.env.NODE_ENV': JSON.stringify('production'),
+    }),
+    inject({
+      process: processShim,
```
mxstbr: Why do we not need the process shim for the dev umd bundle?

Andarist: have no idea why it is here at all for any bundle type - i just have left this untouched, it was already there before my PR

mxstbr: Let's try creating a bundle with and without it and look at the diff? If there's none we should just get rid of it. :grin:

maciej-ka:
:heart: how you improved readability of packages.json and rollup config. I should find a moment for an indepth review tonight.

Andarist:
@mxstbr I've checked outputs with and without process shimming - there was no difference. I've removed virtual module resolver few lines above, it also doesn't seem to have any impact. `process.env` should be replaced in umd bundles (like it is) and `process` generally shouldnt be used for anything else, so it shouldn't appear nowhere in the bundles.

Still unsure why this was configured though in the first place. I see you changed policy of replacing the `process.env` just few days ago - this PR will change that **again**, but I believe it will change it to the more correct usage:
- umd bundles should have it replaced
- any other bundle, no matter if its es, cjs, targeting react-native or primitives should have it untouched and replacing should be done by the consumer's bundle

mxstbr:
Yep, that's the way it was with the last change too except now we have more bundles where it should be left untouched! This LGTM, can we rebase on `master`? Then I'll publish an alpha release so people can try this.

Andarist:
Ah, yeah, forgot that `lib` covered most of those builds before. Rebase coming in a minute

Andarist:
Done, no conflicts so it was even quicker ;) Feel free to squash my commits or whatever, im unsure what flow do you like here. Going out for a Christmas party, so won't be able to tweak this in any way myself till tomorrow.

mxstbr:
Enjoy the party!!

no-parser/package.json
philpl: @mxstbr we could also drop the preprocess/no-parser stuff for the next major release as that was just a proof of concept, but we'd add it back eventually 😅but that might give us some time to think about a better way to implement it

mxstbr: Wait, why? This works perfectly fine now, exactly like it did before except even better in that it resolves to the ES Modules build (so it's treeshakeable) if you bundle with webpack.

Why would we remove this?

mxstbr:
Gotta merge #1371 into master and then rebase this again first though, otherwise the alpha's going to be broken haha :blush:

mxstbr:
Rebased on top of master with #1371, going to publish an alpha 👍

Andarist:
You have push rights to my fork, if you want to rebase yourself

mxstbr:
Done, I've published a first version of this as `styled-components@alpha`! (v2.3.3-0) Going to write a quick post in our community about this so folks can start testing and give feedback whether it works.

mxstbr:
https://spectrum.chat/thread/185e0225-20c8-4e21-a12a-1113101bfcc9

Andarist:
👋 do you want to test this some more or are we ready to merge this?

probablyup:
Don't merge this just yet - might have found an SSR bug. Trying to get my debugging environment to work to verify it >_<

probablyup:
Hmm... not sure why but my server-side build via webpack doesn't seem to be able to process the named exports from `styled-components.cjs.js`. It errors out and when I inspect the webpack require object, only the default export is visible.

Andarist:
I think I know what's going on. Need to check what exact output was produced before the PR, but it's past midnight now, so need to investigate this 2morrow.

Thanks for checking this! I've not verified cjs output good enough - I also thought that this repo is having tests against compiled output.

Andarist:
Ok, lol, made a quick investigation now 😅 Should be fixed now.

probablyup:
@mxstbr want to cut another alpha?

mxstbr:
@probablyup done, published v2.5.0-0 as `styled-components@alpha` with the latest changes!

>  I also thought that this repo is having tests against compiled output.

We should probably run our tests this way...

mxstbr:
Oops, mispublished. Had to undo v2.5.0-0, and republished `alpha` as v2.5.0-1. Test `styled-components@alpha`!

probablyup:
Going to give this another test drive at work tomorrow on one of our larger projects. If all looks good, will merge it 👍

probablyup:
Excellent work @Andarist, thank you!

Andarist:
Not a problem, for some weird reason Im really enjoying tweaking build infra and I've just hated it not so long time ago 😅

PR #1365, opened 2017-12-18
reduce library size

probablyup:
found a bunch of optimizations that bring the bundle size down from `~14.52kB` to `~13.93kB` gzipped


probablyup:
I tried figuring out how to make flow happy but couldn't crack it -_-

src/constructors/styled.js
```diff
@@ -1,14 +1,14 @@
...
 
-  // Shorthands for all valid HTML Elements
-  domElements.forEach(domElement => {
-    styled[domElement] = styled(domElement)
-  })
+  const shorthand = {
+    get: (target, name) => (name in target ? target[name] : baseStyled(name)),
+  }
+
+  const styled = new Proxy(baseStyled, shorthand)
```
mxstbr: Unfortunately we can't do this I think, we have to support IE9+ which don't have `Proxy`?

philpl: nope, we cannot do this. We discussed it before and there's no way to get a polyfill for it. For quite obvious reasons we cannot get rid of the known list and at the same time polyfill Proxies.

src/models/ServerStyleSheet.js
```diff
@@ -24,7 +24,7 @@ class ServerTag implements Tag {
   }
 
   addComponent(componentId: string) {
-    if (this.components[componentId]) {
+    if (process.env.NODE_ENV !== 'production' && this.components[componentId]) {
       throw new Error(`Trying to add Component '${componentId}' twice!`)
```
philpl: I'd rather like us to use `invariant` and use the babel plugin to transpile away the error message in prod. Makes sense? :smile:

Otherwise removing the errors would pose a little bit of confusion if someone is currently relying on these safeguards in prod (even if no one should)

src/models/ThemeProvider.js
```diff
@@ -24,12 +23,18 @@ type ThemeProviderProps = {|
...
 
-const warnChannelDeprecated = once(() => {
-  // eslint-disable-next-line no-console
-  console.error(
-    `Warning: Usage of \`context.${CHANNEL}\` as a function is deprecated. It will be replaced with the object on \`.context.${CHANNEL_NEXT}\` in a future version.`,
-  )
-})
+let warnChannelDeprecated
+if (process.env.NODE_ENV !== 'production') {
+  warnChannelDeprecated = once(() => {
```
philpl: Makes sense to wrap it here :+1: awesome

philpl:
@probablyup so overall, we cannot add the `Proxy` change and must keep our list of valid tags for now. Regarding the other optimisations, they look very good, but maybe we should consider to use `invariant` for the errors, since there's a babel plugin to remove the messages

src/models/ServerStyleSheet.js
mxstbr: > I'd rather like us to use invariant and use the babel plugin to transpile away the error message in prod.

Really? I hate `invariant`, I still don't understand how it works 🙄 So unintuitive. Imo this is pretty clear and if it works, it works?

> Otherwise removing the errors would pose a little bit of confusion if someone is currently relying on these safeguards in prod

We never make any guarantees about those, so I'm don't think that's an argument for not removing them. 99.99999% of our users will benefits, the two that check for this in their tests aren't very smart :wink: Same with #1362, somebody could be using the `lib/` dir directly. Do we care about them? Not really.

probablyup:
Dropped the ES6 proxy refactor. Still have some decent savings though!

mxstbr:
I don't understand why `bundlesize` isn't passing here. The build log says "No env variable provided" but I literally have the env variable in the Travis config:

screen shot 2017-12-18 at 19 06 30

@siddharthkp what's up with that?

siddharthkp:
What is this voodoo magic

Even though the settings has it, I only see one exported variable in the logs 🤔 : [travisCI log](https://travis-ci.org/styled-components/styled-components/jobs/318188583#L444)

How it should look like: [some other repo's logs](https://travis-ci.org/siddharthkp/bundlesize/builds/288293892#L432)

mxstbr:
Wtf that doesn't make any sense.

mxstbr:
Ohhh, Travis doesn't use secure tokens for builds of forks, otherwise people could extract them with `echo $BUNDLESIZE_GITHUB_TOKEN`! :confused: That sucks.

siddharthkp:
Whoa! That makes a lot of sense, but also annoys me at the same time. I'm surprised we haven't had an issue reported for this yet.

Issue #1372, opened 2017-12-19
Prettier pre-commit hook


lukebelliveau:
Now that [we're using prettier](https://github.com/styled-components/styled-components/issues/589) to keep our formatting consistent, is there interest in a pre-commit hook to ensure all commits are Prettified? I've had great success with this in my day job, but I'm not sure how common this is on large open-source projects.

If this _is_ something we want, which direction would we want to go in? There are [a few options](https://prettier.io/docs/en/precommit.html) on the Prettier docs. I've always used the `lint-staged` with `husky` solution, but I'd like to hear thoughts on this.

I'm happy to pick this up myself if we want it implemented.

mxstbr:
Yes yes yes yes! Let's go `lint-staged` with `husky`.

philpl:
This should already work since we’re using the prettier eslint plugin :) but I’m all for switching to husky

lukebelliveau:
Cool, I'll try to get around to this tonight. I'll post back before I get started, so if anybody wants to pick it up beforehand, feel free!

existentialism:
@lukebelliveau really we're just talking about switching from `pre-commit` -> `husky` yeah?

As @philpl already mentioned It's already running `eslint --fix` (and prettier via `eslint-plugin-prettier`) via lint-staged?

glennreyes:
> @lukebelliveau really we're just talking about switching from pre-commit -> husky yeah?

And prettier, which would be a concurrent linting step within lint-staged. If we move prettier over to lint-staged, we can add `eslint-config-prettier` and remove the eslint plugin.

lukebelliveau:
I think I'm almost done here. I'm under the impression we want to run this on the _whole_ repo, based on [this comment](https://github.com/styled-components/styled-components/pull/593#issuecomment-351965625). @mxstbr, can you confirm?

I'm also having some issues between `prettier` and the `flow-typed` definitions:
```es6
flow-typed/lodash_v4.x.x.js
[error] flow-typed/lodash_v4.x.x.js: SyntaxError: default is a reserved word (444:55)
[error]   442 |     conforms(source: Object): Function;
[error]   443 |     constant<T>(value: T): () => T;
[error] > 444 |     defaultTo<T1:string|boolean|Object,T2>(value: T1, default: T2): T1;
[error]       |                                                       ^
[error]   445 |     // NaN is a number instead of its own type, otherwise it would behave like null/void
[...]

PR #1376, opened 2017-12-20
Add pre-commit hook for prettier and eslint --fix

lukebelliveau:
**tl;dr:** All commits must pass `yarn prettier` and `yarn eslint --fix` in the `src/` folder.

This PR adds a `format` script to run `prettier` and `eslint --fix` on the `src/` folder, and uses `lint-staged` and `husky` to run it before each commit.  It also switches the `eslint` config to use `eslint-config-prettier` rather than `eslint-plugin-prettier`, now that `prettier` is run on each commit with this merge and no longer needs a check by `eslint`. Discussed in [#1372](https://github.com/styled-components/styled-components/issues/1372).

I intentionally circumvented the pre-commit hook here because I anticipate feedback on the configuration or targets. Before this is merged, we should push another commit to run `yarn format` on the code.


.eslintrc
```diff
@@ -1,7 +1,10 @@
 {
   "parser": "babel-eslint",
-  "extends": "airbnb",
-  "plugins": ["flowtype", "flowtype-errors", "prettier"],
+  "extends": [
+    "airbnb",
+    "prettier"
```
philpl: Several things here; this is a nice change, but we should use the following config too to ensure smooth compatibility with flow: `"prettier/flowtype"`

Please do not remove the prettier plugin (see below)

package.json
```diff
@@ -24,9 +25,11 @@
     "test:size": "bundlesize",
     "flow": "flow check",
     "flow:watch": "flow-watch",
+    "format": "yarn run prettier && eslint src --fix",
```
philpl: It's a big advantage to run prettier as part of the eslint plugin for multiple reasons:

- Integration of the two
- Avoid duplicated/staggered errors
- Avoid confusing order of prettier & eslint (here eslint runs after prettier, so there's still a chance of conflicts if we change anything in the eslint config)
- Integration into editors that have eslint integration but not prettier integration
- And a big one: get all messages in the same formatted way

lukebelliveau: Thanks for the feedback! To ensure I understand, we want to:

- Re-add `eslint-plugin-prettier`, exactly as was implemented by @existentialism 
- Set the pre-commit hook to run `eslint src --fix`. With `eslint-plugin-prettier`, this script will automatically run `prettier`
- Add `eslint-plugin-flowtype` to .eslintrc

Is this correct?

philpl: > Add eslint-plugin-flowtype to .eslintrc

The plugins should be there, rather we'll want to add the eslint-config-prettier/flow config. There's more info on the available configs here: https://github.com/prettier/eslint-config-prettier

> If so - we still need to keep eslint-config-prettier, right?

That'd be great, yea :+1:

.eslintrc
```diff
@@ -2,9 +2,10 @@
   "parser": "babel-eslint",
   "extends": [
     "airbnb",
-    "prettier"
+    "prettier",
+    "prettier/flowtype"
```
philpl: 👌

existentialism:
Can drop `pre-commit` dep and the pre-commit key from package.json since we're replacing it with husky?

lukebelliveau:
Sorry, I'm new to this collaborator thing :). Looks like my push addressing @existentialism's suggestion automatically dismissed @philpl's review - if this was bad etiquette, I'd really appreciate feedback on how to do this right!

package.json
```diff
@@ -145,11 +148,10 @@
   },
   "lint-staged": {
     "*.js": [
-      "eslint --fix",
+      "format",
```
glennreyes: Will this format/lint the whole codebase before each commit?
I think what we rather want here is to replace this with
```
"prettier --write",
"eslint --fix"
```
... so that only staged files are formatted.

.eslintrc
```diff
@@ -1,6 +1,10 @@
 {
   "parser": "babel-eslint",
-  "extends": "airbnb",
+  "extends": [
+    "airbnb",
+    "prettier",
+    "prettier/flowtype"
+  ],
   "plugins": ["flowtype", "flowtype-errors", "prettier"],
```
glennreyes: IIRC, we can ditch prettier plugin (which runs it as eslint rule) if we use prettier as standalone?

lukebelliveau: That's correct, but it's preferred to run it as part of `eslint` per @philpl's comment [here](https://github.com/styled-components/styled-components/pull/1376#discussion_r157919895).

package.json
lukebelliveau: Hmm.. I just implemented this change locally, and now neither solution is running the lint-staged scripts on commit. Here's what I'm doing:

1. Make a change in a file in the `src` folder that's nonconformant to our prettier rules (like adding an extra tab or a really long line).
2. `git add src && git commit -m "This should run the 'format' command"`
3. I see the files immediately committed, and nothing from `husky`.

I've clearly gone mad, so I'm gonna go to bed :P. Is anybody else seeing this behavior?

package.json
```diff
@@ -145,11 +148,10 @@
   },
   "lint-staged": {
     "*.js": [
-      "eslint --fix",
```
philpl: Can you revert this to just `eslint --fix` please? Otherwise lint-staged is not going to be able to pass the files to eslint, since the format script is already carrying some filenames.

package.json
```diff
@@ -24,9 +24,11 @@
     "test:size": "bundlesize",
     "flow": "flow check",
     "flow:watch": "flow-watch",
+    "format": "eslint src --fix",
```
philpl: Can you change `src` here to `./**/*.js` so that it targets the same files that lint-staged is targeting please?

lukebelliveau:
@philpl I pushed the requested changes in case I'm just having an issue with my machine, but somebody should probably manually verify this actually works before approving - [the hook doesn't actually seem to be working now](https://github.com/styled-components/styled-components/pull/1376#discussion_r157974126).

This _was_ working for me last night, I'm feeling loopy over this! I'll keep trying to chase it down.

philpl:
@lukebelliveau sorry for the (festive) delay 😅 can you run “eslint —fix” on the entire src folder please? Just to be sure

lukebelliveau:
@philpl No worries, happy holidays :smiley::christmas_tree:!

The linter was run before my last commit. I just double checked and there were no changes for `eslint --fix` (or `prettier`, for that matter). Am I missing something?

philpl:
@lukebelliveau no worries, that seems alright 👌 it would also not be bad if we missed a file, since the next person editing it would trigger the hook again on it

lukebelliveau:
Oops @philpl, I totally forgot to add this to CHANGELOG.md. I [pushed a change](https://github.com/styled-components/styled-components/commit/1ac97e343d8e631a79405f33fc4f9d023c72bb65) just now, but want to make sure I'm going about this the right way before opening another PR. Shall I open another?

philpl:
@lukebelliveau sure, let’s just open another PR; sorry for the trouble 😅

lukebelliveau:
Cool, done [here](https://github.com/styled-components/styled-components/pull/1393/). Sorry about that!!

PR #1377, opened 2017-12-20
add back the validAttr regex (corrected)

probablyup:
sorry about that @mxstbr


probablyup:
Happy to add more unit tests too if you can think of any specific use cases

probablyup:
@mxstbr @philpl noticed something while I was fixing this... it looks like `.attrs` doesn't make use of validAttrs. Is that by design? If we're passing through props to a styled component it should probably use the validAttrs filter I think.

PR #1381, opened 2017-12-20
use a simpler hashing strategy when possible 

probablyup:
the execution order of files can differ between build environments
and lead to mismatching hashes. if the babel plugin is being used, we
will usually have a displayName, which will be unique enough for hashing
purposes and we can fall back to the numerical way of doing things as
a fallback

I tested this out in the sandbox and it didn't appear to break anything.

Related: https://github.com/styled-components/babel-plugin-styled-components/issues/105

### DIFF
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -23,17 +23,30 @@ import ServerStyleSheet from './ServerStyleSheet'
 const STATIC_EXECUTION_CONTEXT = {}
 
 export default (ComponentStyle: Function, constructWithOptions: Function) => {
-  /* We depend on components having unique IDs */
   const identifiers = {}
+
+  /* We depend on components having unique IDs */
   const generateId = (_displayName: string, parentComponentId: string) => {
     const displayName =
       typeof _displayName !== 'string' ? 'sc' : escape(_displayName)
 
-    const nr = (identifiers[displayName] || 0) + 1
-    identifiers[displayName] = nr
+    let componentId
 
-    const hash = ComponentStyle.generateName(displayName + nr)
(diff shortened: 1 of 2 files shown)


probablyup:
@mxstbr updated

src/models/StyledComponent.js
```diff
@@ -23,17 +23,30 @@ import ServerStyleSheet from './ServerStyleSheet'
...
 
-    const nr = (identifiers[displayName] || 0) + 1
-    identifiers[displayName] = nr
+    let componentId
+
+    /**
+     * only fall back to hashing the component injection order if
+     * a proper displayName isn't provided by the babel plugin
+     */
+    if (!_displayName) {
```
mxstbr: Why are we using `_displayName` vs. `displayName` in the line below? Can we clean this up?

probablyup: Because `displayName` gets normalized to the string `"sc"` which won't be falsy.

probablyup: I guess I could do `if (displayName === 'sc')` but it's possible that a user could actually name something `"sc"` right?

mxstbr:
Okay so I looked at this on desktop and I got it now. Together with this change (https://github.com/styled-components/babel-plugin-styled-components/pull/112/files) it should mean that hashes are consistent across environments, (dev/test/prod & client/server) given that the Babel plugin is used in all the places.

This makes sense to me overall. I wanna know if @philpl or @geelen have any idea why we weren't doing this in the first place.

probablyup:
My guess is it was just an oversight. Most devs using styled-components probably aren't also doing SSR.

PR #1392, opened 2017-12-27
remove trailing commas for functions

probablyup:
They're not supported in ES5, which is tested against in the node
versions included by the styled-components CI. Also disabled an
eslint rule that was conflicting with what prettier was doing.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,6 +5,7 @@
   "rules": {
     "semi": 0,
     "quotes": 0,
+    "comma-dangle": 0,
     "curly": [2, "multi-line"],
     "arrow-parens": 0,
     "class-methods-use-this": 0,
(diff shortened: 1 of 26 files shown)


philpl:
Cheers for the changes :) I think it’s already covered though by https://github.com/styled-components/styled-components/pull/1376

It adds the configs that should disable conflicting rules. The trailing comma change shouldn’t be necessary though since everything is being transpiled 😊

probablyup:
It matters for the rollup config though, which is evaluated as es5 but
still run through prettier

philpl:
@probablyuphm that’s true; but we weren’t running it through prettier before. 🤔

Mind checking whether only src is being targeted and fixing it up if it isn’t? 😬 I just merged the aforementioned PR

probablyup:
@philpl You know what, I think it's because I have "autofix on save" in my editor. Prettier doesn't seem to have an "ignore" option from a config perspective. I can revert it if you want

PR #1399, opened 2018-01-01
Fix duplicate keyframes being added to the global stylesheet

Blasz:
I'm not sure if you're still accepting PRs for version 1 but this fixes a bug whereby identical keyframes are added multiple times to the global stylesheet rather than only once.

### DIFF
--- a/src/constructors/keyframes.js
+++ b/src/constructors/keyframes.js
@@ -6,14 +6,23 @@ import type { Interpolation, NameGenerator } from '../types'
 
 const replaceWhitespace = (str: string): string => str.replace(/\s|\\n/g, '')
 
+let inserted = {}
+
+export const clearKeyframeCache = () => {
+  inserted = {}
+}
+
 export default (nameGenerator: NameGenerator) =>
   (strings: Array<string>, ...interpolations: Array<Interpolation>): string => {
     const rules = css(strings, ...interpolations)
     const hash = hashStr(replaceWhitespace(JSON.stringify(rules)))
-    const name = nameGenerator(hash)
-    const keyframes = new GlobalStyle(rules, `@keyframes ${name}`)
-    const keyframesWebkit = new GlobalStyle(rules, `@-webkit-keyframes ${name}`)
-    keyframes.generateAndInject()
-    keyframesWebkit.generateAndInject()
-    return name
+    if (!inserted[hash]) {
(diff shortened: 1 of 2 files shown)


philpl:
I can find you the old commit of v1 and push a patch, however I’ll beed to have a look at where it is first 😅

Blasz:
@philpl I branched off this tag - https://github.com/styled-components/styled-components/tree/v1.4.6

We had a look into v2 a while ago but observed some performance slowdowns so we haven't decided to upgrade yet, perhaps the performance issues have since been fixed though.

mxstbr:
> We had a look into v2 a while ago but observed some performance slowdowns so we haven't decided to upgrade yet

Performance slowdowns should definitely not happen, we've worked hard to make it at least as fast as v1—on top of which v2 is about 1/4 the bundle size 😅 I'd much rather you try and upgrade again rather than shipping another v1 release.

src/constructors/keyframes.js
```diff
@@ -6,14 +6,23 @@ import type { Interpolation, NameGenerator } from '../types'
 
 const replaceWhitespace = (str: string): string => str.replace(/\s|\\n/g, '')
 
+let inserted = {}
+
+export const clearKeyframeCache = () => {
```
philpl: Btw it'd be great if this matched what the `ComponentStyle`s are doing, i.e. the `inserted` cache should be part of the closure and the `test/utils` should recreate the `keyframes` helper to reset the cache.

philpl:
@Blasz I mostly agree with @mxstbr here. The difference between v1 and v2 is extraordinarily small, compared to major-bumps in other libraries, and the performance has been tweaked a lot.

I agree that the performance characteristics might be different, but we haven't run into (or received reports of) a real-world case in which the performance was completely insufficient without a workaround in sight.

I'd be grateful if you could provide us with more information on how you tested v2, and what you've found. It'd be a tremendous help :smile: :wink:

kitten:
Sorry for not replying to this with the correct explanation of why this wasn’t merged yet. We have recently merged #1501 and as a result I have a better overview of our caching strategy now.

We are already caching components by IDs and their rules by name. Keyframes are already hashed and due to our caching strategy they shouldn’t be duplicated.

I’m not quite sure whether this was an issue before, but please check whether you can reproduce a cache miss after the release of 3.2.0

Issue #1360, opened 2017-12-16
Replace Uglify with Google Closure Compiler


mxstbr:
Based on [this blogpost by the React team (@gaearon)](https://reactjs.org/blog/2017/12/15/improving-the-repository-infrastructure.html#migrating-to-google-closure-compiler), we should look into replacing Uglify with [Google Closure compiler](https://github.com/google/closure-compiler-js).

We're already using ES modules and Rollup, so we're basically pretty far of the way there. The last step to get this into the realm of possibilities is shipping flat bundles, ref #1359.

ConAntonakos:
Perhaps this PR will be helpful as a reference? https://github.com/facebook/react/pull/10236/files

wardpeet:
I'll have a look at this one

probablyup:
This ends up being more trouble than it's worth. We'll get some nice bundle savings with the various improvements planned for 4.0

PR #1409, opened 2018-01-07
[don't merge] Closure compiler

maciej-ka:
https://github.com/styled-components/styled-components/issues/1360

Waits for:
1. Merging https://github.com/styled-components/styled-components/pull/1362 which is the base for this Closure branch
2. Fix of https://github.com/google/closure-compiler-js/issues/23; My machine couldn't pass the error `JavaScript heap out of memory` and I never saw the result of this uglifier replace.


mxstbr:
> My machine couldn't pass the error JavaScript heap out of memory and I never saw the result of this uglifier replace.

Well that sucks, thanks for trying! Maybe @gaearon also ran into this when switching to Closure Compiler in React?

gaearon:
No, we didn’t run into these issues. Note we’re using a slightly older fork (see our Yarn lockfile). But the issue seems older than that (we had to fork for a different reason).

Have you tried using the same exact options that we use?

maciej-ka:
Same result with https://github.com/facebook/react/blob/master/scripts/rollup/build.js#L62 and permutations of these options (including `assumeFunctionWrapper` and `renaming` flags).

My machine can build current master of React. Both there and in this PR version "1.0.6" of rollup plugin is used.

gaearon:
Any chance you're accidentally processing unnecessary code? e.g. maybe it imports React itself by mistake.

maciej-ka:
Build is passing when `Stylis` is not used 
(https://github.com/styled-components/styled-components/blob/master/src/utils/stringifyRules.js, I removed import, stylis const and return statement)

mxstbr:
Huh, interesting! /cc @thysultan any ideas what's up with that?

maciej-ka:
Looking at Stylis source, problem is present when any of `char`, `format` or `context` variables are changed within part of the code that ignores comments.

The lines that, if commented out, will allow Closure compiler to pass, are:
https://github.com/thysultan/stylis.js/blob/master/stylis.js#L760
https://github.com/thysultan/stylis.js/blob/master/stylis.js#L764
https://github.com/thysultan/stylis.js/blob/master/stylis.js#L767
https://github.com/thysultan/stylis.js/blob/master/stylis.js#L772
https://github.com/thysultan/stylis.js/blob/master/stylis.js#L776-L778

thysultan:
Note sure, could be a bug with the JavaScript version of closure compiler since stylis works with both the basic and advanced mode of the Java version. 

@maciej-ka What do you mean by `Build is passing when Stylis is not used.` what error does it throw with it?

maciej-ka:
@thysultan, when building Styled Components with Stylis, it fails with error: `JavaScript heap out of memory`, just like https://github.com/google/closure-compiler-js/issues/23

This error is gone when Stylis import or few lines of Stylis code are commented out. They are listed in comment above.

Trouble is happening in selector isolation mode. The irony is that this part of Stylis source is never used by Styled Components, because it's disabled by a setting [`cascade: true`](https://github.com/styled-components/styled-components/blob/master/src/utils/stringifyRules.js#L7).

(If this can be any help: with `cascade: true` selectors will be like `.user p a`, with `cascade: false` they will be like `p.user a.user`)

maciej-ka:
Closure compiler build will pass also if just this one line is commented out https://github.com/thysultan/stylis.js/blob/master/stylis.js#L469

maciej-ka:
Solid solving of this Closure Compiler issue is beyond my powers. But here is a little change in Stylis that has no impact on its logic and for a mysterious reason, it makes the build pass. https://github.com/thysultan/stylis.js/pull/90 @thysultan

gaearon:
Can you file a new issue against GCC with a reproducing minimal case/

maciej-ka:
Minimal failing example is my dream here for some time now. I don't have it yet, but maybe it will be possible to distill it from Stylis workaround.

wardpeet:
@maciej-ka thanks for taking this one, wanted to start on it last week but saw you already took this one. 🎉

maciej-ka:
I don't plan to spend more time on extracting minimal repro from [seven hundred parser Stylis function](https://github.com/thysultan/stylis.js/blob/master/stylis.js#L177-L884). I tried. My idea was to cut the code in binary search way, but it seems parts need each other to break closure compiler.

I'm sending my ❤️ to anyone who would like to give this ticket some time.

SavePointSam:
Just ran into this myself. Pretty bummed I have to choose between styled-components or GCC. :(

maciej-ka:
@thysultan any chance to merge this [fix](https://github.com/thysultan/stylis.js/pull/90/files) temporary fix? I understand it's from your perspective a strange and unnecessary change in code, but it: 
- solves the problem,
- has a minimal impact (change is a move of one line variable declaration into broader scope),
- has comes with documentation, describing why this change and linking related ticket.

It's not uncommon to see such walkarounds in Open Source projects
(yeah, I'm trying hard to sell this)

probablyup:
@maciej-ka any news on this PR? I'm open to landing it but has been inactive for a while.

Issue #1333, opened 2017-11-29
Component-based API for global styling


wmertens:
Via https://github.com/styled-components/styled-components/issues/730#issuecomment-320035959

It would be great to have a React component for global styles that does the actual injecting (rendering null), removing the CSS when it is unmounted.

So instead of

```js
    injectGlobal`...`
```

you'd do

```js
const GlobalStyle = styled.global`
… (including prop interpolation)
`
...
return (
  <App>
    <GlobalStyle darkBg={this.props.dark} />
    ...
  </App>
)
```

The only differences with regular styled components would be:

* Renders null
* Does not wrap the CSS with an own class

This way:

* Global styles can be easily modified via props
* It works without issues in SSR
* It's obvious what is happening

mxstbr:
I think this would be a valuable change.

Myself and @geelen originally decided against going down this route because it seemed more tedious to use, but as you said it has a bunch of nice properties. A couple more are:

- You can use the theme dynamically in the global styles
- You can override global styles dynamically and change them
- We can make global styles work with the coming `StyleSheetManager`, ref #1324

I'm open to doing this. My only concern is that I don't like overloading the `styled.x` convention and whether we want to introduce a new API or reuse the old one:

```JS
/* First possibility: injectGlobal now just returns a component */
const GlobalStyles = injectGlobal`
  body {
    background: blue;
  }
`

/* Second possiblity: Different API */
const GlobalStyles = injectGlobalAsComponent`
  body {
    background: blue;
  }
`

/* Third possibility: Radium-like <Style /> component */
<Style>{`
  body {
    background: blue;
  }
`}</Style>
```

I'd personally prefer the first way. I'd rather not we have multiple ways of doing things, and we could break this in a major version without troubles.

philpl:
@mxstbr I'm totally down for replacing `injectGlobal` with sth like your first or second variant. I think both `styled.global` and `<Style>{template string}</Style>` are not favourable options.

I do think though that the name `injectGlobal` should be changed, but maybe we can go for a more succinct name like `globalStyles`.

I don't think this necessarily depends on #1324, but we do need to implement a way to remove styles from our sheet (or alternatively append & replace classes to `body`) which might conflict with refactors in #1208.

mxstbr:
Doesn't depend on #1324 at all, but enables more functionality for the `StyleSheetManager`!

[...]

PR #1416, opened 2018-01-09
Initial work to add createGlobalStyle functionality

JamieDixon:
Implements #1333

This is very much a WIP and open for discussion

### DIFF
--- a/src/constructors/createGlobalStyle.js
+++ b/src/constructors/createGlobalStyle.js
@@ -9,17 +9,19 @@ const _createGlobalStyle = (stringifyRules: Stringifier, css: Function) => {
     return class extends React.Component {
       componentId = ''
 
-      componentDidMount() {
+      componentWillMount() {
         const rules = css(strings, ...interpolations)
         const hash = hashStr(JSON.stringify(rules))
         this.componentId = `sc-global-${hash}`
+
         if (StyleSheet.instance.hasInjectedComponent(this.componentId)) return
 
         StyleSheet.instance.inject(this.componentId, false, stringifyRules(rules))
       }
 
       componentWillUnmount() {
-        // Remove styles
+        if (!StyleSheet.instance.hasInjectedComponent(this.componentId)) return
+        StyleSheet.instance.removeComponent(this.componentId)
       }
 
       render() {
--- a/src/constructors/createGlobalStyle.js
+++ b/src/constructors/createGlobalStyle.js
@@ -6,11 +6,17 @@ import StyleSheet, { CONTEXT_KEY } from '../models/StyleSheet'
 import ServerStyleSheet from '../models/ServerStyleSheet'
 import type { Interpolation, Stringifier } from '../types'
 
-const _createGlobalStyle = (stringifyRules: Stringifier, css: Function) => {
-  const createGlobalStyle = (strings: Array<string>, ...interpolations: Array<Interpolation>) => {
-    return class extends React.Component {
+export default (stringifyRules: Stringifier, css: Function) => {
+  const createGlobalStyle = (
+    strings: Array<string>,
+    ...interpolations: Array<Interpolation>
+  ) =>
+    class extends React.Component {
       static contextTypes = {
-        [CONTEXT_KEY]: PropTypes.oneOfType([PropTypes.instanceOf(StyleSheet), PropTypes.instanceOf(ServerStyleSheet)]),
+        [CONTEXT_KEY]: PropTypes.oneOfType([
+          PropTypes.instanceOf(StyleSheet),
+          PropTypes.instanceOf(ServerStyleSheet),
+        ]),
       }
       componentId = ''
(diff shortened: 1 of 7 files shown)


mxstbr:
Awesome!

In terms of removing styles, we need to add a new method to `StyleSheet.js` which removes tags based on component IDs. That should call `tag.removeComponent(id)`, which we'd need to implement to essentially reverse the work done in `tag.addComponent(id)`. Should be doable maybe?

JamieDixon:
I'll be looking into `removeComponent` today.

philpl:
@mxstbr in a future (!) PR we should probably add interpolation support as well :smile:

src/constructors/createGlobalStyle.js
```diff
@@ -0,0 +1,34 @@
...
+import hashStr from '../vendor/glamor/hash'
+import StyleSheet from '../models/StyleSheet'
+import type { Interpolation, Stringifier } from '../types'
+
+const _createGlobalStyle = (stringifyRules: Stringifier, css: Function) => {
+  const createGlobalStyle = (strings: Array<string>, ...interpolations: Array<Interpolation>) => {
+    return class extends React.Component {
+      componentId = ''
+
+      componentDidMount() {
```
philpl: Here the component should use `componentWillMount` instead :wink: otherwise this wouldn't work correctly for SSR. It's also important as we'll want the global styles to apply immediately before all DOM nodes mount (that might be referred to in or affected by these global styles).

src/constructors/createGlobalStyle.js
```diff
@@ -0,0 +1,34 @@
...
+        const rules = css(strings, ...interpolations)
+        const hash = hashStr(JSON.stringify(rules))
+        this.componentId = `sc-global-${hash}`
+        if (StyleSheet.instance.hasInjectedComponent(this.componentId)) return
```
philpl: This needs to possible grab the stylesheet from the context (for SSR as well, and/or future tag support)

See here: https://github.com/styled-components/styled-components/blob/master/src/models/StyledComponent.js#L94

and here: https://github.com/styled-components/styled-components/blob/master/src/models/StyledComponent.js#L274-L277

mxstbr:
Damnnn df87dd7 :fire: :fire: :fire:

Do you reckon we'll be able to use that for all styled components in the future? That'd be extremely rad.

src/constructors/createGlobalStyle.js
JamieDixon: Thanks! Adjusted in https://github.com/styled-components/styled-components/pull/1416/commits/df87dd75044e19144a80ca6403cae475dce155d7

src/constructors/createGlobalStyle.js
JamieDixon: Sweet. Thanks for the tips. Adjusted in https://github.com/styled-components/styled-components/pull/1416/commits/c6155ec3680a5639673e24073f5e802a2ff2ca3c

philpl:
@mxstbr I don't think we should remove any styles for styled components as that'd be unnecessary work. We don't care whether rules remain around or not since we control their classnames 😁

JamieDixon:
Is there a `printWidth` prop missing from .prettierrc? I think my Atom is mucking up your nice formatting.

philpl:
@JamieDixon that's definitely possible (I haven't had a look) but normally the default should be used. Maybe the Atom plugin falls back to its setting instead? That'd be horrible 😆 Feel free to add it to the `.prettierrc` if you don't mind

JamieDixon:
Thanks @philpl. I've dropped the `Ran yarn format` commit and added a new one updating the `.prettierrc` file. 👍

JamieDixon:
Is there more you'd like me to do on this one guys? I took a quick look into `BaseStyledComponent` but couldn't get it working. I'm happy to take another look if it's stopping this PR being approved?

JamieDixon:
@mxstbr I've dropped the CHANGELOG commit and sent a new one with just the change we want. I need to figure out why prettier changed all the things.

mxstbr:
Hold on, why is that changelog entry under 2.4.1 and not under unreleased? Pretty sure that won't be going out in 2.4.1?

JamieDixon:
Good point. Lemme fix that up.

JamieDixon:
Fixed

JamieDixon:
@kitten Cool. I hope what I did here was still useful 😬

kitten:
@JamieDixon Definitely! Thanks for kicking this work off :+1:

jarodtaylor:
This is sexy.

migueloller:
If would be great if it was possible to inject the styles in a custom document (i.e., an `iframe`). See #659.

probablyup:
Closing in favor of https://github.com/styled-components/styled-components/pull/1493 since it's a continuation and evolution of this one 😸

PR #1418, opened 2018-01-10
expose isStyledComponent utility

probablyup:
Expose `isStyledComponent` utility as a named export. This functionality is useful in some edge cases, such as knowing whether or not to use `innerRef` vs `ref` and detecting if a component class needs to be wrapped such that it can be used in a component selector.


philpl:
Awesome 🙌Can you add it to the native & primitives entrypoint as well please? 😬

PR #1420, opened 2018-01-11
[typescript] Add failing material-ui test

NeoLegends:
Hey!

We're using the v1-beta branch of material-ui together with styled-components (to great success, so far). Recently we have been getting typing errors, though. I already contacted the material-ui guys, but we could not track down the bug so far. I think it seems best if we join efforts to get the typings working properly again.

~~~Please see this https://github.com/mui-org/material-ui/pull/8781#issuecomment-349460247 comment and the associated thread for reference.~~~

Issue in material-ui repo: https://github.com/mui-org/material-ui/issues/9838

I'm currently trying to make the typings work and if I succeed, I will submit a PR with updated typings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,6 +110,7 @@
     "jest-image-snapshot": "^2.2.0",
     "jsdom": "^9.10.0",
     "lint-staged": "^6.0.0",
+    "material-ui": "^1.0.0-beta.27",
     "node-watch": "^0.4.1",
     "prettier": "1.9.2",
     "puppeteer": "^0.13.0",
--- a/typings/tests/material-ui.tsx
+++ b/typings/tests/material-ui.tsx
@@ -1,0 +1,8 @@
+import * as React from "react";
+import Paper from "material-ui/Paper";
+
+import styled from "../..";
+
+styled(Paper)` // Breaks here
+  background-color: black;
+`;
--- a/yarn.lock
+++ b/yarn.lock
@@ -1110,6 +1110,13 @@ babel-runtime@^6.18.0, babel-runtime@^6.22.0, babel-runtime@^6.23.0:
     core-js "^2.4.0"
     regenerator-runtime "^0.10.0"
 
+babel-runtime@^6.26.0:
+  version "6.26.0"
+  resolved "https://registry.yarnpkg.com/babel-runtime/-/babel-runtime-6.26.0.tgz#965c7058668e82b55d7bfe04ff2337bc8b5647fe"
+  dependencies:
+    core-js "^2.4.0"
+    regenerator-runtime "^0.11.0"
+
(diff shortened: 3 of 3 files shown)


philpl:
cc @styled-components/typers: If this is our fault, then why isn't this line being taken into account? https://github.com/styled-components/styled-components/blob/master/typings/styled-components.d.ts#L57

pelotom:
@philpl 

> cc @styled-components/typers: If this is our fault, then why isn't this line being taken into account? https://github.com/styled-components/styled-components/blob/master/typings/styled-components.d.ts#L57

Why isn't the type

```diff
-<P>(component: React.ComponentClass<P>): ThemedStyledFunction<P, T>;
+<P>(component: React.ComponentType<P>): ThemedStyledFunction<P, T>;
```

? Does `styled-components` specifically forbid the use of stateless components?

pelotom:
Ahh, I see that the next overload does allow for stateless components. But for some reason TypeScript can't merge these two overloads to allow for an argument of type `React.ComponentType<P>`. Is there any reason though they couldn't be combined?

```diff
export interface ThemedBaseStyledInterface<T> extends ThemedStyledComponentFactories<T> {
  <P, TTag extends keyof JSX.IntrinsicElements>(tag: TTag): ThemedStyledFunction<P, T, P & JSX.IntrinsicElements[TTag]>;
  <P, O>(component: StyledComponentClass<P, T, O>): ThemedStyledFunction<P, T, O>;
  <P extends { theme: T; }>(component: React.ComponentClass<P>): ThemedStyledFunction<P, T, WithOptionalTheme<P, T>>;
-  <P>(component: React.ComponentClass<P>): ThemedStyledFunction<P, T>;
-  <P extends { [prop: string]: any; theme?: T; }>(component: React.StatelessComponent<P>): ThemedStyledFunction<P, T, WithOptionalTheme<P, T>>;
+  <P extends { [prop: string]: any; theme?: T; }>(component: React.ComponentType<P>): ThemedStyledFunction<P, T, WithOptionalTheme<P, T>>;
}
```

pelotom:
I filed [an issue with TypeScript](https://github.com/Microsoft/TypeScript/issues/21175) about this, but I'm guessing it's something they will consider hard / out of scope. I think the right solution at least in the short term is to have an overload which explicitly supports a `React.ComponentType<P>` argument.

philpl:
@pelotom I’ve seen our own union type at the top of our typings before. Replacing it might be the key.

Do you mind submitting a PR please? 😬

pelotom:
Submitted a PR: #1427

philpl:
@pelotom let's close this in favour of #1427 then 👍it seems the check there will be sufficient to verify this being resolved

PR #1423, opened 2018-01-11
Remove object spread from rollup.config.js

JamieDixon:
Remove object spread from rollup.config.js and replace with Object.assign.

Per @mxstbr's suggestion https://twitter.com/mxstbr/status/951479813910429696

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -33,8 +33,7 @@ const configBase = {
   plugins: commonPlugins,
 }
 
-const umdConfig = {
-  ...configBase,
+const umdConfig = Object.assign({}, configBase, {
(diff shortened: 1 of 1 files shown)


JamieDixon:
This PR came about because I was having trouble running `yarn build` and it was complaining about the spread operator. Having `rm -rf`d `node_modules` and re `yarn`, everything works.

We can probably close this PR. Sorry for the noise.

mxstbr:
As I said on Twitter

> Well, if you're hitting that somebody else might as well, and they might not ask us and just not contribute, so I'd rather err on the side of making it work for everyone 😊

PR #1425, opened 2018-01-13
generate and ship source maps with the flat builds

probablyup:


probablyup:
And it'll help with profiling performance 😄

rollup.config.js
```diff
@@ -31,6 +36,7 @@ const configBase = {
   globals: { react: 'React' },
   external: ['react'].concat(Object.keys(pkg.dependencies)),
   plugins: commonPlugins,
+  sourcemap: true,
```
matej-prokop: Hi folks, 

I believe you have typo here. It should be `sourceMap` instead of `sourcemap` see https://github.com/maxdavidson/rollup-plugin-sourcemaps.

cc @philpl

mxstbr: I'm pretty sure Rollup updated their config to use the lowercase version. If you use `sourceMap` you'll get a warning.

mxstbr: No worries! :blush:

PR #1426, opened 2018-01-14
upgrade tests to react 16

probablyup:


src/test/__snapshots__/ssr.test.js.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`ssr should add a nonce to the stylesheet if webpack nonce is detected in the global scope 1`] = `"<h1 class=\\"sc-a b\\" data-reactroot=\\"\\" data-reactid=\\"1\\" data-react-checksum=\\"197727696\\">Hello SSR!</h1>"`;
+exports[`ssr should add a nonce to the stylesheet if webpack nonce is detected in the global scope 1`] = `"<h1 class=\\"sc-a b\\" data-reactroot=\\"\\">Hello SSR!</h1>"`;
```
probablyup: react 16 removed a bunch of data tags that used to be rendered - that's the reason for the snapshot churn

probablyup:
Looks like flow is failing on random node dependencies... ugh I hate typing software so much.

philpl:
It seems that Flow is not quite happy yet with the flow config/typings :/

PR #1430, opened 2018-01-16
streaming support

probablyup:
~~No idea if this actually works yet in practice. Going to try and test it out on my work's sample SSR project in the next few days...~~ it works 🎉


mxstbr:
`styled-components-server`?

probablyup:
We’d have to make a bunch of changes to Stylesheet to be able to split off,
right?

mxstbr:
Hmm probably, that kinda sucks. How much bigger is the client-side bundle with the server stuff?

probablyup:
Barely bigger, but we have to somehow tell the bundler to not include the
“stream” package.

probablyup:
Maybe we can omit it from the umd build. And provide umd via the
package.json “browser” field. It’s like “main”, but for client side use
only.

Andarist:
I was just about to propose using `browser` field. I think it is not even designed to contain umd bundle, but rather to provide "overrides" for the bundlers - webpack use it by default, I'd suspect browserify is using it too, as for rollup you have to pass `browser: true` to `rollup-plugin-node-resolve` to make it use this field.

You can specify those overrides on file-basis, so you can do something like:
```js
main: 'dist/styled-components.cjs.js', // server bundle
module: 'dist/styled-components.es.js', // server es bundle (i think noone will use it, at least not yet, but imho it makes sense to have clear correlation between all of those)
browser: {
  'dist/styled-components.cjs.js': 'dist/styled-components-browser.cjs.js', // override for cjs
  'dist/styled-components.es.js': 'dist/styled-components-browser.es.js', // override for es
}
```

And you ofc have to provide those additional flat bundles, but I've found it quite easy to do with `rollup-plugin-node-resolve`. Just make additional "overriding" files next to regular ones - i.e make `util.server.js` next to `utils.js` (I think in case of `styled-components` it makes more sense to override for server and not the other way around). And just add `{ extensions: ['.server.js', '.js'] }` to your server config (pass it to the `rollup-plugin-node-resolve`)

gaearon:
That (`browser`) is what we do for `react-dom/server` btw. (You can notice its browser builds don’t contain `renderToStream`)

probablyup:
@mxstbr @philpl this is ready for review - going to work on a tandem PR for the site docs

no-parser/package.json
```diff
@@ -3,5 +3,9 @@
   "private": true,
   "main": "../dist/styled-components-no-parser.cjs.js",
   "module": "../dist/styled-components-no-parser.es.js",
-  "jsnext:main": "../dist/styled-components-no-parser.es.js"
+  "jsnext:main": "../dist/styled-components-no-parser.es.js",
+  "browser": {
+    "../dist/styled-components-no-parser.es.js": "../dist/styled-components-no-parser.browser.es.js",
```
Andarist: when i have last experimented with this, those `browser` entries need to be put in the root package.json, not in the "proxying" ones - that need to be checked (preferrrably on the real example, will be hard to checkout any docs on this)

i **think** the algorithm just traverses up from the module for the nearest `package.json` to check against browser "rewrites", but when we are in the `dist` directory this is the root `package.json` - when this happens `browser-resolve` doesnt really know that we got the `dist` by proxy `package.json`, it knows only the requested module id and the absolute path of the requesting module

rollup.config.js
```diff
@@ -72,16 +79,35 @@ const prodUmdConfig = Object.assign({}, umdConfig, {
   ),
 })
 
-const webConfig = Object.assign({}, configBase, {
+const serverConfig = Object.assign({}, configBase, {
+  external: configBase.external.concat('stream'),
+  output: [
+    { file: 'dist/styled-components.es.js', format: 'es' },
```
Andarist: you could use `pkg.module`/`pkg.main` here

rollup.config.js
```diff
@@ -72,16 +79,35 @@ const prodUmdConfig = Object.assign({}, umdConfig, {
...
+      __SERVER__: JSON.stringify(true),
+    })
+  ),
+})
+
+const browserConfig = Object.assign({}, configBase, {
   output: [
-    { file: pkg.module, format: 'es' },
-    Object.assign({}, cjs, { file: pkg.main }),
+    { file: 'dist/styled-components.browser.es.js', format: 'es' },
```
Andarist: you could use `pkg.browser[pkg.module]` here

rollup.config.js
```diff
@@ -94,21 +120,49 @@ const primitivesConfig = Object.assign({}, configBase, {
...
 const noParserConfig = Object.assign({}, configBase, {
+  external: configBase.external.concat('stream'),
   input: 'src/no-parser/index.js',
   output: [
     { file: 'dist/styled-components-no-parser.es.js', format: 'es' },
     Object.assign({}, cjs, { file: 'dist/styled-components-no-parser.cjs.js' }),
   ],
+  plugins: configBase.plugins.concat(
+    replace({
+      __SERVER__: JSON.stringify(true),
```
Andarist: a little bit bummer that we have to put specify this in each config, but i guess there is not much that can be done about it, except wrapping config creation in a factory function, but im not sure if that would make things clearer, so 🤷‍♂️

src/models/ServerStyleSheet.js
```diff
@@ -149,36 +151,42 @@ export default class ServerStyleSheet {
     return this.instance.toReactElements()
   }
 
-  interleaveWithNodeStream(readableStream: Readable) {
-    const ourStream = new Readable()
+  interleaveWithNodeStream(readableStream: stream.Readable) {
+    if (__SERVER__) {
```
Andarist: for a few bytes less you could wrap this under the class declaration and assign directly to the prototype
```js
if (__SERVER__) {
  ServerTag.prototype. interleaveWithNodeStream = // ...
}
```

but does this file cas to be included in browser bundles altogether? whats the use case? it also might cause problems for cjs consumers, `stream` package might get tree-shaken easily in es bundles, but in cjs its just another whole story

Andarist:
> stream isn’t included in browser bundles at all due to the ignore plugin, I checked the output.

Oh, that's cool! Haven't noticed usage of ignore plugin.

> For cjs, that’s Node and stream is a built in so it’s cool.

I was concerned about browser cjs files (for webpack@1 and other older bundlers) - but it's covered by the ignore plugin, so nevermind :)

src/models/ServerStyleSheet.js
```diff
@@ -139,6 +151,44 @@ export default class ServerStyleSheet {
...
+      })
+
+      readableStream.on('error', err => {
+        ourStream.emit('error', err)
+      })
+
+      return ourStream
+    } else {
+      throw new Error(
+        'streaming is not implemented in browser builds of styled-components.'
```
mxstbr: This should be something like "Streams only work in Node.js, not the browser. Please don't use xyz method in the browser.", since it won't ever be implemented in browser builds of styled-components

probablyup:
@mxstbr I see what you're getting at with the name, but it doesn't fit really. Calling the API `hydrate{Something}` doesn't make sense because it's fundamentally doing something different from `ReactDOM.hydrate`. It's also _only_ meant to be used in a streaming context so that's why I explicitly added the word "streaming" into the API name.

rollup.config.js
probablyup: I ended up not using these because they made scanning for a particular build in the file more difficult.

probablyup:
@philpl do you have an opinion on the naming here?

rollup.config.js
philpl: @Andarist Creating a config creation factory has really worked out well for me before. I'd love a refactor PR for this to make the config more readable again, if you don't mind? :)

philpl:
@probablyup I think "migrateStyles" sounds a little odd, because "migration" is not an established term for filtering stylesheets from a stream (if there's any established term for it at all?).

Similarly I agree that "hydrate" is a misplaced term, as it involves the "filtering" step as well and not only the rehydration.

This might be the solution:

> https://github.com/emotion-js/emotion/issues/341#issuecomment-339539013

`extractStreamStyles` or something might be more clear, since extraction can mean that something is being filtered. Other than that maybe "consume" is fine too? wdyt?

rollup.config.js
Andarist: I won't be able to work on this any time soon, i have just too big OSS backlog at the moment :p but anyway it should be done as a followup PR to this and probably also not the biggest priority. 

That being said I will work on [microbundle](https://github.com/developit/microbundle) and would like to hide most of the stuff that was made here in it. However `microbundle` is "0" configuration tool at the moment and it's good to have a control and possibility to extend a config, so I'm not sure if it will be a fit for you yet.

probablyup:
@philpl @mxstbr updated per our convo

probablyup:
Made a few changes to reclaim bytes since the bundle was at 14.4kB since 3.0. Sitting at 14.39 now... still higher than I want it but at least below the current threshold -_-

mxstbr:
FYI, I published this branch as `styled-components@next` (version `3.1.0-0`) so I can test it in my prod app.

mxstbr:
Damn, I did something wrong in that publish, it says "Module not found: styled-components in src/". :confused:

mxstbr:
I don't know why that doesn't work, it's very annoying. Any clues @probablyup?

probablyup:
@mxstbr I just tried it too and I'm seeing the same thing. very weird...

mxstbr:
Yeah, maybe a problem with the `package.json` changes?

probablyup:
Perhaps it needs relative paths? Like “./dist” instead of “dist/"


On January 23, 2018 at 2:35:21 PM, Max Stoiber (notifications@github.com)
wrote:

Yeah, maybe a problem with the package.json changes?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/styled-components/styled-components/pull/1430#issuecomment-359904528>,
or mute the thread

probablyup:
Actually no, then 3.0.2 wouldn’t work either. Nvm.


On January 23, 2018 at 2:38:38 PM, Evan Scott (probablyup@gmail.com) wrote:

Perhaps it needs relative paths? Like “./dist” instead of “dist/"


On January 23, 2018 at 2:35:21 PM, Max Stoiber (notifications@github.com)
wrote:

Yeah, maybe a problem with the package.json changes?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/styled-components/styled-components/pull/1430#issuecomment-359904528>,
or mute the thread

probablyup:
Moved the changes to error messages over to https://github.com/styled-components/styled-components/pull/1445

probablyup:
@mxstbr made an adjustment to package.json and published a new test build under `3.1.0-3`

rollup.config.js
```diff
@@ -80,16 +87,35 @@ const prodUmdConfig = Object.assign({}, umdConfig, {
   ]),
```
philpl: The UMDs above don't need the streaming support either so we can set `__SERVER__` to false there as well :) it'd be great to just have browser stuff in those

probablyup: It is: https://github.com/styled-components/styled-components/pull/1430/files#diff-ff6e5f22a9c7e66987b19c0199636480R56

probablyup: https://github.com/styled-components/styled-components/pull/1430/files#diff-ff6e5f22a9c7e66987b19c0199636480R53

philpl:
@probablyup before we merge this can you remove the prerelease commit and tag? I’d love to just purge that out to avoid conflicts with my branch

probablyup:
@philpl done - had to make some changes to the rollup config as well to get all the scenarios working

wmertens:
:tada: awesome!

I haven't tried it yet but I noticed two things:

* it always emits style tags, aren't there places that won't accept them, like in a tr or so?
* it tests for nodejs by checking for `NODE_ENV===production`, but in our projects even dev mode does ssr so that it is as close as possible to production behavior

probablyup:
> it always emits style tags, aren't there places that won't accept them, like in a tr or so?

Yeah there are some contexts where the browser might be grouchy for a hot second until consolidation is called. But they're rare and gone before the page is usable since the styles are swiftly relocated to a safer location.

> it tests for nodejs by checking for NODE_ENV===production, but in our projects even dev mode does ssr so that it is as close as possible to production behavior

There are some good reasons to not enable speedy mode in development because it makes debugging the CSS really difficult (the stylesheets _appear_ empty.) You should be able to run your SSR setup in actual prod mode though right? React's perf characteristics are very different with `NODE_ENV=production`.

wmertens:
On Thu, Feb 8, 2018 at 7:15 AM Evan Scott <notifications@github.com> wrote:

> it always emits style tags, aren't there places that won't accept them,
> like in a tr or so?
>
> Yeah there are some contexts where the browser might be grouchy for a hot
> second until consolidation is called. But they're rare and gone before the
> page is usable since the styles are swiftly relocated to a safer location.
>

Aha, but that's only true if javascript is enabled… Something I'd like to
keep optional.

Anyway, probably these cases can be worked around in the CSS, by putting
the CSS in the parent somewhere… OTOH, would it be difficult to track when
we're in a component that does not allow styles, and then buffer the output
till later?

it tests for nodejs by checking for NODE_ENV===production, but in our
> projects even dev mode does ssr so that it is as close as possible to
> production behavior
>
> There are some good reasons to not enable speedy mode in development
> because it makes debugging the CSS really difficult (the stylesheets
> *appear* empty.) You should be able to run your SSR setup in actual prod
> mode though right? React's perf characteristics are very different with
> NODE_ENV=production.
>
Aha. I'll have to experiment. Right now we're not streaming at least until
https://github.com/staylor/react-helmet-async/issues/3 is handled, so I'm
just idle-musing :)

probablyup:
> OTOH, would it be difficult to track when we're in a component that does not allow styles, and then buffer the output till later?

Yes, definitely. We don't have control over how `renderToNodeStream` emits HTML chunks and if we wait to render style blocks it can cause style flashing. I think the better course is to do as you suggested and utilize careful top-down styling of the special cases like `tr`.

PR #1339, opened 2017-12-01
Update support for React Native

jpdriver:
Fix #1339

- [x] update `@types/react-native` dependency
- [x] add SafeAreaView to React Native interface
- [x] remove Navigator from React Native interface

PR #1435, opened 2018-01-20
fix(native typescript): Remove Navigator

iRoachie:
Navigator is no longer included in the react-native definition typings, so it throws an error at compile time.

### DIFF
--- a/native/index.d.ts
+++ b/native/index.d.ts
@@ -35,7 +35,6 @@ export interface StyledInterface extends BaseStyledInterface {
   ListView: ReactNativeStyledFunction<ReactNative.ListViewProperties>;
   MapView: ReactNativeStyledFunction<ReactNative.MapViewProperties>;
   Modal: ReactNativeStyledFunction<ReactNative.ModalProperties>;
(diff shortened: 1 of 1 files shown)


probablyup:
This is a dup of #1339 and I don't think we can change this without a major version bump right @mxstbr?

mxstbr:
Yep, we can't merge this without a semver major bump unfortunately!

philpl:
Yep, this is unfrtounately a dup, so I'll close it for now, since it's missing some changes too compared to #1339

iRoachie:
Is there any reason the other one wasn't merged, saw it open from early December? 

If anything I can suggest the typings be moved on [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped), so that you won't have to update you package version when typings need to be fixed or updated. They also do multiple releases per day.

philpl:
@iRoachie It'd break some people on older versions so we'd like it to go out as a major release.

https://github.com/styled-components/styled-components/pull/1339#issuecomment-348571609
https://github.com/styled-components/styled-components/pull/1339#issuecomment-352291426

PR #1438, opened 2018-01-22
Export secret internals for jest-styled-components 👻

philpl:


philpl:
@mxstbr compiling separately won't be possible due to our new and shiny flat bundles, but we can remove it in the `.min.js` bundle; In the rest of them it will be dead code and removed by all bundlers.

philpl:
@mxstbr I'm now replacing the secretInternals file with a virtual, empty file for the production UMD bundle. Sounds good?

PR #1439, opened 2018-01-22
v3.0.2

philpl:


philpl:
@mxstbr sorry, after merging I just noticed that AppVeyor has been complaining for a couple of commits already. Any idea what's up with it? 😢

PR #1468, opened 2018-01-30
 apply the workaround to enable speedy for IE/Edge

### LINKED ISSUE #1464: speedy injection bugs for styled(StyledComponent) in MS browsers
I don't have a repro just yet, but things are definitely injecting out of order in IE11. Looking at MDN, it seems that IE has a [spotty implementation of `insertRule`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/insertRule) so we might just want to opt out of SPEEDY mode for the IE user agent.

probablyup:
Closes #1464


probablyup:
cc @mitchellhamilton this might be relevant for emotion

mitchellhamilton:
[emotion already inserts an empty text node into all style elements](https://github.com/emotion-js/emotion/blob/master/packages/create-emotion/src/sheet.js#L51) so I think this is already fixed in emotion.

src/models/BrowserStyleSheet.js
```diff
@@ -164,6 +161,9 @@ if (!DISABLE_SPEEDY) {
         )
       }
 
+      // workaround for an IE/Edge bug: https://twitter.com/probablyup/status/958138927981977600
+      newEl.textContent = ' '
```
philpl: Lgtm if it works :+1:

src/models/BrowserStyleSheet.js
```diff
@@ -154,7 +152,6 @@ if (!DISABLE_SPEEDY) {
     replaceElement() {
       // Build up our replacement style tag
       const newEl = this.el.cloneNode(false)
-      newEl.setAttribute(SC_ATTR, '')
```
philpl: Any reason you'd want this to go?

probablyup:
This is good to go 👍

PR #1469, opened 2018-01-30
add jest-styled-components tests

MicheleBertoli:
Hello :)
We were talking about this yesterday in https://github.com/styled-components/jest-styled-components/issues/118
I guess this is the simplest tests we can implement, to make sure the latest version of `jest-styled-components` works on new releases.
What do you think?
Thank you very much!


package.json
```diff
@@ -24,6 +24,7 @@
     "test:primitives": "jest -c .jest.primitives.json",
     "test:primitives:watch": "npm run test:primitives -- --watch",
     "test:integration": "jest -c .jest.integration.json --forceExit",
+    "test:jest-styled-components": "npm install jest-styled-components@next && npm run test:integration",
```
mxstbr: Should we add `jest-styled-components` to the `devDependencies` instead? I think running `npm install` here if people use `yarn` will take a looong time

MicheleBertoli: Thank you very much for your comment, @mxstbr.

As far as I know, there's no way to install `devDependencies` without adding them to the `yarn.lock`.
What I was trying to do here is running the tests against the latest version of `jest-styled-components` (and avoid keeping it manually updated).
I'm not entirely sure why this should be slower for `yarn` users (this project is actually using `yarn`) but since the test doesn't need to be run all the time I wouldn't mind about perf.

That said, I'm more than happy to add the package to the `devDependencies`, stick to a particular version, and keep the dependency updated.

mxstbr: Can't we do something like

```
devDependencies: {
  "jest-styled-components": "*"
}
```

MicheleBertoli: Unless I'm missing something, that's going to add something like this to the `yarn.lock`:

```diff
+jest-styled-components@*:
+  version "4.10.0"
+  resolved "https://registry.yarnpkg.com/jest-styled-components/-/jest-styled-components-4.10.0.tgz#6fcd83208fefc4ceb87b79b55d44697a05d226a8"
+  dependencies:
+    css "^2.2.1"
+
```

mxstbr: Ah god dammit, I forgot about the lock file. Ugh, that sucks. I see what you mean now, that's fine then, sorry for the confusion.

MicheleBertoli:
Thank you very much, and sorry for using the internals without telling you 😘

probablyup:
@MicheleBertoli looks like CI is being a bit grumpy, mind taking a look?

kitten:
@MicheleBertoli another thing about the non install is that it will create a package-lock file and will ignore yarn.lock; this breaking the integrity of the packages in our node_modules folder :(

MicheleBertoli:
Thank you very much for your comments, @probablyup @kitten.
I was hoping someone would give me a hint : )

Just to recap, there's seems to be no way of "temporary" installing a dependency with `yarn`, see [issue](https://github.com/yarnpkg/yarn/issues/1743).
`yarn add X --no-lockfile` leaves the `yarn.lock` untouched but updates the `package.json` 

@kitten with `npm install X  --no-save --no-package-lock` it's possible to avoid generating the `package-lock.json`. However, `npm` still tries to clean up the `node_modules` (e.g. `removed 1328 packages`), which breaks the tests.

In the end, I installed `jest-styled-components` as a dev dependency.
I hope this makes sense for you.

PR #1474, opened 2018-01-31
Append style tags below existing tags

### LINKED ISSUE #1433: New style blocks should inject after the last one
Currently when rendering in the browser, style blocks are automatically appended to the end of the `<head>` tag: https://github.com/styled-components/styled-components/blob/master/src/models/BrowserStyleSheet.js#L169

We should revise this behavior to instead grab a reference to the last known styled-components style block and inject after that, wherever it is in the DOM (falling back to `document.head` if it's the first one.)

chrislaughlin:
Fixes #1433 
Append style tags below existing styled components style tags, if no tags exist append them to the bottom of the head. This prevents side effects e.g. adding a tag after a critical CSS tag and potentially overriding styles. I have updated all instances where the head was appended to now use the new util function. 

I have also updated the gitignore to handle Jetbrains IDE's

Thanks in advance for you feedback


chrislaughlin:
Seen the failing build, based on the contribution guide the person merging the release adds the changelog update. Do I need to do this?

The changes I made have tests, expect for `BrowserStyleSheet.js` due to no existing tests for this class.

src/utils/appendToHead.js
```diff
@@ -0,0 +1,21 @@
...
+ * This prevents any side effects from pre rendered pages or when styles are added after first render.
+ */
+export default function appendToHead(el: HTMLStyleElement) {
+  const scStyleTags = Array.from(
+    document.querySelectorAll('style[data-styled-components]')
+  )
+  if (scStyleTags.length > 0) {
+    const lastScStyleTag = scStyleTags[scStyleTags.length - 1]
+    // $FlowFixMe
+    lastScStyleTag.parentNode.insertBefore(el, lastScStyleTag.nextSibling)
```
probablyup: will this work if the style tag is the last child in its parent context?

src/utils/consolidateStreamedStyles.js
```diff
@@ -21,6 +23,6 @@ export default function consolidateStreamedStyles() {
     }
 
     // $FlowFixMe
-    document.head.appendChild(frag)
```
probablyup: let's keep this code the way it is - there's no time we'd ever want the consolidated styles to not go to the end of `<head>`

src/utils/appendToHead.js
philpl: Probably not; not sure what the alternative is? This maybe: https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentElement

src/utils/consolidateStreamedStyles.js
philpl: Yep, this will break the React rehydration as it would try to remove style tags and then add them back, right?

mxstbr:
> based on the contribution guide the person merging the release adds the changelog update

That's indeed very confusing—every PR that changes code should include a changelog entry. The person who releases a new version has to then add a new section for that version. We should clean those up, PR welcome! :wink:

chrislaughlin:
I have updated appendToHead to use insertAdjacentElement this should cover all use cases. However insertAdjacentElement is not yet supported by jsdom so I had to remove some tests.

kitten:
Please take a look at this PR now that it’s merged since it should also address this problem. Thank you either way for contributing and feel free to reopen it if it is still relevant. Cheers! 👍😊 https://github.com/styled-components/styled-components/pull/1501

PR #1479, opened 2018-02-01
simplify stream emitter code

evan-scott-zocdoc:
fixes a bug where the last tag might end up getting emitted multiple times due to an off-by-1 issue


evan-scott-zocdoc:
Worked fine for my local setup 👍

PR #1482, opened 2018-02-02
fix a few bugs 

evan-scott-zocdoc:
1. on rehydration when SPEEDY is on, certain CSS selector characters were getting HTML entity-ifed due to usage of `innerHTML`

2. fix a streaming edge case when additional styles are appended to an existing element whose tag was already emitted


evan-scott-zocdoc:
Prerelease published as `3.1.6-1` for testing

probablyup:
Pushed `3.1.6-2`.

PR #1484, opened 2018-02-02
Improve ServerStyleSheet performance?

mxstbr:
I don't actually know if this does the same thing, but if it does then it's definitely less code and very likely more engine optimized?


kitten:
Closing in favour of a general WIP refactor #1501

PR #1491, opened 2018-02-05
Introduce StyleSheetManager.target for custom CSS insertion points

marionebl:
Picks up prior work by @paul-veevers and @bteng22 in 

* #1324 
* #1102 

This introduces the `StyleSheetManager.target` API geared towards integration scenarios where it is desirable to apped `<style>` tags to DOM nodes other than `top.document.head`, e.g. when rendering inside an `<iframe>` or `shadowRoot`.

**EDIT(@mxstbr): This is published as `@marionebl/styled-components`, please try that in your apps and report back how it works!**

1. [x] Solve merge conflicts
2. [x] Make the test suite pass
3. [x] Do a scoped publish for real-world testing on some projects

Issue #793, opened 2017-05-15
Unable to change page background via ThemeProvider and injectGlobal


three60five:
My app has a "light" and "dark" theme. Right now, I set the background color of the page using `injectGlobal` and css on the html element, like:

```js
injectGlobal`
  html {
    background-color: #FFFFFF;
  }
`;
```

When trying to use the ThemeProvider component to change themes, I am unable to change the background color via the theme props. Is this use case supported? If not, what workarounds are supported?

k15a:
You can't access the theme with `injectGlobal`. You could either style the highest component or use the `withTheme` helper to set the `background-color` with css:

```jsx
const Root = styled.div`
  background-color: ${props => props.theme.backgroundColor};
`
```

```jsx
import { withTheme } from 'styled-components'

const InjectGlobals = ({ theme }) => {
  document.body.style.backgroundColor = theme.backgroundColor

  return null
}

export default withTheme(InjectGlobals)
```

three60five:
Thanks for the suggestions!

creaux:
I found this as a best solution.

```javascript
import { Children } from 'react';
import { withTheme, injectGlobal } from 'styled-components';

const Global = ({ theme, children }) => {
  injectGlobal`
      font-family: ${theme.typography.secondary};
      color: ${theme.colors.dark}
    }
  `;
  return Children.only(children);
};

export default withTheme(Global);
```

mxstbr:
See #1333 and the WIP PR for `createGlobalStyle` in #1416, this will very soon be much easier! :party:

PR #1493, opened 2018-02-05

marionebl:
Picks up work by @JamieDixon in #1416
* originally motivated by #1333  
* newly motivated by #1491 

**ToDo**

* [x] Solve merge conflicts
* [x] Add tests for new behaviour
* [x] Add support for function interpolation
* [x] Add support for themes
* [x] ~~Add support for `keyframes``` interpolation~~

**Example**

```js
import React from 'react';
import {render} from 'react-dom';
import {ThemeProvider, createGlobalStyle} from 'styled-components'

const GlobalStyle = createGlobalStyle`
  * {
    box-sizing:  border-box;
  }
  body {
    background: ${props => props.theme.background};
    font-family: ${props => props.fontFamily};
  }
`

const App = () => (
  <ThemeProvider theme={{background: 'palevioletred'}}>
     <GlobalStyle fontFamily="Consolas" />
  </ThemeProvider>
)

render(<App />, document.querySelector('[data-root]'))
```

**Demo**
* ~~https://codesandbox.io/s/r43k57q4q~~
* https://codesandbox.io/s/xow4x7p284

**Published versions**

```
npm install @marionebl/styled-components@3.3.0
```


Fer0x:
But what about keyframes?

mxstbr:
Good question @Fer0x. They have the same issue as `injectGlobal`: They live outside of the component tree, which makes things like #1324 hard to implement.

Opened #1496, let's discuss there.

marionebl:
This one bugged be a bit in the previous PR too: 

https://travis-ci.org/styled-components/styled-components/jobs/338113464#L624

```
Invariant Violation: StyleSheetManager.getChildContext(): key "__styled-components-stylesheet__" is not defined in childContextTypes.
```

For some reasone the import of `CONTEXT_KEY` seems to yield undefined in some tests?

https://github.com/styled-components/styled-components/blob/474946d58db59c342de8d6cf8dc715cde592daf7/src/models/StyleSheetManager.js#L4

marionebl:
Went ahead and 

* implemented some tests for this
* also started implementation of interpolation in `createGlobalStyle` strings

I am pretty sure this could be implemented way DRYer, but could not find the reusable bits and pieces for interpolation, theme handling etc. I expected. 

Pointers to the places in the code based would be appreciated!

src/constructors/createGlobalStyle.js
```diff
@@ -0,0 +1,119 @@
...
+import {
+  CHANNEL,
+  CHANNEL_NEXT,
+  CONTEXT_CHANNEL_SHAPE,
+} from '../models/ThemeProvider'
+
+export default (stringifyRules: Stringifier, css: Function) => {
+  const createGlobalStyle = (
+    strings: Array<string>,
+    ...interpolations: Array<Interpolation>
```
kitten: You'll want to run `css()` ASAP on this input so that it runs only once (see `keyframes.js`)

src/constructors/createGlobalStyle.js
```diff
@@ -0,0 +1,119 @@
...
+          this.setState({ theme })
+
+          this.unsubscribeId = context.subscribe(nextTheme => {
+            this.setState({
+              theme: nextTheme,
+            })
+          })
+        }
+
+        const rules = css(
```
kitten: this diverges a lot from what we're doing, you'll probably want to use `ComponentStyle`

src/models/BrowserStyleSheet.js
```diff
@@ -191,6 +191,20 @@ if (!DISABLE_SPEEDY) {
       })
     }
 
+    removeComponent(componentId: string) {
```
kitten: The StyleSheet implementation will change a lot in the future. I'd like to get that refactor PR in before your PR though. Sorry...

I'll make sure to implement the methods you need after it's merged (#1501)

Btw this won't work correctly for the speedy implementation. But no worries I'll make sure that you get the methods you need 🙌

src/models/StyleSheetManager.js
```diff
@@ -1,10 +1,14 @@
...
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
-import StyleSheet, { CONTEXT_KEY } from './StyleSheet'
+import StyleSheet from './StyleSheet'
 import ServerStyleSheet from './ServerStyleSheet'
 import { tagConstructorWithTarget } from './BrowserStyleSheet'
 
+// TODO: Import this from ./StyleSheet again
+// when we figured out why it is undefined during testing sometimes
+const CONTEXT_KEY = '__styled-components-stylesheet__'
```
kitten: It's undefined because of a circular dependency. In #1501 I'm moving it to `src/constants.js`

src/constructors/createGlobalStyle.js
```diff
@@ -0,0 +1,119 @@
...
+          PropTypes.instanceOf(StyleSheet),
+          PropTypes.instanceOf(ServerStyleSheet),
+        ]),
+      }
+      state = { theme: {} }
+      componentId = ''
+      unsubscribeId = -1
+
+      componentWillMount() {
+        const context = this.context[CHANNEL_NEXT]
```
kitten: It doesn't quite look like this is matching our usual theming/subscription logic. We have some code for it here (it's easier to read in `withTheme`): https://github.com/styled-components/styled-components/blob/master/src/hoc/withTheme.js#L40-L60

It'd be cool if we could make a common base class for this stuff. Let's take a look at that in a separate PR though

kitten:
> I am pretty sure this could be implemented way DRYer, but could not find the reusable bits and pieces for interpolation, theme handling etc. I expected.

@marionebl I added some comments with the things that I've found. Unfortunately all this stuff wasn't extracted into a common place yet, even when it became apparent during a refactor of `withTheme` that it'll probably need to be extracted. But let's take a look at that in a separate PR

If you need help to use `ComponentStyle`, `css`, copying the theming stuff, etc, feel free to ping me on Twitter DMs if that helps you to get this done faster :+1:

src/constructors/createGlobalStyle.js
marionebl: Tried to implement on top of `ComponentStyle`, but found in hard to bend it to my use case (no generated classNames, etc). I went ahead and extracted the required parts to a simplified `GlobalStyle`

src/models/BrowserStyleSheet.js
marionebl: Rebuilt the relevant changes on top of your PR, using `StyleSheet.remove` method now

marionebl:
@kitten @mxstbr Could you have another look?

marionebl:
For people reviewing this I'd like to point out two parts:

* [src/models/GlobalStyle](https://github.com/styled-components/styled-components/pull/1493/files#diff-debb77f74057038e175986028f4d543d) - introduced to avoid even higher complexity in `ComponentStyle`

* [src/hoc/withTheme.js#L48-L53](https://github.com/marionebl/styled-components/blob/6483208878e0a85f4651579dcae044db43701489/src/hoc/withTheme.js#L48-L53) - disables the warning in `withTheme` about missing `ThemeProvider` if the decorated Component is found to be a `styled-component`. This enables the usage of `withTheme` in `createGlobalStyle`. Perhaps `withTheme` could be used in `src/models/StyledComponent`, too?

kitten:
@marionebl I’ll try to find some time next week to review this. Generally GlobalStyle might be a good idea, but since it’ll share a lot of code with ComponentStyle I’m not sure whether we shouldn’t find a better abstraction for both. I’d be happy to take on this task in a separate branch though.

Regarding withTheme, I’d like to avoid using it for internal components. It’s true that we need to improve our code sharing here, but the logic that components apply to themes is generally meant to be shared through utilities as we’d like to avoid another level of components for StyledComponents

marionebl:
> [...] Generally GlobalStyle might be a good idea, but [...] I’m not sure whether we shouldn’t find a better abstraction for both. I’d be happy to take on this task in a separate branch though.

👍 for introducing a code sharing abstraction in a separate PR

> Regarding withTheme, I’d like to avoid using it for internal components [...]

Understood. Would `createGlobalStyle` without theme support be an acceptable intermediary step?  I imagine we could add theme support (with its apparent complexity) in a later PR.

kitten:
@marionebl I'd be happy to add theme support and refactor a couple of things myself, if you don't mind :+1: It's been on my todo list either way for a while 😅

marionebl:
> I'd be happy to add theme support and refactor a couple of things myself, if you don't mind 👍 It's been on my todo list either way for a while 😅

Ok. Just trying to find a way to get global styles that honor `StyleSheetManager.target` 😅

marionebl:
That being said I'd like to take part in the refactorings if you provide a rough outline and illustrating pseudo code if applicable. 🙇

chenxsan:
Sorry, didn't expect to toggle the checkbox when clicking on it :(

ermik:
I am actively choking as I'm forced to write the spaghetti from #793 right now, can't wait till this is production-ready. Thank you very much for the hard work — I can't speak for the community, but I will: You're al our heroes (and by that, I obviously mean you guys are major visual assets@2x).

src/models/StyledComponent.js
```diff
@@ -13,16 +13,12 @@ import getComponentName from '../utils/getComponentName'
...
 import escape from '../utils/escape'
 import type { RuleSet, Target } from '../types'
-import { CONTEXT_KEY } from '../constants'
+import { CONTEXT_KEY, STATIC_EXECUTION_CONTEXT } from '../constants'
 
 import { CHANNEL, CHANNEL_NEXT, CONTEXT_CHANNEL_SHAPE } from './ThemeProvider'
 import StyleSheet from './StyleSheet'
 import ServerStyleSheet from './ServerStyleSheet'
 
-// HACK for generating all static styles without needing to allocate
```
wmertens: I see you moved this to constants.js, but you dropped the comment. I don't necessarily think it's a hack, but adding a comment that explains its use would be nice?

wmertens:
@marionebl could you resolve the merge conflicts (very minor)

@kitten @mxstbr I think this is ready for prime time… Can you check?

src/test/utils.js
```diff
@@ -80,6 +80,15 @@ export const stripWhitespace = (str: string) =>
     .replace(/([;\{\}])/g, '$1  ')
     .replace(/\s+/g, ' ')
 
+export const getCSS = (scope?: Document | HTMLElement = document) => {
```
rodocite: Just a question: 

Won't Flow infer `function | object` and let undefined pass for `getCSS`? So wouldn't you get an error if anything but an existing `document` object is passed in? I realize this is just a test and you're completely controlling the execution-- I'm just curious.

marionebl:
@wmertens Solved the conflicts. @kitten requested the internal `withTheme` usage to be removed in order to avoid another layer of components in the tree. 

I held back with implementing this due to @kitten mentioning a pending refactoring for code sharing concerns around internal theme support.

src/constructors/createGlobalStyle.js
```diff
@@ -0,0 +1,62 @@
...
+          PropTypes.instanceOf(ServerStyleSheet),
+        ]),
+      }
+
+      static styledComponentId = id
+
+      componentWillMount() {
+        const { props } = this
+        const sheet = this.context[CONTEXT_KEY] || StyleSheet.master
+        const context = style.isStatic ? STATIC_EXECUTION_CONTEXT : { ...props }
```
kitten: Theming logic is not in place yet? We could mark this as a WIP since it’s been a while and merge this first?

wmertens: @kitten I'm all for merging this now rather than later…

shaunbent:
Do we think this is going to be merged soon? 

I am dependant on this in order to be able to inject global styles within the correct context of `<StyleSheetManger target={someFrame} />`.

mxstbr:
@shaunbent not sure, honestly. We've all been busy, somebody just needs to:

- Add a depreciation warning to `.extend` and `injectGlobal`, publish a v3 minor release with those warnings
- Merge all the big API changes to a `v4` branch and publish a `@next` release

If you have time I'll add you to the repo? :pray:

notatestuser:
I'm not sure if it's relevant but @marionebl's branch appears to break our ssr. not sure if other features are included in there.

wmertens:
@mxstbr hmm, I understand about adding the deprecation, but this PR doesn't break anything does it? So we can add the new API before v4?

mxstbr:
> So we can add the new API before v4?

Yeah sure if we want to increase our bundle size by a couple kB gzip for a feature nobody doesn't even know exists :wink: I understand where you're coming from, but at this point we might as well publish a v4 alpha instead of squeezing this into a v3 release.

wmertens:
I don't think it's even a couple KB non-gzip?

It's just kind of weird to deprecate a feature but not provide the
alternative, no? Then the deprecation should be in v4 and only removed in
v5…

mxstbr:
> It's just kind of weird to deprecate a feature but not provide the alternative, no?

Yeah I guess that's true. 🤔 Good point. So we should add the warnings but also the new APIs.

JamieDixon:
I'm super excited to see this getting towards a release! Amazing work @marionebl 🎉

DieLaughing:
I'm super excited for custom fonts to work everywhere (or at all)!!!

src/constructors/createGlobalStyle.js
```diff
@@ -0,0 +1,62 @@
...
+        style.updateStyles(context, sheet)
+      }
+
+      componentWillUnmount() {
+        const sheet = this.context[CONTEXT_KEY] || StyleSheet.master
+        style.removeStyles(sheet)
+      }
+
+      render() {
+        return null
```
migueloller: This can be an improvement left for the future but it would be nice if `render` just returned `this.props.children`. It would allow for things like:

```js
<Normalize>
  <div />
</Normalize>
```

Instead of:

```js
<>
  <Normalize />
  <div />
</>
```

migueloller: Maybe it's a small enough change that it can be done before merging?

probablyup:
I'd prefer any changes outside the scope of this PR to be made in a separate PR. Thanks!

migueloller:
@probablyup, I'm not sure if this is in reference to my [comment](https://github.com/styled-components/styled-components/pull/1493#discussion_r187933654). If it is, the comment is well within the scope of the PR. If it isn't then disregard this comment! 😄

src/constructors/createGlobalStyle.js
```diff
@@ -0,0 +1,62 @@
...
+      }
+    }
+
+    return withTheme(GlobalStyleComponent)
```
migueloller: There's currently an issue with using `createGlobalStyle` in production builds, where if the created component is not eventually a children of a `ThemeProvider`, you will get an error on [this](https://github.com/marionebl/styled-components/blob/3e2b5e812a1538cbed2365606103377ff3870fba/src/hoc/withTheme.js#L58) line, saying `subscribe` is undefined. It was very difficult to figure out, but it basically prevents us from running a production build using this new feature.

Here's before (not working):
```js
ReactDOM.render(
  <React.Fragment>
    <Normalize />
    <LockBody />
    <ThemeProvider theme={theme}>
      <ApolloProvider client={apolloClient}>
        <Router>
          <App />
        </Router>
      </ApolloProvider>
    </ThemeProvider>
  </React.Fragment>,
  ensure(document.getElementById('root')),
)
```

And after (working):
```js
ReactDOM.render(
  <ThemeProvider theme={theme}>
    <React.Fragment>
      <Normalize />
      <LockBody />
      <ApolloProvider client={apolloClient}>
        <Router>
          <App />
        </Router>
      </ApolloProvider>
    </React.Fragment>
  </ThemeProvider>,
  ensure(document.getElementById('root')),
)
```

I would gladly contribute to the PR with a fix, but unfortunately I'm not comfortable enough in the codebase to know where to make this fix. If someone is willing to help me, I'm happy to contribute, though.

migueloller: Note, this bug only happens when building with `process.env.NODE_ENV` set to `production`.

migueloller:
There seems to be an off-by-one bug in the stylesheet manager. Note, this only happens when `process.env.NODE_ENV` is `production` and a component created with `createGlobalStyle` is used.

Here's a screenshot of the error:

image

kitten:
@migueloller It looks like I messed up here: https://github.com/styled-components/styled-components/blob/master/src/utils/insertRuleHelpers.js#L52

If we delete from `b` and want to delete `3` items. We'd want `b - 2`, `b - 1` and `b`, but the stop index is `b - size (3)` and we compare it using `>=` so we delete one items too many. Can simply be fixed by changing the comparison to `i > lowerBound` instead, I believe.

migueloller:
@kitten, awesome. Thanks for the quick response! Will open up a PR.

src/constructors/createGlobalStyle.js
marionebl: Good idea, let's do this

marionebl: Adressed via 438e79e257daefda9bfff0239cae929acda2d519

src/models/StyledComponent.js
marionebl: Adressed via d9d6cd48e8b4a6611917f0e33406f70809b08f0b

src/test/utils.js
marionebl: Changed it to be more explicit about the scope. The current typing appears to be necessary though as `document` is not compatible to `HTMLElement`

src/constructors/createGlobalStyle.js
marionebl: Currently this uses `withTheme` - can we go forward with the current solution and replace it with a util-based one later?

marionebl: Added a comment noting the `withTheme` usage should be removed when theming is refactored

src/constructors/createGlobalStyle.js
marionebl: Released via `@marionebl/styled-components@3.3.0`
I also updated the demo at https://codesandbox.io/s/30m80qjwy1 to show this in action

src/constructors/createGlobalStyle.js
marionebl: Adressed this via a64f2272d6d3d41e32baa0c747bfdcb6ffca1158
Released via `@marionebl/styled-components@3.3.0`
I also updated the demo at https://codesandbox.io/s/30m80qjwy1 to show this in action

src/constructors/createGlobalStyle.js
migueloller: Awesome work! Thanks @marionebl!

marionebl:
> @mxstbr what was the reasoning behind allowing this component to accept children? Looking at the demo, it's kind of confusing what scope the styles are being applied at. I think it'd make more sense as a standalone component.

I followed @migueloller's reasoning. This is mainly about ergonomics - I felt being able to use it either way (instead of swallowing children like before) is the least surprising behavior for everyone.

probablyup:
> instead of swallowing children like before

I'd rather add a dev warning to not give it children and keep it as a meta component. The alternative is way more confusing IMO.

migueloller:
@probablyup, I think that’s a fair point and I didn’t consider that when making the suggestion. I mostly focused on terseness, but it’s true that it might cause some confusion for devs.

Even in the example that I give, `Normalize` still applies to everything, no just its children.

marionebl:
Ok circled back on the `props.children` change and added a development warning as @probablyup suggested.

probablyup:
Nice! Ok this is looking good to me. Would like @kitten to do one last look-see to make sure we're not missing anything re: stylesheet lifecycles.

Could we also add an SSR test for this to make sure things are emitted properly?

marionebl:
Added the relevant test as suggested by @probablyup

marionebl:
@mxstbr @kitten  @probablyup Any chance we could get this into a release this week? I'll be unavailable very soon (parental leave ❤️ ) and would like to see this through before then.

willwull:
Tried out `@marionebl/styled-components@3.3.2 `, but I'm still getting this error, production mode with create-react-app:

image

migueloller:
@willwull, this should've been fixed by [this](https://github.com/styled-components/styled-components/commit/36345f7fd484bd931806f510b22a7b40e8f788d1).

@marionebl, I'm guessing you rebased v3.3.0?

willwull:
@migueloller I tried it in another test project, but I can't seem to reproduce the error, it only occurs in the first project I tried it in. Why does it only happen in production mode? I'm trying to pinpoint if I did something strange, but it's a bit slow since I have to build the project each time 😅

willwull:
@marionebl @migueloller 

Update: seems this error only happens when trying to switch the theme for the `ThemeProvider`.
Here is a minimal example, where clicking the button will crash the app:

`package.json`
```json
{
  "name": "styled-components-test",
  "version": "0.1.0",
  "private": true,
  "dependencies": {
    "@marionebl/styled-components": "^3.3.2",
    "react": "^16.4.1",
    "react-dom": "^16.4.1",
    "react-scripts": "1.1.4"
  },
  "scripts": {
    "start": "react-scripts start",
    "build": "react-scripts build",
    "test": "react-scripts test --env=jsdom",
    "eject": "react-scripts eject"
  }
}
```

`src/index.js`

```js
import React, { Component } from "react";
import ReactDOM from "react-dom";
import { createGlobalStyle, ThemeProvider } from "@marionebl/styled-components";

const GlobalStyle = createGlobalStyle`
  h1 {
    color: ${props => props.theme.color};
  }
`;

class App extends Component {
  state = {
    theme: {
      color: "red"
    }
  };

  toggleTheme = () => {
    this.setState({
      theme: {
        color: "blue"
      }
    });
  };

  render() {
    return (
      <ThemeProvider theme={this.state.theme}>
        <React.Fragment>
          <GlobalStyle />
          <div className="App">
            <h1 className="App-title">Welcome to React</h1>
            <button onClick={this.toggleTheme}>Click me</button>
          </div>
        </React.Fragment>
      </ThemeProvider>
    );
  }
}

ReactDOM.render(<App />, document.getElementById("root"));
```

Disclaimer: I'm extremely new to styled-components, so tell me if I'm doing something weird.

probablyup:
Going to pick this up Monday or Tuesday to finish out any remaining items.

migueloller:
@probablyup,

I can confirm that the `deleteRule` error that @willwull is mentioning is still there. This bug only happens in production.

In the process of trying to figure out what is wrong I came across a few inconsistencies between dev and prod. One of them is the fact that [`safeInsertRule`](https://github.com/styled-components/styled-components/blob/36345f7fd484bd931806f510b22a7b40e8f788d1/src/utils/insertRuleHelpers.js#L34-L40) is swallowing 2 errors in production caused by `normalize` from https://github.com/styled-components/polished. Could this be the cause for the `deleteRule` size and index mismatch?

Here's a screenshot of the errors logged to the console:
image

During debugging we commented out [this line](https://github.com/yjimk/styled-components/blob/67652fa377ccff4ae3d77eaa8d7e6d902491527c/src/constructors/createGlobalStyle.js#L45) and it seemed to remove the issue, so it could be a problem with `updateStyles` making it be in a inconsistent state. Some of the `deleteRule` errors thrown come from `componentWillRecieveProps` as well, it's not all from `componentWillUnmount`.

Another issue we're having is duplication of styled from global style components. In the server we have a 2-render pass because of Apollo, this causes global styles to be applied twice. Then, once React hydrates in the client, the styles are applied once more, resulting in triple styles coming from global style components. It would seem that the IDs aren't stable, removing idempotency from these components. We're using the Babel plugin, too. This might be a different issue (style duplication), but I thought maybe the lack of idempotency might contribute to the issue at hand.

The way we ended up fixing the issue for now was by putting our global style components all the way outside of our app so that they're only ever rendered once. Before they used to be in our app wrapped by things like react-router's BrowserRouter and react-apollo's ApolloProvider. Whenever a route changed and caused a re-render, the `deleteRule` error would happen. This is very bizarre, given that the render method for global style components doesn't actually do anything, and shouldn't be the cause for side effects.

Thoughts?

probablyup:
#1807 supercedes this closing (it's a cumulative PR)

PR #1498, opened 2018-02-06
Proposal: Always rehydrate into document.head

kitten:
Some interesting insights:

- The `target` doesn't actually affect the rehydration; We would always replace the SSR elements in place
- Some people might rely on the `replaceChild` behaviour of rehydration (how sensible is it to just put everything into the head?)

But this would remove the need of a separate Streaming SSR rehydration phase. If we do not want to break anything we could apply this behaviour only on streamed style tags (special attribute?)


kitten:
Obsolete due to #1501

PR #1501, opened 2018-02-07
Refactor and rewrite StyleSheet logic

kitten:
This is a complete rewrite of our stylesheets, however the goal is to make it work 90% as it did before. (If not more :wink:)

No substantial test or snapshot changes as far as I can tell.

- Merge BrowserStyleSheet and StyleSheet
- Switch automatically at lowest point in code between implementations
- Rename `StyleSheet.instance` to `.master` for clarity (`.instance` is still present for compatibility w/ jest-styled-components)
- Remove concept of "local vs global" StyleSheets (They'll be merged instead of split)
- Rewrite streaming implementation to not rely on tag/sheet properties
- Add a rehydration tag as a composed browser/speedy tag
- Improved a couple of error messages
- Rehydrate into a single stylesheet (Important change to merge all styles after streaming)
- Remove stream consolidation (handled by rehydration, cc @probablyup)
- Ensure that all style tags are consecutive in the target (new ones will be appended to the last)

I played around with minimising the caching behaviour (do we need hash *AND* names?) but that
should be refactored in a separate PR.


src/constants.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+
+declare var __DEV__: ?string
+
+export const SC_ATTR = 'data-styled-components'
+export const CONTEXT_KEY = '__styled-components-stylesheet__'
+
+export const IS_BROWSER = typeof window !== 'undefined'
+
+export const DISABLE_SPEEDY =
```
probablyup: What if we made this an instance variable of the StyleSheet such that it could be overridden if desired by an external composer?

src/models/ServerStyleSheet.js
```diff
@@ -2,218 +2,104 @@
...
+    : ''
+
+const streamBrowserErr =
+  process.env.NODE_ENV !== 'production'
+    ? 'Streaming SSR is only supported in a Node.js environment; Please do not try to call this method in the browser.'
+    : ''
 
 export default class ServerStyleSheet {
   closed: boolean
+  global: StyleSheet
```
probablyup: `s/global/singleton/` ? `global` reminds me too much of `injectGlobal`

probablyup: Or perhaps "master"

src/models/ServerStyleSheet.js
```diff
@@ -2,218 +2,104 @@
...
+    if (!__SERVER__ || IS_BROWSER) {
+      throw new Error(streamBrowserErr)
     }
-  }
 
-  static create() {
-    return new StyleSheet(isLocal => new ServerTag(isLocal))
+    /* the tag index keeps track of which tags have already been emitted */
+    const { instance } = this
+    let instanceTagIndex = 0
```
probablyup: this might cause problems with streaming - will verify tomorrow

src/models/ServerStyleSheet.js
kitten: Typically singletons would indicate that there’s only a single instance... at least that’s what I think of them.

I chose “global” as an opposite to localised Stylesheets since soon we might need to introduce more sophisticated Stylesheet management for the manager’s rehydration.

Totally up to discuss this for sure, but it’d also be no problem to do this in a separate issue/PR :)

src/models/ServerStyleSheet.js
kitten: As far as I can tell since I always refer to the latest tag and am forcing the Stylesheet to create a new one after this routine, an index should suffice to keep track of which tags haven’t been seen before

But I might be wrong 😅

src/constants.js
kitten: Let’s discuss this separately as well. We have another issue where we were discussing to add it to the StyleSheetManager. We previously have been against exposing any kind of global function/flags to change settings like these though

evan-scott-zocdoc:
Tested this in our project and it looks like the style blocks aren't being removed from `<body>` fast enough before `React.hydrate()` kicks in. Seeing the typical "unexpected `<style>` in `<div>`" type messages...

src/models/StyleTags.js
```diff
@@ -0,0 +1,399 @@
...
+    names,
+    insertMarker,
+    insertRules,
+    css,
+    toHTML: wrapAsHtmlTag(css, names),
+    toElement: wrapAsElement(css, names),
+    clone() {
+      return {
+        ...tag,
+        names: [...names],
```
probablyup: nit: `names.slice()` might be slightly faster and less bytes

src/models/StyleTags.js
```diff
@@ -0,0 +1,399 @@
...
+  immediateRehydration: boolean
+): Tag<any> => {
+  let isReady = false
+
+  /* rehydration function that adds all rules to the new tag */
+  const rehydrate = () => {
+    /* only rehydrate once */
+    if (isReady) {
+      return
+    } else {
```
probablyup: nit: don't need the else block

probablyup:
We'll also need to add a note to the website that `consolidateStreamedStyles()` isn't needed anymore after this is released.

src/models/StyleTags.js
kitten: Yea, I thought that when I was writing this as I copied it over. Not sure what it transpiles to, but good idea, I’ll replace it

src/models/StyleTags.js
kitten: Yep, it’s more of a stylistic choice but not in line obv

kitten:
@probablyup maybe we should also add a message to it when it is called, pointing the user to the section that explains that it’s not needed anymore

src/models/ServerStyleSheet.js
mxstbr: This still needs to be changed to `master` right?

src/models/StyleTags.js
```diff
@@ -0,0 +1,396 @@
...
+    ? `
+Trying to insert a new style tag, but the given Node is unmounted!
+- Are you using a custom target that isn't mounted?
+- Does your document not have a valid head element?
+- Have you accidentally removed a style tag manually?
+`.trim()
+    : ''
+
+/* this error is used for tags */
+const throwCloneTagErr = () => {
```
mxstbr: Why is this a function, but all the other errors are static strings? This doesn't really need to be a function right?

src/utils/generateAlphabeticName.js
```diff
@@ -1,17 +1,25 @@
...
 
-  for (x = code; x > charsLength; x = Math.floor(x / charsLength)) {
-    name = chars[x % charsLength] + name
+  /* get a char and divide by alphabet-length (bitwise zero for flooring the number) */
+  for (x = code | 0; x > charsLength; x = (x / charsLength) | 0) {
+    name = getAlphabeticChar(x % charsLength) + name
   }
 
-  return chars[x % charsLength] + name
+  return getAlphabeticChar(x % charsLength) + name
```
mxstbr: Is this refactor going to change existing classes? If so, that'd be very annoying for any existing tests :confused:

src/utils/consolidateStreamedStyles.js
```diff
@@ -1,26 +1,18 @@
...
+ * the use of streamed style tags and will perform the rehydration earlier when needed
+ * this function will not be needed anymore */
+export default function consolidateStreamedStyles() {
+  if (process.env.NODE_ENV !== 'production') {
+    // eslint-disable-next-line no-console
+    console.warn(
+      'The consolidateStreamedStyles function for streamed SSR rehydration was removed.\n' +
+        'The rehydration will now be automatically performed at an earlier point in time ' +
+        'on the client so that this function is now not needed anymore and is thus obsolete.\n' +
+        '- Consider removing this call on the client and use React + styled-components as you would expect'
```
mxstbr: Slight rephrase of this entire error message:

```
styled-components automatically does streaming SSR rehydration now, so manually calling consolidateStreamedStyles is no longer necessary and a noop now.
- Please remove the consolidateStreamedStyles call from your client.
```

src/test/ssr.test.js
```diff
@@ -202,10 +202,8 @@ describe('ssr', () => {
     )
     const elements = sheet.getStyleElement()
 
-    expect(elements).toHaveLength(2)
-
+    expect(elements).toHaveLength(1)
```
mxstbr: Why do we have fewer elements here? Shouldn't it still be one style tag for global styles + one style tag for local styles?

src/test/rehydration.test.js
```diff
@@ -252,40 +247,20 @@ describe('rehydration', () => {
...
-      /* Note: any future additions don't replace the style tag */
-      const D = styled.div.withConfig({ componentId: 'TWO' })`color: tomato;`
-      shallow(<D />)
-
-      expect(Array.from(document.querySelectorAll('style'))[1]).toBe(styleTagsAfterAddition[1])
-
-      /* The point being, now we have a style tag we can inject new styles in the middle! */
+
+      /* the order stays correct and the styles are unharmed
+        * NOTE: during rehydration the empty rules are stripped out however */
```
mxstbr: Why does rehydration strip the empty rules now? I'm pretty sure we still want those, at least in development.

src/models/StyleTags.js
kitten: That’s for convenience; look at where it’s used in the style tag creation. I’m avoiding writing some anonymous functions there repeatedly

src/test/rehydration.test.js
kitten: That’s because we’re using stylis to split the rules, so for instance “.A { }” isn’t sent over during SSR. It’s not really worth keeping it around if that implies that we need to meddle with what stylis does currently.

You can target those components regardless of whether the rule is present, since targeting the Button’s component ID for example depends on their className props, not on whether a rule for it is present

src/test/ssr.test.js
kitten: There is no difference between local and global styles anymore (see commit msgs / PR description). SSR now produces a single style tag and it’s simply not worth or necessary to differentiate between local and global styles, especially since we’re getting rid of global styles soon.

src/utils/consolidateStreamedStyles.js
kitten: Sure thing

src/utils/generateAlphabeticName.js
kitten: Nope, all tests still pass & and the output is the exact same

kitten: Btw I’ve done this for gzip :) turns out the small list of the alphabet is really bad for gzip since that’s really impossible to compress.

src/utils/consolidateStreamedStyles.js
kitten: done

mxstbr:
(haven't tested in prod though)

mxstbr:
Also want to decrease bundlesize again? :hearty_eyes:

kitten:
@mxstbr I’ll send a second PR soon that shaves off some bytes again 😛 then I can decrease it I guess

PR #1514, opened 2018-02-11
Add StyleSheet#remove for rule deletion and refactor caching strategy

kitten:
This adds a `remove(id: string)` method to `StyleSheet` which is important for #1493 to work.

The modifications in the aforementioned PR to the `BrowserStyleSheet` weren't safe before, since they incorrectly deleted the entire component from the sheet's registry. This would cause the original position of the component in the StyleSheet to change, which is not desirable when rules are deleted for an `id` that is still relevant.

Imagine a global component being created whose id is stable (which is the case since its rules will never change). If we don't keep the id stable or change its position in the stylesheet, then its effect in the cascade aren't predictable.

This will also provide the `remove` method after the StyleSheet refactor PR (#1501) since it would otherwise leave #1493 stranded.

The caching strategy was changed, so that `hash` is not used anymore. This is because our name generator should operate on a rather high performance. The cost of checking both `hash` and then `name` is theoretically slower than just generating `name`. This also saves us a couple of lines of code and a massive headache


src/constructors/keyframes.js
```diff
@@ -13,23 +13,18 @@ export default (
...
-  const hash = hashStr(replaceWhitespace(JSON.stringify(rules)))
+  const name = nameGenerator(hashStr(replaceWhitespace(JSON.stringify(rules))))
+  const id = `sc-keyframes-${name}`
 
-  const existingName = StyleSheet.master.getNameForHash(hash)
-  if (existingName !== undefined) {
-    return existingName
+  if (!styleSheet.hasNameForId(id, name)) {
+    styleSheet.inject(
+      `sc-keyframes-${name}`,
```
probablyup: reuse the `id` variable here?

src/models/ComponentStyle.js
```diff
@@ -60,11 +62,9 @@ export default (
       this.isStatic = !isHRMEnabled && isStaticRules(rules, attrs)
       this.componentId = componentId
 
-      if (!StyleSheet.master.hasInjectedComponent(componentId)) {
```
probablyup: This check is no longer needed?

src/models/ComponentStyle.js
```diff
@@ -60,11 +62,9 @@ export default (
...
-        const placeholder =
-          process.env.NODE_ENV !== 'production' ? [`.${componentId} {}`] : []
-        StyleSheet.master.deferredInject(componentId, placeholder)
-      }
+      const placeholder =
+        process.env.NODE_ENV !== 'production' ? [`.${componentId} {}`] : []
+      StyleSheet.master.deferredInject(componentId, placeholder)
```
probablyup: Seems like we don’t need to call deferredInject here at all in production mode since there’s no rule to add?

src/models/ComponentStyle.js
```diff
@@ -49,6 +49,8 @@ export default (
   flatten: Flattener,
   stringifyRules: Stringifier
 ) => {
+  const generateName = (str: string) => nameGenerator(hashStr(str))
```
probablyup: This is somewhat confusing with the static version further in the file.

src/models/StyleSheet.js
```diff
@@ -197,26 +203,30 @@ class StyleSheet {
     return (this.tagMap[id] = tag)
   }
 
-  /* optimal caching: hash is known and name can be returned */
-  getNameForHash(hash: string) {
-    return this.hashes[hash]
+  /* mainly for injectGlobal to check for its id */
+  hasId(id: string) {
+    return this.tagMap[id] !== undefined
```
probablyup: Should we switch to an es6 Map now? Might be faster.

src/test/ssr.test.js
```diff
@@ -46,6 +46,22 @@ describe('ssr', () => {
     expect(css).toMatchSnapshot()
   })
 
+  it('should respect removed rules', () => {
+    const Heading = styled.h1`
```
probablyup: It’s fine to have a test I guess, but this should never happen. In the SSR use case, you’d never reuse a stylesheet and once render is complete, you’re done.

src/models/StyleSheet.js
mxstbr: Not enough browser support, it barely works in IE11 and we support IE9 and up: http://kangax.github.io/compat-table/es6/#test-Map :cry:

src/models/ComponentStyle.js
mxstbr: Seems like it should be fine to just wrap that whole `placeholder = bla; deferredInject()` block into a production check

src/constructors/keyframes.js
kitten: Oops, good catch!

src/models/ComponentStyle.js
kitten: I’ll rename it

src/models/ComponentStyle.js
kitten: No, it’s performed as part of deferredInject now

src/models/ComponentStyle.js
kitten: deferredInject is our secret sauce to keep the order of injection correct. It actually adds a marker and adds deferred rules to inject at a later point in time.

So it’s still super important

src/test/ssr.test.js
kitten: Yea, this is really just to test the remove method, because I wasn’t sure where I had to do the least amount of work to write the test 😛

I’ll see if there’s a better place

src/constructors/keyframes.js
kitten: done

src/models/ComponentStyle.js
kitten: done

src/test/ssr.test.js
kitten: moved

src/models/ComponentStyle.js
kitten: Sorry, you're right. It's still needed for streaming in some edge cases 😅

src/models/StyleSheet.js
evan-scott-zocdoc: Oh, we shouldn't be using Array.from then.

PR #1524, opened 2018-02-15
Disable marker comment in NODE_ENV = produciton

bakunyo:
Recently in production ( especially [AMP Site](https://www.ampproject.org/) ), response sizes from servers are wanted as short as possible.
[AMP restrict css sizes to 50,000 bytes.](https://www.ampproject.org/docs/guides/responsive_amp)

So I'd like to delete `makeTextMarker` comment in production.
I'm not sure what kind of effects exist in rehydration, so could you please confirm about it.

### DIFF
--- a/src/models/StyleTags.js
+++ b/src/models/StyleTags.js
@@ -46,7 +46,10 @@ The clone method cannot be used on the client!
 }
 
 /* this marker separates component styles and is important for rehydration */
-const makeTextMarker = id => `\n/* sc-component-id: ${id} */\n`
+const makeTextMarker = id =>
+  process.env.NODE_ENV !== 'production'
+    ? `\n/* sc-component-id: ${id} */\n`
+    : ''
 
 /* retrieve a sheet for a given style tag */
 const sheetForTag = (tag: HTMLStyleElement): CSSStyleSheet => {


mxstbr:
This is only tangentially related: How are you using styled-components with AMP? I thought AMP needed only a single `style` tag with some special attribute?

kitten:
This would break SSR unfortunately. #1541 is going to delete the ones we don’t need once it’s merged, and we don’t care about them when the browser creates them during runtime.

All in all it’s unfortunately not safe to always delete them. Sorry!

bakunyo:
@mxstbr @kitten 
Thanks for your reviews. I've already used styled-components with AMP in production environment.
First, I get all styles of components by using `ServerStyleSheet#getStyleTags`.
Next I force change string of styles like below rules.
- replace from `<style` to `<style amp-custom`
- delete comments in styles like `/* sc-component-id xxx */` ( to size down bytes )
- delete newlines in styles ( same reason )

Last, I insert this changed string to inner `header` tag.

Is pr 1514 going to delete these comments and newlines? If so, it's very good news for me!
I'm looking forward to it will be merged.

kitten:
@bakunyo we can’t delete these comments as they’re markers for our rehydration. As that’s now an integral part of our library (since it’s first and foremost about dynamic styles) it would probably not make sense to offer the option to disable these markers as that’s essentially just a huge foot gun 😶

mxstbr:
@kitten I could see an "AMP mode" or maybe "Static mode" though where we only insert a single style tag, automatically set `amp-custom` on it and don't care about any of the rehydration stuff since you don't rehydrate AMP pages anyway.

We might need that for Spectrum 😅

kitten:
@mxstbr Maybe it'd make sense though to build a library for that around styled-components, right? Because essentially if you'd minify the output from `ServerStyleSheet` you'd also minify away other CSS rules that you don't care about. That might be a preferable approach in my eyes here.

But for what it's worth we could also add `ServerStyleSheet.getStyleTags({ static: true })`

mxstbr:
What about `ServerStyleSheet.getStaticStyleTags()` to mimic `renderToStaticString`?

kitten:
@mxstbr I’d be happy with that 👍

PR #1528, opened 2018-02-18
Add failing test for nested media queries

pziemkowski:
Nested media queries generate incorrect CSS.

Such component:
```js
const Comp = styled.div`
  @media (min-width: 10px) {
    @media (min-height: 20px) {
        color: red;
     }
  }
`
```

Will generate this CSS:
```
.sc-a {}
      
@media (min-width: 10px) {
  @media (min-height: 20px) {
    .b {
      color: red;
    }
  }
      
```

Notice the missing closing curly brace character (`}`)

I would love to fix this myself, but unfortunately, I don't know the code enough.

styled-components version: 3.1.6

### DIFF
--- a/src/test/css.test.js
+++ b/src/test/css.test.js
@@ -34,6 +34,28 @@ describe('css features', () => {
     `)
   })
 
+  it('should generate styles for nested media queries', () => {
+    const Comp = styled.div`
+      @media (min-width: 10px) {
+        @media (min-height: 20px) {
+          color: red;
+        }
+      }
+    `
+    shallow(<Comp />)
+    expectCSSMatches(`
+      .sc-a {}
+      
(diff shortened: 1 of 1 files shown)


mxstbr:
@thysultan

thysultan:
Possibly related to [https://github.com/zeit/styled-jsx/issues/370](https://github.com/zeit/styled-jsx/issues/370). What version of the plugin is in use?

kitten:
@thysultan sounds indeed like that’s the culprit. So a simple upgrade should do it, I suppose?

thysultan:
@kitten Yes, an upgrade should probably fix it.

kitten:
Addressed by #1529 :tada:

PR #1572, opened 2018-03-05
v3.2.0

kitten:


kitten:
@probablyup I think we did. But we need to be more aggressive about removing it since otherwise PRs will add it back 😅

PR #1573, opened 2018-03-05
Fix defaultProps warning fired wrongly

DefoyPhilip:
Fix the bug about the warning `'[withTheme] You are not using a ThemeProvider nor passing a theme prop or a theme in defaultProps'` being fired even if a defaultProp is used.

Change the code to check this.props.theme (which cover both defaultProps and props.theme) to validate the good usage of the withTheme hoc.

Also check if styledContext is undefined before trying to use it.

Edit: Using this pr in my personal project, a bug occur but I'm not able to reproduce it in the test environment. 
Here's the test reproducing my problem but that pass correctly in your test environement
```javascript
it('should use a theme from ThemeProvier even if it has a defaultTheme', () => {
    const defaultTheme = { color: 'green' }
    class MyDiv extends React.Component {
      render() {
        return <div>{this.props.theme.color}</div>
      }
    }
    MyDiv.defaultProps = { defaultTheme }
    const MyDivWithTheme = withTheme(MyDiv)
    MyDivWithTheme.defaultProps = { defaultTheme }

    const theme = { color: 'red' }
    const newTheme = { color: 'blue' }

    const Theme = ({ theme }) => (
      <ThemeProvider theme={theme}>
        <MyDivWithTheme />
      </ThemeProvider>
    )

    const wrapper = mount(<Theme theme={theme} />)

    expect(wrapper.find('div').text()).toBe('red')

    // Change theme
    wrapper.setProps({ theme: newTheme })

    expect(wrapper.find('div').text()).toBe('blue')
  })
```


src/hoc/withTheme.js
```diff
@@ -93,6 +93,10 @@ const wrapWithTheme = (Component: ReactClass<any>) => {
     }
   }
 
+  WithTheme.propTypes = {
```
kitten: This is also missing from the `StyledComponent` itself. So either we don't add it at all, or at least we add it to everything in a separate PR, I'd say, so we can discuss it there :wink:

src/hoc/withTheme.js
```diff
@@ -42,7 +42,7 @@ const wrapWithTheme = (Component: ReactClass<any>) => {
       const themeProp = determineTheme(this.props, undefined, defaultProps)
```
kitten: We won't need this anymore, I'd say

src/hoc/withTheme.js
```diff
@@ -51,7 +51,7 @@ const wrapWithTheme = (Component: ReactClass<any>) => {
         )
       } else if (styledContext === undefined && themeProp !== undefined) {
         this.setState({ theme: themeProp })
```
kitten: Since `themeProp` will never default to `this.props.theme` i.e. `defaultProps.theme` actually (see `determineTheme`) this won't actually apply a theme provided by `defaultProps`.

src/hoc/withTheme.js
DefoyPhilip: Yeah I've noticed it wasn't described anywhere. The problem is the the pre-commit run eslint which fire an error if it's not provided.

src/hoc/withTheme.js
DefoyPhilip: Removing it causes a bug. The hoc will use the theme from defaultProp instead of the one from the themeProvider since it won't be able to tell the difference between a theme from defaultProp and a theme passed as prop

src/hoc/withTheme.js
DefoyPhilip: themeProp will be this.props.theme if the prop theme is used because `props.theme === defaultProps.theme` (from determineTheme) will make isDefaultTheme false

kitten: yep, but in case the theme prop is not being passed and `defaultProps.theme` exists this will cause us to not use it and not log a warning.

Instead it'd be great to just use it, as described in the original issue.

I don't think it'd make sense to hide the warning and then only use `this.props.theme` when it's not defaulting

DefoyPhilip: Considering the scenario where theme prop is not being passed and `defaultProps.theme` exist, defaultProps will be used by the component. The test `should not break without a ThemeProvier if it has a defaultTheme` validate this with the first expect.
I'm I completely wrong about this?

johnnyBira: I'm currently seeing this warning in one of my packages, which is a HOC.
https://www.npmjs.com/package/responsive-props

The HOC is using `withTheme`. I tried suppressing the warning by setting the `defaultProp.theme` of the component wrapping the component to an empty object without any luck.

The warning doesn't appear when wrapped with a `ThemeProvider`.

PR #1577, opened 2018-03-06
Move @import rules to an extra/separate browser tag inside tags

### LINKED ISSUE #1574: import external fonts is broken in 3.2.0
## Version
styled-components: 3.2.0
babel-plugin-styled-components: 1.5.1

## Reproduction

Sorry I tried to reproduce it [here](https://codesandbox.io/s/z2zz2x0jp) but it works fine, however you can see the issue on [this repo](https://github.com/piuccio/styled-bug) it's a clone of react-boilerplate.

There's a file called `global-styles.js` that injects global CSS, including the following

```js
import { injectGlobal } from 'styled-components';

/* eslint no-unused-expressions: 0 */
injectGlobal`
  @import url(https://fonts.googleapis.com/earlyaccess/notosansjapanese.css);
`
```

## Steps to reproduce

`npm install` with styled-component `3.2.0` and `npm start` it'll load a local server. Inspect the styles and you'll see that `Noto Sans` is not used.

Kill the server, change package.json to `3.1.6`, `npm install && npm start` and the fonts are now `Noto Sans`

## Expected Behavior

Always use `Noto Sans`

screen shot 2018-03-06 at 11 24 38

## Actual Behavior

Uses fallback font

screen shot 2018-03-06 at 11 25 34

kitten:
Fixes #1574 

When a BrowserTag or SpeedyTag detect an @import rule, which is expected
to be the first rules in the array to arrive due to Stylis' reordering,
then an extra tag is injected, and the @import rules are injected there
instead.

This ensures that it's *impossible* for an @import rule to not be at the
beginning of a stylesheet.

Published under `styled-components@dev` version `3.2.1-0` for testing.


src/constructors/test/__snapshots__/injectGlobal.test.js.snap
```diff
@@ -0,0 +1,14 @@
...
+"<style data-styled-components=\\"\\">
+/* sc-component-id: sc-global-3616276198 */
+html{padding:1px;}
+/* sc-component-id: sc-a */
+.sc-a {} .b{color:green;}
+/* sc-component-id: sc-global-1433896746 */
+html{color:blue;}</style>
+<style data-styled-components=\\"\\">
+/* sc-component-id: sc-global-1433896746-import */
+@import url('bla');</style>"
```
mxstbr: Shouldn't this style tag be above the other style tags? Or is it only about the first thing in a certain style tag?

kitten: @mxstbr I could enforce it to come before the style tag that contains the `@import` actually. I'm not sure whether we'd want that, since it's mostly meant for fonts only anyway.

But I could modify the `makeStyleTag` method to allow me to inject elements before another

mxstbr: Not sure, I mean if it works that's totally fine!

PR #1595, opened 2018-03-10
Upgrade stylis & stylis-rule-sheet to fix nested at-rules

### LINKED ISSUE #1576: v3.2.0 Breaks some styles
## Version
styled-components: 3.2.0

## Reproduction
I believe this is partially related to #1454, which was closed in favour in #1473 and fixed in #1529. However, some issues still persist. Most notably it's the inconsistency between production and development build and ignored styles in development build. None of this used to be an issue before 3.1.0

## Steps to reproduce
I wasn't able to isolate the exact place or cause where the issue happens. I plaid around with this file in my app to get all I know so far about the issue -> https://github.com/dmitrizzle/Analog.Cafe/blob/develop/src/app/components/Picture/styles.js

Basically, adding conditional styles to my `<Figure />` component in this fashion:
```
[top styles]
{props => props.feature && '
  [conditional styles]
'}
```
...tends to remove all of the [top styles] from it. There are some [media queries](https://github.com/roast-cms/react-sugar-styled/blob/develop/src/utils.js) involved (i.e. `props.theme.size.breakpoint.min.xxl`) but removing them does not seem to solve the issue.

## Expected Behavior
In production I run 3.0.2, this is how [one of the views](https://www.analog.cafe/submit/compose) looks with some sample content (below). Please note the the size of the "Test heading" as well as location of the image below (and box-shadow on it, if you can see it):

Production server with 3.0.2

## Actual Behavior

On my local machine I am trying 3.2.0, and in this case the styles that dictate the image to float left, have a shadow and other properties are thrown out (they do not appear in browser's console on "inspect element"):

Development server with 3.2.0

I tried compiling the code generated with 3.2.0 into production package. This created some strange effects. The styles I mentioned above (shadow, float) are no longer ignored and are properly added to the stylesheet, however the heading text size has changed:

Package compiled for production with 3.2.0

kitten:
Resolves #1576 

This will simply upgrade to stylis@^3.5.0 and stylis-rule-sheet@^0.0.10

See: https://github.com/thysultan/stylis.js/issues/99
And: https://github.com/thysultan/stylis.js/blob/master/CHANGELOG.md#350-march-08-2018

### DIFF
--- a/src/test/basic.test.js
+++ b/src/test/basic.test.js
@@ -176,5 +176,18 @@ describe('basic', () => {
 
       expectCSSMatches('.sc-a {} .d { color:red; } .sc-b {} .c { color:blue; }')
     })
+
+    it('handle media at-rules inside style rules', () => {
+      const Comp = styled.div`
+        > * {
+          @media (min-width: 500px) {
+            color: pink;
+          }
+        }
+      `
+
+      shallow(<Comp />)
+      expectCSSMatches('.sc-a{ } @media (min-width:500px){ .b > *{ color:pink; } } ')
+    })
   })
 })
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file. If a contri
 
 ## Unreleased
 
-- n/a
+- Fix nested media at-rules by upgrading to stylis@^3.5.0 and stylis-rule-sheet@^0.0.10 (see [#1595](https://github.com/styled-components/styled-components/pull/1595))
 
 ## [v3.2.1] - 2018-03-07
 
--- a/package.json
+++ b/package.json
@@ -75,8 +75,8 @@
     "is-plain-object": "^2.0.1",
     "opencollective": "^1.0.3",
     "prop-types": "^15.5.4",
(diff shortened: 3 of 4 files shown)


src/test/basic.test.js
```diff
@@ -176,5 +176,18 @@ describe('basic', () => {
...
+      const Comp = styled.div`
+        > * {
+          @media (min-width: 500px) {
+            color: pink;
+          }
+        }
+      `
+
+      shallow(<Comp />)
+      expectCSSMatches('.sc-a{ } @media (min-width:500px){ .b > *{ color:pink; } } ')
```
probablyup: Is this right? Shouldn’t it be .sc-a > * inside the media query?

kitten: It's indeed correct 😬 

Unfortunately at-rule are supposed to wrap normal style rules, but often browsers are quite lenient with these kinds of restrictions, since CSS is supposed to be interpreted in a non-restrictive way.

But `insertRule` will be more strict with these kinds of mistakes, so we have to reverse the order. This was already in place by stylis-rule-sheet, but it'd insert an extra closing curly brace before

Issue #1598, opened 2018-03-11
Better IS_BROWSER check


danieldunderfelt:
I am working on a server-side rendered project (with Razzle) that uses [node-pixi](https://github.com/timknip/node-pixi) for rendering Pixi scenes to an image on the server. When I tried to start using styled.components, I got the error `TypeError: document.querySelectorAll is not a function` as styled-components tried to rehydrate the styles... on the server.

I investigated, and it turns out that styled-components uses a very simple `typeof window !== 'undefined'` to determine whether or not the current environment is a browser. It also turns out that, in order to make PIXI work server-side, node-pixi assigns a shimmed Window object onto `global.window`, making styled-components think that the server is a browser.

You may argue that node-pixi should change. There are a few reasons why it's better to change styled-components:

- It's much easier, since we can just change one line of code. Who knows to what extend PIXI uses the window object.
- styled-components is arguably the more "general" library, and I assume that the maintainers want it to work in a variety of weird situations.

So I propose the following change, in `src/constants.js` on [line 9](https://github.com/styled-components/styled-components/blob/19c98575466218b8f7a0110bc13ed2844519da91/src/constants.js#L9):

```
// Instead of:
export const IS_BROWSER = typeof window !== 'undefined'

// Do this:
export const IS_BROWSER = typeof window !== 'undefined' && window.window === window
```

To shim the window, which may be required in certain cases like PIXI's, you hardly need to do the circular assignment. That's why this is a much safer test that will make styled-components work more reliably.

What do you think? I might make a PR since I'm gonna fork SC now and get my app working again.

EDIT: As I was implementing this, the original `this.window === this` didn't pass the tests, so I settled on a similar solution that does pass the tests and works as intended.

mxstbr:
I like that!

danieldunderfelt:
@mxstbr Cool! I'm trying to implement a better check now, but it's not as easy as I thought it'd be when I wrote the issue. I'll update the issue and make a PR when I have something that works!

kitten:
Resolved by #1599 and released in https://github.com/styled-components/styled-components/releases/tag/v3.2.2

PR #1599, opened 2018-03-11
Skip 'this' for a better window check that actually works and passes tests

danieldunderfelt:
This PR implements the safer `IS_BROWSER` test described in #1598.

In a nutshell, if the `window` object is shimmed for some reason in node environments, styled-components will try to rehydrate the styles on the server. This PR implements a more robust browser check to ensure that no browser code will run server-side, even in shimmed environments.


danieldunderfelt:
@mxstbr Dang... I would think so, but I'll find out to be 100% sure!

mxstbr:
Thanks, much appreciated!

danieldunderfelt:
@mxstbr Yes, it's a thing on IE9! Tested in BrowserStack.

danieldunderfelt:
@kitten Good point. Will this work better?

danieldunderfelt:
@kitten Ha! At first I thought I was being an idiot for making such an obvious thinking error, and I was, but I simultaneously found the answer! We'll just switch the && to a ||, which is what I was THINKING of, and that should work in all cases. Right?

src/constants.js
```diff
@@ -6,7 +6,10 @@ export const SC_ATTR = 'data-styled-components'
 export const SC_STREAM_ATTR = 'data-styled-streamed'
 export const CONTEXT_KEY = '__styled-components-stylesheet__'
 
-export const IS_BROWSER = typeof window !== 'undefined'
+export const IS_BROWSER =
+  typeof window !== 'undefined' ||
+  // eslint-disable-next-line no-new-func
+  new Function('try {return this===window;}catch(e){ return false;}')()
```
kitten: Sorry it’s me again ;) so we have to get rid of Function and the OR.

With the OR the same false positives will occur since the second check won’t be applied when the first succeeds.

danieldunderfelt: Damn, you're right. I had tested this in my application, but maybe it used cached files or somehow not picking up the changes.

Many good suggestions in the commit thread, I'll test them.

danieldunderfelt:
@mxstbr awesome, thank you!! 😄

PR #1606, opened 2018-03-13
Added failing test for SSR

edqd:
There is a new failing test to show whether the styles generated
for different request are shared as discussed in https://spectrum.chat/thread/536fcaf2-138c-41bd-aa9b-165dcf2569ef


mxstbr:
Ref #1605 which should fix this test, /cc @kitten

kitten:
Yep, that'll be taken care of by #1605

PR #1631, opened 2018-03-23
Remove deprecated theme channel

### LINKED ISSUE #1618: Remove deprecated ThemeProvider channel
We have had this deprecated theme channel for a while now, and it should be removed.

It's quite a simple task since nothing should be depending on it anymore, but it does involve some "housekeeping" task of checking whether there's any leftovers, renaming the new `CHANNEL`, and checking libraries that might depend on it still.

https://github.com/styled-components/styled-components/blob/master/src/models/ThemeProvider.js#L80-L88

matedeilo:
Resolve #1618

Hi, This is my first PR. Trying to remove channel deprecated.

### DIFF
--- a/src/hoc/withTheme.js
+++ b/src/hoc/withTheme.js
@@ -5,7 +5,6 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import hoistStatics from 'hoist-non-react-statics'
 import {
-  CHANNEL,
   CHANNEL_NEXT,
   CONTEXT_CHANNEL_SHAPE,
 } from '../models/ThemeProvider'
@@ -29,7 +28,6 @@ const wrapWithTheme = (Component: ReactClass<any>) => {
     static styledComponentId = 'withTheme'
 
     static contextTypes = {
-      [CHANNEL]: PropTypes.func,
       [CHANNEL_NEXT]: CONTEXT_CHANNEL_SHAPE,
     }
 
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -15,7 +15,7 @@ import escape from '../utils/escape'
 import type { RuleSet, Target } from '../types'
 import { CONTEXT_KEY } from '../constants'
 
-import { CHANNEL, CHANNEL_NEXT, CONTEXT_CHANNEL_SHAPE } from './ThemeProvider'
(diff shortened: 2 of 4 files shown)


src/hoc/withTheme.js
```diff
@@ -5,7 +5,6 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import hoistStatics from 'hoist-non-react-statics'
 import {
-  CHANNEL,
   CHANNEL_NEXT,
```
kitten: We can safely rename this to `CHANNEL` now, since the old one is removed

kitten:
@matedeilo Hiya, thanks for picking this up; see the above comment

Also we need to update withTheme as well: https://github.com/styled-components/styled-components/blob/master/src/hoc/withTheme.js

Can you also include a changelog entry please?

matedeilo:
Hi @kitten  yes, no problem.

probablyup:
@matedeilo hi hi, are you able to pick this up to finish it out?

alexandernanberg:
I could pick this up if @matedeilo is unable. Do I create a new PR in that case or do I layer the commits on this PR somehow?

matedeilo:
@alexandernanberg feel free to finish this.

alexandernanberg:
Cool! @probablyup How do you think will be the best way to accomplish this? The absolute easiest way would to just create a new PR without the existing commits but then they would obviously be lost. Otherwise I guess I need write access on the fork?

mxstbr:
> Otherwise I guess I need write access on the fork?

Pull these commits, then do a `git checkout` to a new branch—which'll preserve the existing commits, but you can then push them to your own fork!

probablyup:
Now that we’re edging toward 4.0 work this should just be superseded by
switching to the new context api I think

alexandernanberg:
Okay, yea seems smarter to jump on the new Context API directly in that case. I've begun migrating SC to that on a local branch but I've hit some problems with Enzyme and `StyledComponent.js`, is it worth opening a PR even though I'm not sure I'll be able to tackle this alone?

probablyup:
That’s up to you! If you’ve started the work we definitely won’t say no to
getting the ball rolling. The final product might look a bit different. I’m
going to make a “develop” branch for 4.0 sometime today

probablyup:
Superceded by the new context API work merged into the develop (4.0) branch

Issue #1032, opened 2017-07-25
Two instances of styled-components conflict when loaded on the same page


ryanheathers:
## Background
I develop a React app that gets installed as a 3rd-party plugin. That means multiple instances of my app may be running on a single page at any time. Along the way I've run into issues with how styled-component's shares the generated hashed css class amongst components on the page. If the same component is used twice, even if they are part of different React apps, they share the same class. This inevitably leads to style conflicts and specificity issues. [This issue](https://github.com/styled-components/styled-components/issues/633) has more background.

I've attempted - and largely succeeded - in working around this issue by the use of a style helper that wraps each styled-component in an ID selector. That way my generated styles look something like this:

```
#carousel-f75e1d05 .jaaqKx,
#carousel-f75e1d05-modal .jaaqKx {
    color: inherit;
    fill: currentColor;
    width: inherit;
    height: inherit;
}
```

Here is my helper function:

```
// Note: I've previously added a rootId UUID to my theme via code not shown here

export function rootIdWrapper(cssToUse) {
  return (props) => {
    return css`
      #${props.theme.rootId} &,
      #${props.theme.rootId}-modal & {
        ${cssToUse}
      }
    `;
  };
}
```

And here is how I use it:
```
const Svg = styled.svg`
  ${rootIdWrapper(css`
    color: inherit;
    fill: currentColor;
    width: inherit;
    height: inherit;
  `)}
`;
```

While it would be amazing to have first-class support for this kind of namespacing in styled-components itself, this slightly laborious workaround solves the issue.

## Expected Behavior
The styles for each of my app instances will be properly added to the page at run time.

## Actual Behavior
This code works great, that is, **when using styled-components v2.0.0-13.**

If I upgrade to a later version of styled-components, or revert to the base v2.0.0 version, I get buggy behavior:
- If **only one** of my React apps is loaded on the page, things are fine
- However if two or more instances are loaded, only the styles from one of the instances are loaded into the `<head>`. The other styles do not appear on the page at all.

## Steps to reproduce
It's difficult to provide an easily shareable replication of the issue since it involves concurrently running two React instances on a page. In my troubleshooting, I have produced a minimal example that uses the exact same code with the only difference being the version of styled-component's being used. **The version running 2.0.0-13 works great, any other 2.x version exhibits buggy behavior.**

Appreciate any insights out there on what the issue might be!

philpl:
[...]

PR #1632, opened 2018-03-23
Add ability to change SC_ATTR via process.env

Fer0x:
This PR will fix https://github.com/styled-components/styled-components/issues/1032

I have similar problem when embedding our widget to sites using styled-components.
Even when StyleSheetManager used inside iframe for isolating styles, styled-components inserts empty `<style />` into main `window.head`. As @prevostc mentioned, providing custom `SC_ATTR` fix conflict and styles no longer break.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -2,7 +2,7 @@
 
 declare var __DEV__: ?string
 
-export const SC_ATTR = 'data-styled-components'
+export const SC_ATTR = process.env.SC_ATTR || 'data-styled-components'
 export const SC_STREAM_ATTR = 'data-styled-streamed'
 export const CONTEXT_KEY = '__styled-components-stylesheet__'
 
--- a/src/constants.js
+++ b/src/constants.js
@@ -2,7 +2,9 @@
 
 declare var __DEV__: ?string
 
-export const SC_ATTR = process.env.SC_ATTR || 'data-styled-components'
+export const SC_ATTR =
+  (typeof process !== 'undefined' && process.env.SC_ATTR) ||
+  'data-styled-components'
 export const SC_STREAM_ATTR = 'data-styled-streamed'
 export const CONTEXT_KEY = '__styled-components-stylesheet__'
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file. If a contri
 
 ## Unreleased
 
-- n/a
+- Added ability to override `SC_ATTR` via `process.env.SC_ATTR`
 
 ## [v3.2.3] - 2018-03-14
 
--- a/src/test/constants.test.js
+++ b/src/test/constants.test.js
@@ -1,0 +1,43 @@
+
+import React from 'react'
+import { shallow } from 'enzyme'
+
+import { expectCSSMatches } from './utils'
+import { SC_ATTR as DEFAULT_SC_ATTR } from '../constants';
+
+
+function renderAndExpect(expectedAttr) {
+  const SC_ATTR = require('../constants').SC_ATTR
+  const styled = require('./utils').resetStyled()
+
+  const Comp = styled.div`
+      color: blue;
+    `
+
+  shallow(<Comp />)
+
+  expectCSSMatches('.sc-a { } .b { color:blue; }')
+
+  expect(SC_ATTR).toEqual(expectedAttr)
+  expect(document.head.querySelectorAll(`style[${SC_ATTR}]`)).toHaveLength(1)
+}
+
+describe('constants', () => {
+  it('should work with default SC_ATTR', () => {
+    renderAndExpect(DEFAULT_SC_ATTR)
+  })
+
+  it('should work with custom SC_ATTR', () => {
+    const CUSTOM_SC_ATTR = 'data-custom-styled-components'
+    process.env.SC_ATTR = CUSTOM_SC_ATTR
+    jest.resetModules()
+
+    renderAndExpect(CUSTOM_SC_ATTR)
+
+    delete process.env.SC_ATTR
+  })
+
+  afterEach(() => {
+    jest.resetModules()
+  })
+});
--- a/src/test/constants.test.js
+++ b/src/test/constants.test.js
@@ -1,18 +1,16 @@
-
 import React from 'react'
 import { shallow } from 'enzyme'
 
 import { expectCSSMatches } from './utils'
-import { SC_ATTR as DEFAULT_SC_ATTR } from '../constants';
-
+import { SC_ATTR as DEFAULT_SC_ATTR } from '../constants'
 
 function renderAndExpect(expectedAttr) {
   const SC_ATTR = require('../constants').SC_ATTR
   const styled = require('./utils').resetStyled()
 
   const Comp = styled.div`
-      color: blue;
-    `
+    color: blue;
+  `
 
   shallow(<Comp />)
 
@@ -40,4 +38,4 @@ describe('constants', () => {
   afterEach(() => {
     jest.resetModules()
   })
-});
+})


Fer0x:
Is issue https://github.com/styled-components/styled-components/issues/1032 clear for maintainers?

mxstbr:
How would you use this to fix #1032?

Fer0x:
@mxstbr if second instance of styled-components has different `SC_ATTR`, then both styled-components work without breaking styles of each other.

mxstbr:
I'm not sure how that helps: you have two instances in your build somewhere, how are you going to set a different environment variable for each of them

Fer0x:
@mxstbr 
Look, I have React-based application, which uses styled-components. That application is embeddable communication widget (like Intercom or Facebook's Messenger Widget).

It looks like that: 

image


So, widget can be inserted in any site, written with any technology. To isolate widget styles from site's own, StyleSheetManager is used together with local frame, created by React Portal.

But some sites also use React with styled-components. And here comes problem with two instances of styled-components - after widget injected, some site's styles disappeared. And that problem solved if widget uses custom `SC_ATTR`, which can be given in build stage.

ryanheathers:
I was the one who originally opened https://github.com/styled-components/styled-components/issues/1032.

Happily, the solution in this PR appears to be working! **Huge thanks,** @Fer0x! 🙌  After passing a different `process.env.SC_ATTR` for the second instance of our app, I was able to run both instances on the same page without issue. 💥 

**For background on why this has been a issue:**
My company makes a widget that gets run on 3rd-party sites. We have no control over the code that the host site is running. There's no way for us to use webpack magic to ensure that only one instance of SC is present on the page. We're currently stuck at styled-components `v2.0.0-13` beta build and can't upgrade beyond that point because it breaks our ability to concurrently run two instances of styled-components. We're currently running the risk that one of the sites that embed our app is also running a new-ish version of styled-components, in which case I think everything would break, but so far that has not happened thankfully.

While I understand that the [official position of styled-components is that multiple instances of styled-components is not a best practice](https://www.styled-components.com/docs/faqs#why-am-i-getting-a-warning-about-several-instances-of-module-on-the-page), that perspectives ignores the use case of embeddable widgets, which cannot control the code of the host site.

Getting this PR merged would be amazing. If there's anything I can do to help, let me know.

mxstbr:
That use case makes sense to me. /cc @kitten @probablyup

src/constants.js
```diff
@@ -2,7 +2,7 @@
 
 declare var __DEV__: ?string
 
-export const SC_ATTR = 'data-styled-components'
+export const SC_ATTR = process.env.SC_ATTR || 'data-styled-components'
```
kitten: I think this might lead to minor issues in some setups where not all `process.env` variables are replaced; specifically setups where only `NODE_ENV` is set up.

@mxstbr We might want to either test this against such a setup or just use `typeof process !== 'undefined'` for now. Agreed?

Fer0x: @kitten rollup already defines `process.env.NODE_ENV`, so `process.env` can't be undefined.

kitten: We’re using the replace plugin there. So it’ll just search for it as a string and replace it. The same won’t go for this new env variable. Thus I’m pretty sure we can’t make the assumption that process will be defined

Fer0x: @kitten ok, I see it now.

But, if we will use something like this:
```js
export const SC_ATTR = (process && process.env && process.env.SC_ATTR) || 'data-styled-components'
```

after processing with application bundler this code may look like that:

```js
var SC_ATTR = (process && process.env && 'data-some-custom-styled-components') || 'data-styled-components'
```

And in that case we rely on the fact that `process` defined in runtime, which is not guaranteed with bundlers like rollup. Now I think, that custom `process.env` is not good for browser libraries.

Another question: Is it necessary to have `data-` within `SC_ATTR`?

Fer0x:
@mxstbr @kitten 

Ok, there is another option, more complicated in terms of implementation.

```js
// constants.js
const customizable = {
  SC_ATTR: 'data-styled-components'
};

export const SC_ATTR = () => customizable.SC_ATTR

// ...

export const configure = options => Object.assign(customizable, options);


// usage in app
import { configure } from 'styled-components';

configure({
  SC_ATTR: 'data-custom-styled-components'
})
```

This will allow us to avoid possible issues with `process.env`, but requires minor refactoring due to `SC_ATTR` will became function.

kitten:
@Fer0x We're not going to go down this route since we don't believe in configurability in this way. We have actually discussed this multiple times before and have decided against it every single time 😅 Sorry

I'd be happy to go forward with `SC_ATTR` however.

ryanheathers:
> We're not going to go down this route since we don't believe in configurability in this way. We have actually discussed this multiple times before and have decided against it every single time

@kitten, any links to prior discussion on this? I think runtime configuration could be helpful. Adopting `process.env.SC_ATTR` is already a step toward config, so having a standard approach seems beneficial over one-off config hooks.

> That use case makes sense to me

@mxstbr great to hear that! Further on this point, while I think that adding `process.env.SC_ATTR` is a great step, I still believe that styled-components should move to an approach that doesn't touch any globals (or allows configuration so that SC can be easily scoped). Other browser libraries do this because it opens up more use cases. For example, React only touches the DOM child nodes of the `reactroot` node that it gets fed, jQuery has `noConflict()`, etc etc.

A couple weeks back I started looking at how to retrofit SC to not touch any globals. I started with using the `<StyleSheetManager target={my-root-node}` approach and changing any `document.querySelectorAll` references to `element.querySelectorAll` that were scoped to the same node as the `target`. However I was still running into conflicts between multiple instances of SC that were hard to trace. Probably because the `SC_ATTR` namespace was being used in ways I wasn't accounting for, which is why @Fer0x's fix sidesteps the issue.

I was hoping to be able to submit a PR that handles the globals issue completely but no luck so far. I may keep trying, but I wanted to drop a note here as food for thought to you and the other maintainers, because I think that this "global safe" approach for SC would allow more people to use your great library.

kitten:
@ryanheathers It mostly comes down to us not believing in configurability. This is already a case that we'd love to support, but that's not part of our major userbase. I'm not going to or trying to say that your usecase is not equally important to us, but I don't believe that warrants an individual config hook that's basically a workaround and inelegant on its own. We might supports this when or if the CSS pipeline is separated from the main package.

So, the tl;dr being said, this is my own opinion but–as you requested-there's more here: https://github.com/styled-components/styled-components/pull/1143 (There were some more issues on this but I'm unable to find them, sorry 😢)

> For example, React only touches the DOM child nodes of the reactroot node that it gets fed

That'd be great, but I believe we could easily solve this with a "rehydration provider". I was thinking about this but I'll open a separate RFC for it soon :wink:

> I was hoping to be able to submit a PR that handles the globals issue completely but no luck so far. I may keep trying

Truly sorry, the StyleSheet logic is unfortunately quite complex and I've recently refactored it to simplify it for new contributors. I'd love to brainstorm this, but I think I'll propose an RFC so that we can have a longer discussion on this.

So as said above, I think we can solve multiple problems at one by modifying style injection an rehydration to use the `StyleSheetManager` (optionally). This would also solve the problem that the manager doesn't use the `target` prop quite right. It's quite a hacky workaround (doesn't handle component creation order, doesn't work with SSR, etc).

ryanheathers:
@kitten, cool thanks for the link on the config discussion. Enlightening. I agree that config is very challenging, and fwiw I was not suggesting opening a config pipeline just to support my use case of running multiple instances of SC on a single page. A config pipeline would only make sense if you had other things to toggle, which it sounds like you might based on the discussion in https://github.com/styled-components/styled-components/pull/1143, but I understand why you've moved away from that direction. For my short-term purposes, I'll be quite happy to just have the `process.env.SC_ATTR` support merged in. 😛 

> So as said above, I think we can solve multiple problems at one by modifying style injection an rehydration to use the StyleSheetManager (optionally).

That sounds promising. I'll keep an eye out for that discussion.

Fer0x:
@kitten Can you review this PR again? Your stale review was dismissed by cleaning code and resolve merge conflicts.

mxstbr:

Fer0x:
@mxstbr Thanks a lot! Next I will try to look at refactor `keyframes` to better work with `StyleSheetManager`.

mxstbr:
:100:

kitten:
@Fer0x please take a look at https://github.com/styled-components/styled-components/issues/1333 before you jump onto that :wink: and maybe don't make `StyleSheetManager` your goal immediately, as it's a half-baked feature (especially with the `target` prop) as of now.

PR #1635, opened 2018-03-23
Eliminate the short-circuiting of the ComponentStyle stylesheet-registration check

### LINKED ISSUE #1634: Style Cache not invalidated when StyleSheetManager target is removed
## System:
 - CPU: x64 Intel(R) Core(TM) i7-2860QM CPU @ 2.50GHz
 - Free Memory: 494.87 MB
 - Total Memory: 8.00 GB
 - Shell: /usr/local/bin/fish - 2.7.1
## Binaries:
 - Node: 8.9.4
 - Yarn: 1.5.1
 - npm: 5.7.1
 - Watchman: 4.7.0
### styled-components:
 - wanted: ^3.2.3
 - installed: 3.2.3
## Reproduction

I'm using styled-components within an iframe on a PowerBI visual. These are instantiated pretty frequently as you switch around report pages. I'm using the StyleSheetManager to target my iframe and inject styles into the head. This works the first time, but when this frame re-renders the styles are gone from the iframe document, but styled-components is still holding onto the cached names.

## Steps to reproduce
Working on a test to reproduce this

## Expected Behavior
Styled-components should inject styles into newly created iframes if the old one disappears.

Perhaps ComponentStyle checks that the StyleSheetManager instance is the same instance it registered its styles with before?

## Actual Behavior
styled-components has cached stylenames, but those style names are no longer available.

darthtrevino:
Fixes #1634

When `ComponentStyle.generateAndInjectStyles` is invoked, the stylesheet instance should be checked even if the style name is already defined. If the component is being rendered in a sibling iframe  with a separate StyleSheetManager, this allows the second iframe to receive the injected styles.


src/models/test/StyleSheetManager.test.js
```diff
@@ -86,6 +100,69 @@ describe('StyleSheetManager', () => {
...
+        this.props.resolve()
+      }
+      render() {
+        return <Title />
+      }
+    }
+
+    const Structure = ({ resolve }) => (
+      <Frame>
+        <SheetInjector>
```
kitten: So you're expecting this to run with the styles in the main window?

I'm not sure whether I can get this to work without understanding what `react-frame-component` does precisely. I probably won't have time soon to look at it, but I'll try.

Generally, are you trying to write `react-frame-component` for styled-components?

darthtrevino:
Nope, I haven't seen that PR before. So it looks like the cache-check line is here: https://github.com/styled-components/styled-components/blob/master/src/models/ComponentStyle.js#L81 

Do you think there's a way to inject the active stylesheet into this check to see if the style is already registered?

src/models/test/StyleSheetManager.test.js
darthtrevino: So react-frame-component is a common library to use to put iframes into a React tree. I want to style things inside of the iframe, and to make that work you need to use a StyleSheetManager and target the iframe's head. The end-result should be that the iframe head has the styles on each render.

kitten:
@darthtrevino 

> Do you think there's a way to inject the active stylesheet into this check

That'd be unexpected behaviour already; so I'm not sure what that'd achieve? 😅 

>  to see if the style is already registered

Do you mean the generated rule? That's already checked by `styleSheet.inject`, which uses the StyleSheet's cache to check whether the component and styles are known

darthtrevino:
So when I debug this, the cache-check line is passing because lastClassName is defined. (it was defined in the first render), but really, we want to know "is a class name defined, and this class name registered on the stylesheet"

darthtrevino:
Huh, it looks like it's doing that here https://github.com/styled-components/styled-components/blob/master/src/models/ComponentStyle.js#L88

kitten:
@darthtrevino so, a changing stylesheet during the runtime is not supported. I think I'm missing some pieces of information here of what is supposed to happen.

> I'm using the StyleSheetManager to target my iframe and inject styles into the head. This works the first time, but when this frame re-renders the styles are gone from the iframe document, but styled-components is still holding onto the cached names.

Ok, so your iframe unmounts, why does the StyleSheetManager stay around? Is it not part of the iframe? How will it receive the styles from the iframe? Do they share a js context, etc?

> Styled-components should inject styles into newly created iframes if the old one disappears.

So, here is where I think we're misunderstanding each other. Since this is not expected behaviour, your comment here confuses me a little:

> So when I debug this, the cache-check line is passing because lastClassName is defined

You have to make sure that the entire stylesheet changes. The `StyleSheetManager` should be doing that as long as the sheet is passed down in the context. Otherwise you'll be falling back to the global `StyleSheet.master`. Could this be what is happening?

darthtrevino:
@kitten Can you take a look at the latest commit on ComponentStyle? I think this should solve it.

darthtrevino:
So PowerBI is an analytics application. You can install custom data visualizations to have unique views into your data. These visuals are isolated into iframes to prevent their styles from colliding with the app and each other. As you interact with the application, the visuals are reconstructed repeatedly in their lifecycles. This means iframe sandboxes are recreated as you switch between report pages. 

As far as styled-components is concerned, if it's hosted in an iframe it needs to use the StyleSheetManager. This makes sure the styles flush out to the correct location. The problem I was running into is that once they flushed the first time correctly, new iframes would not have their styles loaded. This problem would affect any application using the same component in different iframes, so my test case here is basically two sibling iframes with the same react tree.

darthtrevino:
I'll cut a new PR that cleans up the whitespace changes

PR #1636, opened 2018-03-23
Fix ComponentStyle caching strategy to take StyleSheet cache into account

### LINKED ISSUE #1634: Style Cache not invalidated when StyleSheetManager target is removed

darthtrevino:
Fixes #1634, Whitespace-change free version of PR #1635

When ComponentStyle.generateAndInjectStyles is invoked, the stylesheet instance should be checked even if the style name is already defined. If the component is being rendered in a sibling iframe with a separate StyleSheetManager, this allows the second iframe to receive the injected styles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file. If a contri
 
 ## Unreleased
 
-- Fix issue with multiple iframes and StyleSheetManager instances. @darthtrevino (see [#1634])
+-  Fix ComponentStyle caching strategy to take StyleSheet cache into account. @darthtrevino (see [#1634])
 
 ## [v3.2.3] - 2018-03-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file. If a contri
 
 ## Unreleased
 
-- n/a
+- Fix issue with multiple iframes and StyleSheetManager instances. @darthtrevino (see [#1634])
 
 ## [v3.2.3] - 2018-03-14
 
--- a/src/models/ComponentStyle.js
+++ b/src/models/ComponentStyle.js
@@ -1,13 +1,7 @@
 // @flow
 import hashStr from '../vendor/glamor/hash'
 
-import type {
-  RuleSet,
-  NameGenerator,
-  Flattener,
-  Stringifier,
-  Interpolation,
-} from '../types'
+import type { RuleSet, NameGenerator, Flattener, Stringifier } from '../types'
 import StyleSheet from './StyleSheet'
 import { IS_BROWSER } from '../constants'
 import isStyledComponent from '../utils/isStyledComponent'
@@ -84,23 +78,20 @@ export default (
      * */
     generateAndInjectStyles(executionContext: Object, styleSheet: StyleSheet) {
       const { isStatic, componentId, lastClassName } = this
-      let flatCSSResult: ?Array<Interpolation> = null
-      const flatCSS = (): Array<Interpolation> => {
-        if (!flatCSSResult) {
-          flatCSSResult = flatten(this.rules, executionContext)
-        }
-        return flatCSSResult
+      if (
+        areStylesCacheable &&
+        isStatic &&
+        lastClassName !== undefined &&
+        styleSheet.hasNameForId(componentId, ((lastClassName: any): string))
+      ) {
+        return lastClassName
       }
 
-      const useLastClassName: boolean =
-        areStylesCacheable && isStatic && lastClassName !== undefined
-
-      const name: string = useLastClassName
-        ? ((lastClassName: any): string)
-        : generateRuleHash(this.componentId + flatCSS().join(''))
+      const flatCSS = flatten(this.rules, executionContext)
+      const name = generateRuleHash(this.componentId + flatCSS.join(''))
 
       if (!styleSheet.hasNameForId(componentId, name)) {
-        const css = stringifyRules(flatCSS(), `.${name}`)
+        const css = stringifyRules(flatCSS, `.${name}`)
         styleSheet.inject(this.componentId, css, name)
       }
 
--- a/src/models/ComponentStyle.js
+++ b/src/models/ComponentStyle.js
@@ -1,7 +1,13 @@
 // @flow
 import hashStr from '../vendor/glamor/hash'
 
-import type { RuleSet, NameGenerator, Flattener, Stringifier } from '../types'
+import type {
+  RuleSet,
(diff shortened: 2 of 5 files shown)


kitten:
@darthtrevino Changing the StyleSheet while the component is mounted — essentially hot swapping the sheet — is not something we should support, just to be clear again. However, this is a reasonably small change that I can maybe accept, but we'll have to regression test it before release. Hence a small disclaimer: Sorry in advance if this change doesn't land :wink:

The current state of the code breaks an important caching optimisation, as it runs a bunch of operations that we're trying to avoid just to fix this. I'll leave some comments on the code itself since there's another way.

src/models/test/StyleSheetManager.test.js
```diff
@@ -116,6 +119,69 @@ describe('StyleSheetManager', () => {
     expect(document.body.innerHTML).toMatchSnapshot()
   })
 
+  // https://github.com/styled-components/styled-components/issues/1634
+  it('should flush styles to an iframe when the iframe is re-rendered', async () => {
```
kitten: This test is extremely specific to your environment and additionally this conflicts with it as well:

> Changing the StyleSheet while the component is mounted — essentially hot swapping the sheet — is not something we should support,

Also it's a little unnecessarily complex. Instead we can use enzyme to swap out the the stylesheet itself together with a prop, and expect to see a different class.

Edit: I'd say it's ok to not test this for now. I can write a test eventually, but one that is specific to all caching guarantees instead

src/models/ComponentStyle.js
```diff
@@ -78,12 +78,14 @@ export default (
      * */
     generateAndInjectStyles(executionContext: Object, styleSheet: StyleSheet) {
       const { isStatic, componentId, lastClassName } = this
-      if (areStylesCacheable && isStatic && lastClassName !== undefined) {
-        return lastClassName
-      }
-
```
kitten: Here you're leading the code into expensive operations that we're trying to avoid

src/models/ComponentStyle.js
```diff
@@ -78,12 +78,14 @@ export default (
...
       const flatCSS = flatten(this.rules, executionContext)
-      const name = generateRuleHash(this.componentId + flatCSS.join(''))
+
+      const useLastClassName: boolean =
```
kitten: This goes through all operations even if it's not necessary too

darthtrevino:
I don't believe this will hot-swap stylesheets for mounted and styled components. This is kind of analogous to a Java thread-local - it's two components of the same type in totally different stylesheet contexts.

src/models/ComponentStyle.js
```diff
@@ -78,12 +78,14 @@ export default (
      * */
     generateAndInjectStyles(executionContext: Object, styleSheet: StyleSheet) {
       const { isStatic, componentId, lastClassName } = this
-      if (areStylesCacheable && isStatic && lastClassName !== undefined) {
```
kitten: This is the crux of it, so instead you can replace this: 

`if (areStylesCacheable && isStatic && lastClassName !== undefined) {`

with this:

`if (areStylesCacheable && isStatic && lastClassName !== undefined && styleSheet.hasNameForId(componentId, lastClassName)) {`

a lot simpler, doesn't break our caching and optimisation, and it should have the same effect :+1:

so essentially this is the only change that is needed

darthtrevino:
I'm guessing the flatten() is what you're trying to avoid?

kitten:
@darthtrevino 

> I don't believe this will hot-swap stylesheets for mounted and styled components. This is kind of analogous to a Java thread-local - it's two components of the same type in totally different stylesheet contexts.

No, this code change won't. Essentially you're just invalidating the cache for a new stylesheet. Whether that's hot swapped through the context, or because you're suddenly in a new context with the same ComponentStyle doesn't matter that much, because it won't occur during normal usage.

I think it's a fair addition though, so I've left some comments above and I think we can ship this

darthtrevino:
No problem, I get that it's a weird use case. I put some additional context at the tail of the last PR

kitten:
@darthtrevino just seen your last changes. Haven’t given it a quick run yet but see you sure that the change I proposed here won’t be enough? https://github.com/styled-components/styled-components/pull/1636#comment-176869902

Because essentially we only have to prevent returning the lastClassName if it’s not on the StyleSheet so that one line change should be enough

darthtrevino:
@kitten Switching to your suggestion - I think that will be fine

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file. If a contri
 
 ## Unreleased
 
-- n/a
+- Fix issue with multiple iframes and StyleSheetManager instances. @darthtrevino (see [#1634])
```
kitten: Can we make this generic as well like the test?

I.e. fix last className caching strategy to take StyleSheet cache into account

Or sth

kitten:
@darthtrevino thanks for making this change this quickly 👏 hope it will help more ppl with similar edge cases.

I’ll give it a couple of tests tomorrow and will push a prerelease. If that looks fine I’ll publish a patch :)

bliitzkrieg:
This fix should resolve a similar issue I am having with a different use case. I created an embedded component system that mounts a react component (bundled with styled components) on html markup on any given client. The components wrap a StyleSheetManager that use the element its mounting as the target and when I programically unmount/mount the component for advanced usage, the styles disappear.

kitten:
@bliitzkrieg yes, that might be fixed as well by this. Quick note that what you're doing will break specificity and certain guarantees (see previous comments) :wink:

mxstbr:

PR #1639, opened 2018-03-25
Fix Typescript SSR Typings

Lavoaster:


typings/styled-components.d.ts
```diff
@@ -102,7 +102,7 @@ interface StylesheetComponentProps {
 }
 
 interface StyleSheetManagerProps {
-  sheet?: ServerStyleSheet;
+  sheet?: StyleSheet;
```
kitten: This is definitely correct, as opposed to what it was before.

However, it seems we don't have any typings for our stylesheet class.
It's probably falling back to the browser's StyleSheet's TS typings?

It's technically not a public API but would you mind to submit some rudimentary typings please? If not I'd be happy to add them myself later

Lavoaster: Yeah, this is falling back to browsers StyleSheet in TS. I only updated it to match the return type of ServerStyleSheet.instance.

I wasn't aware it was using an internal class, after having a quick gander at it I'm not comfortable writing that one. I could however implement a stub interface so that it doesn't fallback to the browser stylesheet typings. Is that something you'd prefer over the current state?

kitten: Both is fine; either way, let’s make that part of a different PR maybe

mxstbr:

PR #1656, opened 2018-04-03
Add support for react-tree-walker

ewolfe:
In certain situations such as react-tree-walker, it's possible for `this.context` to be undefined. This simply adds a conditional check before proceeding.


kitten:
First of all, thanks for taking the time to dig deeper into this issue. However, I don’t see why we would be merging a patch for a library that doesn’t behave like (P)React?

We can go ahead with this and I’ll leave a note on the PR but as of now I’m a little unsure whether this is truly the right place to submit a fix 😅 Maybe it’d be more fitting to fix treewalker?

cc @mxstbr

src/models/StyledComponent.js
```diff
@@ -121,43 +121,45 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
 
     componentWillMount() {
       const { componentStyle } = this.constructor
-      const styledContext = this.context[CHANNEL_NEXT]
+      if (this.context) {
```
kitten: This will have to be more explicit so v8 doesn’t deopt this check. “!== undefined” will do

src/models/StyledComponent.js
```diff
@@ -121,43 +121,45 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
...
-      if (componentStyle.isStatic) {
-        const generatedClassName = this.generateAndInjectStyles(
-          STATIC_EXECUTION_CONTEXT,
-          this.props
-        )
-        this.setState({ generatedClassName })
-        // If there is a theme in the context, subscribe to the event emitter. This
-        // is necessary due to pure components blocking context updates, this circumvents
-        // that by updating when an event is emitted
-      } else if (styledContext !== undefined) {
```
kitten: We already have a check here so it might be better to just expand it

ewolfe:
Thanks for reviewing @kitten -- you're right, this probably isn't truly the right place for it. I'll go ahead and close.

kitten:
@ewolfe alright, feel free to keep the branch around if it becomes necessary after all :)

PR #1658, opened 2018-04-04
Use react-is.isValidElementType to validate React elements in constructWithOptions

martahj:
This PR changes the way tags are validated in `constructWithOptions` from checking that they are either a string or function to using the validator exposed by [react-is](https://www.npmjs.com/package/react-is), `isValidElementType`. In support of that goal, the PR also adds react-is as a dependency and adds a small amount of config for it in rollup.config.js.

The immediate motivation behind this PR is that React's forwardRef api [returns an object](https://github.com/facebook/react/issues/12453) rather than a function, so was being rejected by `constructWithOptions`. Using the officially supported element validator may also help future-proof against further React api changes.


martahj:
Added a changelog entry 👍

src/constructors/constructWithOptions.js
```diff
@@ -1,4 +1,5 @@
 // @flow
+import * as ReactIs from 'react-is'
```
kitten: Can you destructure this import please to make it more explicit? We really only need to import `isValidElementType`

mxstbr:
I don't think this deprecates `innerRef`? It just makes it so `innerRef` will keep working even with React 16 afaiu

kitten:
@mxstbr Oh, that makes sense. I misunderstood parts of the roadmap then. :sweat_smile: Either way, let's not deprecate it then, but let's try to answer the remaining questions to be able to land this PR 👍

petegleeson:
@kitten I believe the usage of `isValidElementType` is correct here. The method returns false for things like `<MyComponent />`. I put together a little codesandbox that shows the behaviour of `isValidElementType` https://codesandbox.io/s/v3v29wqkv7

martahj:
I've added in the changes @petegleeson made in https://github.com/martahj/styled-components/pull/1 (unless something got lost in the merge, the diff made it kind of hard to tell).

I'll add in those tests 👍

src/constructors/constructWithOptions.js
```diff
@@ -7,7 +8,7 @@ export default (css: Function) => {
     tag: Target,
     options: Object = {}
   ) => {
-    if (typeof tag !== 'string' && typeof tag !== 'function') {
+    if (!isValidElementType(tag)) {
```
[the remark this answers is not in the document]
martahj: My understanding is that, roughly,`React.isValidElement` returns true if it's a valid element, while `isValidElementType` returns true if it's a valid component or tag. Comparing the [React.isValidElement test cases](https://github.com/facebook/react/blob/master/packages/react/src/__tests__/ReactJSXElement-test.js) to the examples from the [sandbox](https://codesandbox.io/s/v3v29wqkv7) @petegleeson shared, their behavior is almost opposite:

```js
isValidElementType(ClassComponent) // true
expect(React.isValidElement(Component)).toEqual(false);

isValidElementType(<ClassComponent />) // false
expect(React.isValidElement(<Component />)).toEqual(true);
```

martahj:
Added tests to verify that `styled` does not throw when passing in components returning true from `isValidElementType` but that the function continues to throw when passed elements

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ All notable changes to this project will be documented in this file. If a contri
 
 ## Unreleased
 
+- Use react-is.isValidElementType to check whether elements passed to constructWithOptions are valid (see [#1658]https://github.com/styled-components/styled-components/pull/1658)
```
kitten: Small markdown typo in the link here. Can we also describe what this achieves and not what it changes please? In this case it’d be great to say that the styled constructor has been improved to accept ref forwarding components :)

martahj:
Removed the yarn-error.log and updated the changelog

mxstbr:

Issue #1655, opened 2018-04-02
Remove usages of unsafe lifecycle methods


Hypnosphi:
React team recently published a blogpost: https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

TL;DR: `cWM`, `cWRP`, and `cWU` methods are considered unsafe for future async rendering and will be deprecated soon. The recommended way for open source maintainers to deal with it is to use [`react-lifecycles-compat`](https://github.com/reactjs/react-lifecycles-compat) package and migrate to new `static getDerivedStateToProps` method

camjackson:
Adding the full lifecycle names here for searchability: `componentWillMount`, `componentWillReceiveProps`, `componentWillUpdate`.

mxstbr:
Would love a PR for this change!

I think (correct me if I'm wrong) we should be able to get by without `react-lifecycles-compat` by moving the subscriptions and unsubscription logic to componentDidMount and componentWillUnmount, and then computing the theme in `render()` instead of caching it in state, right?

Hypnosphi:
the goal of `react-lifecycles-compat` is to keep compatibility with older React versions when using `static getDerivedStateFromProps` and `getSnapshotBeforeUpdate`

Hypnosphi:
The goal of 16.3 release is to allow contributors to remove the unsafe methods before they are formally deprecated (with warning messages) in 16.x:

https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#gradual-migration-path

janhoogeveen:
When using the `<StrictMode>` component,  my console is being blasted by warnings like these:

`componentWillMount: Please update the following components to use componentDidMount instead: Anchor__Link, Footer__AuthLinks, Footer__Links, Footer__Logo, Footer__LogoContainer, Footer__StyledFooter, Footer__StyledLink`

Of course I understand that these are just warnings and I'm happy to see the migration path already being discussed here 😄 . But seeing as React is already exposing the API in 16.3 and showing these warnings in the console makes me believe the new lifecycle methods are pretty much locked down already.

tsiq-swyx:
also posting for searchability. my warnings look like these

```
Warning: Unsafe lifecycle methods were found within a strict-mode tree:

componentWillMount: Please update the following components to use componentDidMount instead: styled.div

componentWillReceiveProps: Please update the following components to use static getDerivedStateFromProps instead: styled.div
```

yonatanmn:
[...]

PR #1664, opened 2018-04-06
Remove usages of async-unsafe lifecycles

Hypnosphi:
See #1655


src/models/StyledComponent.js
```diff
@@ -60,9 +61,41 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
...
 
+    static getDerivedStateFromProps(
+      nextProps: { theme?: Theme, [key: string]: any },
+      prevState
+    ) {
+      if (prevState.isStatic) {
+        return prevState.generatedClassName === ''
+          ? // First run
+            {
+              generatedClassName: prevState.generateAndInjectStyles(
```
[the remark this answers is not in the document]
mxstbr: I think this is what the `instanceMounted` static method is for, but it's currently only a RFC. https://github.com/reactjs/rfcs/pull/8 I'm not sure how we can work around that not existing, cc @bvaughn

src/hoc/withTheme.js
```diff
@@ -33,45 +39,53 @@ const wrapWithTheme = (Component: ReactClass<any>) => {
...
+
+    styledContext = this.context[CHANNEL_NEXT]
+
+    state: State = {
+      theme:
+        this.styledContext !== undefined
+          ? this.styledContext.getTheme()
+          : undefined,
+      // We need it in state so that gDSFP can access it
+      defaultProps: this.constructor.defaultProps,
```
gaearon: Why does a component need to read its own default props?

src/models/StyledComponent.js
```diff
@@ -150,37 +192,32 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
...
       }
 
+      // Only react to props change to avoid infinite update loop
+      if (prevProps === this.props) {
+        return
+      }
+
+      // eslint-disable-next-line react/no-did-update-set-state
       this.setState(oldState => {
         const theme = determineTheme(
```
gaearon: setState in cDU typically leads to re render. Why isn’t this using gDSFP instead?

src/models/StyledComponent.js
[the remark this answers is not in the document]
mxstbr: It's collecting them on the server, and later exposing them as `styleSheet.getStyleTags`.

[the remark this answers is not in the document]
gaearon: Yeah you’ll probably want that server render RFC.

src/hoc/withTheme.js
kitten: I’d suspect it’s to determine which theme to use... but it could really be extracted there maybe?

gaearon: The reason I'm confused is that React already takes `defaultProps` into account (and copies them into `props` on element creation). It’s odd for a component to have to “think about” default props explicitly.

src/models/StyledComponent.js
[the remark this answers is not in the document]
gaearon: Yes, it’s supported. (Not very common but supported.)

src/hoc/withTheme.js
kitten: (I hope I’m going to get this right) the most important theme for us is one that’s passed in as a prop, the “theme” prop. This is so that themes can be overridden for a specific component (rarely used and barely documented though). The standard is that we use the theme from the themeprovider i.e. context however. But if that doesn’t exist we’d like to fall back to the default props. Hence we’re comparing the theme prop to the defaultProps’ theme prop

Hypnosphi: AFAIK styled-components wants to enforce "explicit props > context > defaultProps" priority, there's a comment about it somewhere

src/models/StyledComponent.js
Hypnosphi: Just forgot to remove it in https://github.com/styled-components/styled-components/pull/1664/commits/ed7efd2c019eba0817d42bd6bf82c7ba0543f704, will do

src/hoc/withTheme.js
kitten: Yep, that’s the one I was thinking of! Thanks for linking it

gaearon: Eh. Not a fan. From React's point of view default props *are* props, and an optimizing compiler would feel free to "inline" default values at the call site.

gaearon: I can see why you’d want it to work like this though.

src/models/StyledComponent.js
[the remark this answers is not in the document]
Hypnosphi: > detect server environment and use the constructor on the server

Sounds good to me. Will try to use [`IS_BROWSER`](https://github.com/styled-components/styled-components/blob/master/src/constants.js#L11) check for that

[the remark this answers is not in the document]
sebmarkbage: How does this work with streaming rendering where part of the document has already been sent before you get here?

src/models/StyledComponent.js
```diff
@@ -60,13 +61,53 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
...
+        ...this.getInitialState(),
+        defaultProps: this.constructor.defaultProps,
+        generateClassName: this.generateClassName.bind(this),
+        isStatic: this.constructor.componentStyle.isStatic,
+      }
+
+      // We need to inject styles in SSR
+      const styleSheet = this.context[CONTEXT_KEY] || StyleSheet.master
+      if (!IS_BROWSER || styleSheet.forceServer) {
+        this.injectStyles()
```
sebmarkbage: This has the same problem as gDSFP since the constructor also need to be pure.

So with things like suspense on the server, this will get called multiple times.

Hypnosphi:
@sebmarkbage I like that approach. But wouldn't env check in render break rehydration?

sebmarkbage:
Yea, I suppose it'd get deleted. :/

src/models/StyledComponent.js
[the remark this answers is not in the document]
Hypnosphi: There are some insights here: https://www.styled-components.com/docs/advanced#streaming-rendering

Hypnosphi:
Maybe it could actually work good on client as well, but it looks like quite a huge change to me

src/hoc/withTheme.js
kitten: We can discuss it separately maybe. I never liked the complexity it introduces and “just happened” more organically than consciently. It’s giving us some trouble in GlobalStyles since there the contributor wanted to reimplement it as a component, since there’s no generic “StyledComponent” abstraction yet. The logic proved unintuitive code-wise

src/models/StyledComponent.js
kitten: How can we make sure to play nicely with side effects right now before the RFC goes in? I realise our logic here is quite specific.

The StyleSheet can handle multiple injection (caching prevents it on multiple levels to happen anyway)

Also, what do you geese think of abstracting this as a simple reducer-like fn so that we can reuse it more easily without a component (might be worth replacing ComponentStyle in the process)

kitten:
@sebmarkbage SSR is indeed not the most elegant implementation on our end but the advantage right now is that it works (streaming & non streaming) in the same ways and doesn’t require a newer version of a React.

I’m sure we’d be willing to look at alternatives but I think we found the fragment approach a little limiting before. It’d also not match our rehydration logic and the idempotency of the StyleSheet right now, but itd be great for an improved rehydration process

Hypnosphi:
> it raises the question whether styled-components is ready to deprecate React 15

OK looks like it doesn't: https://github.com/jamiebuilds/create-react-context/pull/15

src/models/StyledComponent.js
```diff
@@ -119,22 +153,52 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
...
+      const styledContext = this.context[CHANNEL_NEXT]
+
+      this.injectStyles()
+
+      // If there is a theme in the context, subscribe to the event emitter. This
+      // is necessary due to pure components blocking context updates, this circumvents
+      // that by updating when an event is emitted
+      if (!componentStyle.isStatic && styledContext !== undefined) {
         const { subscribe } = styledContext
         this.unsubscribeId = subscribe(nextTheme => {
```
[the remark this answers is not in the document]
Hypnosphi: Theme may change between initial read (in constructor) and mount, so I think the subscribtion logic can be kept the same. We can add `sCU` or extend `PureComponent` to prevent unneeded rerenders. This would make sense anyway, as right now context theme changes trigger a rerender even if they're overriden by explicit props

See also: https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#adding-event-listeners-or-subscriptions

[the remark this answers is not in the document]
kitten: Good call 👍 let’s try to avoid unnecessary rerenders by checking whether the theme hasn’t changed. (sCU / PureComponent would be the wrong tools though since it’s introduce a huge slowdown in components like with SC that are spread throughout an app. Also I’d love not to introduce any unexpected behaviour) It’d be nice if we can just introduce a small check in the subscription that’d compare whether it’s execution is necessary.

tl;dr themes are expected to be immutable, (shallow) props not necessarily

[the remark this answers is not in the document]
Hypnosphi: Should we rather check that `generatedClassName` has changed then? There's a chance that theme has changed in a way that doesn't affect this particular component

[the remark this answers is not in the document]
Hypnosphi: Oh, I don't think it's possible: `prevState.theme` should be up to date when `gDSFP` is called next time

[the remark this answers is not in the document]
Hypnosphi: Or we can intoduce a `shouldUpdate` field on state which could be the only thing checked in `sCU`. Then `gDSFP` would always set it to `true`, and subscription would set it to `false` if `generatedClassName` hasn't changed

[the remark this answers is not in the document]
kitten: @Hypnosphi I don't think a `shouldUpdate` field makes sense, sorry 😅 We can trust that the theme is pure, since that's an assumption we already made a while ago (I believe). If not I think we should make it now either way.

Regenerating styles, reinjection, etc, that's something we'd like to avoid together with an unnecessary rerender. The former is theoretically often more expensive even than the latter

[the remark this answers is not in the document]
Hypnosphi: Or maybe even simpler:

```
shouldComponentUpdate(nextProps, nextState) {
  // React passes a new props object on each external rerender
  return nextProps !== prevProps || nextState.generatedClassName !== this.state.generatedClassName
}
```

The first part of this check is extra cheap, and it guarantees that component keeps being rerendered when its parent rerenders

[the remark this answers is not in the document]
Hypnosphi: > Regenerating styles, reinjection, etc, that's something we'd like to avoid together with an unnecessary rerender. The former is theoretically often more expensive even than the latter

OK, got your point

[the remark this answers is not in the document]
kitten: Oh, I think we shouldn't try to guess how the theme has changed. At that point it's pretty cheap to say that the theme has changed and perform the style generation & injection again. Then it's up to the stylesheet to deduplicate and running the render on the classname should be cheap enough so that we don't have to decide whether to do it or not.

Theme changes should be rare anyway 💯

src/hoc/withTheme.js
```diff
@@ -81,6 +81,12 @@ const wrapWithTheme = (Component: ReactClass<any>) => {
         const { subscribe } = this.styledContext
         this.unsubscribeId = subscribe(nextTheme => {
           const theme = determineTheme(this.props, nextTheme, defaultProps)
+
+          // Don't perform any actions if the actual resolved theme didn't change
+          if (theme === this.state.theme) {
```
kitten: :100:

src/hoc/withTheme.js
```diff
@@ -93,6 +113,8 @@ const wrapWithTheme = (Component: ReactClass<any>) => {
     }
   }
 
+  polyfill(WithTheme)
```
kitten: @mxstbr out of scope for this PR, but do you think we should ship a bundle in the future that drops these polyfills for decreased bundlesizes if someone really needs that? could be quite trivial to achieve using rollup

mxstbr: My assumption was that for React v16.3+ the polyfill is turned into a noop, and you can't run React 16.2- without it, so who would ever use that extra bundle? Not sure if that's the case

kitten: @mxstbr the concern was more around whether we'd want to reduce the bundlesize if possible, but I'm fine with always including it personally

mxstbr: - User on React v16.3+: Polyfill is a noop and gets optimized out, no bundle size increase
- User on React v16.2-: Polyfill works to mask the new API, if you remove the polyfill from the bundle your app stops working

Maybe I'm misunderstanding what you're saying, there is no way to use that bundle, right?

kitten: Ah really? If it's being optimised out that'd be perfect. Sorry for raising this then :)

gaearon: I don't think it could get optimized out—how would that work? Tools like webpack can't do static analysis like this, you'd need something like Prepack.

gaearon: That said it's likely that many libs will use the polyfill so as long as the version is in the same range, the bundle size cost will get amortized between them.

probablyup:
@Hypnosphi is there anything you need from us to finish this out?

Hypnosphi:
@probablyup I think it's ready for review

probablyup:
@Hypnosphi we might need to update our react native install to 0.54+ for the lifecycle stuff to be compatible? idk based on the test fails

probablyup:
@kitten might have a better idea

Hypnosphi:
@probablyup looks like `react-lifecycles-compat@3.0.3` is broken, see https://github.com/reactjs/react-lifecycles-compat/pull/24#issuecomment-388315328

I've pinned it to `3.0.2` for now

Hypnosphi:
Now tests fail because of an increased bundle size, see related discussion:
https://github.com/styled-components/styled-components/pull/1664#discussion_r180044170

kitten:
@Hypnosphi I’d say let’s bump it for now. @probablyup has a change in progress that’ll bring it back down

bvaughn:
> @probablyup looks like react-lifecycles-compat@3.0.3 is broken, see reactjs/react-lifecycles-compat#24 (comment)

Sorry for the regression. Bugfix release of `react-lifecycles-compat` just went out as version 3.0.4. Please let me know if you run into any more problems!

src/hoc/withTheme.js
```diff
@@ -33,45 +39,59 @@ const wrapWithTheme = (Component: ReactClass<any>) => {
...
+      nextProps: {
+        theme?: ?Object,
+        [key: string]: any,
+      },
+      prevState: State
+    ) {
+      const theme = determineTheme(
+        nextProps,
+        prevState.theme,
+        prevState.defaultProps
```
thymikee: `defaultProps` are static, so you should be good to obtain them from `WithTheme.defaultProps` – this way you'll at least avoid weird constructs like putting defaultProps in state.

Hypnosphi: Makes sense, thanks

Hypnosphi: Unfortunately, I can't do the same for `Styled(Native)Component`, because inheritance is used there:

https://github.com/JetBrains/styled-components/blob/fb2e97f8e08fb660a4453cdc79040d133a2bfe1b/src/models/StyledComponent.js#L317

So should I do it for `withTheme` only, or keep it as is for consistency?

thymikee: I'd still do it for WithTheme, even when it's troublesome (but probably doable) to get it working in other places

Hypnosphi: Maybe I could just move `gDSFP` from `BaseComponent` to `StyledComponent`, will try later today

gaearon: Inheritance 😭😭😭

src/models/StyledComponent.js
satya164: Also, AFAIK even if the constructor is called, the component might not mount.

src/hoc/withTheme.js
kitten: We can also just get rid of `BaseComponent` for now. It's a perf optimisation for v8, but it's probably not worth the mess with this refactor anymore. I'll think about a babel transpilation that can do this automatically maybe. Either way, it's not worth keeping if that means that it complicates this PR and leads to worse code.

Hypnosphi: Plus, there is a `withComponent` case, I'm not sure what would be a better soulution to handle it

Hypnosphi: We wouldn't recreate the methods in `Object.assign`, only create new references to them

Hypnosphi: > Plus, there are withComponent and extend case, I'm not sure what would be a better solution to handle those

This is a more tough part. In this case `ParentComponent` is not `BaseComponent` but rather the one being extended

kitten: As far as I remember, extend doesn’t imply an actual inheritance. We just create a new config object based on the previous (parent) component and concatenate all rules.

If we are actually applying inheritance then I don’t see it being actually necessary

satya164: `Object.assign(StyledComponent, Base)` is still inheritance though, just written differently.

Hypnosphi: @kitten OK, so which changes are you suggesting now?

Hypnosphi:
After seeing a [tweet about gDSFP](https://twitter.com/brian_d_vaughn/status/987443099537309696) by @bvaughn and reading through https://github.com/facebook/react/issues/12898, I have an idea of using a memoized version of `generateClassName` in `render` instead, just likein twitter example

@kitten @gaearon WDYT?

probablyup:
@Hypnosphi I'm fine with that if it's fast enough. Once you're rebased up, let's benchmark via https://github.com/cssinjs/css-in-js-perf-tests

probablyup:
Thanks for this work! After moving to the new context API in the 4.0 dev branch, usage of cWM, cDU, and cWRP ended up going away.

PR #1677, opened 2018-04-11
Fix splitByRules using `cascade: false` with stylis

### LINKED ISSUE #1676: Different styles on server vs client when using 3rd party component lib with styled-components
## Issue
Thank you for all your work with styled-components, it's a pleasure to use, but unfortunately we are experiencing an issue with it currently.

The problem is that we use a component library that's created with react 16.3 and styled-components 3.2.5 (+ babel sc plugin). (Very similar to https://github.com/diegohaz/reas - bug exists with both libraries, but reas is using styled-components 3.1.6)

When we use this library in our nextjs app, some styles do not work client side, from the server everything looks good, but client side the style is wrong as you can see here:

```css
/* SERVER */
/* sc-component-id: sc-kafWEX */
.sc-cIShpX + * .HJRRB{ stuff }

/* CLIENT */
/* sc-component-id: sc-kafWEX */
.sc-cIShpX+*.HJRRB{ stuff }
```

The issue is that it removes the space between `*` and `.`.

The source (link below) component:
``` javascript
  // This works on server but not client
  ${OuterWrapper} + * & {
    background: palevioletred;
    h1 {
      color: papayawhip;
    }
  }
```

I have provided two links to the app and links to two different branches below because as you can see there, the issue is only there if we import and use the external library, the syntax above works fine otherwise. We have tried with two different external libraries using styled-components and it happens with both. And it works server-side.

In our own app we also use (but not in the reproduction app):
```javascript
config.resolve.alias = {
	'styled-components': path.resolve(__dirname, 'node_modules', 'styled-components')
}
```
to be sure we don't have two versions of styled-components.

Should also note that this bug was not there when both app and library used styled-components 3.1.6

I appreciate everything you do and hope someone can help.

## System:
 - CPU: x64 Intel(R) Core(TM) i7-6567U CPU @ 3.30GHz
 - Free Memory: 357.23 MB
 - Total Memory: 16.00 GB
 - Shell: /bin/zsh - 5.3
## Binaries:
 - Node: 9.8.0
 - Yarn: 1.5.1
 - npm: 5.6.0
 - Watchman: Not Found
### babel-plugin-styled-components:
 - wanted: ^1.5.1
 - installed: 1.5.1
### styled-components:
 - wanted: ^3.2.5
 - installed: 3.2.5

### Apps:
* Example with lib (not working): https://sc-next-bug-mizjydjbep.now.sh/
* Example without lib (working): https://sc-next-bug-jfmvebfqne.now.sh/

### Source: 
* https://github.com/Fumler/sc-next-bug/tree/master (not working)
* https://github.com/Fumler/sc-next-bug/tree/working (working) 

## Expected Behavior
Expect CSS to be same on server and client even if we use styled-components in the external library as well as in our app.

## Actual Behavior
CSS is different on client vs server

kitten:
Fix #1676


probablyup:
@kitten is there any way to add a test for this?

kitten:
@probablyup not with our current normaliser as it tries to compensate for white space changes since that used to differ between stylis versions a lot :/ have to look into that in detail

kitten:
@mxstbr it seems to “disable” space combinators although that just mangoes and removes spaces in the selector...

PR #1680, opened 2018-04-12
update node typings

### LINKED ISSUE #1640: Wrong typings: 'ReadableStream from NodeJS namespace which is not present'
## System:
 - OS: Windows 10
 - CPU: x64 Intel(R) Core(TM) i5-4340M CPU @ 2.90GHz
 - Free Memory: 3.31 GB
 - Total Memory: 15.69 GB
 - Shell: ¯\_(ツ)_/¯
## Binaries:
 - Node: 8.9.4
 - Yarn: 1.3.2
 - npm: 5.6.0
 - Watchman: Not Found
### styled-components:
 - wanted: ^3.2.3
 - installed: 3.2.3

The system cannot find the path specified.

## Reproduction

in [at-loader] ./node_modules/styled-components/typings/styled-components.d.ts:116:51 
    TS2694: Namespace 'NodeJS' has no exported member 'ReadableStream'.

## Steps to reproduce

Just install styled-components in Typescript project.

## Expected Behavior

It should run without errors.

## Actual Behavior

Typings are using ReadableStream from NodeJS namespace which is not present in current node typings.

probablyup:
hopefully should close #1640

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "supports-color": "^3.2.3"
   },
   "devDependencies": {
-    "@types/node": "9.4.4",
+    "@types/node": "9.6.4",
     "@types/react": "^16",
     "@types/react-dom": "^16",
     "@types/react-native": "^0.50.7",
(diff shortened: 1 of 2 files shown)


probablyup:
@styled-components/typers is there a good reason to not use the semver range for types? seems they're pinned to a specific version

Issue #1683, opened 2018-04-12
Improve helper type operators through TypeScript 2.8


miZyind:
At that time before TypeScript 2.8, there wasn't a clean solution about how to "Omit" some keys from an existing interface.

But now we can use the new conditional type "Exclude" to improve type safety!

We can drop these helper type operators and rewrite some definition:
```typescript
// Line 73 ~ 76 in node_modules/styled-components/typings/styled-components.d.ts
// Helper type operators
type Diff<T extends string, U extends string> = ({ [P in T]: P } & { [P in U]: never } & { [x: string]: never })[T];
type Omit<T, K extends keyof T> = Pick<T, Diff<keyof T, K>>;
type WithOptionalTheme<P extends { theme?: T; }, T> = Omit<P, "theme"> & { theme?: T; };
```

After:
```typescript
// Helper type operators
type WithOptionalTheme<P extends { theme?: T; }, T> = Pick<P, Exclude<keyof P, "theme">> & { theme?: T; };
```

And now we get the benefit (Go to current props' definition & props' type checking)!
![improve](https://user-images.githubusercontent.com/21041375/38709653-b4f0ef44-3eee-11e8-9588-15e2a016c53e.png)

leebenson:
Perhaps related: In Typescript v2.9.0-dev.20180519:

> node_modules/styled-components/typings/styled-components.d.ts:76:48 - error TS2344: Type 'keyof T' does not satisfy the constraint 'string'.
>   Type 'string | number | symbol' is not assignable to type 'string'.
>     Type 'number' is not assignable to type 'string'.
> 
> 76 type Omit<T, K extends keyof T> = Pick<T, Diff<keyof T, K>>;

probablyup:
Please see https://github.com/styled-components/styled-components/issues/1778. Would love a fix contributed to that effort as y'all are able 👍

PR #1684, opened 2018-04-13
Improve helper type operators through TypeScript 2.8

miZyind:
Mentioned in issue #1683

### DIFF
--- a/typings/styled-components.d.ts
+++ b/typings/styled-components.d.ts
@@ -54,7 +54,7 @@ type ThemedStyledComponentFactories<T> = {
 export interface ThemedBaseStyledInterface<T> extends ThemedStyledComponentFactories<T> {
   <P, TTag extends keyof JSX.IntrinsicElements>(tag: TTag): ThemedStyledFunction<P, T, P & JSX.IntrinsicElements[TTag]>;
   <P, O>(component: StyledComponentClass<P, T, O>): ThemedStyledFunction<P, T, O>;
-  <P>(component: React.ComponentType<P>): T extends undefined ? StyledFunction<P> : ThemedStyledFunction<P, T, WithOptionalTheme<P, T>>;
+  <P>(component: React.ComponentType<P>): ThemedStyledFunction<P, T, WithOptionalTheme<P, T>>;
 }
 export type BaseStyledInterface = ThemedBaseStyledInterface<any>;
 
@@ -72,7 +72,7 @@ export interface ThemedCssFunction<T> {
 }
 
 // Helper type operators
-type WithOptionalTheme<P extends { theme: T; }, T> = Pick<P, Exclude<keyof P, "theme">> & { theme?: T; };
+type WithOptionalTheme<P, T> = P extends { theme: T; } ? Pick<P, Exclude<keyof P, "theme">> & { theme?: T; } : P;
 
 export interface ThemedStyledComponentsModule<T> {
   default: ThemedStyledInterface<T>;
--- a/typings/styled-components.d.ts
+++ b/typings/styled-components.d.ts
@@ -72,7 +72,7 @@ export interface ThemedCssFunction<T> {
 }
 
 // Helper type operators
-type WithOptionalTheme<P extends { theme?: T; }, T> = Pick<P, Exclude<keyof P, "theme">> & { theme?: T; };
+type WithOptionalTheme<P extends { theme: T; }, T> = Pick<P, Exclude<keyof P, "theme">> & { theme?: T; };
 
 export interface ThemedStyledComponentsModule<T> {
   default: ThemedStyledInterface<T>;
--- a/typings/styled-components.d.ts
+++ b/typings/styled-components.d.ts
@@ -72,7 +72,7 @@ export interface ThemedCssFunction<T> {
 }
 
 // Helper type operators
-type WithOptionalTheme<P extends { theme?: T; }, T> = Exclude<P, "theme"> & { theme?: T; };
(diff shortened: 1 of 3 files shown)


clayne11:
This looks great. Can we get this reviewed?

typings/styled-components.d.ts
```diff
@@ -54,7 +54,7 @@ type ThemedStyledComponentFactories<T> = {
 export interface ThemedBaseStyledInterface<T> extends ThemedStyledComponentFactories<T> {
   <P, TTag extends keyof JSX.IntrinsicElements>(tag: TTag): ThemedStyledFunction<P, T, P & JSX.IntrinsicElements[TTag]>;
   <P, O>(component: StyledComponentClass<P, T, O>): ThemedStyledFunction<P, T, O>;
-  <P extends { [prop: string]: any; theme?: T }>(component: React.ComponentType<P>): ThemedStyledFunction<P, T, WithOptionalTheme<P, T>>;
+  <P>(component: React.ComponentType<P>): T extends undefined ? StyledFunction<P> : ThemedStyledFunction<P, T, WithOptionalTheme<P, T>>;
```
Igorbek: I think a cleaner return type would be `T extends { theme: T; } ? ThemedStyledFunction<P, T, WithOptionalTheme<P, T>> : ThemedStyledFunction<P, T>`

typings/styled-components.d.ts
```diff
@@ -72,9 +72,7 @@ export interface ThemedCssFunction<T> {
 }
 
 // Helper type operators
-type Diff<T extends string, U extends string> = ({ [P in T]: P } & { [P in U]: never } & { [x: string]: never })[T];
-type Omit<T, K extends keyof T> = Pick<T, Diff<keyof T, K>>;
-type WithOptionalTheme<P extends { theme?: T; }, T> = Omit<P, "theme"> & { theme?: T; };
+type WithOptionalTheme<P extends { theme?: T; }, T> = Pick<P, Exclude<keyof P, "theme">> & { theme?: T; };
```
Igorbek: if it's not going to be used anywhere else, then we can have only constraint `P extends { theme: T; }` (not optional)

Igorbek:
@miZyind I'm actually not sure if my suggestions will work. Try them if you don't mind to see if it is better.

miZyind:
@Igorbek  Sure! I'll try it later. Thanks for your advice.

miZyind:
I've tried this:
```typescript
// Line 54~ in typings/styled-components.d.ts
  // ...
  // <P>(component: React.ComponentType<P>): T extends undefined ? StyledFunction<P> : ThemedStyledFunction<P, T, WithOptionalTheme<P, T>>;
  <P>(component: React.ComponentType<P>): T extends { theme: T; } ? ThemedStyledFunction<P, T, WithOptionalTheme<P, T>> : ThemedStyledFunction<P, T>;
}
```

But the test case **issue1068.tsx** won't pass:
```typescript
// Line 7 ~ 9 in typings/tests/themed-tests/issue1068.tsx
interface ThemedOptionalProps extends OptionalProps {
  theme: MyTheme; // <- theme is necessary prop here
}

// Line 36 ~ 48 in typings/tests/themed-tests/issue1068.tsx
function optionalPropsWithRequiredTheme() {
  const Component = (props: ThemedOptionalProps) => <div>{props.text}</div>;

  const StyledComponent = styled(Component) ``;
  const StyledStyledComponent = styled(StyledComponent) ``;

  <Component theme={theme} />; // theme is required
  <Component text="test" theme={theme} />; // theme is required
  // Error: Property 'theme' is missing in type '{ text: string; }'.
  <StyledComponent text="test" />; // theme is optional
  <StyledComponent text="test" theme={theme} />; // theme is allowed
  <StyledStyledComponent text="test" />; // theme is optional
  // Error: Property 'theme' is missing in type '{ text: string; }'.
  <StyledStyledComponent text="test" theme={theme} />; // theme is allowed
}
```

That's because the type of this **StyledComponent** now inferred to:
```typescript
React.Component<{
  text?: string;
  theme: MyTheme
}>
```

instead of:
```typescript
React.Component<{
  text?: string;
  theme?: MyTheme
}>
```

I have no idea how to make this better :(

But this works fine(not optional theme prop in WithOptionalTheme):
```typescript
// Line 75 ~ 76 in typings/styled-components.d.ts
// Helper type operators
type WithOptionalTheme<P extends { theme: T; }, T> = Pick<P, Exclude<keyof P, "theme">> & { theme?: T; };
```

Igorbek:
Let me look at this. I'll get back shortly.

Igorbek:
@miZyind I tried another way, see commit: https://github.com/Igorbek/styled-components/commit/0b54a58be9fc64c7c2e85d699ab7337940604191
That makes `WithOptionalTheme` a conditional type that only removes `theme` if it was required in the original props.
It seems to work fine. Please take it for your PR.

probablyup:
@Igorbek is this considered safe for a minor release, or should it wait until 4.0?

Igorbek:
For TypeScript users it is definitely a breaking change as it now requires a higher version (2.8).
To me it seems that should be ok to go in minor release version with documenting that change is breaking.

probablyup:
@kitten what's your opinion? If it's a breaking change, my inclination is it should wait for 4.0.

clayne11:
I think it's OK to introduce new minor TS versions in a minor version bump. It's only a minor TS version change.

probablyup:
Hey folks - so we've decided to [move the TS typings out of this repo](https://github.com/styled-components/styled-components/issues/1778) to DefinitelyTyped for 4.0. Please feel free to copy over the existing typings along with this change there!

PR #1696, opened 2018-04-19
extend PureComponent rather than Component

rothfels:
We found a [thread about using PureComponent](https://github.com/styled-components/styled-components/pull/619#issuecomment-290774283) and think it's a good idea.

In this PR, `BaseStyledComponent` and `BaseStyledNativeComponent` now extend `PureComponent` rather than `Component`. I've made the same change to `StyleSheetManager` and `ThemeProvider` (though it is less impactful to performance).

We're running this fork in production and haven't observed negative consequences.

### DIFF
--- a/src/models/StyleSheetManager.js
+++ b/src/models/StyleSheetManager.js
@@ -1,5 +1,5 @@
 // @flow
-import React, { Component } from 'react'
+import React, { PureComponent } from 'react'
 import PropTypes from 'prop-types'
 import StyleSheet from './StyleSheet'
 import ServerStyleSheet from './ServerStyleSheet'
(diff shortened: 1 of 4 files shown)


kitten:
This is something that we explicitly do not do as we cannot guarantee pureness of props in a shallow comparison as library authors. Unfortunately we have to back off and let the users decide when it's right to apply these practices.

Overall what I mean is that such a check on such a low level is actually harmful. The cost of the check which is basically run on a styling scale, so every single leaf component in your app, is high and can actually slow down your app tremendously. We have a couple of optimisations which operate on a lower and cheaper level, which @schwers was working on a while ago.

We cannot accept this with solid benchmarks, which I believe will show that this slows things down on a high level of stress (i.e. loads of individual leaf elements with loads of changes). We have a couple of benchmarks that you might be able to adapt, but even then the first point stands: We cannot make a pureness assumption for the user :/

Sorry, I hope this makes sense. I'd be happy to point you to other PRs that have helped with performance optimisations, or even the benchmarks we've been writing. 😸

PR #1698, opened 2018-04-20
Adding testIDs

charliewilco:
React Native has this nifty helper [`testID`](https://facebook.github.io/react-native/docs/view.html#testid) on its components. I work in a React Native Web codebase and have lots of components that use this convention that bubble out like [this](https://github.com/necolas/react-native-web/blob/83a8758f688067b2545568c2292a4d8417734e4b/packages/react-native-web/src/modules/createDOMProps/index.js#L136)

```
createdProps['data-testid'] = testID
``` 

Since I'm trying to port some of those components over I was thinking this could be nice for this project to use them too!


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ All notable changes to this project will be documented in this file. If a contri
 
 ## Unreleased
 
+## [v3.2.7] - 2018-04-20
```
probablyup: Don't add a version header please, should stay under "Unreleased".

probablyup:
I'm not sure how I feel about this one. Since it's a first-party pattern, I'm comfortable adding it for RN. But for web I'm not so sure. It's entirely possible that someone is already using a pattern like this for their own ends.

probablyup:
@charliewilco Could you reduce this PR to just add it for StyledNativeComponent? Happy to accept it then.

Thanks!

charliewilco:
@probablyup the purpose of the change is the bring that react native specific pattern to the larger library as it's a clearer pattern.

evan-scott-zocdoc:
(this is @probablyup's work account) I'm aware, but this is only an established pattern for RN and people might be using this pattern in the FE for other cases. That's why I'm only willing to accept it for RN.

charliewilco:
that's okay!

ReshmiCode:
@charliewilco could you reduce this PR to just RN so it can get added? It's a feature I could use

Issue #1496, opened 2018-02-06
[RFC] Move keyframes inside the component-tree


mxstbr:
Similar to `injectGlobal` (ref #1333) we need to move the ` keyframes`` ` helper inside the component-tree for two reasons: You should be able to use the theme in your animations and it makes appending by target possible. (ref #1324)

----

There's three main options I can think of here:

1. **Deprecate and remove `keyframes` entirely**: The `keyframes` helper isn't really necessary to style an app with styled-components, you could also just make global `@keyframes` declaration in your global styles.
2. **Introduce a component-based API for `keyframes`**: Similar to the change from #1333 you'd have a ` const FadeIn = createKeyframesComponent`` ` or something API that returns a component which you can render `<FadeIn />`. (note that this wouldn't apply that animation to it's children or anything, like the `<GlobalStyle />` component it would only inject the `@keyframes` declaration)
3. **Make `keyframes` "lazy"**: Rather than returning the name of the keyframes from the ` keyframes`` ` call, we return a custom Keyframes object. Then, when folks interpolate said object, we check for instanceof and only then inject the `@keyframes` declaration into the DOM. We should also be able to walk the ruleset to resolve any interpolated functions etc.

Personally, 1. feels weird to me: The whole reason we introduced the `keyframes` helper initially was because it was weird to never have to worry about a global namespace, just to then use a global namespace again for animations. 2. also doesn't fit very well imo, I feel like that takes the component-based thing too far.

To me, 3. is the best of all worlds: No big API change (we'd probably still need to cut a major version if people rely on the name being returned), docs all stay the same and you can use e.g. the theme in your keyframes. (and makes #1324 possible)

Any concerns, other ideas or comments? First WIP PRs demonstrating what an implementation _could_ look like would be appreciated!

marionebl:
I lean towards variant `2`, primarily to keep the APIs in symmetry:

```js
const KeyFrame = createKeyframe``;
const Globals = createGlobalStyle``;

const StyledDiv = styled.div`
  animation-name: ${props => props.animationName};
`;

const Comp = () => (
  <StyledDiv animationName={KeyFrame.name}>
    <KeyFrame/>
    <Globals/>
  </StyledDiv>
);
```

One way to think about #1493 and #1491 is establishing a way to render `styled-components` output inside the HTML of your application. A nice benefit of variant `2` is that it enables to specifiy output targets in the DOM arbitrarily via `ReactDOM.createPortal`:

```
```js
[...]

PR #1700, opened 2018-04-20
Lazy keyframes injection

Fer0x:
Implement third option from https://github.com/styled-components/styled-components/issues/1496

ToDo

- [ ] fix existed tests
- [ ] write new tests


src/constructors/keyframes.js
```diff
@@ -5,24 +5,41 @@ import StyleSheet from '../models/StyleSheet'
...
+  }
+
+  inject(styleSheet: StyleSheet) {
+    if (!styleSheet.hasNameForId(this.id, this.name)) {
+      styleSheet.inject(this.id, this.rules, this.name)
+    }
+  }
+
+  toString() {
+    return this.name
```
Fer0x: This is important for https://github.com/styled-components/styled-components/blob/master/src/models/ComponentStyle.js#L91 to insert animation name

src/utils/flatten.js
```diff
@@ -58,6 +59,10 @@ const flatten = (
         : ruleSet.concat(chunk)
     }
 
+    if (chunk instanceof Keyframes) {
+      return ruleSet.concat(chunk)
```
kitten: An explicit string instead of coercion has a better chance of not being deoptimised

src/constructors/keyframes.js
kitten: See below; let’s keep it explicit without coercion

src/models/ComponentStyle.js
```diff
@@ -99,6 +102,14 @@ export default (
       return name
     }
 
+    findAndInjectKeyframes(styleSheet: StyleSheet) {
+      this.rules.forEach(rule => {
+        if (rule instanceof Keyframes) {
```
kitten: This is highly inefficient unfortunately; we need to handle this either on a StyleSheet level (Not preferable) or extract it once we can via your class combined with componentstyle, which I think would be more elegant :)

kitten:
Thanks for picking this up; it’ll be a longer way to get it to where we planned it to be (connected discussions to the global style component) but I’d be happy to help and advise 💖👍

src/utils/flatten.js
kitten: It’d actually be more interesting to pass on the current StyleSheet I to flatten from ComponentStyle, evaluate the keyframes, inject it, and return its name :)

kitten: Essentially less efficient and introduces side effects but it can be sharped out as well

Fer0x:
@kitten Rewrote code just before you new comments:) Look at new changes, maybe it nicer than pass actual `Stylesheet` to flatten.

Fer0x:
@kitten The main problem  that `flatten()` is called for first time before  `ComponentStyle` instance was created. Because of this `keyframes` instance should be returned "as is" for further calls of `flatten()`.

Fer0x:
Closing this PR because of rewriting code in https://github.com/styled-components/styled-components/pull/1930 based on develop

PR #1711, opened 2018-04-25
Add ReactRef as a type option for innerRef

codepunkt:
The new `React.createRef` syntax (more information in ["Refs and the DOM"](https://reactjs.org/docs/refs-and-the-dom.html)) is not reflected in the typescript typings.

To do so, i had to update `@types/react` and then adjust the `styled-components` typings by allowing the `innerRef` to be one of two variations: Either the functional reference that was already reflected in typings before, or `React.RefObject<HTMLElement | SVGElement>` 

Can someone cc this to @styled-components/typers please?

### DIFF
--- a/typings/styled-components.d.ts
+++ b/typings/styled-components.d.ts
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { StatelessComponent, ComponentClass, PureComponent, ReactElement } from "react";
+import { StatelessComponent, ComponentClass, PureComponent, ReactElement, RefObject } from "react";
 
 type Component<P> = ComponentClass<P> | StatelessComponent<P>;
 
@@ -12,7 +12,7 @@ export type StyledProps<P> = ThemedStyledProps<P, any>;
 
 export type ThemedOuterStyledProps<P, T> = P & {
   theme?: T;
-  innerRef?: (instance: any) => void;
+  innerRef?: ((instance: any) => void) | RefObject<HTMLElement | SVGElement>
 };
 export type OuterStyledProps<P> = ThemedOuterStyledProps<P, any>;
 
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
   },
   "devDependencies": {
     "@types/node": "9.6.4",
-    "@types/react": "^16",
+    "@types/react": "16.3.12",
     "@types/react-dom": "^16",
     "@types/react-native": "^0.50.7",
     "babel-cli": "^6.22.2",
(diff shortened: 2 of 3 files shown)


codepunkt:
I also had to update `@types/react-dom` and `@types/react-native` to the latest versions because they both rely on `@types/react` and there'd be a conflict of typings otherwise.

typings/styled-components.d.ts
```diff
@@ -12,7 +12,7 @@ export type StyledProps<P> = ThemedStyledProps<P, any>;
 
 export type ThemedOuterStyledProps<P, T> = P & {
   theme?: T;
-  innerRef?: (instance: any) => void;
+  innerRef?: ((instance: any) => void) | RefObject<HTMLElement | SVGElement>
```
Igorbek: Since styled-components can style any component that has `className` property, it may happen that ref would not be only a dom element.
The previous callback-based ref assumed that by `any`. The new one can be `object`, in my opinion.

codepunkt: It's a React component instance or HTML DOM element - so `object` should be fine. I'm not sure if there's anything in the `react` or `react-dom` typings that is even more specific.

For now, i'll update it to object.

typings/styled-components.d.ts
```diff
@@ -12,7 +12,7 @@ export type StyledProps<P> = ThemedStyledProps<P, any>;
 
 export type ThemedOuterStyledProps<P, T> = P & {
   theme?: T;
-  innerRef?: (instance: any) => void;
+  innerRef?: ((instance: object) => void) | RefObject<HTMLElement | SVGElement>
```
Igorbek: I think we still need it be 
```ts
innerRef?: ((instance: object) => void) | RefObject<object>
```
as it can be a ref to react component instance

codepunkt: You are correct. Still, `object` is very unspecific. Same as in `(instance: object) => void`.

There's got to be something from `@types/react` that we can use that is more specific, right? Something along the lines of `RefObject<HTMLElement | SVGElement | Component<{}, {}, any>>`?

Igorbek: yep, `Component` should work 👍

codepunkt: Done! 😃 👍

KagamiChan:
Sorry for making something noisy, but I wonder if a new release could be published soon because the issue blocks TypeScript users to use new ref API

probablyup:
@KagamiChan probably tonight if I have time

Issue #1675, opened 2018-04-10
Forwarding Ref


hankandre:
With React now introducing the ability to use [ref forwarding](https://reactjs.org/docs/forwarding-refs.html) it seems like this would be a wonderful opportunity to update styled-components to use, in certain instances. 

An example of this is a component I'm working on where I'm using the `styled()` HOF. When I try to access the ref of the child component I'm getting the style-component class reference instead. Instead I've had to implement a different solution, using `childRef` and then consuming that in my child component.

It _seems_ like if we were able to just pass in a `ref` and styled-components forwarded that on, at least in the instance of using the `styled()`HOF, it would simplify this problem quite a bit.

kitten:
Luckily we were involved in the RFC itself and we didn’t have to do anything to support this apart from a little fix which has just been merged and will land soon :) https://github.com/styled-components/styled-components/pull/1658

We’re now interested though in replacing our existing solution, the innerRef prop entirely with either a new API are manual ref forwarding

PR #1718, opened 2018-05-02
Add createRef() tests, fix for react-native.

simonbuchan:
Using a `React.createRef()`-created ref as `innerRef` would break under `react-native` (but not react-dom) as `StyledNativeComponent` (unlike `StyledComponent`) is already using `ref` and didn't know how to forward to the new type.

AFAIK, primitives didn't have the same issue, but I updated the test there, since they were copy pasted. I attempted to add tests for `react-dom`, but ran into what looks like an `enzyme` bug, see the follow-on #1720 for more details.

Also, I had to test under jest 22 on my machine for some reason - jest 20 just hung on (at least) test:native? I avoided including that change here.

### DIFF
--- a/src/models/StyledNativeComponent.js
+++ b/src/models/StyledNativeComponent.js
@@ -130,7 +130,11 @@ export default (constructWithOptions: Function, InlineStyle: Function) => {
 
       if (typeof innerRef === 'function') {
         innerRef(node)
-      } else if (typeof innerRef === 'object' && innerRef) {
+      } else if (
+        typeof innerRef === 'object' &&
+        innerRef &&
+        innerRef.hasOwnProperty('current')
+      ) {
         innerRef.current = node
       }
     }
--- a/src/models/StyledNativeComponent.js
+++ b/src/models/StyledNativeComponent.js
@@ -130,6 +130,8 @@ export default (constructWithOptions: Function, InlineStyle: Function) => {
 
       if (typeof innerRef === 'function') {
         innerRef(node)
+      } else if (typeof innerRef === 'object' && innerRef) {
+        innerRef.current = node
       }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file. If a contri
 
 ## Unreleased
 
+- Fix `React.createRef()` values for `innerRef` being ignored in React Native, by @simonbuchan (see [#1718](https://github.com/styled-components/styled-components/pull/1718))
 
 ## [v3.2.6] - 2018-04-17
 
--- a/package.json
+++ b/package.json
@@ -127,7 +127,7 @@
     "prettier": "1.9.2",
     "puppeteer": "^0.13.0",
     "raf": "^3.4.0",
-    "react": "^16.0.0",
+    "react": "^16.3.0",
     "react-dom": "^16.0.0",
     "react-frame-component": "^2.0.2",
     "react-native": "^0.46.0",
--- a/src/native/test/native.test.js
+++ b/src/native/test/native.test.js
@@ -288,7 +288,7 @@ describe('native', () => {
   })
 
   describe('innerRef', () => {
-    it('should pass the ref to the component', () => {
+    it('should pass a callback ref to the component', () => {
       const Comp = styled.View``
       const ref = jest.fn()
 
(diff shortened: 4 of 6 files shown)


simonbuchan:
Damn, new error output wasn't due to jest 22. OK, I can revert the react upgrade and switch to using `const ref = { current: null }` in the tests, which is [exactly what react does](https://github.com/facebook/react/blob/master/packages/react/src/ReactCreateRef.js), to get this passing for now.

Fixing the messages requires bumping `react-native` (it's due to it's built-in mocking `View => 'View'`), but that breaks `babel-plugin-flow-react-proptypes`, so you need to [bump that and update the config](https://github.com/brigand/babel-plugin-flow-react-proptypes/issues/152) - which I wanted to avoid fattening this PR up with. 😢

simonbuchan:
Nevermind - looks like it's just complaining about CHANGELOG.md in the CI too - I didn't see changelog updates in previous PRs, nor is it covered by CONTRIBUTING.md - is that a new requirement?

src/models/StyledNativeComponent.js
```diff
@@ -130,6 +130,8 @@ export default (constructWithOptions: Function, InlineStyle: Function) => {
 
       if (typeof innerRef === 'function') {
         innerRef(node)
+      } else if (typeof innerRef === 'object' && innerRef) {
```
probablyup: Hmm this might be too loose. Is there a ReactRef detector exposed by `react` or `react-is`?

simonbuchan: Sorry for the delay!

I've looked over the documented APIs, and didn't find anything. Digging deeper, they don't seem to have considered the case of *both* the component and the outer needing to ref the same element. They provide `React.forwardRef((props, ref) => node)`, but from testing it simply forwards the ref as-is - e.g. function to function, object to object, though invalid values are validated and rejected before the component is called.

I checked the react source code for refs, and [`createRef()` is nearly literally `return { current: null }`](https://github.com/facebook/react/blob/25dda90c1ecb0c662ab06e2c80c1ee31e0ae9d36/packages/react/src/ReactCreateRef.js#L12) and they seem to only bother with `typeof ref` - they are validated with `coerceRef(any) => null | FnRef | RefObject`](https://github.com/facebook/react/blob/25dda90c1ecb0c662ab06e2c80c1ee31e0ae9d36/packages/react-reconciler/src/ReactChildFiber.js#L101) and updated by [`commitAttachRef()`](https://github.com/facebook/react/blob/25dda90c1ecb0c662ab06e2c80c1ee31e0ae9d36/packages/react-reconciler/src/ReactFiberCommitWork.js#L321), which does have a dev mode warning with `ref.hasOwnProperty('current')`, but that seems like an implementation detail? I wouldn't want to assume that couldn't change at some point (e.g., made read-only by client code)

Seems like this "shared ref" use-case could be created as an issue, I didn't find an existing one.

probablyup: @gaearon could you weigh in on this? Wondering if `react-is` needs a new API to detect refs...

simonbuchan: Keep in mind any realistic changes to `createRef()` that could break the type check would probably also break the actual update `innerRef.current = node` below!

gaearon: The check looks fine to me. React only cares that it’s a plain object with `current` property. No plans to change this.

gaearon:
Can you explain in more detail why `forwardRef` doesn’t help you? I didn’t get that.

simonbuchan:
@gaearon The styled component wants to do its own thing with the wrapped element *and* forward it. That means it needs its own ref that forwards onto the outer ref (somewhat confusingly called `innerRef`) - `forwardRef` would allow calling the outer ref simply `ref` (and this is the perfect use for it), but since it doesn't normalise it, this still has to detect which kind of ref it got, regardless of how it got it.
That is, `forwardRef` works fine, in just doesn't solve *this* problem.

simonbuchan:
Because that didn't exist when this was created 😉

gaearon:
Makes sense, thanks for explaining. Longer term it might be a good idea to migrate to ForwardRef fully.

src/models/StyledNativeComponent.js
probablyup: Ok @simonbuchan can you do a simple check for if `ref.current` is defined then? If duck typing is sufficient, then adding that should square off this PR.

simonbuchan: @probablyup OK, added, though be aware react itself is perfectly happy to stamp `current` on any old `object` you give it. This change means your object will be silently ignored, I'd be happier with a warning similar to react's, but I don't know if there's a __DEV__ equivalent in styled-components?

simonbuchan:
@gaearon Yeah, if any component is a use-case for `forwardRef()`, styled-components is. It was requested in #1675, it looks like, but that looks like it was closed as a misunderstanding of working *on* `forwardRef` components. Should that be re-opened @probablyup?

probablyup:
@simonbuchan yeah we are going to move to all the new React 16 goodness by default in 4.0

mxstbr:

PR #1720, opened 2018-05-02
[WIP] Support `.attrs({ ref: ... })`

simonbuchan:
As a follow on to #1718, created as a separate PR due to both not being complete and the callback ref part (which is working) not depending on it.

I'm happy to rebase away the createRef() stuff if #1718 is blocking this.

Basically the goal is implementing tab order like behavior in react-native:

```js
const refList = [React.createRef(), React.createRef()]
const handleFocus = refList.map(ref => () => ref.current.focus())
const common = { refList, handleFocus, ... }

<MyInput {...common} refIndex={0} />
<MyInput {...common} refIndex={1} />

const MyInput = styled.TextInput.attrs({
  key: props => props.refList[props.refIndex],
  onSubmitEditing: props => props.handleFocus[props.refIndex + 1],
})
```

Starting with web first, since it's easier to test, but I'm running into an issue getting object refs to work. The new callback refs tests pass, and the object refs tests fail (with `ref.current` being null still), but it looks like it's an enzyme bug, since the following enzyme only test fails:

```js
test('enzyme should mount object refs', () => {
  const ref = React.createRef()
  const wrapper = mount(<div ref={ref} />)
  const div = wrapper.find('div').first()
  expect(div).toBeTruthy()
  expect(ref).toHaveProperty('current', div.instance())
})
```

Creating this as a WIP to check interest before I dig further into fixing and expanding tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,7 +127,7 @@
     "prettier": "1.9.2",
     "puppeteer": "^0.13.0",
     "raf": "^3.4.0",
-    "react": "^16.0.0",
+    "react": "^16.3.0",
     "react-dom": "^16.0.0",
     "react-frame-component": "^2.0.2",
     "react-native": "^0.46.0",
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -209,15 +209,17 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
         .filter(Boolean)
         .join(' ')
 
+      const { ref: attrRef, ...attrs } = this.attrs
+
       const baseProps = {
-        ...this.attrs,
+        ...attrs,
         className,
       }
 
       if (isStyledComponent(target)) {
-        baseProps.innerRef = innerRef
+        baseProps.innerRef = innerRef || attrRef
       } else {
-        baseProps.ref = innerRef
+        baseProps.ref = innerRef || attrRef
       }
 
       const propsForElement = Object.keys(this.props).reduce(
--- a/src/models/StyledNativeComponent.js
+++ b/src/models/StyledNativeComponent.js
@@ -130,6 +130,8 @@ export default (constructWithOptions: Function, InlineStyle: Function) => {
 
       if (typeof innerRef === 'function') {
         innerRef(node)
+      } else if (typeof innerRef === 'object' && innerRef) {
+        innerRef.current = node
       }
     }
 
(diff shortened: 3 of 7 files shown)


simonbuchan:
Well the problem is the directly wrapped element gets either key or innerKey depending on its type. Otherwise I'm fine with either.
One thing I ran into is `styled(Inner)` will stomp the attrs ref of Inner if it's also a styled component in react native - Inner.extend seems fine. Not sure what to do about that other than have *another* ref so Inner can tell this isn't from the user?

simonbuchan:
Ah, the ref error is https://github.com/airbnb/enzyme/issues/1504
Easy enough to work around.

simonbuchan:
Description cleaned up, rebased and squashed, tests working (was actually just `react-dom` needed to be updated and the errors were being swallowed!?)

@kitten I'm going to stick with `ref` for now as @probablyup implied in #1718 4.0 will move to using `forwardRef` for styled components, which would make/allow the outer element usage `ref` instead of `innerRef` as well.

The confusion is basically over whether you feel `attrs()` is closer to `defaultProps` (should match outer attributes) or the `render()` implementation (should match what the wrapped component receives) - I've gone with the latter here - especially since `attrs()` can have logic based on the outer props.

simonbuchan:
Fair enough. I haven't looked into it with v4, but I suppose you could also keep the current `innerRef` implementation as well and pass-through `.attrs({ ref })` that way, but it's a bit overkill.

When you say "another component layer above", you mean a raw component, not a styled component; something like this?

```js
const A = styled.input`color: red`;
const B = (props) => (
  <A
    {...props}
    ref={props.refList[props.refIndex]}
    onKeyPress={props.handleFocus[props.refIndex + 1]}
  />
);
```

It always feels like I'm doing something wrong when I do that, but I suppose it can't be helped here.

probablyup:
Well so the issue with forwardRef is the `ref={}` prop has to be put on the uppermost level for it to be forwarded. `attrs` is applied one level down in the StyledComponent class, so we couldn't apply a ref from there.

simonbuchan:
Sure, but the inner component *also* has access to the real component to give to `ref`, `forwardRef()` just gives you a way to get the literal `ref` parameter passed.

probablyup:
I mean I tried redoing your PR against develop and it just didn't work with forwardRef. You're welcome to give it another try though if you can get it working.

PR #1729, opened 2018-05-08
[DO NOT MERGE] background-image: image-set not prefixed

G07cha:
For some reason `background-image: image-set` is not being prefixed for Chrome with `-webkit` although `inline-style-prefixer` v4.0.0 able to prefix this value:

```javascript
var Prefixer = require("inline-style-prefixer")

const style = {
  'background-image': 'image-set(url(test.jpg) 2x)'
}

const prefixer = new Prefixer()
prefixer.prefix(style)
// ->
{ 
  background-image: [
    "-webkit-image-set(url(test.jpg) 2x)", 
    "image-set(url(test.jpg) 2x)"
  ] 
}
```

Here is also [codepen](https://codepen.io/anon/pen/yjpzQx) with the setup to reproduce the bug.


probablyup:
Can you move this to the stylis repo? That library handles the prefixing.

PR #1732, opened 2018-05-10
First-class object support

mxstbr:
With this patch you can now do

```JS
const Component = styled.div({
  color: 'blue'
})
```

This is in response to https://mobile.twitter.com/kentcdodds/status/994254700651339776, /cc @kentcdodds


Gabri3l:
Now my life is complete. Thank you!

src/constructors/css.js
```diff
@@ -4,6 +4,11 @@ import flatten from '../utils/flatten'
 import type { Interpolation, RuleSet } from '../types'
 
 export default (
-  strings: Array<string>,
+  strings: Array<string> | Object,
```
probablyup: We should probably rename this property. Perhaps "styles"?

mxstbr: Yeah that makes sense, incoming.

ghengeveld:
Can you also do this? 

```js
const Component = styled.div({
  color: 'blue'
}, {
  width: 100
}, props => ({
  background: props.bg
}))
```

A contrived example but the idea is to have multiple style objects stacked on top of each other, done even being the callback style.

PR #1738, opened 2018-05-15
chore: upgrade Jest to 23

thymikee:
**Summary**

Bumps Jest to 22 (20 is pretty old now). Had to introduce some minor changes, as I've upgraded related deps by the way (e.g. `puppeteer`).

This PR makes it possible to test under Node 10 (it hanged on react-native test for reasons, no time to investigate)


src/test/warnTooManyClasses.test.js
```diff
@@ -8,8 +8,8 @@ import styleSheet from '../models/StyleSheet'
 let styled
 
 describe('warn too many classes', () => {
-  const nativeWarn = console.warn
```
probablyup: Hmm why rename this? It was correctly described before, meaning native console.warn vs a mocked console.

test-utils/setupTestFramework.js
```diff
@@ -0,0 +1,18 @@
...
+  ;(console: any).error = message => {
+    if (!message.includes('Error: Could not parse CSS stylesheet')) {
+      consoleError(message)
+    }
+  }
+})
+
+afterEach(() => {
+  // eslint-disable-next-line flowtype-errors/show-errors
+  ;(console: any).error = consoleError
```
probablyup: `;(console: any).error` what is this magic?

src/test/warnTooManyClasses.test.js
thymikee: Can revert this, it's a leftover actually.

test-utils/setupTestFramework.js
thymikee: Ask Prettier :D

kitten: ha! It'd make sense if the previous line required a semicolon to be put there but better safe than sorry ?! 😆

thymikee:
Seems it's not complaining :)

probablyup:
Thanks!

PR #1750, opened 2018-05-22
hoist non react & sc statics

### LINKED ISSUE #1411: Hoist non react statics
### What will it allow you to do that you can't do today?

Solve https://github.com/mui-org/material-ui/issues/9736 out of the box. We add a `muiName` static property to some of the Material-UI components. However, this static property is lost after wrapping the component with styled-components.

### How will it make current work-arounds straightforward?

You are already solving the problem for withTheme:
https://github.com/styled-components/styled-components/blob/1dfc7a64ff8ff3706424d68bbe021b8f973ff9db/src/hoc/withTheme.js#L6

### What potential bugs and edge cases does it help to avoid?

It's making the styled component higher-order component transparent to static properties.

probablyup:
Closes #1411

### DIFF
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -251,8 +251,10 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
     const {
       isClass = !isTag(target),
       displayName = isClass
-        ? `Styled(${getComponentName(target)})`
-        : `styled.${target}`,
+        ? // $FlowFixMe
+          `Styled(${getComponentName(target)})`
+        : // $FlowFixMe
+          `styled.${target}`,
       componentId = generateId(options.displayName, options.parentComponentId),
       ParentComponent = BaseStyledComponent,
       rules: extendingRules,
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -3,22 +3,21 @@
 import hoist from 'hoist-non-react-statics'
 import PropTypes from 'prop-types'
 import { Component, createElement } from 'react'
-
-import type { Theme } from './ThemeProvider'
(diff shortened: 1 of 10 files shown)


src/test/basic.test.js
```diff
@@ -215,7 +236,31 @@ describe('basic', () => {
       `
 
       shallow(<Comp />)
-      expectCSSMatches('.sc-a{ } @media (min-width:500px){ .b > *{ color:pink; } } ')
+      expectCSSMatches(
+        '.sc-a{ } @media (min-width:500px){ .b > *{ color:pink; } } '
+      )
+    })
+
+    it('should hoist non-react static properties', () => {
```
probablyup: these are the new tests - the rest is just prettier

src/models/StyledComponent.js
```diff
@@ -248,9 +249,12 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
...
     const {
-      displayName = isTag(target)
-        ? `styled.${target}`
-        : `Styled(${getComponentName(target)})`,
+      isClass = !isTag(target),
+      displayName = isClass
+        ? // $FlowFixMe
+          `Styled(${getComponentName(target)})`
+        : // $FlowFixMe
+          `styled.${target}`,
```
mxstbr: Why do we need the `FlowFixMe`s here even though we didn't previously? Smelly

probablyup:
Not sure, I spent a solid 30 mins trying to appease flow with no joy.
Nothing was working, not even the inline comment syntax to explicitly type
the variable in each part of the ternary.

probablyup:
@mxstbr ok I isolated the check into a util and the flow error went away... don't ask me why

PR #1751, opened 2018-05-22
Allow ES6 class attrs in styled-component

valerybugakov:
A small fix to allow passing of the ES6 classes using `.attrs()` 
I was trying to pass `component` prop to `redux-form` `Field` component, but got the error message:
```
Uncaught TypeError: Class constructor TextInput cannot be invoked without 'new'
```
It was mistakenly treated like a `prop-factory`.


src/utils/isClass.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+
+export default function isClass(target: any) {
+  return typeof target === 'function' && /^\s*class\s+/.test(target.toString())
```
probablyup: I don't think this will work for transpiled classes?

valerybugakov: Right, forgot about that and according to https://github.com/babel/babel/issues/5640 it's not possible right now.

src/models/StyledComponent.js
```diff
@@ -83,7 +84,8 @@ export default (ComponentStyle: Function, constructWithOptions: Function) => {
       this.attrs = Object.keys(attrs).reduce((acc, key) => {
         const attr = attrs[key]
         // eslint-disable-next-line no-param-reassign
-        acc[key] = typeof attr === 'function' ? attr(context) : attr
+        acc[key] =
+          typeof attr === 'function' && !isClass(attr) ? attr(context) : attr
```
probablyup: Instead let's use one of the methods from the `react-is` package. I think it'll work better.

valerybugakov: It seems that it's not possible too. With this package we can only check if `attr` is a valid element type for React but [it will accept functions too](https://github.com/facebook/react/blob/fc3777b1fe295fd2661f1974f5587d214791f04b/packages/shared/isValidElementType.js#L23). 

Also: https://github.com/facebook/react/pull/12578

probablyup: Well if we’re just concerned with classes, can we simply check if `React.Component` is in the inheritance chain for the function?

valerybugakov: Yeah, it's better than nothing. Will update PR.

src/models/StyledNativeComponent.js
```diff
@@ -42,8 +43,12 @@ export default (constructWithOptions: Function, InlineStyle: Function) => {
 
       this.attrs = Object.keys(attrs).reduce((acc, key) => {
         const attr = attrs[key]
+        const extendsReactComponent = hasInInheritanceChain(attr, Component)
```
probablyup: let's inline this, no need to do the work unless `attr` is confirmed to be a function

valerybugakov: updated

valerybugakov:
I haven't changed anything related to types in this PR. Flow should be happy as it is in master branch. 
@probablyup do you have an idea why it might fail?

valerybugakov:
@probablyup deleted `test-results.json` and merged master. Still getting some errors from flow.

probablyup:
Dang, ok. If you can't figure out what flow wants you to add a type for, feel free to throw a `// $FlowFixMe` above the offending line.

valerybugakov:
Yeah, flow is pretty mysterious sometimes, bypassed errors for now.

probablyup:
@valerybugakov try merging from master - I turned off danger for the moment so travis should stop being cranky

probablyup:
ha no doubt

mxstbr:

probablyup:
Deployed in https://github.com/styled-components/styled-components/releases/tag/v3.3.2!

PR #1755, opened 2018-05-25
fix generated component id in non-babel scenario

### LINKED ISSUE #1715: Different components defined with the same display name end up having same per-component class name
## Environment

`styled-components@3.2.6`

## Reproduction

```jsx
const Named1 = styled.div.withConfig({ displayName: 'Name' })`
  color: blue;
`;
const Named2 = styled.div.withConfig({ displayName: 'Name' }) `
  color: red;
`;
```

outputs

```html
<div class="Name-liQCRp iWZgJe">Named1</div>
<div class="Name-liQCRp fGgVaz">Named2</div>
<!--        ^^^^^^^^^^^ same per-component class name -->
```

Code: https://codesandbox.io/s/92o6vrym2w

## Expected Behavior

I expect that even if `displayName` is the same, two components should have different class names like `Name-ibxdXY` and `Name-RefOD`.

## Actual Behavior

Both components have the same per-component class name `Name-liQCRp`.

## Note

The bug was initially filed to `typescript-plugin-styled-components` https://github.com/Igorbek/typescript-plugin-styled-components/issues/10
Please see it for rationale

probablyup:
Specifically when two components share the same `displayName`.

Closes #1715


probablyup:
@Igorbek published this under `styled-components@dev` if you want to test it out.

Igorbek:
Verified that it is working correctly under `styled-components@dev`

probablyup:
@kitten I’m away for a week, but you’re welcome to merge this whenever 👍

PR #1768, opened 2018-05-30
 reimplement streaming as a transform stream

### LINKED ISSUE #1757: interleaveWithNodeStream not piping data chunks
(Using v 3.2.6)

`interleaveWithNodeStream()` doesn’t seem to pipe chunks of data to the response as expected, defeating the performance benefit of React’s `renderToNodeStream()`.

We noticed the issue when trying to implement error handling:
```
const jsx = sheet.collectStyles(<App />);
const stream = sheet.interleaveWithNodeStream(renderToNodeStream(jsx)).
    on(‘error’, () => {
        // redirect user to Error page
    });
stream.pipe(res, { end: false });
```

We would expect the React stream returned by `renderToNodeStream(jsx)` to pass chunks of streamed data to the Styled Components stream, which interleaves styles into these chunks and writes them to `res` as soon as each is available. That’s what the `data` event handler in `interleaveWithNodeStream()` looks as though it ought to do:

```
   readableStream.on('data', chunk => {
      const { tags } = instance
      let html = ''

      /* retrieve html for each new style tag */
      for (; instanceTagIndex < tags.length; instanceTagIndex += 1) {
        const tag = tags[instanceTagIndex]
        html += tag.toHTML(streamAttr)
      }

      /* force our StyleSheets to emit entirely new tags */
      instance.sealAllTags()
      /* prepend style html to chunk */
      ourStream.push(html + chunk)
    })
```
That’s not the behaviour we see. If an error is thrown by <App />, we can see that _no_ chunks of HTML have been streamed to the browser before the Error page redirect. If we remove Style Components from the equation...

```
const stream  = renderToNodeStream(jsx).
    on(‘error’, () => {
        // redirect user to Error page
    });

stream.pipe(res, { end: false })
```
...we _do_ see chunks of data flushed to the browser before the error is thrown and the user is redirected.

Moreover, if I add `.on('data', () => {console.log('[React/Styled Component] stream - data chunked')})` handlers to the React stream and the interleaving Styled Components stream, we can see that all the React stream data events are called before the Styled Component events, rather than each taking turns.

Is there a way to force `interleaveWithNodeStream()` to pipe chunks as it receives them to ensure we get the performance benefits of streaming?

probablyup:
Closes #1757 

This is how it should have been implemented in the first place, my stream knowledge at the time was just a bit lacking!

### DIFF
--- a/src/models/ServerStyleSheet.js
+++ b/src/models/ServerStyleSheet.js
@@ -98,7 +98,12 @@ export default class ServerStyleSheet {
     })
 
     readableStream.on('end', () => this.complete())
-    readableStream.on('error', () => this.complete())
+    readableStream.on('error', err => {
+      this.complete()
+
+      // forward the error to the transform stream
+      transformer.emit('error', err)
+    })
 
     return readableStream.pipe(transformer)
   }
--- a/src/models/ServerStyleSheet.js
+++ b/src/models/ServerStyleSheet.js
@@ -76,36 +76,30 @@ export default class ServerStyleSheet {
     let instanceTagIndex = 0
 
     const streamAttr = `${SC_STREAM_ATTR}="true"`
-    const ourStream = new stream.Readable()
-    // $FlowFixMe
-    ourStream._read = () => {}
 
-    readableStream.on('data', chunk => {
-      const { tags } = instance
-      let html = ''
+    const transformer = new stream.Transform({
+      transform: function appendStyleChunks(chunk, /* encoding */ _, callback) {
+        const { tags } = instance
+        let html = ''
 
-      /* retrieve html for each new style tag */
-      for (; instanceTagIndex < tags.length; instanceTagIndex += 1) {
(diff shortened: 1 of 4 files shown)


probablyup:
@jonsanders101 would you be able to test this on your end?

mxstbr:
Going to test in Spectrum!

src/models/ServerStyleSheet.js
```diff
@@ -76,36 +76,30 @@ export default class ServerStyleSheet {
...
+
+        /* prepend style html to chunk */
+        this.push(html + chunk)
+        callback()
+      },
     })
 
-    return ourStream
+    readableStream.on('end', () => this.complete())
+    readableStream.on('error', () => this.complete())
```
kitten: Don’t we have to pipe errors through?

probablyup:
@kitten good call! had to do a little extra work to forward the errors correctly. Should be good now 👍

PR #1769, opened 2018-05-30
Update ci setup

probablyup:
cherrypicking some of the stuff I was doing in https://github.com/styled-components/styled-components/pull/1768 so I can clean up that PR


.travis.yml
```diff
@@ -6,6 +6,8 @@ addons:
     packages:
       - libnss3
       - google-chrome-beta
+branches:
+  only: master
```
mxstbr: Why are we running only on master?

appveyor.yml
```diff
@@ -1,4 +1,7 @@
-version: '{build}'
+version: "{build}"
+branches:
+  only:
+  - master
```
mxstbr: Same here, why would we only run on master?

PR #1771, opened 2018-05-31
Make typings work with new TS 2.9 keyof behavior

Swatinem:
Instead of using a copy-pasted version of `Diff`/`Omit`, use the `Omit`
implementation as recommended by the official TypeScript Release Notes here:
https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#predefined-conditional-types

That version works correctly with the new `keyof` behavior introduced in
TS 2.9

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,9 +83,9 @@
   },
   "devDependencies": {
     "@types/node": "9.6.4",
-    "@types/react": "^16.3.12",
+    "@types/react": "^16.3.14",
     "@types/react-dom": "^16.0.5",
-    "@types/react-native": "^0.55.7",
+    "@types/react-native": "^0.55.16",
     "babel-cli": "^6.22.2",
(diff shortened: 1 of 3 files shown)


Swatinem:
Ah, that would be even better.
Please note, as that was discussed in the linked PR, that the current behavior prevents an upgrade to TS 2.9, unless you use `keyofStringsOnly`. So I would prefer to have this ASAP to be able to upgrade our own project to TS 2.9.

Igorbek:
Both PRs are in the same position, unfortunately, by introducing a breaking change.
It is a breaking change to require 2.9. BTW, `Exclude` was introduced in 2.8.

Swatinem:
I opened #1772 just so you have documentation on what is currently breaking. I will close this for now and try to find a different workaround.

PR #1773, opened 2018-05-31
Make TypeScript definition compatible with TS 2.9.1

Igorbek:
This is a minor fix that makes TS typings work on older and the latest version of TypeScript (up to 2.9.1).
Should supersede #1771 and fix #1772.
This is a non-breaking change, so can be released as a patch.

PR #1774, opened 2018-05-31
[v2] Make TypeScript definition compatible with TS 2.9.1

Igorbek:
I am not sure how `2.x` versions are supported.
It looks like `v2` branch is out of sync with the latest 2.x release (2.4.0), but I believe it is where it should go.

So this change is a port of #1773 for 2.x. It is required for v2 users (I am one of these) to upgrade to TS 2.9.

I think we can make a patch release 2.4.1.


Igorbek:
@probablyup @kitten @mxstbr please advise how to make patches for 2.x, and if it is correct branch to target.
My changes is just one commit of top v2.4.0 tag.

probablyup:
@Igorbek Uh I guess you'd find the last commit for 2.x and make a branch for it, then switch your PR to point at it as HEAD

probablyup:
But why are we releasing patches for 2.x? It's quite old.

Igorbek:
My commit is already based on the latest 2.x. The problem, though, is that there's no branch for 2.x releases. My understanding was that it would be `v2` branch which I targeted here. Should I create another branch for 2.x? something named like `2.x`?

Igorbek:
> But why are we releasing patches for 2.x? It's quite old.
half a year? :) I don't think all the users upgraded to 3.x. The patch would allow to upgrade TypeScript.

probablyup:
@Igorbek you can switch the HEAD for your branch to [legacy-v2](https://github.com/styled-components/styled-components/tree/legacy-v2)

Igorbek:
close in favor of #1779

PR #1775, opened 2018-05-31
Improve CSS object support for first argument

### LINKED ISSUE #1762: Improving CSS-object support
Howdy! I'm coming from Glamorous, and I would like to improve styled-components CSS-object support. The actual implementation shouldn't be too difficult, but I wanted to talk about the use case a little before opening a PR.

Current functionality possible via #1732 is `styled.div({ color: 'blue' })`, etc. Hooray!

A nice improvement on this would be to allow for a recursive array of style objects that are deeply merged: 

```js
styled.div([
  { color: "blue", fontSize: 22, fontWeight: "bold" },
  [{ color: "red" }],
  [[[{ color: "pink", fontSize: 11 }]]]
]);

// Would generate a component of: 
// => { color: "pink", fontSize: 11, fontWeight: "bold" }
```

Allowing interpolation functions into the mix would be even nicer:

```js
styled.div([
  { color: "blue", fontSize: 22, fontWeight: "bold" },
  [ (props) => ({ color: props.theme.colorBlindMode ? "red" : "black" }) ], // <- Nested Interop fn 
]);

// Would generate a component of: 
// => { color: "red or black", fontSize: 22, fontWeight: "bold" }
```

Finally, it would be great if the returned value of the interop functions could recursively nest, too:

```js
 (props) => [{ ...someCss}, (props) => [{...moreCss}, anotherObjOrFunction, etc] ]
```

If we were able to accept recursive arrays of objects or interop functions, it would enable really great compositional patterns for plain-object styles. Also, it types pretty well with TypeScript: https://github.com/paypal/glamorous/blob/master/typings/css-properties.d.ts#L24-L39

It'd be happy to explore adding this feature, and it shouldn't take too much work.

An open question I have: I'm new to styled-components, and am not clear on if there's a difference between static styles and interpolated style functions. I saw in Emotion's code that there's a difference so that they may extract static CSS. Is this something to look out for in this library?

Happy to hear thoughts on implementing this feature. It enables some really nice patterns! If any of my examples aren't clear, I can make up some more.

quicksnap:
Closes #1762 

This checks for the `raw` property on a tagged template. If it does not exist, styled takes the CSS Object interpolation path.

All other described functionality described in #1762 already exists in the flattened/interleaved interop arguments! 

Examples:

```js
styled.div([ {color: 'red'}, {fontSize: 24}, () => [...moreInterops]) // Works as expected
styled.div`color: red;` // Also works as expected
```


quicksnap:
Not sure why `yarn danger` is failing.. something about an API request?

probablyup:
Yeah. Trying to get a hold of Max to get an updated token

probablyup:
@quicksnap try merging from master - I turned off danger for the moment.

quicksnap:
Good catch! Would it be a reasonable approach instead to check the first element to be `typeof === "string"`? 

I appreciate you considering this PR! I feel it would be a really nice DX to be able to pass in an array of CSS objects (or a function).

quicksnap:
To further clarify what I hope to accomplish:

If the first argument is not an array of strings, i.e. tagged template, then simply treat that argument as if it were the second argument.

probablyup:
```
styled.div([ {color: 'red'}, {fontSize: 24}, () => [...moreInterops]) // Works as expected
```

That one actually doesn't seem to work. Only the first object appears to be honored: https://codesandbox.io/s/j4n6nr7k49

quicksnap:
@probablyup if you switch the first argument to `null` it will work as described, since you are referencing existing styled-components which doesn't have my PR merged

probablyup:
Going to close since there hasn't been action on this in some time. Feel free to make a new PR off master if you want to take this across the finish line.

wagerfield:
I'd like to see support for arrays of CSS objects as the first argument. This is something Emotion supports and for those of us that like to write styled using objects, it's a bit of a pain not being able to do this. Currently the only work around is to spread an array of style objects:

```jsx
const styleList = [{ background: "black" }, { color: "white" }]

const result1 = css(styleList) // only { background: "black" } gets applied
const result2 = css(...styleList) // this works
const result3 = css({}, styleList) // so does this

const Div1 = styled.div(styleList) // only { background: "black" } gets applied
const Div2 = styled.div(...styleList) // this works
const Div3 = styled.div({}, styleList) // so does this
```

Is there any particular reason that a flat array of style objects can't be passed as the first argument?

PR #1777, opened 2018-06-01
Fix type error of tagged template literal in TypeScript 2.9

utatti:
With TypeScript 2.9, we can provide type arguments to tagged template literals.

- https://blogs.msdn.microsoft.com/typescript/2018/05/31/announcing-typescript-2-9/#type-arguments-tagged-template-strings

It means that we no more need extra library for typing `ThemedStyledFunction`.

```typescript
const withTypeArgs = styled.h1<{ hello: string }>`
  color: ${props => props.hello};
`;
```

However, with the current typings, an error occurs:

image

The reason is that `ThemedStyledFunction` has overridden function types in a wrong order.

```typescript
interface ThemedStyledFunction<P, T, O = P> {
  (...): StyledComponentClass<P, T, O>;
  <U>(...): StyledComponentClass<P, T, O>;

  ...
}
```

When doing like above, the first type is applied, so type argument is not actually used and the error occurs. To fix this, we can just change the order, but I think using a default type argument is better.

```typescript
interface ThemedStyledFunction<P, T, O = P> {
  <U = {}>(...): StyledComponentClass<P, T, O>;

  ...
}
```

I've also added a test case for this. Please let me know if there is anything unclear.

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,7 +146,7 @@
     "rollup-plugin-uglify": "^3.0.0",
     "rollup-plugin-visualizer": "^0.1.5",
     "tslint": "^4.3.1",
-    "typescript": "^2.4.1"
+    "typescript": "^2.9.1"
   },
   "peerDependencies": {
     "react": ">= 0.14.0 < 17.0.0-0"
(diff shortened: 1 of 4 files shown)


utatti:
Rebased with master to fix CI failure.

probablyup:
@Igorbek is this safe to merge as a minor release?

probablyup:
Please see https://github.com/styled-components/styled-components/issues/1778. Moving TS types out of this repo and would be great for you to contribute this fix to DefinitelyTyped when it's moved over.

PR #1780, opened 2018-06-01
feat: don't force plain themes

### LINKED ISSUE #1776: Webpack 4.10 breaks theme modules ("isPlainObject" is too conservative).
Hey folks! I just noticed that Webpack 4.10 breaks one's ability to import theme objects from modules located elsewhere in the code base when those objects are constructed with wildcard imports. The object produced by `import * as theme` now has `Symbol.toStringTag` set to `"Module"`; so even though it's _effectively_ a plain object, Styled Components doesn't interpret it as one.

Would it be possible to make the check in `isPlainObject` a little more forgiving to accommodate this case? Thanks for considering it, and for all your work on Styled Components!

### System:
 - OS: macOS High Sierra 10.13.4
 - CPU: x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
 - Memory: 275.31 MB / 16.00 GB
 - Shell: 5.5.1 - /usr/local/bin/zsh
### Binaries:
 - Node: 10.3.0 - /usr/local/bin/node
 - Yarn: 1.7.0 - /usr/local/bin/yarn
### npmPackages:
 - babel-plugin-styled-components: ^1.5.1 => 1.5.1
 - styled-components: ^3.3.1 => 3.3.1

## Reproduction

https://glitch.com/edit/#!/grave-fowl

## Steps to reproduce
1. Import a theme from another module using Webpack 4.10.
2. Attempt to pass the imported theme to the `ThemeProvider` component.
3. :boom:

## Expected Behavior
Styled Components should recognize a namespaced module as, effectively, a plain object.

## Actual Behavior
Styled Components calls `isPlainObject`, which reads Webpack 4.10's new `Symbol.toStringTag` as `"Module"` and rejects the object.

phyllisstein:
`is-plain-object` is still required as an additional dependency, in order to guarantee that objects _within_ rulesets are `toString`ed.

Fixes #1776.


src/models/ThemeProvider.js
```diff
@@ -117,7 +114,7 @@ class ThemeProvider extends Component<ThemeProviderProps, void> {
       }
       return mergedTheme
     }
-    if (!isPlainObject(theme)) {
+    if (!isObject(theme)) {
```
probablyup: Looks good. Could you add a test that simulates what webpack does to the best of your ability so we know it'll work?

probablyup:
@phyllisstein try merging from master - I turned off danger for the moment. Please do add a CHANGELOG entry though 😃

kitten:
I think we should use this opportunity to get rid of such a small dependency that will likely never change. I think it’d make sense to check for:

- not equal null
- not an array
- is type of object

Without taking on a small dependency. I’m not even sure if we still need this constraint even?

phyllisstein:
Since this test just passed, I take it SC can pretty well handle non-plain objects:

```js
  it('should not allow other complex objects to be passed as themes', () => {
    class Theme {
      constructor(borderRadius) {
        this.borderRadius = borderRadius
      }
    }

    const theme = new Theme('2px')

    const Comp1 = styled.div`
      border-radius: ${ ({ theme }) => theme.borderRadius };
    `

    const wrapper = mount(
      <ThemeProvider theme={ theme }>
        <Comp1 />
      </ThemeProvider>
    )

    expectCSSMatches(`.sc-a {} .b {border-radius:${theme.borderRadius};}`)
  })
```

I'd be happy to remove the check---and the dependency---altogether. If you want to keep it, though, `isobject` isn't the right module.

phyllisstein:
Consider it done! With a few extra tests to boot.

probablyup:
@phyllisstein great! could you please merge from master to fix CI and add a changelog entry?

phyllisstein:
With pleasure. Should be ready to go once tests pass.

phyllisstein:
Looks like there were some Flow issues that I'll work on correcting, but also a problem with the bundle size check that I'm less sure how to resolve.

probablyup:
Heh, Flow is doing it's job and preventing you from making the mistakes! Since these tests are intentional, you can put a `// $FlowFixMe` above the lines in question to disable Flow for that particular case.

phyllisstein:
Thanks! With that added and a small bump to the maximum bundle size I think I'm set.

package.json
```diff
@@ -180,7 +180,7 @@
   "bundlesize": [
     {
       "path": "./dist/styled-components.min.js",
-      "maxSize": "16kB"
+      "maxSize": "16.5kB"
```
probablyup: Hmm why'd we have to bump this? Aren't we using less code now?

src/models/ThemeProvider.js
```diff
@@ -1,7 +1,6 @@
 // @flow
 import React, { Component, type Element } from 'react'
 import PropTypes from 'prop-types'
-import isPlainObject from 'is-plain-object'
```
probablyup: Can this dependency be removed entirely now?

phyllisstein: It's also used in `utils/flatten.js` to determine whether to try flattening the object or to just call `toString` on it. I think in that case it's still necessary that the object be "plain."

package.json
phyllisstein: The check was failing pretty consistently, but I'm not sure where the extra code came from. I guess the typecheck @kitten suggested (`(mergedTheme === null || Array.isArray(mergedTheme) || typeof mergedTheme !== 'object')`) is slightly longer than the original `isPlainObject(mergedTheme)`, but I'm not sure if that would've been sufficient to break the test.

mxstbr:

probablyup:
Released in https://github.com/styled-components/styled-components/releases/tag/v3.3.2!

phyllisstein:
Neat-o! Thanks for all your help!

PR #1784, opened 2018-06-01
Fix typings for TypeScript@2.9 and upgrade TypeScript dev dependency

### LINKED ISSUE #1783: TypeScript 2.9.1 makes typings break
The following error is thrown, when using with TypeScript `2.9`:
```/node_modules/styled-components/typings/styled-components.d.ts
(76,48): Type 'keyof T' does not satisfy the constraint 'string'.
  Type 'string | number | symbol' is not assignable to type 'string'.
    Type 'number' is not assignable to type 'string'.
```

Environment: Windows

Reproduction: install these:
```
    "styled-components": "^3.3.0"
    "typescript": "^2.9.1"
```
and try to compile your app

Actual Behavior: compiler error related to typings file of styled-components, not the app code

Expected Behavior: compiler passes styled-components source typings file.

mohsen1:
Fixes #1783

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,7 +146,7 @@
     "rollup-plugin-uglify": "^3.0.0",
     "rollup-plugin-visualizer": "^0.1.5",
     "tslint": "^4.3.1",
-    "typescript": "^2.4.1"
+    "typescript": "2.9"
   },
   "peerDependencies": {
     "react": ">= 0.14.0 < 17.0.0-0"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -146,7 +146,7 @@
     "rollup-plugin-uglify": "^3.0.0",
     "rollup-plugin-visualizer": "^0.1.5",
     "tslint": "^4.3.1",
-    "typescript": "^2.4.1"
+    "typescript": "2.9"
```
Igorbek: this is a breaking change. a non breaking fix is here #1773

mohsen1: That's better. Closing this

PR #1787, opened 2018-06-02
re-enable danger

probablyup:


probablyup:
Bleh, still not working. Guess we need to look into Peril.

Issue #630, opened 2017-03-30
How do I provide props to styled elements in Typescript?


tal:
I looked at the docs and other than creating a whole empty component purely for providing a definition of props I'm not sure of if it's possible to provide custom props to a styled element other than theme.

## Example
![image](https://cloud.githubusercontent.com/assets/18906/24527219/3e3b97c2-156f-11e7-83b5-5a110ca9d387.png)

For the above example I don't know how to make it not error.

philpl:
@tal Assuming that TypeScript infers the generics correctly — Disclaimer :wink: I haven't used SC with TypeScript yet — you'll just need to type your interpolation argument:

```ts
const Input = styled.input`
  border: ${(p: YourProps) => p.invalid ? 'red' : 'blue'};
`
```

tal:
Unfortunately it also errors at call site. I need a way to put generics in
I think

jupl:
Right now I do something like this for the time being:

```ts
const Input = (props: YourProps) => <input {...props} />
const StyledInput = styled(Input)`
  border: ${p => p.invalid ? 'red' : 'blue'};
`
```

giladaya:
@philpl 's approach worked for me

IIIristraM:
the helper I use for this case

export function styledWithProps<TProps>() {
  const Div = styled.div``;

  class Wrap extends React.PureComponent<TProps & React.HTMLProps<{}>, {}> {
    public node: HTMLElement;

    public render() {
      return <Div {...this.props} innerRef={n => this.node = n} />;
    }
  }

  return styled(Wrap);
}

beshanoe:
I'm using v2 and found this way more convinient:
```typescript
function stuc<T>(El: string = 'div') {
  type PropType = T & { children?: JSX.Element[], className?: string }
  return (props: PropType) => <El className={props.className}>{props.children}</El>
}

const Container = stuc<{customProp?: boolean}>()

const StyledContainer = styled(Container)`
  background: ${p => p.customProp ? 'red' : 'blue'};
`
```

patrick91:
@beshanoe hopefully we'll have a way to do this in TS (see: https://github.com/Microsoft/TypeScript/issues/11947)

alexf101:
@beshanoe 's solution working great for me, thank you! I renamed it to 'withCustomProps' so I can write things like:

```
interface EditAreaProps {
    maxHeight?: number;
}
[...]

PR #1798, opened 2018-06-08
Allow TypeScript 2.9.1 accept tagged template type argument

Igorbek:
This change is a workaround for TypeScript issue (Microsoft/TypeScript#24449) that prevents passing type argument to tagged template.
The workaround is simply consolidating overloads into one.

With this PR and TypeScript 2.9.1 it will be possible to write:
```ts
const Title = styled.h1<{ size: number }>`
  font-size: ${p => p.size}em;
`
```
and it will be fully type-safe,

Addresses #630

### DIFF
--- a/typings/styled-components.d.ts
+++ b/typings/styled-components.d.ts
@@ -40,8 +40,7 @@ export interface StyledComponentClass<P, T, O = P> extends ComponentClass<Themed
 }
 
 export interface ThemedStyledFunction<P, T, O = P> {
-  (strings: TemplateStringsArray, ...interpolations: Interpolation<ThemedStyledProps<P, T>>[]): StyledComponentClass<P, T, O>;
-  <U>(strings: TemplateStringsArray, ...interpolations: Interpolation<ThemedStyledProps<P & U, T>>[]): StyledComponentClass<P & U, T, O & U>;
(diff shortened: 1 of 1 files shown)


Huvik:
Can't wait :)

probablyup:
So @Igorbek is this backward compatible with older TS or does it require 2.9?

Igorbek:
It is fully backward compatible.

kachkaev:
Hi guys 👋 

Could this PR break `p.theme` by a chance? Here's what I get in 3.3.3:

```ts
Looks like I have a similar issue in 3.3.3:

```ts
const Statement = styled.div`
  margin: 0 auto;
  padding: 0 ${(p) => p.theme.gutter}px;
`;
```

> [ts] Property 'theme' does not exist on type 'never'.

↑ reported in https://github.com/styled-components/styled-components/issues/1821#issuecomment-399667542

kachkaev:
Similar case:

```ts
const MyComponent = (p: {flag?: boolean}) => ...;

const MyStyledComponent = styled(MyComponent)`
  background: ${p => p.flag ? 'red' : 'white' };
`;
```

> [ts] Property 'flag' does not exist on type 'never'.

Replacing `styled(MyComponent)` with `styled(MyComponent)<{}>` fixed the issue with typings.

PR #1801, opened 2018-06-12
Add human readable error to insertRuleHelpers' sheetForTag fn

Fer0x:
Add human-readable error description.

Also, I noticed that half of throwing errors has no text for production env. What is the purpose in this?

### DIFF
--- a/src/utils/insertRuleHelpers.js
+++ b/src/utils/insertRuleHelpers.js
@@ -17,7 +17,7 @@ export const sheetForTag = (tag: HTMLStyleElement): CSSStyleSheet => {
   }
 
   /* we should always be able to find a tag */
-  throw new Error()
+  throw new Error('Cannot find sheet for given tag')
 }
 
 /* insert a rule safely and return whether it was actually injected */


kitten:
> Also, I noticed that half of throwing errors has no text for production env. What is the purpose in this?

That's since we don't want to pollute the bundle with errors that are unlikely to even occur after debugging.

Cheers, I never got to add a message here. The problem is that this error might sufficient for `production`, but it's insufficient for `development`. Our new practice is that every new log message or error should explain what's going on in `development`. It's more useful to give the user an idea of what _might_ be the issue.

See: https://github.com/styled-components/styled-components/blob/master/src/models/StyleTags.js#L46-L68

So let's use your original error here in `production`, since this might still occur when multiple instances of SC are included, but write a more detailed message for `development`.

Fer0x:
@kitten I'm agree that we should provide more helpful error text for `development`, but I have no idea how to reproduce this error. I saw this error only in Sentry production logs.

probablyup:
> Also, I noticed that half of throwing errors has no text for production env. What is the purpose in this?

Keeping bundle size small. Free text is hard to compress so you'd be surprised how much bloat it adds.

We could consider moving to an error ID-style system though if we want to start providing more useful feedback in production mode.

probablyup:
Gonna close this and fold it into my error system PR!

PR #1806, opened 2018-06-13
[TypeScript] Fix: Diff removes props from indexed prop types

### LINKED ISSUE #1804: [Typescript Definitions] WithOptionalTheme appears to do nothing, but causes errors
We're having trouble using `styled()` to wrap components. None of the original component's props appear on the resulting styled component. I think I've narrowed it down to this line:

https://github.com/styled-components/styled-components/blob/34d18292e6cc97b508b7adb572f1d357232bccd5/typings/styled-components.d.ts#L78

This type appears to take a type that contains on optional `theme` property, and then take it off (using a hacky pre-TS 2.8 method) and then adds it right back.

Changing the line to the following appears to fix the issue:

```ts
type WithOptionalTheme<P extends { theme?: T; }, T> = P;
```

At that point, though, we might as well get rid of `WithOptionalTheme` altogether, which would simplify some other types as well.

More specifically, I see this when I try to wrap a component that has an index signature in its props:

```tsx
const MyComponent: React.ComponentType<{ [prop: string]: any }> = props => (
  <div className={props.className}>
    {/* do something fancy with all the other props */}
  </div>
);

const StyledImplementer: React.SFC = () => (
  <StyledMyComponent foo={1} className="bar" />
                     ^^^^^^^^^^^^^^^^^^^^^^^   Type '{ foo: 1; className: "bar"; }' has no properties in common with type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<ThemedOuterStyledProps<WithOptionalTheme...'
);
```

devrelm:
Fixes #1804

Calling `styled()` with a component whose props have an index signature, results in a component with no props besides `theme`.

### DIFF
--- a/typings/styled-components.d.ts
+++ b/typings/styled-components.d.ts
@@ -73,7 +73,7 @@ export interface ThemedCssFunction<T> {
 
 // Helper type operators
 type KeyofBase = keyof any;
-type Diff<T extends KeyofBase, U extends KeyofBase> = ({ [P in T]: P } & { [P in U]: never } & { [x: string]: never })[T];
+type Diff<T extends KeyofBase, U extends KeyofBase> = ({ [P in T]: P } & { [P in U]: never })[T];
 type Omit<T, K extends keyof T> = Pick<T, Diff<keyof T, K>>;
 type WithOptionalTheme<P extends { theme?: T; }, T> = Omit<P, "theme"> & { theme?: T; };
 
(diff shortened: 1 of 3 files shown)


probablyup:
Thanks for reviewing @Igorbek!

mxstbr:

Igorbek:
@probablyup this one can be merged too. It's also backwards compatible.

Igorbek:
ha, you were faster :)

probablyup:
Fingers of fire over here 😆

PR #1807, opened 2018-06-14

yjimk:
Picks up work by @marionebl in #1493

Which was originally, originally picked up from @JamieDixon in #1416

Original issues: #1333, #1491

Seeing as it looks like there was only a little bit left to go before this bit of functionality can be merged, I decided to pick it up to try to get it over the line. Looks like @marionebl may have indeed gone on parental leave.
https://github.com/styled-components/styled-components/pull/1493#pullrequestreview-125501038

I didn't want to retread old ground so haven't added an example, demo or scoped published version, please check them out in #1493.

As a side note, thanks for everyones great work on this project!


probablyup:
Thank you! Going to take a look at this Monday or Tuesday.

src/models/StyledComponent.js
```diff
@@ -1,12 +1,15 @@
 // @flow
 
+'no babel-plugin-flow-react-proptypes'
```
probablyup: What's this?

src/constructors/test/createGlobalStyle.test.js
```diff
@@ -0,0 +1,274 @@
...
+
+    expect(html).toBe('');
+    expect(stripWhitespace(stripComments(style.textContent))).toBe('[data-test-inject]{ color:red; } ');
+  });
+
+  it(`supports interpolation`, () => {
+    const {cleanup, render} = setup()
+    const Component = createGlobalStyle`div {color:${props => props.color};} `
+    render(
+      <ThemeProvider theme={{ color: 'orange' }}>
```
probablyup: I don't think the `ThemeProvider` is relevant to this test and makes it confusing.

mxstbr:
We should keep track of the bugs @migueloller pointed out in #1493, those seem pretty grave?

- [ ] Duplication of styles from global style components while server-side rendering (no consistent IDs?)

probablyup:
Alright this is the last blocker for 3.4 release. Anyone available to pick up and finish this out?

U-4-E-A:
I am pretty new to styled-components. I am looking for a way where I can inject global styles in a manner that they are global only to the component and it's children when it is mounted so as to not interfere with other themes. Can I confirm the new feature suggested here will provide that?

probablyup:
@U-4-E-A no, this is `injectGlobal` with a lifecycle + ability to update.

If you want to introduce scoped styles to your component, just write them like you would normal CSS.

```js
styled.div`
  color: red;
  a {
    color: blue;
  }
`
```

now all `a` under that component will be red

imbhargav5:
@probablyup I can wrap this up. Can you tell me what's left?

probablyup:
See Max’s comment

imbhargav5:
So 3/4 seem to be already fixed without me working on them. Not sure if issue 1 should be fixed as part of 3.4 or not from the discussion in the thread so far. Am I missing anything? 

- [ ] Duplication of styles from global style components while server-side rendering (no consistent IDs?)
- [x] Fixed Dynamically switching the theme throws "Failed to execute deleteRule" error
- [x] Docs update
- [x] Add warning to injectGlobal, if it isn't there yet

imbhargav5:
There seems to be a new bug in this branch. Not sure how but my local styles are getting duplicated. Works fine on 3.3, but this branch breaks it. Each time I reload the page, one duplicate style rule gets added. But the duplicate styles are not only for global, but for local components. 

Global rules were duplicated once, local rules are leaking. This is on production.  Does anyone know what may be causing this?


**Initial load** 

[Image from Gyazo](https://gyazo.com/e369be148a14f9f31f4312dc9572b9d9)

**After 4 refreshes**

[Image from Gyazo](https://gyazo.com/e0e1743b35e2afecc8f7ffdb9d47664d)

imbhargav5:
@migueloller  Update: I found why it was throwing the `deleteRule` error. In production mode, the GlobalStyle component seems to be unmounting unnecessarily and it was calling `removeStyles`. `updateStyles` also calls `removeStyles` internally so that makes it two calls, and the index goes out of bounds in the second call. It works fine in dev.

imbhargav5:
https://github.com/styled-components/styled-components/blob/v3.3.0/src/models/StyleTags.js#L211

I dug through the code today, and one change to the code made the `deleteRule` bug go away. 
```javascript
 deleteRules(sheet, removalIndex - 1, size)
```
All the test cases passed and createGlobalStyle with dynamic themes works fine on production. @probablyup @migueloller. I understand that this particular line might be used in a lot of places, but not sure what's going on here or whether it was overlooked. What do you think?

imbhargav5:
Closing this since all the work from this PR was later moved into #1867 and was merged successfully into develop.

PR #1823, opened 2018-06-22
add preact prop variants to whitelist

### LINKED ISSUE #1822: Add preact prop variants to whitelist
Preact uses `for`, `class` and `autofocus` instead of `htmlFor`, `className` and `autoFocus`.

These should be added to the attribute regex

probablyup:
Closes #1822


probablyup:
cc @developit are there any I missed?

PR #1824, opened 2018-06-22
upgrade rollup

probablyup:
before:

```
 PASS  ./dist/styled-components.min.js: 16.12KB < maxSize 16.5KB (gzip) 
 PASS  ./dist/styled-components.cjs.min.js: 11.31KB < maxSize 11.5KB (gzip) 
 PASS  ./dist/styled-components.es.min.js: 11.2KB < maxSize 11.5KB (gzip)
```

after:

```
 PASS  ./dist/styled-components.min.js: 16.03KB < maxSize 16.5KB (gzip) 
 PASS  ./dist/styled-components.cjs.min.js: 11.02KB < maxSize 11.5KB (gzip) 
 PASS  ./dist/styled-components.esm.min.js: 10.91KB < maxSize 11.5KB (gzip) 
```

need to verify nothing broke


package.json
```diff
@@ -134,17 +134,17 @@
...
     "rollup-plugin-ignore": "^1.0.3",
-    "rollup-plugin-json": "^2.1.0",
-    "rollup-plugin-node-resolve": "^2.0.0",
+    "rollup-plugin-json": "^3.0.0",
+    "rollup-plugin-node-resolve": "^3.3.0",
     "rollup-plugin-replace": "^1.1.1",
     "rollup-plugin-sourcemaps": "^0.4.2",
-    "rollup-plugin-uglify": "^3.0.0",
-    "rollup-plugin-visualizer": "^0.1.5",
+    "rollup-plugin-terser": "^1.0.1",
```
probablyup: terser is the new fork of uglify-es that's supported

rollup.config.js
```diff
@@ -174,7 +185,7 @@ const noParserConfig = Object.assign({}, configBase, {
   external: configBase.external.concat('stream'),
   input: 'src/no-parser/index.js',
   output: [
-    { file: 'dist/styled-components-no-parser.es.js', format: 'es' },
+    Object.assign({}, esm, { file: 'dist/styled-components-no-parser.es.js' }),
```
TrySound: How about shorter `getESM('dist/styled...')`?

probablyup: Ah good idea 🙂

rollup.config.js
```diff
@@ -5,56 +5,65 @@ import commonjs from 'rollup-plugin-commonjs'
...
 ]
 
+const globals = { 'prop-types': 'PropTypes', react: 'React' }
+
 const configBase = {
   input: 'src/index.js',
-  globals: { 'prop-types': 'PropTypes', react: 'React' },
   external: ['react', 'prop-types'].concat(
     Object.keys(pkg.dependencies),
     Object.keys(pkg.peerDependencies)
```
TrySound: Safer external can be achieved with this function
```js
// \0 is rollup convention for generated in memory modules
const external = id => !id.startsWith('\0') && !id.startsWith('.') && !id.startsWith('/');
```

rollup.config.js
```diff
@@ -5,56 +5,65 @@ import commonjs from 'rollup-plugin-commonjs'
...
   output: {
+    exports: 'named',
     file: 'dist/styled-components.js',
     format: 'umd',
+    globals,
     name: 'styled',
-    exports: 'named',
+    sourcemap: true,
   },
   external: ['react'],
```
TrySound: To mirror globals (they are mismatched now) it's better to use `Object.keys(globals)`

TrySound:
Would be also good to migrate to `.esm.js` suffix to be more descriptive what that file for. `es` for example is used by `material-ui` to ship es6 syntax. Do you think it's a breaking change?

rollup.config.js
probablyup: Hmm this would be a breaking change maybe... since `PropTypes` was getting packaged before.

TrySound: Then just remove `prop-types` from globals.

probablyup:
> Would be also good to migrate to .esm.js suffix to be more descriptive what that file for

Should be safe I think. If anyone's explicitly referring to them then they're depending on implementation details and we explicitly don't recommend that.

TrySound:
Agreed

rollup.config.js
probablyup: Hmm would this let `devDependencies` through?

rollup.config.js
mxstbr: Wait why are we packaging PropTypes? Let's not do that?

probablyup: Not sure, but I added it to the 4.0 roadmap since it would be a breaking change.

mxstbr: Ace, thanks!

rollup.config.js
TrySound: Nope. Is it requirement?

probablyup: Need to double-check. This rollup file is a little hairy as you've seen 😵

package.json
```diff
@@ -187,12 +187,12 @@
...
+      "path": "./dist/styled-components.esm.min.js",
       "maxSize": "11.5kB"
     }
   ],
   "collective": {
     "type": "opencollective",
     "url": "https://opencollective.com/styled-components"
   }
-}
+}
```
TrySound: Is this editor autoformatting?

probablyup: Probably prettier yeah

probablyup:
Published under `styled-components@dev` if people could test out. @mxstbr I'm interested in you testing in spectrum specifically since you have the SSR use cases

probablyup:
Alright it's been a while and no peeps of issues. Shall we ship this @mxstbr @kitten?

mxstbr:
Sorry, we didn't actually ship this because e2e tests failed 😆

probablyup:
@mxstbr that's not because of s-c though right?

mxstbr:
Probably not, no, so I assume this is safe to ship. Let's just try and revert if necessary.

Issue #1756, opened 2018-05-25
Missing TypeScript type definitions for v3.3.0 style objects


Wund3r:
Seems like there are no corresponding TypeScript definitions for the first-class object support mentioned in the released notes of 3.3.0

#### System:
 - OS: Windows 10
 - CPU: x64 Intel(R) Core(TM) i3-4150 CPU @ 3.50GHz
 - Memory: 8.22 GB / 15.66 GB
#### Binaries:
 - Yarn: 1.6.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
 - npm: 5.0.3 - C:\Program Files\nodejs\npm.CMD

#### Packages used:
 - styled-components: 3.3.0
 - typescript: 2.9.0-rc

## Steps to reproduce

The problem can be reproduced with a few lines of code like
```
import styled from "styled-components";

const Container = styled.div({
  color: "white",
});

export default Container;
```

## Expected Behavior

No TypeScript error shown

## Actual Behavior

TypeScript throws an error
```
Error:(4, 3) TS2345: Argument of type '{ color: string; }' is not assignable to parameter of type 'TemplateStringsArray'.
  Object literal may only specify known properties, and 'color' does not exist in type 'TemplateStringsArray'.
```

probablyup:
Want to add them?

probablyup:
cc @styled-components/typers

davidtkramer:
I'm not sure if this is the right approach, but something like the following might work.  I'm able to get type safety with these modifications but no intellisense.

```typescript
// new type for style objects
type StyleObject<P, T, U> = {
  [K in keyof P]: ((props: ThemedStyledProps<P & U, T>) => P[K]) | P[K];
}

export interface ThemedStyledFunction<P, T, O = P> {
  <U = {}>(strings: TemplateStringsArray, ...interpolations: Interpolation<ThemedStyledProps<P & U, T>>[]): StyledComponentClass<P & U, T, O & U>;
  // new call signature for style objects
  <U = {}>(styles: StyleObject<P, T, U>): StyledComponentClass<P & U, T, O & U>;
  attrs<U, A extends Partial<P & U> = {}>(attrs: Attrs<P & U, A, T>): ThemedStyledFunction<P & A & U, T, O & U>;
}
```

@probablyup I'd be happy to open a PR if this seems like the right direction.

probablyup:
@Igorbek would be a good person to look over this. I don't use typing stuff personally.

[...]

PR #1837, opened 2018-07-06
Add TypeScript support for v3.3.0 style objects

davidtkramer:
This PR is a work in progress for addressing #1756 by adding TypeScript typings for the first-class object support introduced in v3.3.0.

Todo:
- [ ] Add tests
- [ ] Verify compatibility with React Native
- [ ] Discuss ambiguities between `TemplateStringsArray` and `StyleObject` types

Regarding the last todo, there's some ambiguity between the call signatures for tagged template literals and style objects. If there is a type error, the compiler infers that you are providing an invalid `TemplateStringsArray`. I'm not sure if there is a way around this, or if it's even something we should be concerned about. @Igorbek, do you have any input?

Example:

```typescript
const StyledDiv = styled.div({
  flexDirection: 'invalid value for flex direction'
})
```

> Argument of type '{ flexDirection: string; }' is not assignable to parameter of type 'TemplateStringsArray'.
  Object literal may only specify known properties, and 'flexDirection' does not exist in type 'TemplateStringsArray'.

### DIFF
--- a/typings/styled-components.d.ts
+++ b/typings/styled-components.d.ts
@@ -26,6 +26,10 @@ export interface InterpolationFunction<P> {
   (props: P): Interpolation<P>;
 }
 
+type StyleObject<P extends { style?: object; }, U, T, S = P['style']> = {
+  [K in keyof S]: ((props: ThemedStyledProps<P & U, T>) => S[K]) | S[K];
+}
+
 type Attrs<P, A extends Partial<P>, T> = {
   [K in keyof A]: ((props: ThemedStyledProps<P, T>) => A[K]) | A[K];
 };
@@ -38,6 +42,7 @@ export interface StyledComponentClass<P, T, O = P> extends ComponentClass<Themed
 }
 
 export interface ThemedStyledFunction<P, T, O = P> {
+  <U = {}>(styles: StyleObject<P, U, T>): StyledComponentClass<P & U, T, O & U>;
   <U = {}>(strings: TemplateStringsArray, ...interpolations: Interpolation<ThemedStyledProps<P & U, T>>[]): StyledComponentClass<P & U, T, O & U>;
   attrs<U, A extends Partial<P & U> = {}>(attrs: Attrs<P & U, A, T>): ThemedStyledFunction<DiffBetween<A, P & U>, T, DiffBetween<A, O & U>>;
 }


quicksnap:
@davidtkramer this is some typings I'm using right now: https://gist.github.com/quicksnap/c536f59d18d85ca3e0883296b4f4359d

This code is mostly copied from Glamorous' typings that I've contributed in the past: https://github.com/paypal/glamorous/blob/master/typings/css-properties.d.ts

The gist of it is that the style object maybe a CSSObject, or an function that returns a CSSObject

The CSSObject may be a recursive structure of arrays of objects/functions.

I typed them separately as "Strict" and "Loose" since the escape hatch properties cannot be typed:

```
const someObj = { 
  fontSize: '200px',
  '& > .escapeHatch' : {...} // Doesn't type unless it's a "Loose" object
}
```

Also, this structure is only applicable as the second parameter in the following:

```
const myCssObj: Style | StyleLoose = {...whatever...}

styled.div(null!, myCssObj); // <-- Only the second argument is the recursive structure for now.
```

I have a PR open #1775 for allowing the nested structure as the first argument.

Hope this brain dump helps at all!

quicksnap:
Also, my linked typings differ from Glamorous in that `styled-components` do not accept `number` literal shorthand, so the generic of `CSS.Properties` is left as-is, instead of being `CSS.Properties<string | number>` like Glamorous allows.

For some more context, these are some helpers I wrote to write CSS with Objects the way I like, similar to how I used to use Glamorous:

```ts
type DivProps = { css?: Style; cssLoose?: StyleLoose }

export const Div: SFC<
  DivProps &
    // JSX.IntrinsicElements.div
    React.DetailedHTMLProps<
      React.HTMLAttributes<HTMLDivElement>,
      HTMLDivElement
    >
> = ({ css, cssLoose, ...rest }) =>
  React.createElement(
    styledOrig('div')(null!, [css as any, cssLoose as any]),
    rest,
  )

export const styledObj = (css?: Style | null, cssLoose?: StyleLoose | null) =>
  styledOrig.div(null!, [css as any, cssLoose as any])```

probablyup:
Ok [`@types/styled-components`](https://www.npmjs.com/package/@types/styled-components) is now out, so please move this over to DefinitelyTyped!

aaronmcadam:
Hi @davidtkramer, did you ever get a chance to move this PR to DefinitelyTyped? It still appears to be an issue with the latest types for v4.

PR #1840, opened 2018-07-09
[TypeScript] Fix #1821

### LINKED ISSUE #1821: [Typescript] css mixin - problem with theme
After updating to styled-components 3.3.3, i'm getting errors in all components which use mixins.

## Reproduction
 
https://codesandbox.io/s/m7lqkkw6l8

## Expected Behavior

Should not require ```theme``` prop.

## Actual Behavior

```
Property 'theme' is missing in type '{ children: string[]; }'.
```

Igorbek:
This fixes #1821 by reverting #1798.

- [x] added test cases

Please note, that new syntax that was allowed by #1798 is still allowed but requires TypeScript 2.9.2 which fixed overload resolution.

### DIFF
--- a/typings/styled-components.d.ts
+++ b/typings/styled-components.d.ts
@@ -38,7 +38,8 @@ export interface StyledComponentClass<P, T, O = P> extends ComponentClass<Themed
 }
 
 export interface ThemedStyledFunction<P, T, O = P> {
-  <U = {}>(strings: TemplateStringsArray, ...interpolations: Interpolation<ThemedStyledProps<P & U, T>>[]): StyledComponentClass<P & U, T, O & U>;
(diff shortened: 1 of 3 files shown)


kachkaev:
How about releasing 3.3.4? 😉

19majkel94:
@Igorbek 
Am I the only one whose code stopped working due to this change after upgrading from `3.3.3` to `3.4.9`? This is a simple reproduction example:

image

kachkaev:
@19majkel94 it may be that your editor is still using old (cached) typings. Reloading it might help.

19majkel94:
@kachkaev Of course I've reloaded the TS server but even Webpack complains about types:

image

Can TypeScript v2.9.1 be the cause? Fresh project with 3.1.1 works ok 😕

PR #1843, opened 2018-07-11
performance improvements

probablyup:
The broad strokes:

- use mutation instead of allocation wherever possible
- unpacking closures so the VM can cache things better (important for SSR performance)

So far:

### before

---Mount deep tree---
20.82 ±02.67 ms
(S/L) 16.45/04.37 ms

---Mount wide tree---
31.97 ±03.22 ms
(S/L) 25.45/06.51 ms

---Update dynamic styles---
06.90 ±01.10 ms
(S/L) 06.29/00.61 ms

### after

---Mount deep tree---
17.09 ±01.77 ms
(S/L) 13.01/04.08 ms

---Mount wide tree---
24.34 ±02.19 ms
(S/L) 18.25/06.09 ms

---Update dynamic styles---
05.56 ±01.08 ms
(S/L) 05.00/00.56 ms


src/hoc/withTheme.js
```diff
@@ -89,11 +81,9 @@ const wrapWithTheme = (Component: ComponentType<any>) => {
         delete props.innerRef
       }
 
-      return <Component {...props} />
+      return React.createElement(Component, props)
     }
```
kitten: This should already transpile to no spread in loose mode :) but it’s always good to be explicit 👌

src/models/ComponentStyle.js
```diff
@@ -9,7 +9,7 @@ import isStyledComponent from '../utils/isStyledComponent'
 const areStylesCacheable = IS_BROWSER
 
 const isStaticRules = (rules: RuleSet, attrs?: Object): boolean => {
-  for (let i = 0; i < rules.length; i += 1) {
+  for (let i = 0, len = rules.length; i < len; i += 1) {
```
kitten: Nice catch! I thought we got all of them 🙌

src/models/StyledComponent.js
```diff
@@ -29,224 +30,225 @@ type BaseState = {
...
+
+  render() {
+    // eslint-disable-next-line react/prop-types
+    const { innerRef } = this.props
+    const { generatedClassName } = this.state
+    const { styledComponentId, target } = this.constructor
+
+    const isTargetTag = isTag(target)
+
+    const className = [
```
kitten: We could replace this one with a simpler concatenation, especially since extra white spaces don’t matter we don’t have to do this too correctly

src/models/StyledComponent.js
```diff
@@ -29,224 +30,225 @@ type BaseState = {
...
     }
+  }
 
-    componentWillUnmount() {
-      this.unsubscribeFromContext()
+  componentWillReceiveProps(nextProps: { theme?: Theme, [key: string]: any }) {
+    // If this is a statically-styled component, we don't need to listen to
+    // props changes to update styles
+    const { componentStyle } = this.constructor
+    if (componentStyle.isStatic) {
```
kitten: I’m wondering whether we could eventually roll this “isStatic” check into the css helper so that we save ourselves an extra iteration

src/utils/flatten.js
```diff
@@ -38,31 +38,36 @@ const flatten = (
...
     }
 
     /* Either execute or defer the function */
     if (typeof chunk === 'function') {
-      return executionContext
-        ? ruleSet.concat(
-            ...flatten([chunk(executionContext)], executionContext)
-          )
-        : ruleSet.concat(chunk)
+      if (executionContext) {
```
kitten: 👌

Issue #1442, opened 2018-01-23
[v3] Typescript needs to switch typings over to RN automatically


IljaDaderko:
## Version
styled-components: 3.0.1
babel-plugin-styled-components: ---

## Reproduction

https://www.webpackbin.com/bins/-KeeZCr0xKfutOfOujxN
https://codesandbox.io/s/rkmNRByE4

## Steps to reproduce
After adopting new import for native i.e. omitting `/native` all types are missing for things like `styled.Text`, `styled.ImageBackground` etc..

## Expected Behavior
These should exist in native on `styled` with new import.

## Actual Behavior
```
import styled from "styled-components";

const Test = Styled.Text``;

[ts] Property 'Text' does not exist on type 'ThemedBaseStyledInterface<any>'. Did you mean 'text'?
```

philpl:
cc @styled-components/typers any idea if we can do an automatic switch for TS? Otherwise we’ll just have to add a caveat note to the website I guess

Igorbek:
Unfortunately, I don't see any way how we can statically tell TypeScript which types to use depending on compilation target. To me, deprecating `/native` module will lead to very inconvenient development experience for TypeScript users.

mxstbr:
Let's have one set of types with all aliases then?

Igorbek:
Do you mean that we should merge all the component aliases? Then TS would allow to use both web and native components at the same time. That does make much sense I think.

Igorbek:
correction: does not make much sense :)

Igorbek:
We can file an issue in the TypeScript repo, asking how we can differentiate type modules resolution based on application target. The problem is that in react native compilation modules get resolved by package.json's hint in `react-native` field and TS does not know about this and uses either `typings` field or just `./index.tsx?` file.

tjoskar:
I have not update styled-components to 3.x in projects yet, due to this change. But I’m a little confused; according to the [release page](https://github.com/styled-components/styled-components/releases/tag/v3.0.1) we can now import styled-components for RN by just importing `styled-components` and not `/native`. In fact, importing `/native` is deprecated. 

[...]

PR #1844, opened 2018-07-12
[Typescript] Migrate native types with keeping native dir exports left

tkow:
I found that using typescript, completion doesn't work and build failed because not resolving native type definitions due to ambient file located  in /native directory if you use `import from 'styled-components'` though `import from 'styled-components/native'` works well. I fixed them with keeping compatibility of styled-components/native because original code already implements native feature codes.

I saw the discussion in https://github.com/styled-components/styled-components/issues/1442 so, close this request if you don't need this.

### DIFF
--- a/native/index.d.ts
+++ b/native/index.d.ts
@@ -1,8 +1,7 @@
-import * as ReactNative from "react-native";
-import * as React from "react";
-import { StatelessComponent, ComponentClass } from "react";
-
 export {
+  default,
+  ReactNativeStyledFunction,
+  StyledInterface,
   ThemeProps,
   ThemeProvider,
   Interpolation,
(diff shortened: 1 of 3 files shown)


Igorbek:
In my opinion, this would be very undesirable solution. As I pointed out already in #1442, merging all shortcuts would not make any sense. It would so much confuse people who just use web styled-components (vast majority).

tkow:
@Igorbek Thank you for reply, and I think that's right, too, but [document says now /native path is deprecated](https://www.styled-components.com/docs/basics#react-native) , and I have trouble that no intellisense suggested and build failed if /native path is used or definite some types explained by documents. I think the best is that apis for native are not implemented in styled-components and rewrite docs though they have already implemented to standard apis. Is anyone going on that ?

PR #1866, opened 2018-07-26
Fix for IE10-11

Anber:
IE doesn't support `appendData` for an already attached node (at least in Win7/8), but we can reattach the node by calling `appendChild`. According to [specification](https://developer.mozilla.org/en-US/docs/Web/API/Node/appendChild), repeated calls of `appendChild` just move an attached node to the end and also force IE to update content of the node.


src/models/StyleTags.js
```diff
@@ -305,6 +305,11 @@ const makeBrowserTag = (
       // $FlowFixMe
       getImportRuleTag().insertRules(`${id}-import`, importRules)
     }
+
+    if (marker.replaceWholeText) {
```
Anber: It's a deprecated method for replacing content of the text node.  It was removed from all modern browsers.

probablyup: Actually I think i'd like to move this up into the `else` portion of the loop so it's more clear what this is an alternative to. @kitten do you remember why we are using `appendData` vs a more mainstream DOM API call?

Anber: @probablyup It’s not an alternative of ‘appendData’, it just says to IE that the node was updated and we don’t need to do it in the loop.

kitten: @probablyup i think this stuff is quite old and it’s only used in development so it didn’t really matter. No time was spent to investigate alternatives 😛

kitten: So reading through this some things need to be documented:

- where does “replaceWholeText” stem from?
- what does this force IE to do? “Update text node” is inaccurate

That’d be great 👍 also, have your checked that “appendChild” here never appends the child twice in both positions? That’d probably break a lot of ordering guarantees.

probablyup: @Anber could you take a look at these? ^

Anber: Yes, I had checked appendChild behavior before.
I’ll add documentation in two weeks when I return from vacation :)

Anber: @kitten, I've added a comment. I hope, it has become more clear :)

Anber:
@kitten, yes, it may affect specificity ordering in some specific cases, but the alternative is completely broken behavior.

kitten:
@Anber unfortunately we do care about the order of components’ styles as a whole, since we’re explicitly keeping them in the order of creation 😅

Anber:
@kitten oh…  I will try to rerender whole style tag but it may be too slow solution. Anyway slow working IE will be better then non-working IE.

kitten:
@Anber yea, that’d be great ✨ maybe there’s other ways to force IE to evaluate text nodes? I can try to give it a shot if you’d like?

Either way, this is development mode and it’ll always be slower than prod mode where we use insertRule.

Anber:
@kitten, don’t waste your time :) I’ve tried different solutions and unfortunately, there are no other simple ways. We can’t even reattach the node because it will break future updates, so the trick with appenChild is the only working solution.

kitten:
@Anber this might sound stupid, but what if we’d just call “appendChild” on all text nodes in reverse order? It shouldn’t be horrendously bad as we limit the number of components that we add to a style tag by a lot in development. Since you already have some Ie detection in there this workaround should also only affect IE, I suppose

kitten:
@Anber: sorry, Yea, messed up the order there in my head 😅 if it’s fine to append anything to the styletag to force a recalc, couldn’t we append and then remove an empty text node?

Anber:
@kitten, the problem is much deeper than I thought. `appendData` works fine, but in some cases (and I still can't understand when exactly) IE joins all text nodes into one node. It happens immediately after filling a style tag by markers. If I schedule a microtask in `makeStyleTag` it shows that we already have a style tag with joined text nodes. Nodes which are appended later work fine.
This behaviour can be prevented by adding comment nodes between text nodes, but it looks ugly. I'll continue investigation later this week.

Issue #1546, opened 2018-02-22
Deprecate .extend in favour of only styled(Component)


pke:
The FAQ explains the difference between `styled()` and `.extend`. However I wonder if it would be possible to take the decision out of the devs hand and have `styled` check for the handed in Component if its already a styled component. If it is `.extend` it otherwise create a new class.

Was this considered at one point?

kitten:
@mxstbr Totally agreed! We've discussed removing it before—and there was some general concensus—so I think it's time to pull the trigger.

kitten:
@mxstbr Do you have any reason to believe that we need to keep it around for another major version? 😅 I think that won't actually make a huge difference

mxstbr:
I don't think folks will be particularly happy about us suddenly throwing warnings in a minor version upgrade.

pke:
Thanks for considering this simplification of the API surface. I agree such removal of API should not happen in a minor update. Please stay true to semver.
The behaviour of `styled` can be changed in a minor version but the deprecation warning could be introduced in the next major RC.

On the other hand why not release a new major version right away. What to wait for?

mxstbr:
Yeah exactly.

kitten:
Yea, I don't have a problem with issuing a major version asap. However, semver doesn't restrict deprecations to major releases. That'd be quite extreme as you'd then during a typical release flow have to release two major bumps—one potentially without breaking changes—just to get new APIs in and remove old ones.

Quote:

> When you deprecate part of your public API, you should do two things: (1) update your documentation to let users know about the change, (2) issue **a new minor release with the deprecation in place**. Before you completely remove the functionality in a new major release there should be at least one minor release that contains the deprecation so that users can smoothly transition to the new API.

https://semver.org/

But anyhow, I'm fine with both, but I'd like to coordinate a new release with #1493 and all related deprecations (`injectGlobal`). So I'd love to push out a major only when #1493 is done, and deprecate in a minor, but I'll leave the call up to you, @mxstbr :+1:

Btw I'll rename the issue to represent the new plan.

mxstbr:
Yep, sounds very reasonable to batch that together with #1493! Maybe also #1171? I'd love to land that, personally.

probablyup:
We had a similar convo last year: https://github.com/styled-components/styled-components/issues/1087

I'm definitely in favor of getting rid of `.extend` from the public API.

kitten:
[...]

PR #1867, opened 2018-07-26

imbhargav5:
Picks up work by @yjimk in #1807

Which was originally picked up by @marionebl in #1493

Which was originally, originally picked up from @JamieDixon in #1416

- [x] Fix merge conflicts
- [x] Fix bug relating to `deleteRule` when `createGlobalStyle` was used with dynamic themes
- [x] Remove unnecessary ThemeProvider in a test case

**Note**: I am happy to merge my commits into either of the PRs above. But, I also want to check if the `deleteRule` bug, the only blocking issue for 3.4 is fixed without breaking anything else, by running the code against Appveyor and Travis.


imbhargav5:
https://github.com/styled-components/styled-components/blob/v3.3.0/src/models/StyleTags.js#L211

I dug through the code today, and one change to the code made the `deleteRule` bug go away. 
```javascript
 deleteRules(sheet, removalIndex - 1, size)
```
Won't `addUpUntilIndex` always give a number equals to cumulative size of markers? Index would always be at least one less than size right?

Test cases via`npm test` passed and createGlobalStyle with dynamic themes works fine on production. @probablyup @migueloller. I understand that this particular line might be used in a lot of places, but not sure what's going on here or whether it was overlooked. What do you think?

probablyup:
Could you switch the base branch for this to the new "develop" pre-4.0 branch?

imbhargav5:
Done.

probablyup:
@kitten could you comment on https://github.com/styled-components/styled-components/pull/1867/#issuecomment-408119666? I have less context

src/constructors/injectGlobal.js
```diff
@@ -1,5 +1,6 @@
 // @flow
 import hashStr from '../vendor/glamor/hash'
+import warnOnce from '../vendor/postcss/warn-once'
```
probablyup: You can drop the warning now that we're going directly into 4.0 with this new API.

probablyup:
> I’d say, let’s maybe merge context first and pick changes over from here if no one objects?

Yes definitely.

probablyup:
@imbhargav5 needs some merging/cleanup after the context API stuff went in

imbhargav5:
@probablyup Got it. Will do. 👍

CHANGELOG.md
```diff
@@ -13,6 +13,19 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this
     color,
   }))
   ```
+  
+* Add `createGlobalStyle` that returns a component which, when mounting, will apply global styles. It can be updated, replaced, removed, etc like any normal component and the global scope will update accordingly, by @JamieDixon @marionebl and @yjimk (see #1416)
+
+  ```jsx
+  const GlobalStyles = createGlobalScope`
```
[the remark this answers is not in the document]
imbhargav5: Done.

probablyup:
>  I am little confused about StyleSheet.master and StyleSheetManager context value. Which one should be used for createGlobalStyle?

I believe you can use the new `StylesheetConsumer` to get the stylesheet. It's one of the new internal exports from https://github.com/styled-components/styled-components/pull/1894/files#diff-e98b682885adddb6438755cb735d7dbeR17

probablyup:
> createGlobalStyle is the only function relying on React lifecycle methods like cwm and cwrp

Yes that will definitely need to change. Ideally anything in `cWM` can go in `constructor` instead and `cWRP` can probably move to `render`? Depends on what it's doing.

imbhargav5:
Regarding StyleSheetConsumer, a test case that was passing before isn't passing anymore if I remove `|| Stylesheet.master`. Any idea why that is ?

imbhargav5:
Regarding StyleSheetConsumer, a test case that was passing before isn't passing anymore if I remove `|| Stylesheet.master`. Any idea why that is ? https://github.com/styled-components/styled-components/pull/1867/commits/5b1b11ba8c1416ac114a3b50f9eca7a61ee58c9b#diff-bce2bc1e99348463230dfa6bc68a3cffR92

mxstbr:
I think `StyleSheet.master` only applies when client-side rendered and is a singleton, whereas the `styleSheet` is passed from the `ServerStyleSheet` or the `StyleSheetManager` via context—but @kitten is the one who knows all that code

probablyup:
@imbhargav5 can you remove the package-lock.json that got added and make sure it's in gitignore?

src/constructors/createGlobalStyle.js
```diff
@@ -0,0 +1,110 @@
...
+  const createGlobalStyle = (
+    strings: Array<string>,
+    ...interpolations: Array<Interpolation>
+  ) => {
+    const rules = css(strings, ...interpolations)
+    const id = `sc-global-${hashStr(JSON.stringify(rules))}`
+    const style = new GlobalStyle(rules, id)
+
+    class GlobalStyleComponent extends React.Component<*, *> {
+      componentWillMount() {
```
probablyup: This can probably become `constructor`

src/constructors/createGlobalStyle.js
```diff
@@ -0,0 +1,110 @@
...
+        const { sheet, context } = this.props
+        style.createStyles(context, sheet)
+      }
+
+      componentWillReceiveProps(props) {
```
probablyup: This can probably go in `componentDidUpdate`? What do you think @mxstbr

src/constructors/createGlobalStyle.js
```diff
@@ -0,0 +1,110 @@
...
+        style.removeStyles(sheet)
+      }
+      render() {
+        return null
+      }
+    }
+
+    class GlobalStyleComponentManager extends React.Component<*, *> {
+      static defaultProps: Object
+      // static contextTypes = {
```
probablyup: delete

src/constructors/createGlobalStyle.js
```diff
@@ -0,0 +1,110 @@
...
+      //   ]),
+      // }
+
+      static styledComponentId = id
+
+      render() {
+        if (process.env.NODE_ENV !== 'production') {
+          if (typeof this.props.children !== 'undefined') {
+            // eslint-disable-next-line no-console
+            console.warn(
```
probablyup: Let's make this an error and use the new error system

src/constructors/injectGlobal.js
probablyup: This still needs to be done.

src/models/StyleTags.js
```diff
@@ -186,7 +186,7 @@ const makeSpeedyTag = (
     const size = sizes[marker]
     const sheet = sheetForTag(el)
     const removalIndex = addUpUntilIndex(sizes, marker)
-    deleteRules(sheet, removalIndex, size)
+    deleteRules(sheet, removalIndex - 1, size)
```
probablyup: Let's double check that this is still necessary, feel like this could cause all sorts of weird bugs.

src/models/StyledComponent.js
```diff
@@ -1,10 +1,13 @@
 // @flow
 
+'no babel-plugin-flow-react-proptypes'
```
probablyup: I removed this plugin so you can take this pragma out

src/models/StyleTags.js
imbhargav5: It was necessary to remove that one (index out of bounds)  bug for global styles. But I am not sure what other bugs this might cause. 🤔

src/constructors/createGlobalStyle.js
imbhargav5: Ok. I managed to move the `insert` and `update` of the styling into the render method. So we are only left with `componentWillUnmount` which, I think, is safe to use. Right?

imbhargav5:
@probablyup What do you think I should change in `injectGlobal`? Should I remove the code and tests for injectGlobal?

src/models/StyleTags.js
imbhargav5: Hmmm. Could be. I will need to see what's going on internally. But shouldn't more tests fail if that's the case? Can you think of a test case that I can write to test around this scenario?

probablyup:
There's a `package.lock` in sandbox that needs to go too

imbhargav5:
Would `StyleSheetManager` be mandatory as a top level component for styled components to work on both server and browser?

imbhargav5:
@probablyup  @mxstbr So I removed `injectGlobal` and have been trying to replace it with `createGlobalStyle` for the last few hours and I have a few questions mainly around `StyleSheet`, `StyleTag` and other classes. Can we chat about this somewhere? I feel discussing it in this thread might bloat it and spam everyone.

mxstbr:
:point_right: https://spectrum.chat/styled-components

imbhargav5:
https://spectrum.chat/thread/7c76d241-aeed-447c-b933-b8d0c9f8859b

imbhargav5:
Have been digging through the code to see why Global Styles were always being pushed to the end of the style sheet. From what I understand, local styles are registered during `styled.div` and are applied during render. Global styles don't have this registration right now, which is why they are getting added late into the stylesheet. (Local styles have already been added into the markers. Hence, they are appended to the css stylesheet first). Anyway, so my question is should we always have global styles stringified before any of the local styles or do we have any exceptions?

imbhargav5:
@probablyup I got most of the test cases except for 7 test cases which I skipped. I think they only make sense for injectGlobal and not createGlobalStyle. Can you take a look?

kitten:
@imbhargav5 I left some replies on your Spectrum thread :) Hope this helps

I think it'd be great to have a separate package for `StyleTags` and I'll see if I can carve some time out for that. It'd be great to have more testing and benchmarks around it that are separate from styled-components, and to be able to reuse the implementation for other component-based CSS-in-JS libraries.

imbhargav5:
@kitten Yes! I just read them. And just to reiterate. If someone can take a look at the test cases I skipped in rehydration.test.js and ssr.test.js and suggest the next steps. That would be awesome.

imbhargav5:
@kitten Personally for me, `StyleTags` was the module I found toughest to wrap my head around. It makes sense if we can somehow simplify its behaviour, define Input-Output for all of its internals and write test cases around them.

imbhargav5:
@kitten @probablyup I think there are 3 skipped tests that we need to fix and perhaps add a few more for createGlobalStyle to be finalized. Let's review that and close this? I can't wait to start using createGlobalStyle in my projects 😄

src/models/GlobalStyle.js
```diff
@@ -0,0 +1,45 @@
...
+    componentId: string
+    isStatic: boolean
+
+    constructor(rules: RuleSet, componentId: string) {
+      this.rules = rules
+      this.componentId = componentId
+      this.isStatic = isStaticRules(rules)
+      if (!StyleSheet.master.hasId(componentId)) {
+        const placeholder =
+          process.env.NODE_ENV !== 'production' ? [`.${componentId} {}`] : []
```
kitten: I think you don’t need this “marker class”. Since the global styles won’t be referring to it it’ll be pretty useless 😅

src/models/GlobalStyle.js
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    createStyles(executionContext: Object, styleSheet: StyleSheet) {
+      const flatCSS = flatten(this.rules, executionContext)
+      const css = stringifyRules(flatCSS, '')
+      styleSheet.inject(this.componentId, css, '')
+    }
+
+    renderStyles(executionContext: Object, styleSheet: StyleSheet) {
+      if (styleSheet.hasId(this.componentId)) {
```
kitten: Let’s move this check to remove itself just to be sure.

src/test/rehydration.test.js
```diff
@@ -316,13 +319,15 @@ describe('rehydration', () => {
       `)
     })
 
-    it('should not change styles if rendered in the same order they were created with', () => {
-      injectGlobal`
+    it.skip('should not change styles if rendered in the same order they were created with', () => {
```
kitten: This should still be there I guess to check the injection order. Let me know what you need to fix this up :) if you think another test already covers injection order, we can remove this

src/test/rehydration.test.js
```diff
@@ -340,21 +345,23 @@ describe('rehydration', () => {
       `)
     })
 
-    it('should still not change styles if rendered in a different order', () => {
+    it.skip('should still not change styles if rendered in a different order', () => {
```
kitten: See above; this is critical to hold true as per the test description

src/test/ssr.test.js
```diff
@@ -155,8 +167,8 @@ describe('ssr', () => {
     expect(cssTwo).toMatchSnapshot()
   })
 
-  it('should allow global styles to be injected during rendering', () => {
-    injectGlobal`html::before { content: 'Before both renders'; }`
+  it.skip('should allow global styles to be injected during rendering', () => {
```
kitten: This is probably covered somewhere else so we can delete this one

src/models/GlobalStyle.js
imbhargav5: Oops!  😅

src/test/rehydration.test.js
imbhargav5: So I was thinking that while `injectGlobal` would only inject styles once, `createGlobalStyle` will inject for each rendered component. Actually, I am a bit confused here.

kitten: Ah I See, so it’s related to my later comment. Let’s delete this test for this PR then or keep it slipped with a TODO comment

imbhargav5: Yes. So these two tests seem to be failing massively and they are related to how `name` in general is handled or not-handled with local and global rules. So I will skip them and add comments that they should eventually work before we release.

imbhargav5:
@kitten Also I realise now that `createGlobalStyle` is going to depend on `ComponentStyle` so I added that within the constructor and I added `resetCreateGlobalStyle` util. We aren't using that yet to generate `name` but I commented out that part with a comment that we should be. That's fine right?

imbhargav5:
An integration test seems to be failing. And the test is also failing in the develop branch actually.

kitten:
@imbhargav5 Yea, it seems that something is generally wrong in the integration:

> Expected image to match or be a close match to snapshot but was 24.770355224609375% different from snapshot (194802 differing pixels).

Maybe worth running & updating it locally, looking at the image to check whether something's off, and committing the change.

src/models/GlobalStyle.js
```diff
@@ -15,29 +15,28 @@ export default (stringifyRules: Stringifier) => {
...
-
-        StyleSheet.master.deferredInject(componentId, placeholder)
+        StyleSheet.master.deferredInject(componentId, [])
       }
     }
 
     createStyles(executionContext: Object, styleSheet: StyleSheet) {
       const flatCSS = flatten(this.rules, executionContext)
       const css = stringifyRules(flatCSS, '')
+      // TODO: We will need to figure out how to do this before 4.0
```
kitten: Nice comment :+1:

imbhargav5:
@kitten It's weird. 

[Image from Gyazo](https://gyazo.com/75575aa9c02b72b2776b829d50ba7493)

This bug is in develop branch too I think. So not sure what the cause of the bug is.

kitten:
@imbhargav5 Seems that there's just some minor conflicts :+1: @probablyup if you don't have anything else, I'll merge this and will see what changes the stylesheet will need :)

imbhargav5:
@kitten Just fixed the merge conflicts 😃

example/example.js
```diff
@@ -1,8 +1,8 @@
 import React from 'react'
-import styled, { injectGlobal, keyframes } from '..'
+import styled, { createGlobalStyle, keyframes } from '..'
 
 export default () => {
-  injectGlobal`
+  const GlobalStyle = createGlobalStyle`
```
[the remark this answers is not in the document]
imbhargav5: It is outside a component's render method. So the default export from this file is not rendered but invoked. 
https://github.com/styled-components/styled-components/blob/def31b7381150fab09b790472459e239a53721b1/example/devServer.js#L42
So just like other `h1` and `section` tags in the `Example`, `createGlobalStyle` is fine where it is right now.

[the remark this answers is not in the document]
alexandernanberg: Sorry, you're right. I misread it! Looks OK to me 👍

probablyup:
Let's get this in and start testing scenarios! Thank you for your hard work @imbhargav5! I think we can cut out first beta this week!

imbhargav5:
@probablyup Yes!!

migueloller:
@imbhargav5, thanks so much for the hard work on this PR! I'll be testing the `styled-components:develop` branch extensively and reporting back with any bugs.

kitten:
@migueloller please be aware of the current limitation that you should only mount you global component once or multiple times with identical props, when testing :)

@imbhargav5 thanks for your hard work and welcome to the team! :tada:

imbhargav5:
Thank you @migueloller and @kitten!

janosh:
Very excited to start using this. Thanks for the effort @imbhargav5 and those who came before.

One question: Will the [docs continue to advise against injecting global styles](https://www.styled-components.com/docs/api#injectglobal) and if so, why?

> We do not encourage the use of this. Try to use it once per app at most, if you must, contained in a single file. This is an escape hatch. Only use it for the rare @font-face definition or body styling.

imbhargav5:
@janosh Thank you. Also injectGlobal will now be deprecated from 4.0. createGlobalStyle will now be the only way to add global styles going forward.

rainboxx:
Coming here through #1546, esp. because @kitten mentioned in https://github.com/styled-components/styled-components/issues/1546#issuecomment-367689079 that `injectGlobal` will be deprecated with v4. Is there already a plan for a deprecation warning similar to https://github.com/styled-components/styled-components/pull/1909 for `.extend`?

PR #1877, opened 2018-08-02
 shallow merge style object if present in both props and attrs

### LINKED ISSUE #1849: Specific `attrs` cannot be merged inside component that is wrapped with `styled()`
## System:
 - OS: macOS High Sierra 10.13.5
 - CPU: x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
 - Memory: 135.84 MB / 16.00 GB
 - Shell: 5.3 - /bin/zsh
## Binaries:
 - Node: 10.6.0 - /usr/local/bin/node
 - Yarn: 1.7.0 - /usr/local/bin/yarn
## npmPackages:
 - styled-components: ^3.3.0 => 3.3.3 

## Reproduction

https://codesandbox.io/s/885yo82z19

## Steps to reproduce

Please see the codesandbox. Note that this is a highly contrived example - the inline styles in a real project are ones that change frequently, hence the reason for putting them in the style attribute in the first place (as per #134).

Essentially:

- set up a component which uses a styled-component that defines a `style` key as one of its `attrs`
- set up another component which wraps this first component with `styled` and attempts to set different inline styles

## Expected Behavior

- The `...props.style` spread inside the `Paragraph` component should merge the styles from the `BlueText` component

## Actual Behavior

- Inline styles are not merged.

I believe this is related to #1343, however in that example `extend` is used directly on the styled-component, in this example we are wrapping the styled-component in a regular component, and using the `styled()` wrapper instead.

probablyup:
Fixes #1849

### DIFF
--- a/src/test/attrs.test.js
+++ b/src/test/attrs.test.js
@@ -189,7 +189,7 @@ describe('attrs', () => {
     )
   })
 
-  it('should merge nested "style" attr', () => {
+  it('should shallow merge "style" prop + attr instead of overwriting', () => {
     const Paragraph = styled.p.attrs({
       style: props => ({
         ...props.style,
@@ -224,7 +224,7 @@ describe('attrs', () => {
     })``
 
     expect(shallow(<BlueText>Hello</BlueText>).html()).toEqual(
-      '<p style="color:blue;font-size:4em;" class="sc-b c sc-a d">Hello</p>'
+      '<p style="color:blue;font-size:4em" class="sc-b c sc-a d">Hello</p>'
     )
   })
 })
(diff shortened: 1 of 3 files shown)


probablyup:
I tried to figure how to do this for RN as well but couldn't figure out how "style" is handled from attrs. It doesn't seem like we create a new StyleSheet using that injection vector, so does it even work?

src/test/attrs.test.js
```diff
@@ -86,78 +96,135 @@ describe('attrs', () => {
...
-      children: <span>Amazing</span>
+      children: <span>Amazing</span>,
     })``
-    expect(shallow(<Comp>Something else</Comp>).html()).toEqual('<div class="sc-a b">Something else</div>')
+    expect(shallow(<Comp>Something else</Comp>).html()).toEqual(
+      '<div class="sc-a b">Something else</div>'
+    )
+  })
+
+  it('should shallow merge "style" prop + attr instead of overwriting', () => {
```
probablyup: this is the new test, the rest of the changes are just autoprettier

PR #1876, opened 2018-08-02
Fix broken isStyledComponent type

MayhemYDG:
Fixes #1875

PR #1879, opened 2018-08-03
Fix 'Cannot find name Component'.

willbrazil:


typings/styled-components.d.ts
```diff
@@ -95,7 +95,7 @@ export function withTheme<P extends { theme?: T; }, T>(component: React.Componen
 export function keyframes(strings: TemplateStringsArray, ...interpolations: SimpleInterpolation[]): string;
 export function injectGlobal(strings: TemplateStringsArray, ...interpolations: SimpleInterpolation[]): void;
 export function consolidateStreamedStyles(): void;
-export function isStyledComponent(target: string | Component<object>): boolean;
+export function isStyledComponent(target: string | React.Component<object>): boolean;
```
kachkaev: Can this be `ReactComponent` instead of `React.Component` given that it's already imported?

probablyup:
Dup of https://github.com/styled-components/styled-components/pull/1876. The other one is already approved, so I'm going to close this one. Thanks!

PR #1881, opened 2018-08-03
add error system

probablyup:
Rich errors in development AND production, but production ones will be a link to the error markdown file.

Also includes the error message added in #1801


src/utils/error.js
```diff
@@ -0,0 +1,61 @@
...
+  })
+
+  return a
+}
+
+/**
+ * Create an error file out of errors.md for development and a simple web link to the full errors
+ * in production mode.
+ */
+export default function throwError(
```
kitten: I'm not sure whether I like this `throwError` function actually as it'd deepen the stack unnecessarily. Do we maybe want to create a lean Error class that generates the message and is thrown explicitly?

probablyup: oh hmm sure, that makes sense

src/utils/error.js
```diff
@@ -0,0 +1,61 @@
...
+  code: string | number,
+  ...interpolations: Array<any>
+) {
+  if (process.env.NODE_ENV === 'production') {
+    throw new Error(
+      `An error occurred. See https://github.com/styled-components/styled-components/blob/master/src/utils/errors.md#${code} for more information. ${
+        interpolations
+          ? `Additional arguments: ${interpolations.join(', ')}`
```
kitten: If we'd add `interpolations.map(x => `${x}`).join(', ')` we might be able to avoid having to cast to string before calling this function

probablyup: Hmm? I'm not casting anywhere

sompylasar:
Related: https://twitter.com/acdlite/status/1017881577601187840
https://github.com/facebook/react/tree/f8456b2ecbf194d19d70ff057f251cbc5814bc34/scripts/error-codes

kitten:
✨👏👏👏

PR #1883, opened 2018-08-04
Check if React element was mistakenly used within styles

### LINKED ISSUE #1846: Need Error handling when referencing to none styled components
When you mistakenly reference to a none styled component like so:

```jsx
const Bar = () => (<p>bar</p>)

const Foo = styled.div`
  & ${Bar} {
    //some styles
  }
`
```

you get an `Uncaught TypeError: Cannot convert a Symbol value to a string` error.

It would be helpful if you got an error that said that you can only reference styled components.

imbhargav5:
This PR attempts to fix this issue #1846.

The main idea behind this PR is to throw an error if a React element is mistakenly considered to be a plain object and is converted to `css`. 

Closes #1846


probablyup:
@imbhargav5 could you rebase this to use the new error system? See https://github.com/styled-components/styled-components/pull/1881 for examples of usage.

src/utils/flatten.js
```diff
@@ -52,6 +52,11 @@ const flatten = (
       return ruleSet
     }
 
+    /* Throw if a React Element was given styles */
+    if (React.isValidElement(chunk)) {
+      throw new Error('Cannot reference React Elements within styles')
```
probablyup: Let's also beef up this message to make it more useful:

> A React Element (JSX) was interpolated in your styles, probably as a [component selector](https://www.styled-components.com/docs/advanced#referring-to-other-components). Only styled-component classes can be targeted in this fashion.

imbhargav5: Done.

src/utils/flatten.js
```diff
@@ -52,6 +53,20 @@ const flatten = (
       return ruleSet
     }
 
+    /* Throw if a React Element was given styles */
+    if (React.isValidElement(chunk)) {
```
probablyup: Hmm this seems to be happening too late. We should be doing this check before the interpolated function is executed so we can get the assigned name of the actual component class.

For example:

```js
const NestedComp = () => <button />
const Thing = styled.div`
  ${NestedComp} {
    color: blue;
  }
`
```

Should result in the error being like this:

A plain React class (NestedComp) was interpolated in your styles [...]

src/utils/flatten.js
```diff
@@ -52,6 +53,20 @@ const flatten = (
...
+      const elementName =
+        /* $FlowFixMe TODO: flow for chunk of type React Element. Not sure if Element should be added to Interpolation */
+        typeof chunk.type === 'string'
+          ? chunk.type
+          : chunk.type.displayName || chunk.type.name
+      throw new StyledError(
+        1,
+        `A plain React class (${elementName})
```
kitten: This is a little inaccurate :) it says “class” here instead of “element “

I’d also either remove “your styles” to “has been interpolated into styles” or just change it to an active voice (“ You have interpolated a React element into your styles”)

probablyup: Yeah I don't want it to occur so late. We want to catch the original class being interpolated so the resulting error is useful

imbhargav5: True. I felt the same way about catching the error sooner. Let me know if you know a good place (file/line number) for me to put this. I will see if I can find it out myself.

imbhargav5: Done. Please take a look.

src/utils/flatten.js
```diff
@@ -55,7 +56,16 @@ const flatten = (
...
     if (typeof chunk === 'function') {
       if (executionContext) {
-        ruleSet.push(...flatten([chunk(executionContext)], executionContext))
+        const nextChunk = chunk(executionContext)
+        /* Throw if a React Element was given styles */
+        if (React.isValidElement(nextChunk)) {
+          const elementName = chunk.displayName || chunk.name
+          throw new StyledError(
+            1,
+            `A plain React element (${elementName}) has been interpolated into styles, probably as a component selector (https://www.styled-components.com/docs/advanced#referring-to-other-components). Only styled-component classes can be targeted in this fashion.`
```
probablyup: Also this message should be moved to errors.md. Please see how the other ones are set up.

imbhargav5: Ok. Just did this. Let's go!

src/utils/errors.md
```diff
@@ -56,3 +56,7 @@ Missing document `<head>`
 ## 10
 
 Cannot find sheet for given tag
+
+## 11
+
+A plain React element (%s) has been interpolated into styles, probably as a component selector (https://www.styled-components.com/docs/advanced#referring-to-other-components). Only styled-component classes can be targeted in this fashion.
```
probablyup: This is a component class, not an element. Sorry for the conflicting responses from @kitten and I.

probablyup:
@imbhargav5 yeah I'm not sure. It seems pretty random

src/utils/flatten.js
```diff
@@ -55,7 +56,13 @@ const flatten = (
     /* Either execute or defer the function */
     if (typeof chunk === 'function') {
       if (executionContext) {
-        ruleSet.push(...flatten([chunk(executionContext)], executionContext))
+        const nextChunk = chunk(executionContext)
+        /* Throw if a React Element was given styles */
+        if (React.isValidElement(nextChunk)) {
+          const elementName = chunk.displayName || chunk.name
```
kitten: Small nit; I’m not sure whether it’ll actually catch components here since they’re not elements so the displayName access is potentially dead code?

If I’m mistaken. We also have a utility for component names: https://github.com/bhargav-kitchen/styled-components/blob/aabb2bb0abc90b9c7a2d9822085be05d7505a07c/src/utils/getComponentName.js

Anyway, this might just be a thing for future PRs

mxstbr:

src/utils/flatten.js
probablyup: The babel react preset includes https://babeljs.io/docs/en/babel-plugin-transform-react-display-name which sets displayName on the classes

PR #1886, opened 2018-08-05
#1885: fix missing generic type arguments in .d.ts

### LINKED ISSUE #1885: d.ts: Generic type 'StyledComponentClass<P, T, O>' requires between 2 and 3 type arguments
Just updated to version 3.4.1 and I now get following error when trying to compile my project:

`Generic type 'StyledComponentClass<P, T, O>' requires between 2 and 3 type arguments`

Changing line 179 to below fixes it:
```
export function isStyledComponent(target: any): target is StyledComponentClass<{}, {}>
```

I guess that woud be ok as the generic types aren't relevant when simply checking if a component is a styled component.

PzYon:
fixes #1885 

Another option would be to set the default type params on the `StyledComponentClass`-type itself. But I guess that would mean that when using the `StyledComponentClass`-interface the generic types could be omitted and hence some type-safety would be lost.

### DIFF
--- a/typings/styled-components.d.ts
+++ b/typings/styled-components.d.ts
@@ -176,7 +176,7 @@ export function injectGlobal(
 
 export function consolidateStreamedStyles(): void
 
-export function isStyledComponent(target: any): target is StyledComponentClass
+export function isStyledComponent(target: any): target is StyledComponentClass<{}, {}>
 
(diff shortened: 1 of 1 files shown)


mikewheaton:
Will this be included in a 3.4.2 patch release?

probablyup:
Yes

PR #1892, opened 2018-08-06
Fix rule-injection with deferred rules being mutated while used across clones

### LINKED ISSUE #1890: SSR duplicates stylesheets on refresh
### System:
 - CPU: x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
 - Memory: 111.95 MB / 8.00 GB
 - Shell: 3.2.57 - /bin/bash
### Binaries:
 - Node: 9.10.1 - /usr/local/bin/node
 - Yarn: 1.6.0 - /usr/local/bin/yarn
### npmPackages:
 - babel-plugin-styled-components: ^1.5.1 => 1.5.1 
 - styled-components: 3.4.1 => 3.4.1 

## Reproduction

I cloned the official example of styled-components with NextJS: https://github.com/zeit/next.js/tree/master/examples/with-styled-components

It works as expected with initial version *^2.1.0*.
It works as expected with version *3.3.3*.
__It does not work as expected with *3.4.0* and *3.4.1*__

Using different versions of NextJS or the Babel plugin did not impact the behavior.

## Steps to reproduce
Run `npm run dev`, view page source and hit refresh with >= 3.4.0.

## Expected Behavior
The created stylesheet exists once.

## Actual Behavior
The created stylesheet gets duplicated with every browser refresh.
~It does *not* occur after a hot reload.~ 
It also happens in production mode. [See comment](https://github.com/styled-components/styled-components/issues/1890#issuecomment-410644078) by @elgubbo

---

~It's by no means a show blocker, since you usually don't refresh when using HMR, but~ I think it still might be a bug.
As far as I saw in the NextJS repo this occured as well before styled-components SSR was stable at some point.
If you'd like me to try to isolate the problem out of NextJS, shout.
Thank you!

kitten:
Fix #1890 

A regression was introduced and not caught in time in #1843 which has gone out recently.

What happens is:

- Components are created and call `deferredInject` on a `StyleSheet`
- Some clones are created from StyleSheet.master which inherit all deferred arrays
- The prior two steps might also be in reverse order; doesn't matter
- `inject` is called on render which takes deferred rules into account
- During the injection the `deferred[id]` array is mutated (to avoid `concat`)
- The mutated array is then used in the same manner in other stylesheets, which add the rules that need to be injected again

Hence the outcome is that rules are duplicated for each subsequent clone that injects them.

### DIFF
--- a/src/models/StyleSheet.js
+++ b/src/models/StyleSheet.js
@@ -275,13 +275,16 @@ export default class StyleSheet {
     const tag = this.getTagForId(id)
 
     /* add deferred rules for component */
-    if (this.deferred[id]) {
-      this.deferred[id].push(...cssRules)
-
-      // $FlowFixMe
-      tag.insertRules(id, this.deferred[id], name)
+    if (this.deferred[id] !== undefined) {
+      // Combine passed cssRules with previously deferred CSS rules
+      // NOTE: We cannot mutate the deferred array itself as all clones
+      // do the same (see clones[i].inject)
+      const rules = this.deferred[id].concat(cssRules)
+      tag.insertRules(id, rules, name)
       this.deferred[id] = undefined
-    } else tag.insertRules(id, cssRules, name)
+    } else {
+      tag.insertRules(id, cssRules, name)
+    }
   }
 
   /* removes all rules for a given id, which doesn't remove its marker but resets it */
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 ## Unreleased
 
-* placeholder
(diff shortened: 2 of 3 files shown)


src/models/StyleSheet.js
```diff
@@ -275,13 +275,16 @@ export default class StyleSheet {
     const tag = this.getTagForId(id)
 
     /* add deferred rules for component */
-    if (this.deferred[id]) {
-      this.deferred[id].push(...cssRules)
-
-      // $FlowFixMe
-      tag.insertRules(id, this.deferred[id], name)
+    if (this.deferred[id] !== undefined) {
```
probablyup: Can this be one-lined?

kitten: I suppose it could but ... readability 😅

probablyup: Yeah I was just thinking code size. But it's cool, LGTM

PR #1894, opened 2018-08-06
Use new context API

### LINKED ISSUE #1459: Switch to new context API
React will soon introduce [a new API for context](https://github.com/reactjs/rfcs/pull/2). We should start using the latest version of [`react-broadcast`](https://github.com/ReactTraining/react-broadcast), which is a polyfill of that API. When we first built styled-components we inlined an implementation of `react-broadcast` to avoid introducing an unnecessary layer of components, so that needs to be removed too.

PRs very welcome! 💯 

/cc @gaearon

### LINKED ISSUE #1655: Remove usages of unsafe lifecycle methods

alexandernanberg:
This is just to get the ball rolling with the new context API

Fixes #1459 
Fixes #1655

Pretty happy with how the `ThemeProvider` turned out, a lot cleaner. I'll continue working on this and hopefully fully landing this PR. 

~~Enzyme does not support the react > 16.3 yet so it's hard to test this until that gets fixed~~. We probably want to use a polyfill for this as well, like [create-react-context](https://github.com/jamiebuilds/create-react-context) unless we drop React < 16.3 support

No hard feelings if you decide not to use this!


probablyup:
Good start! Let's remove that memoize-one dep that looks to have slipped in.

At this point for 4.0 I think we will just plan to drop old React support and just target 16.4+ to keep things easy.

alexandernanberg:
Oh I actually just forgot to use the `memoize` function, we need to either memoize the `getContext` method or move the context to `state` because otherwise we'll provide a new context value each render which will make consumers update when they might not need to. 

Cool, nice to save some KBs as well! 👍

probablyup:
Let's use `state` and recalculate upon prop change by hard equality

alexandernanberg:
Okay, although I think using the memoized approach is a bit cleaner and simpler, because that you don't need to recalculate with eg `getDerivedPropsFromState`. But either way is fine

alexandernanberg:
I think the `ThemeProvider` should work now 😮 

Switched the test to `react-testing-library` because `enzyme` wasn't working (no react 16.3+ compability). If you don't want to switch over entirely to `react-testing-library` we could migrate back later. 

Left the memoization for now but should be fairly simple to change if you'd still like that

src/models/StyledComponent.js
```diff
@@ -110,7 +87,8 @@ class BaseStyledComponent extends Component<*, BaseState> {
 
   generateAndInjectStyles(theme: any, props: any) {
     const { attrs, componentStyle, warnTooManyClasses } = this.constructor
-    const styleSheet = this.context[CONTEXT_KEY] || StyleSheet.master
+    // const styleSheet = this.context[CONTEXT_KEY] || StyleSheet.master
+    const styleSheet = StyleSheet.master
```
alexandernanberg: I'm unsure what this previously was for, but I think it will be fine just assigning it to `Stylesheet.master`?

src/models/StyledComponent.js
```diff
@@ -137,118 +115,80 @@ class BaseStyledComponent extends Component<*, BaseState> {
...
+      target,
+      defaultProps,
+      componentStyle,
+    } = this.constructor
+    const isTargetTag = isTag(target)
 
-    return createElement(target, propsForElement)
+    return (
+      <ThemeConsumer>
+        {({ theme } = {}) => {
```
probablyup: Can you use the empty object from utils/empties?

probablyup: Can we also cache this function somehow instead of recreating it every render?

src/models/StyledComponent.js
probablyup: @kitten can you comment on this? I think it's okay

src/models/StyledComponent.js
alexandernanberg: If I use `EMPTY_OBJECT` I get a flow error though, so the best way might create a `type` for the context and not destructure it, right? I've not really worked with flow before so I might need some help on certain bits

Yeah good point, although I don't it's that expensive to recreate it

probablyup: It's more about memory pressure. This is a hot path, so we want it to be as maximally-efficient as possible.

alexandernanberg: Makes sense, will refactor to take this into account!

src/models/StyledComponent.js
alexandernanberg: Oh just realized that that context comes from the `StyleSheetManager`, so I'll have to wrap it in a `StyleSheetConsumer` or similar to get the correct value

src/models/ThemeProvider.js
```diff
@@ -133,18 +53,35 @@ export default class ThemeProvider extends Component<ThemeProviderProps, void> {
...
-    return { ...this.outerTheme, ...(theme: Object) }
+    return { ...outerTheme, ...(theme: Object) }
   }
 
-  publish(theme: Theme | ((outerTheme: Theme) => void)) {
-    this.broadcast.publish(this.getTheme(theme))
+  getContext(theme: (outerTheme: ?Theme) => void, outerTheme?: Theme) {
+    return {
+      theme: this.getTheme(theme, outerTheme),
+    }
```
alexandernanberg: What if we instead of using setting context to `{ theme }` we set it to `theme` without a wrapping object, we won't be able to add any other properties on the context but I don't think we will want that anyways?

alexandernanberg:
Okay so I've hit a problem that I don't really get why it's occurring. The `BaseStyledComponent` doesn't seem to get any static properties, like `attrs`, `target` and so on. They will be `undefined` in the `render` method which causes all sorts of issues. 

Would appreciate if someone could take a look!

alexandernanberg:
Tests are passing 🎉  

Would love to get some second opinions and feedback on this now @kitten @mxstbr  @probablyup

probablyup:
Nice! Ok so the last thing I want to do here is some performance analysis.

Specifically making sure we aren't regressing against current master. Could you rig up a test using ThemeProvider in the benchmark suite (it's in this repo under benchmarks.)

probablyup:
Let's also switch this branch to be pointing at "develop" which is the 4.0 development branch I just made.

alexandernanberg:
Yeah I thought about that too, but the current test-suit won't work with react 16 right? Any ideas on how we'd solve that?

probablyup:
@alexandernanberg It works with React 16 https://github.com/styled-components/styled-components/blob/e3cd52ef93c53d9c5923f91b20552334a857b67f/benchmarks/package.json#L18

https://github.com/styled-components/styled-components/tree/e3cd52ef93c53d9c5923f91b20552334a857b67f/benchmarks

src/models/StyleSheetManager.js
```diff
@@ -29,27 +26,34 @@ export default class StyleSheetManager extends Component<Props, void> {
...
 
-  sheetInstance: StyleSheet
+  getContext: (sheet: ?StyleSheet, target: ?HTMLElement) => Object
+
+  constructor(props: Props) {
+    super(props)
 
-  getChildContext() {
-    return { [CONTEXT_KEY]: this.sheetInstance }
+    this.getContext = memoize(this.getContext.bind(this))
```
kitten: small nit, I'd prefer it if instead of a constructor, we'd just wrap `getContext` inline, e.g.:


```js
  getContext = memoize((...) => {...});
```

flow-typed/npm/memoize-one_vx.x.x.js
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Fill this stub out by replacing all the `any` types.
+ *
+ * Once filled out, we encourage you to share your work with the
+ * community by sending a pull request to:
+ * https://github.com/flowtype/flow-typed
+ */
+
+declare module 'memoize-one' {
+  declare module.exports: any
```
kitten: This should probably be sufficient here; haven't tested it but it should be vaguely correct 😅 

```js
declare function memoize<R, A>(fn: (A) => R): (A) => R;
declare function memoize<R, A, B>(fn: (A, B) => R): (A, B) => R;
declare function memoize<R, A, B, C>(fn: (A, B, C) => R): (A, B, C) => R;
declare function memoize<R, A, B, C, D>(fn: (A, B, C, D) => R): (A, B, C, D) => R;
declare function memoize<R, A, B, C, D, E>(fn: (A, B, C, D, E) => R): (A, B, C, D, E) => R;

declare var exports: typeof memoize;
declare export default typeof memoize;
```

alexandernanberg:
Master branch
```
---Mount deep tree---
19.60 ±02.24 ms
(S/L) 15.73/03.86 ms

Trace written to mount-deep-tree-trace.json

---Mount wide tree---
28.67 ±02.19 ms
(S/L) 22.87/05.80 ms

Trace written to mount-wide-tree-trace.json

---Update dynamic styles---
24.36 ±05.49 ms
(S/L) 08.58/15.78 ms

Trace written to update-dynamic-styles-trace.json
```

This PR
```
---Mount deep tree---
17.13 ±01.80 ms
(S/L) 13.14/03.99 ms

Trace written to mount-deep-tree-trace.json

---Mount wide tree---
23.17 ±02.92 ms
(S/L) 17.31/05.85 ms

Trace written to mount-wide-tree-trace.json

---Update dynamic styles---
24.16 ±05.72 ms
(S/L) 08.35/15.81 ms

Trace written to update-dynamic-styles-trace.json
```

src/models/StyleSheetManager.js
alexandernanberg: Yeah I agree that this looks a lot better, but after reading [articles like this](https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1) I had my concerns with perf among other things. But this might not be relevant anymore, not sure.

alexandernanberg:
Okay so I’m not sure that I interpret the benchmark results correctly, is a high or low number better? First thought it was high=better but I just noticed the ms suffix which made me uncertain

Anything you want me to fix before this can get merged?

probablyup:
Lower ms is better. I think @kitten is right though, there are a bunch of other changes we're going to make in this branch so perf analysis might be premature.

package.json
```diff
@@ -150,7 +148,7 @@
     "typescript": "^2.4.1"
   },
   "peerDependencies": {
-    "react": ">= 0.14.0 < 17.0.0-0"
+    "react": ">= 16.3.0 < 17.0.0-0"
```
probablyup: Let's just make this `>= 16.3` for simplicity.

src/hoc/withTheme.js
```diff
@@ -18,70 +19,43 @@ export default (Component: ComponentType<any>) => {
...
-
-      if (!shouldSetInnerRef) {
-        props.ref = props.innerRef
-        delete props.innerRef
-      }
-
-      return React.createElement(Component, props)
+      return (
+        <ThemeConsumer>
+          {(theme?: Theme) => {
```
probablyup: Let's cache the function on the class so it's not getting recreated per render.

src/models/StyledComponent.js
```diff
@@ -137,127 +113,103 @@ class BaseStyledComponent extends Component<*, BaseState> {
...
-            : this.props[key]
-      }
-    }
-
-    return createElement(target, propsForElement)
+    // TODO: Refactor to use cached functions
+
+    return (
+      <StyleSheetConsumer>
+        {(styleSheet?: StyleSheet) => (
```
probablyup: Same here, let's cache it on the class.

src/models/StyledNativeComponent.js
```diff
@@ -144,31 +84,49 @@ class BaseStyledNativeComponent extends Component<*, State> {
...
-      propsForElement.ref = this.onRef
-      delete propsForElement.innerRef
-    } else {
-      propsForElement.innerRef = this.onRef
-    }
-
-    return createElement(target, propsForElement, children)
+    return (
+      <ThemeConsumer>
+        {(theme?: Theme) => {
```
probablyup: And here

src/models/ThemeProvider.js
```diff
@@ -1,121 +1,38 @@
...
+ */
+export default class ThemeProvider extends Component<Props> {
+  getContext: (
+    theme: Theme | ((outerTheme: Theme) => void),
+    outerTheme?: Theme
+  ) => Theme
+
+  constructor(props: Props) {
+    super(props)
+    this.getContext = memoize(this.getContext.bind(this))
```
probablyup: Don't need to do `constructor` if you use this pattern on the class itself:

```js
class ThemeProvider extends Component<Props> {
  getContext = memoize(
    (theme: (outerTheme: ?Theme) => void, outerTheme?: Theme) => this.getTheme(theme, outerTheme)
  )
}
```

probablyup:
This is looking great! Let's do this final cleanup pass and get it in!

src/models/ThemeProvider.js
alexandernanberg: Yeah I actually wrote it similar to that first, but refactored after reading this [thread](https://github.com/facebook/react/issues/9851). I was worried about booth perf and that the output isn't the same. Example [here](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=PTAEBEFMGMBsEMBOlT1AW0gFwBYHsATVAZ1QDtQBLM4reM6FAB0TyckSwE8AoOeYqQBiePKADePUKABmo0AF5QACgCUigHyhoeGnliQAdLDwBzZQHI5eC6p4BfHk5CgAQtQLVTqDNnxEBcioaOgZmVnZOXn5BNyQJKW1dWkQAV2gsPEQ1BOlpXEpiQwAjeKUCotLEEo9lCtUAbkTHaUSqnMk8pL0DYzNLKtsm6Ud7IA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=6.26.0&envVersion=).

However if you feel that that's not a problem I'll happily change

alexandernanberg:
Cool, in that case this PR actually outperforms the master branch, correct? 

So I tested to use class properties instead of binding the function in the constructor and it performed worse :/ Feel free to try this yourself as well, in case I did something wrong.

Also tested to break out the inlined functions inside the render method, and that didn't seem to make any notable difference other than we run into the same problem described above.

It might be easier to get this in the develop branch and then iterate on these things?

probablyup:
@alexandernanberg Yeah that's possibly true! We did make a lot of changes.

src/models/ThemeProvider.js
probablyup: We use loose mode, so ends up looking more like this: https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=true&code_lz=PTAEBEFMGMBsEMBOlT1AW0gFwBYHsATVAZ1QDtQBLM4reM6FAB0TyckSwE8AoOeYqQBiePKADePUKABmo0AF5QACgCUigHyhoeGnliQAdLDwBzZQHI5eC6p4BfHk5CgAQtQLVTqDNnxEBcioaOgZmVnZOXn5BNyQJKW1dWkQAV2gsPEQ1BOlpXEpiQwAjeKUCotLEEo9lCtUAbkTHaUSqnMk8pL0DYzNLKtsm6Ud7IA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&sourceType=module&lineWrap=true&presets=react%2Cstage-2%2Cenv&prettier=false&targets=&version=6.26.0&envVersion=1.6.2

probablyup: You can leave it though, I'll experiment with it both ways later

src/models/test/ThemeProvider.test.js
```diff
@@ -2,117 +2,93 @@
 /* eslint-disable react/no-multi-comp */
 import React from 'react'
 import { shallow, render, mount } from 'enzyme'
-import ThemeProvider, { CHANNEL_NEXT, CONTEXT_CHANNEL_SHAPE  } from '../ThemeProvider'
+import ThemeProvider, { ThemeConsumer } from '../ThemeProvider'
```
probablyup: Maybe instead of `ThemeConsumer` we should test this with the `withTheme` HOC and normal `styled` constructors instead. That way we are ensuring the external API doesn't change.

probablyup:
I'm happy with this once the last testing point is resolved. Excellent work 👊

src/models/test/ThemeProvider.test.js
alexandernanberg: Okay so would this be OK or do we want to use `expectCSSMatches` function as it's done in `theme.test.js`?
```js
  it('should merge its theme with an outer theme', () => {
    const outerTheme = { main: 'black' }
    const innerTheme = { secondary: 'black' }

    const childrenSpy = jest.fn()
    const MyDiv = ({ theme }) => {
      childrenSpy(theme)
      return null
    }
    const MyDivWithTheme = withTheme(MyDiv)

    render(
      <ThemeProvider theme={outerTheme}>
        <ThemeProvider theme={innerTheme}>
          <MyDivWithTheme />
        </ThemeProvider>
      </ThemeProvider>
    )

    expect(childrenSpy).toHaveBeenCalledTimes(1)
    expect(childrenSpy).toHaveBeenCalledWith({
      ...outerTheme,
      ...innerTheme,
    })
  })
```

alexandernanberg:
Superb! Thank you 😄

probablyup:
It's going in!

mxstbr:

Issue #1899, opened 2018-08-08
Object.getPrototypeOf(StyledComponent).isReactComponent is false


privard:
## Steps to reproduce
https://codesandbox.io/s/0m23286p7v

## Actual Behavior
Object.getPrototypeOf(StyledComponent).isReactComponent is false

## Expected Behavior
Seems to me like Object.getPrototypeOf(StyledComponent).isReactComponent should be true, since a StyledComponent is an extension of a ReactComponent. 

## System:
 - OS: macOS Sierra 10.12.6
 - CPU: x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
 - Memory: 124.21 MB / 16.00 GB
 - Shell: 3.2.57 - /bin/bash
## Binaries:
 - Node: 8.11.2 - ~/.nvm/versions/node/v8.11.2/bin/node
 - Yarn: 1.2.1 - /usr/local/bin/yarn
 - npm: 5.6.0 - ~/.nvm/versions/node/v8.11.2/bin/npm
 - Watchman: 4.9.0 - /usr/local/bin/watchman
## npmPackages:
 - styled-components: ^3.4.2 => 3.4.2

privard:
isReactComponent is not a public React API and shouldn't be used to check if an object is a valid react component. See closed PR for more details.

PR #1900, opened 2018-08-08
Fix: #1899 StyledComponent.prototype.isReactComponent is false

privard:
- Matched BaseStyledComponent.isReactComponent to BaseStyledComponent.prototype.isReactComponent
- Added unit test

### DIFF
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -252,6 +252,9 @@ class BaseStyledComponent extends Component<*, BaseState> {
   }
 }
 
+BaseStyledComponent.isReactComponent =
+  BaseStyledComponent.prototype.isReactComponent
+
 export default (ComponentStyle: Function, constructWithOptions: Function) => {
   const createStyledComponent = (
     target: Target,
--- a/src/models/test/StyledComponents.test.js
+++ b/src/models/test/StyledComponents.test.js
@@ -1,0 +1,17 @@
+// @flow
+/* eslint-disable react/no-multi-comp */
+import { Component } from 'react'
+import { resetStyled } from '../../test/utils'
+
+let styled
+
+describe('StyledComponents', () => {
+  beforeEach(() => {
+    styled = resetStyled(true)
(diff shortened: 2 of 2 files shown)


src/models/StyledComponent.js
```diff
@@ -252,6 +252,9 @@ class BaseStyledComponent extends Component<*, BaseState> {
   }
 }
 
+BaseStyledComponent.isReactComponent =
```
probablyup: Why does this need to be a static property instead of an inherited one? I thought the recommended way to check for this stuff now was `react-is` also..

privard: Hey thanks for the comment!

I'm not sure I see another way of making it "inherited", but I'm all ears. What do you propose?

Didn't know about react-is, but I'll run another test to see if it correctly identifies the prototype of a styled-component as a React component.

kitten:
I'm not quite sure what this is supposed for, since it's not part of React's public API (top-level API) as documented? Can you maybe elaborate on this a little?

Why would someone check `X.isReactComponent` instead of checking prototypal inheritance? It doesn't seem like a usual API to me, like `isReactElement`.

Sorry for the comment here. Usually I'd put that on the issue, but in the presence of both a PR and an issue I'll just put it here :)

privard:
You guys are right. I had looked at the Facebook/react repo before creating the PR to make sure that isReactComponent was used in React's source code, and it was. 

I was using other libraries, e.g. React Table, that used it as well to [check for the best method to instantiate an object](https://github.com/react-tools/react-table/blob/9fe806d8a7e11e78e5b063cb143499023ece1bb7/src/utils.js#L208), and this created some conflicts with styled components.

But after a bit more digging, I found this [tweet by Dan Abramov](https://twitter.com/dan_abramov/status/980510354810732544) saying it's not a public API and essentially a "quirk".

I will close the PR and the issue, and reach out to the React Table developers to convince them to switch to react-is. Thanks!

PR #1907, opened 2018-08-11
remove jsnext:main package.json entry point

### LINKED ISSUE #1825: Remove jsnext:main in major release
`jsnext:main` was initiated by rollup and then replaced to enabled by default `module` field. It's better to remove it so nobody will copy it in their projects.

probablyup:
It's been superceded by "module"

Closes #1825


package.json
```diff
@@ -196,4 +195,4 @@
     "type": "opencollective",
     "url": "https://opencollective.com/styled-components"
   }
-}
+}
```
TrySound: This should not be changed I think

probablyup: Yeah it's ok, prettier keeps doing it but it doesn't break anything

PR #1909, opened 2018-08-11
add warning for the removal of extend in v4

probablyup:


probablyup:
That sounds like a lot of backbending. The React 16 upgrade was pretty easy and it's been out long enough that I'm not concerned.

mxstbr:
Sure, then let's just warn about extend!

src/models/StyledNativeComponent.js
```diff
@@ -17,6 +18,16 @@ type State = {
   generatedStyles: any,
 }
 
+let warnExtendDeprecated
+if (process.env.NODE_ENV !== 'production') {
+  warnExtendDeprecated = once(() => {
+    // eslint-disable-next-line no-console
+    console.error(
+      'Warning: The "extend" API has been removed in the upcoming v4.0 release. Use styled(StyledComponent) instead. You can find more information here: https://github.com/styled-components/styled-components/issues/1546'
```
mxstbr: Maybe change the wording to `The "extend" API will be removed... ...due to issues with server-side rendering.`?

probablyup: There were other reasons too. Linking to the issue isn't enough?

mxstbr: Oh yeah I didn't see that, but we should still say "will be"?

probablyup: Sure

probablyup:
@mxstbr done

mxstbr:
Yesss

PR #1911, opened 2018-08-12
Upgrade dependencies

alexandernanberg:
Just upgrade the dependencies to the latest version. Left out typescript deps because they will be removed soon anyways. Also didn't touch `react-native` because tests began to fail after I upgraded to `v0.56.0`.


alexandernanberg:
Okay so there seems to be an issue with node 6 and `jest-styled-components` 6. Do we want to keep running travis with node 6 or should switch to just running with 8 and 10 instead?

probablyup:
@alexandernanberg We can do 8 + 10 in the 4.0 branch yeah

PR #1912, opened 2018-08-12
add "no-tags" entry to be used in conjunction with babel plugin

probablyup:
This allows babel users to drop the element whitelist and save some bytes.

Related to https://github.com/styled-components/babel-plugin-styled-components/issues/125

### DIFF
--- a/integration-test/styled-components-no-tags.test.js
+++ b/integration-test/styled-components-no-tags.test.js
@@ -1,0 +1,20 @@
+import React from 'react'
+import styled, { css } from 'styled-components'
+import renderer from 'react-test-renderer'
+
+jest.mock('styled-components', () =>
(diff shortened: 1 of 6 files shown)


probablyup:
I also went ahead and lowered the bundlesize thresholds now that a bunch of code has been taken out for various things 😄

alexparish:
Is this ready to use with styled-components 4.0 and if so is it documented anywhere?

probablyup:
It was an experiment that we decided not to move forward with and removed

alexparish:
> It was an experiment that we decided not to move forward with and removed

OK fair play. I just found your comment saying that `react-dom` ships this list anyway so there is negligible size saving once the bundle has been compressed.

PR #1913, opened 2018-08-12
cache functions & use bind 

probablyup:
Does seem to have a small perf bonus over class binding

_before_
```
---Mount deep tree---
17.49 ±02.15 ms
(S/L) 13.58/03.90 ms

---Mount wide tree---
24.80 ±03.00 ms
(S/L) 18.80/06.00 ms

---Update dynamic styles---
23.78 ±05.86 ms
(S/L) 08.34/15.43 ms
```

_after_
```
---Mount deep tree---
17.34 ±01.85 ms
(S/L) 13.28/04.06 ms

---Mount wide tree---
23.02 ±02.82 ms
(S/L) 17.03/05.99 ms

---Update dynamic styles---
22.66 ±05.02 ms
(S/L) 07.77/14.88 ms
```


probablyup:
cc @alexandernanberg

alexandernanberg:
Nice, I also ran the benchmarks and can confirm that there is an improvement! Only concern I have is that the render method in `BaseStyledComponent` is not pure (correct me if I'm wrong) because we set `this.styleSheet`  and that could cause problems when async rendering is introduced in the future. I've not read up too much about async rendering yet so I could be wrong about this. 

Maybe we could do something like this to avoid setting `this.styleSheet`? 
```js
  renderOuter(styleSheet?: StyleSheet) {
    return <ThemeConsumer>{this.renderInner(styleSheet)}</ThemeConsumer>
  }
```

probablyup:
It’s okay because the stylesheet is not likely to ever change for a
particular component instance.

alexandernanberg:
Okay, LGTM then 👍

mxstbr:
> It’s okay because the stylesheet is not likely to ever change for a particular component instance.

I've never thought about this, but with the new `StyleSheetManager` folks could theoretically dynamically swap out the style sheet. 😕 Should we maybe add some validation to it to make sure folks don't do that?

src/models/StyledComponent.js
```diff
@@ -58,8 +58,102 @@ class BaseStyledComponent extends Component<*> {
...
+    this.renderOuter = this.renderOuter.bind(this)
+    this.renderInner = this.renderInner.bind(this)
+  }
+
+  render() {
+    return <StyleSheetConsumer>{this.renderOuter}</StyleSheetConsumer>
+  }
+
+  renderOuter(styleSheet?: StyleSheet) {
+    this.styleSheet = styleSheet
```
mxstbr: Does this need an `if (!this.styleSheet) this.styleSheet = styleSheet`?

mxstbr:
I also realized that we could potentially be `WeakMap`ping incoming props and the generated class name? Something like this, pseudo code:

```JS
constructor() {
  this.cache = new WeakMap();
}

render() {
  let className = this.cache.get(this.props);
  // Only generate and inject if the props changed to something we haven't seen before
  if (!className) {
    className = this.generateAndInject();
    cache.set(this.props, className);
  }

  /* ... */
}
```

Could that work theoretically? This is obviously not prod ready by any measure, we probably want to implement [this `map.clear` method on top](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap#Implementing_a_WeakMap-like_class_with_a_.clear()_method) and call that in `componentWillUnmount` to avoid ballooning memory, but I think this could make us a lot faster. I also think that this is what @threepointone did in glamor maybe?

threepointone:
you probably won't need to `.clear()` it manually, props should get gced naturally. 

you should make certain you get the same instance of props on every render tho. I don't think react reuses a props object if the individual values are the same. so something like - 
`<Box x={1} y={2}/>` might actually miss the weakmap cache every time. please verify!

mxstbr:
> you should make certain you get the same instance of props on every render tho. I don't think react reuses a props object if the individual values are the same. so something like -
<Box x={1} y={2}/> might actually miss the weakmap cache every time.

Ah crap I didn't consider that :confused: I assume they don't, since why would they, so this might not work after all. Ugh, that's annoying.

probablyup:
I’ve tried caching this stuff in previous branches and things broke in
bizarre ways

mxstbr:
@tkh44 do you have some solution for caching generated styles based on props?

tkh44:
> @tkh44 do you have some solution for caching generated styles based on props?

Nothing other than what has been proposed above. Ignoring the theme prop, most props that are "flags" for styles are primitives (booleans usually) so maybe a regular object hash map could work for most cases. There are so many edge cases and it requires quite a bit of discipline from users so its a hard sale.

As for this PR in general, the numbers are so close that I would hesitate to say one is faster than the other under any real world circumstances. Have you happened to check the differences in memory usage @probablyup? That might be a more objective representation of "better" in this case due to the similarites in speed.

probablyup:
It’s somewhat hard to say because the profiler isn’t granular enough to
give a strong signal. But I do think that the late binding appears to be
less intensive to the VM. I consistently see 1.5-2ms faster on the hot path
with this change.

probablyup:
> I've never thought about this, but with the new StyleSheetManager folks could theoretically dynamically swap out the style sheet.

Since the instance variable is updated inline during render and only referenced within the same frame, there's no risk.

src/models/StyledComponent.js
probablyup: Nope, we want to continually update it. This is just a place to stash the variable so we can reference it in `renderInner` without making new functions every render.

PR #1919, opened 2018-08-15
 add a new [data-styled-version] attr to styles

probablyup:
This will be hardcoded into the bundles at each release. Basically, it allows multiple versions of styled components to peacefully coexist on a page. There will still be a warning though since this isn't recommended.

This PR also changes the primary attribute from `data-styled-components` to `data-styled` to avoid conflicts with legacy versions of the library.


probablyup:
Because old SC will is not paying attention to the version attribute and
will think those style tags are its own and potentially modify them.

kitten:
@mxstbr @probablyup Yep, I think changing the SC attribute’s name is absolutely necessary to avoid conflicts. The old stylesheets might still rehydrate the new styles away indeterministically otherwise. Not sure if we need both attributes around however

src/test/globals.js
```diff
@@ -1,3 +1,4 @@
 // @flow
-// eslint-disable-next-line no-underscore-dangle
+/* eslint-disable no-underscore-dangle */
 global.__SERVER__ = typeof document === 'undefined'
+global.__VERSION__ = '0.0.0'
```
kitten: Can we maybe set this to sth that makes it clear that Jest is using a “mocked” global here? Possibly “JEST_MOCK_VERSION” or sth?

probablyup: Sure, makes sense

jackrobertscott:
@probablyup In this example, I notice you've changed the "data-styled-components" to "data-styled". In practice, how would you be able to change that value when using styled-components? Or alternatively, how would this prevent conflicts with multiple instances? Thanks!

probablyup:
You can control it via process.env.SC_ATTR at build time:
https://github.com/styled-components/styled-components/blob/master/src/constants.js

My change allows multiple versions to coexist, but if you need multiple
instances of the same version then the build variable is the way to go.
That’s actually available now in v3.

PR #1921, opened 2018-08-15
fix removal of PropTypes import in production builds

probablyup:
Frustrating, but because rollup doesn't have an internal concept
of prod vs dev, it doesn't set NODE_ENV accordingly and babel
uses the same config for everything. So we are able to eliminate
the references to prop types, but the module import itself is not
being eliminated without this change.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -12,7 +12,6 @@
 .*/benchmarks/.*
 .*/dist/.*
 .*/lib/.*
-.*/node_modules/babel-plugin-flow-react-proptypes/src/__test__/.*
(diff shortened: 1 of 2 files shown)


.flowconfig
```diff
@@ -12,7 +12,6 @@
 .*/benchmarks/.*
 .*/dist/.*
 .*/lib/.*
-.*/node_modules/babel-plugin-flow-react-proptypes/src/__test__/.*
```
probablyup: Got rid of this plugin last week

PR #1923, opened 2018-08-16
use React.forwardRef

probablyup:
~~This probably doesn't work yet, but I'm waiting for an enzyme adapter release so I can actually test the functionality.~~


flow-typed/react_v16.x.x.js
```diff
@@ -0,0 +1,347 @@
+// from https://github.com/facebook/flow/pull/6510
```
probablyup: The current flowtyped typings don't include `forwardRef`.

probablyup:
@gaearon when you get some time, could you take a look at this PR? I'm feeling a bit conflicted at how the ref forwarding is implemented. Since the `React.forwardRef` API introduces an interim component, there's some backflips happening to maintain things like letting the consumer set `defaultProps` and then doing referential checks against them, etc.

It seems like this API maybe should have been implemented as a decorator.

src/hoc/withTheme.js
```diff
@@ -57,5 +39,11 @@ export default (Component: ComponentType<any>) => {
     }
   }
 
-  return hoistStatics(WithTheme, Component)
+  const ForwardRef = React.forwardRef((props, ref) => (
+    <WithTheme {...props} forwardedRef={ref} />
```
mxstbr: Couldn't we call this `innerRef`, like we did before? Then we wouldn't even have to change the above code, and old code would keep working as-is I think.

src/utils/isDerivedReactComponent.js
```diff
@@ -0,0 +1,22 @@
+// @flow
+import { Component } from 'react'
+
+const cachedResults = new Map()
```
mxstbr: browser support isn't awesome on map, this will probably break old browsers—need to fall back to no caching in old browsers most likely?

probablyup: It's required for React 16, so no big deal.

src/hoc/withTheme.js
probablyup: I guess, but with `forwardedClass` also being added I like keeping the recommended name for similarity

src/utils/isDerivedReactComponent.js
mxstbr: Oh shit it is? Are you sure? That'd be huge, them dropping IE 9 and 10!

probablyup: Well, just requiring polyfills anyway 😛

src/hoc/withTheme.js
mxstbr: Not sure that's worth much, given it's only internal? I'd prefer if the migration path was nicer, especially if we can warn in dev mode if `innerRef` was explicitly passed and warn to change to `ref`.

probablyup: Well the concept of `innerRef` is totally gone now in general. Just use `ref`. Very easy codemod or grep+replace

probablyup: I'll add a dev warning for innerRef though

src/utils/isDerivedReactComponent.js
kitten: @probablyup You probably would use a `WeakMap`. I've seen it's used for attrs which sticks around since it's never not present (and thus GC'd), but just to be sure I'd still use it :)

probablyup:
WeakMap isn’t guaranteed to be polyfilled

src/utils/isDerivedReactComponent.js
probablyup: So only `Map` and `Set` are guaranteed to be polyfilled due to React 16's environment requirements. If we wanted to use a `WeakMap` we'd have to ship a polyfill for it and bloat the library size.

src/utils/isDerivedReactComponent.js
```diff
@@ -0,0 +1,22 @@
...
+
+export default function isDerivedReactComponent(fn: any) {
```
gaearon: What is this for?

If you want to check if function is a React component, you need to do `fn.prototype && fn.prototype.isReactComponent`. That's what React does.

probablyup: Ah okay, I'll change it in a followup PR. Thanks!

gaearon: You probably also won't want to cache it because I'd expect two property lookups to be faster than what this is doing

probablyup: Done, thanks for the feedback! https://github.com/styled-components/styled-components/pull/1946/files

PR #1930, opened 2018-08-20
Keyframes refactoring

Fer0x:
Re-implement code from https://github.com/styled-components/styled-components/pull/1700 which ideas was taken from https://github.com/styled-components/styled-components/issues/1496

Based on current develop branch (I hope to ship this with 4.0 together with new global styling API)

Changelog:

 - keyframes`` helper  now returns Keyframes model
 - keyframes styles now inserting only with component rendering
 - To get animation name use `.getName()` method
 - some tests was changed to proper work with new `keyframes` helper

### DIFF
--- a/src/models/Keyframes.js
+++ b/src/models/Keyframes.js
@@ -1,0 +1,25 @@
+// @flow
+import StyleSheet from '../models/StyleSheet'
+
+export default class Keyframes {
+  id: string
+  name: string
+  rules: Array<string>
+
+  constructor(name: string, rules: Array<string>) {
+    this.name = name
+    this.rules = rules
+
+    this.id = `sc-keyframes-${name}`
+  }
+
+  inject = (styleSheet: StyleSheet) => {
+    if (!styleSheet.hasNameForId(this.id, this.name)) {
+      styleSheet.inject(this.id, this.rules, this.name)
+    }
+  }
+
+  getName() {
+    return this.name
+  }
+}
--- a/src/test/rehydration.test.js
+++ b/src/test/rehydration.test.js
@@ -383,20 +383,34 @@ describe('rehydration', () => {
     })
 
     it('should not regenerate keyframes', () => {
-      keyframes`
+      const fadeIn = keyframes`
         from { opacity: 0; }
       `
+
+      const A = styled.div`
+        animation: ${fadeIn} 1s both;
+      `
+      mount(<A />)
+
       expectCSSMatches(`
         @-webkit-keyframes keyframe_880 {from {opacity: 0;}}@keyframes keyframe_880 {from {opacity: 0;}}
+        .sc-a{ } .b{ -webkit-animation:keyframe_880 1s both; animation:keyframe_880 1s both; }
       `)
     })
 
     it('should still inject new keyframes', () => {
-      keyframes`
+      const fadeOut = keyframes`
         from { opacity: 1; }
       `
+
+      const A = styled.div`
+        animation: ${fadeOut} 1s both;
+      `
+      mount(<A />)
+
       expectCSSMatches(`
         @-webkit-keyframes keyframe_880 {from {opacity: 0;}}@keyframes keyframe_880 {from {opacity: 0;}}
+        .sc-a{ } .b{ -webkit-animation:keyframe_144 1s both; animation:keyframe_144 1s both; }
         @-webkit-keyframes keyframe_144 {from {opacity:1;}}@keyframes keyframe_144 {from {opacity:1;}}
       `)
     })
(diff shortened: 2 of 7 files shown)


Fer0x:
This PR has breaking changes with return value of `keyframes` helper. My previous thoughts was to add `.toString()` method in order to achieve backward compatibility.  But coercion may cause very unpredictable results, like using typeof or strict equal. So, here now method `.getName()` which can give actual name of animation. 
Eventually it's not a big change, I don't think too many people used this feature.

src/models/Keyframes.js
```diff
@@ -0,0 +1,25 @@
...
+    this.id = `sc-keyframes-${name}`
+  }
+
+  inject = (styleSheet: StyleSheet) => {
+    if (!styleSheet.hasNameForId(this.id, this.name)) {
+      styleSheet.inject(this.id, this.rules, this.name)
+    }
+  }
+
+  getName() {
```
kitten: Just an idea; if we implement “valueOf” and “toString” to inject the keyframes CSS, then the effects of this breaking change would be minimal :) cc @probablyup

Fer0x: I wrote about it https://github.com/styled-components/styled-components/pull/1930#issuecomment-414242846

And in previous PR you said that explicit coercion is better than implicit ;)

mxstbr: Honestly, I feel like it's fine for this to be a breaking change in v4—let's not do too much hacky stuff?

kitten: Sorry, just thinking about breaking changes again, since we already have quite a lot 😅 In this case, just refer to my old comment lol. Sorry about that.

@mxstbr @probablyup If we can get this change in, we can probably deprecate a couple of old uses of the StyleSheet and get rid of cloning soon, since we won’t have non-component usages anymore

src/utils/flatten.js
```diff
@@ -55,12 +57,23 @@ const flatten = (
...
       return ruleSet
     }
 
+    if (chunk instanceof Keyframes) {
+      if (styleSheet) {
+        chunk.inject(styleSheet)
+        return ruleSet.concat(chunk.getName())
+      }
+
+      return ruleSet.concat(chunk)
```
kitten: Looking at the function interpolation above, it seems that calling push and returning ruleSet is equally save and it has probably been changed by @probablyup as a micro optimisation, so let’s do the same here :)

src/test/rehydration.test.js
```diff
@@ -421,8 +435,33 @@ describe('rehydration', () => {
...
+      `
+      /* Purposely rendering out of order to make sure the output looks right */
+      mount(<B animation={fadeOut} />)
+      mount(<A animation={fadeIn} />)
+
+      expectCSSMatches(`
+        @-webkit-keyframes keyframe_880 {from {opacity: 0;}}@keyframes keyframe_880 {from {opacity: 0;}}
+        .sc-a { } .d { -webkit-animation:keyframe_880 1s both; animation:keyframe_880 1s both; }
         .sc-b { } .c { -webkit-animation:keyframe_144 1s both; animation:keyframe_144 1s both; }
+        @-webkit-keyframes keyframe_144 {from {opacity:1;}}@keyframes keyframe_144 {from {opacity:1;}}
```
probablyup: Is it a problem that this is injected after the usage?

Fer0x: @probablyup Look at example 3 https://www.w3.org/TR/css-animations-1/#at-ruledef-keyframes

It's ok to define keyframes elsewhere from it usage.

probablyup:
@Fer0x mind doing a quick followup PR to add a changelog entry for your efforts? :)

Fer0x:
@probablyup https://github.com/styled-components/styled-components/pull/1935
Done, but I'm afraid that my English can blow someone's mind 😃

Fer0x:
@probablyup @mxstbr @kitten Also, these changes need to be described on the website.
Does website repo have something like 4.0 branch?

stonebk:
I've been trying out the 4.0 beta, and it appears keyframes break when you try to use them in a nested template string:

```
const fadeIn = keyframes`
    0% {                                                                                                                                                                                                                                    
        opacity: 0;                                                                                                                                                                                                                         
    }                                                                                                                                                                                                                                       
    100% {                                                                                                                                                                                                                                  
        opacity: 1;                                                                                                                                                                                                                         
    } 
`;

const Foo = styled.div`
    ${props => props.enableAnimations ? `animation: ${fadeIn} 1s linear;` : ''}
`;
```

This ends up adding the animation: `animation: [object Object] 1s linear;`

Is this expected behavior?

I have worked around it by putting the animation in a separate css like so:

```
const fadeIn = keyframes`
    0% {                                                                                                                                                                                                                                    
        opacity: 0;                                                                                                                                                                                                                         
    }                                                                                                                                                                                                                                       
    100% {                                                                                                                                                                                                                                  
        opacity: 1;                                                                                                                                                                                                                         
    } 
`;

const fadeInAnimation = css`
    animation: ${fadeIn} 1s linear;
`;

const Foo = styled.div`
    ${props => props.enableAnimations ? fadeInAnimation : ''}
`;
```

probablyup:
Can you try using the css helper?

probablyup:
Oh never mind I see you did. Yes that’s correct. Interpolating without the
helper only works for simple values.

alansouzati:
is it a good idea to list this as a breaking changing here? https://www.styled-components.com/docs/faqs#what-do-i-need-to-do-to-migrate-to-v4

probablyup:
It was the case before I believe. Not a new change from v3

alansouzati:
I don't think so, I'm in the process of migrating and if I change nothing in my code and use styled-components v3 I don't get `animation: [object Object] 1s linear;` now I'm getting it. Let me try to put two codesandboxes together.

alansouzati:
Here is a working example using styled-components v3:

https://codesandbox.io/s/3qrz117n95

Here is the same example using styled-components v4 (does not animate):

https://codesandbox.io/s/nrq5owl63l

alansouzati:
I'm my case it is slightly tricker to move things around to make it work, since the animations are dynamic and I use props to generate the final animation string.

alansouzati:
for example I can pass an array of animation names to a component and it will generate multiple keyframes calls.

alansouzati:
Here is a more complex example on what I'm trying to achieve:

Working fine with v3: https://codesandbox.io/s/0o8k69vlow

Broken in v4: https://codesandbox.io/s/ll1o5xwqkl

probablyup:
Okay, I think this was working before on a technicality but wasn't explicitly documented so yeah I'll update the migration guide to call it out. In general we recommend always using the css helper when creating CSS partials with more than a simple value to be interpolated into a styled component.

probablyup:
@alansouzati and here's your v4 example tweaked with the css helper to get it working: https://codesandbox.io/s/74765zlom6

alansouzati:
How do you solve for this example? I’m failing to identify how to put this together for arrays

https://codesandbox.io/s/ll1o5xwqkl

I would appreciate if you fork this and try to make this work in v4

probablyup:
@alansouzati https://codesandbox.io/s/v84yyp97yy

probablyup:
@alansouzati ok got it https://codesandbox.io/s/v84yyp97yy

alansouzati:
yeah I thought about using reduce as well. the code readability is slightly reduced here. But I understand the refactor. thanks for chiming in so quick.

alansouzati:
hey I think something is up with the latest beta 5 v4 release. 

Take a look at this codesandbox (it does not fade if I put the opacity in a variable):

https://codesandbox.io/s/m4116pw31p

Using beta 1, this works fine, check this codesandbox:

https://codesandbox.io/s/jv3q8z9o65

Fer0x:
@alansouzati @probablyup 
This can be happened after https://github.com/styled-components/styled-components/pull/2006

probablyup:
Thanks taking a look

alansouzati:
thanks for chiming in @Fer0x 

on a side note, a work around was to move that interpolation to a separate variable, and use `css` from styled-components...

check this codesandbox:

https://codesandbox.io/s/o426lv949y

probablyup:
This should fix it: https://github.com/styled-components/styled-components/pull/2013

Sorry about that!

PR #1937, opened 2018-08-21
Alternate prop validator

quantizor:
It's lighter and faster since it's memoized. Saw some nice perf gain in benchmarking.

before:

---Mount deep tree---
17.14 ±01.82 ms
(S/L) 13.06/04.08 ms

---Mount wide tree---
22.89 ±02.67 ms
(S/L) 16.91/05.98 ms

---Update dynamic styles---
23.47 ±05.14 ms
(S/L) 07.98/15.49 ms

after:

---Mount deep tree---
16.07 ±01.72 ms
(S/L) 12.03/04.05 ms

---Mount wide tree---
21.45 ±02.74 ms
(S/L) 15.59/05.85 ms

---Update dynamic styles---
22.49 ±05.28 ms
(S/L) 07.50/14.99 ms

Need https://github.com/emotion-js/emotion/pull/814 & yarn update before merging

cc @Andarist 🚀


flow-typed/npm/@emotion/is-prop-valid_vx.x.x.js
```diff
@@ -0,0 +1,3 @@
+declare module '@emotion/is-prop-valid' {
```
Andarist: the package is written using flow & ships a siblings files to entry points: `dist/is-prop-valid.cjs.js.flow` & `is-prop-valid.esm.js.flow`

flow should pick those typings automatically 🤔 hasnt it?

probablyup: It was raising a module not found error... dunno

Andarist: Thanks for the info, need to investigate it later

src/utils/test/validAttr.test.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-import validAttr from '../validAttr'
+import validAttr from '@emotion/is-prop-valid'
 
 describe('validAttr', () => {
```
kitten: Let’s maybe remove these tests.

I assume that the package on its own is sufficiently tested

probablyup: I waffled back and forth on that. In case we switch again in the future, I was thinking we should keep these. They did reveal some missing props 😄

Andarist:
Im wondering. Do u consider this a breaking change or not? Such cases are always tricky in this regard

probablyup:
@Andarist nah, all tests are passing other than the PR against emotion so it's not breaking

src/utils/test/validAttr.test.js
Andarist: Actually our list might not be tested, gonna shamelessly copy ur tests later to cover that ;)

Andarist: Actually our list might not be tested, gonna shamelessly copy ur tests later to cover that ;)

flow-typed/npm/@emotion/is-prop-valid_vx.x.x.js
mitchellhamilton: You’re probably ignoring everything in node_modules. You might want to use the new [declarations option](https://flow.org/en/docs/config/declarations/) in flow which ignores errors in the specified folders but still uses the type declarations. (You might be on an older version of flow that doesn’t support it though so you might need to update flow)

mitchellhamilton:
Just published `@emotion/is-prop-valid@0.6.6` with the changes.

mxstbr:
Yay!

PR #1942, opened 2018-08-22
refactor away from enzyme

probablyup:
Using the React-provided test renderer and test utilities.

I couldn't figure out how to deal with the innerRef tests yet for native / primitives, but since that API is going away in one of my open PRs, I figured it wasn't worth the time to get working.

This unblocks #1923


alexandernanberg:
Really nice work! We could also look into [`react-testing-library`](https://github.com/kentcdodds/react-testing-library), but looking at the diff there might not be a need for it anymore.

PR #1962, opened 2018-08-31
implement implicit element folding, "as" polymorphic prop

probablyup:


src/models/StyledComponent.js
```diff
@@ -219,6 +219,8 @@ export default (ComponentStyle: Function) => {
     options: Object,
     rules: RuleSet
   ) {
+    const targetIsSC = isStyledComponent(target)
```
kitten: nit: Shall we improve the variable name here? e.g. `isTargetStyledComp` or sth?

src/models/StyledComponent.js
```diff
@@ -236,7 +238,27 @@ export default (ComponentStyle: Function) => {
         ? `${escape(options.displayName)}-${options.componentId}`
         : options.componentId || componentId
 
-    const componentStyle = new ComponentStyle(rules, attrs, styledComponentId)
+    let finalAttrs = attrs
+
+    // fold the underlying StyledComponent attrs up (implicit extend)
+    // $FlowFixMe
+    if (targetIsSC && target.attrs) finalAttrs = { ...target.attrs, ...attrs }
```
kitten: I'm starting to think it'd make sense to abstract `attrs` into a shareable set of utilities or sth... Maybe a future thing.

nit: Btw, I don't think having the `let`s here makes sense since a ternary would also do the trick and would imply that it won't change later in this fn

src/models/StyledComponent.js
```diff
@@ -236,7 +238,27 @@ export default (ComponentStyle: Function) => {
...
+
+    const componentStyle = new ComponentStyle(
+      targetIsSC
+        ? // fold the underlying StyledComponent rules up (implicit extend)
+          // $FlowFixMe
+          target.componentStyle.rules.concat(rules)
```
kitten: Should this be moved to the `ComponentStyle` constructor itself? i.e. 

```js
new ComponentStyle(rules, attrs, id, child)
```

where the child is another `ComponentStyle`; This would also move the `attrs` merging, I guess? Not quite sure why we don't move the bulk of `buildExecutionContext` there anyway.

src/models/StyledComponent.js
```diff
@@ -271,6 +293,7 @@ export default (ComponentStyle: Function) => {
 
       const newOptions = {
```
kitten: I think we might not need this anymore, i.e. we can avoid `options` altogether.

- `attrs` is called before the StyledComponent is created and never again
- `extend` will go
- `withComponent` will go

So any "cloning" is gone

src/models/StyledComponent.js
probablyup: Yeah makes sense

src/models/StyledComponent.js
probablyup: Could go either way. Currently ComponentStyle isn't concerned with folding which I think is probably a good thing to separate concerns?

src/models/StyledComponent.js
probablyup: Let's wait to refactor this out for v5 when withComponent is gone?

src/models/StyledComponent.js
kitten: Yea, I was thinking of a "combined" refactor of StyledComponent and ComponentStyle, but that can certainly wait :+1:

probablyup:
> In the current version color doesn't affect styles of StaticStyledLink, but it will in v4.

The main change in v4 is the styles of `StaticStyledLink` will be copied over to `DynamicStyledLink` and then that color style will be applied last.

probablyup:
It all happens in one go. It’s not slow, we test thoroughly.

Anber:
@probablyup what happens in one go? I don't mean an initial render, I speak about dynamic styles. If I later change `color` props, new implementation will rerender all styles, and not only styles of DynamicStyledLink how it is now. It's ok if you have just a few styles, but if StaticStyledLink has a complex stylesheet it will definitely affect performance.

mxstbr:
No, `StaticStyledLink` will not be recomputed I'm 99% sure.

Anber:
@mxstbr but all it styles will be recomputed. After changing just one prop you need to process all styles of DynamicStyledLink and StaticStyledLink, instead of processing just DynamicStyledLink styles.

mxstbr:
Yeah, `<DynamicStyledLink />` will recompute its styling that's true, but any instance of `<StaticStyledLink />` won't recompute its style. That's the same as before, any component that has a props interpolation will recompute its styling when re-rendered? I'm still not sure I understand.

Anber:
@mxstbr but now it's possible to reduce the number of styles which will be recomputed by splitting component into two (or more) components: one with static styles (and it will be never recomputed after initial render) and one with interpolated styles. v4, on the other hand, will optimize this two components by merging into one.
For example, `DynamicStyledLink` now is rendered like `<a class="StaticStyledLinks-zxcv DynamicStyledLink-asdf …"></a>` and only `DynamicStyledLink-asdf` will be recompute in case of props changing. In v4 it will be rendered as `<a class="DynamicStyledLink-asdf"></a>` and any props changing will trigger rerender of whole `DynamicStyledLink-asdf` which also contains styles from `StaticStyledLinks-zxcv`.

mxstbr:
@Anber now I understand what you're saying, and I'm happy to say you're way overthinking this: it doesn't matter.

Here's what conceptually happened before before:

```JS
// This CSS is static and will never be recomputed
const staticLinkStyles = css`
  color: blue;
  background: red;
`

// This CSS is recomputed every render
const dynamicLinkStyles = css`
  color: ${props => props.color};
`;
```

Here's what conceptually happens after:

```JS
// This CSS is static and will never be recomputed
const staticLinkStyles = css`
  color: blue;
  background: red;
`

// This CSS is recomputed every render
const dynamicLinkStyles = css`
  color: blue;
  background: red;
  color: ${props => props.color};
`;
```

Sure, a couple extra rules are re-parsed on every render but our parser is so fast it basically doesn't matter. If you're parsing anyway, parsing 5 extra lines is ~0 overhead.

Anber:
@mxstbr if the new parser in v4 is as fast as you say it will be great, but in v3 the trick with component splitting makes sense because render of a complex component with hundred lines of styles takes a noticeable period of time.

mxstbr:
@Anber do you have a benchmark showing the parser to be the bottleneck? I'd be very interested in seeing that. /cc @thysultan

Anber:
@mxstbr I've made a simple [example](https://github.com/Anber/styled-components-benchmark) for three cases: static, dynamic and combo dynamic + static. Despite the fact that dynamic + static 15% faster, in absolute numbers the difference is just 10-15 ms. Maybe there are cases where the difference will be bigger, but yes, usually it doesn't matter.

PR #1967, opened 2018-09-04
unskip & fix cGS tests

probablyup:


imbhargav5:
I think not using `name` while injecting rules will cause issues when we use interpolation within cGS. No?

mxstbr:
@imbhargav5 can you add a (failing?) test case for that to make sure we cover it?

probablyup:
@kitten @mxstbr @imbhargav5 added a warning if you compose & render the same cGS component multiple times. I think this should be good enough messaging to cut our first beta and see if it ends up being an actual issue.

PR #1968, opened 2018-09-04
should throw earlier if styleSheets.length = 0

sabrinaluo:
We just accidentally encountered a case that `styleSheets.length === 0` caused by some unknow reasons (still under investigation, but pretty sure it has nothing with styled-components). Thought it would be more reasonable to handle such case rather than letting it falls into the for loop section.

image


probablyup:
As far as I know, this only happens if multiple copies of styled components are loaded at once

sabrinaluo:
maybe throw a different error if size===0 ?

probablyup:
Beefed up the error in 3.4.8 to be more helpful and offer a potential fix

PR #1994, opened 2018-09-09
use PureComponent

probablyup:
With React 16 and all the APIs they implemented to ensure things
rerender properly deep in the tree, using PureComponent should be
safe now and it gives a pretty nice speed gain when updating
dynamic styles.

Before:

```
---Update dynamic styles---
22.14 ±05.96 ms
(S/L) 07.89/14.26 ms
```

After:

```
---Update dynamic styles---
19.25 ±06.16 ms
(S/L) 04.95/14.30 ms
```

_Updated, based on best of 3 runs for both cases._


mxstbr:
Can you publish a test version of this so I can deploy it to Spectrum and test it?

probablyup:
@mxstbr try `styled-components@experimental`

PR #2005, opened 2018-09-12
Add test for correct rendering order

Fer0x:
Rebased test from https://github.com/styled-components/styled-components/pull/1194
Now test passes which means bug with ordering no longer exists


probablyup:
Glad to hear it and thank you!

Issue #1969, opened 2018-09-04
v4 beta constructive feedback


probablyup:
Please provide as much detail as possible when commenting here so we can help! A huge amount of time and thought has been put into the API changes. We know it won't be a "flip a switch" situation, but the ergonomics of the library going forward should be markedly better and more flexible.

migueloller:
When doing server-side rendering, webpack `resolve.module` and `resolve.alias` don't work as expected. And while [these lines](https://github.com/styled-components/styled-components/blob/9d72abe32754ae9dd0bf9edf8a652ea592bdd06a/src/index.js#L42-L50) only check for a browser environment, there's some broken behavior in SSR.

Because of the new use of `React.createContext` it's imperative that the same instance of styled-components is used, otherwise theming will not work. We're having an issue where `resolve.module` and `resolve.alias` isn't helping (SSR) but we still have hoisted and non-hoisted styled-component packages in our monorepo, causing SSR to fail due to an empty theme.

I've spent a bit trying to come up with a solution for this, but so far I haven't been able to get around Node's resolution algorithm to get it to resolve to the right instance.

probablyup:
@migueloller Do you use yarn by any chance? They have a nifty package.json field called "resolutions" you can use to dedupe styled-components at the package manager level:

```json
{
  "resolutions": {
    "styled-components": "beta"
  }
}
```

migueloller:
We do! Unfortunately it’s not an issue of different versions. We have the beta version in the hoisted node_modules as well as the no-hoisted one. The issue is because of how the modules are resolved.

We have a component library that imports from the hoisted package and we have application code that imports from the no-housed package. The application also imports from the component library. When webpack works as expected (i.e., front end build), it will properly use `resolve.alias` to always pick the no-hoisted dependency. Unfortunately, for SSR that’s not the case, causing the instance of styles-components to be different.

kitten:
@migueloller since this is a dedup issue that's more related to tooling, there's not much we can do 😅

There's an alternative solution here: https://github.com/styled-components/styled-components/issues/1941#issuecomment-417862021 

But in your case it sounds like you might want to use babel-plugin-module-alias or module-alias for node

diegohaz:
Thank you and congratulations for the awesome work, guys. 

I have an issue here.

I was trying to test the behavior of the `as` prop (I wanted to know whether it conflicts with existing `as` props in custom components), but it's throwing this error:

```
ModuleNotFoundError
[...]

PR #2007, opened 2018-09-13
fix cGS + HMR

probablyup:
This was a really weird one. I still don't know why this works,
but it does and it's the only impl that did.

Fixes the issue noted here: https://github.com/styled-components/styled-components/issues/1969#issuecomment-421142583


probablyup:
That’s probably a good call. The count is only relevant for the browser
anyway

PR #2015, opened 2018-09-17
Noop on injectGlobal deprecation warning by default

### LINKED ISSUE #2014: v3.4.7 - warnInjectGlobalDeprecated is not a function
## Environment
```
 - styled-components: ^3.4.4 => 3.4.7
```

## Reproduction
In 3.4.7 a deprecation warning for `injectGlobal` was introduced that throws a `TypeError` in production mode. Running a build in with `NODE_ENV=production` that uses `injectGlobal` causes the build to fail. 

This is caused by #2000 here: 
https://github.com/styled-components/styled-components/pull/2000/files#diff-095484d1d0538c1c11e49a50c9475a20R33

Since `warnInjectGlobalDeprecated` has no value in production mode, calling it causes the build to fail. Either it should check that it has a value, check the env. or noop. 

## Expected Behavior
No warning is thrown when `injectGlobal` gets used with `NODE_ENV=production`

## Actual Behavior
Build fails with 

```js
TypeError: warnInjectGlobalDeprecated is not a function
```

BenMagyar:
This resolves #2014 where the deprecation warning for `injectGlobal` fails production builds. Followed the noop pattern from [warnExtendDeprecated](https://github.com/styled-components/styled-components/blob/104fc5565d5a9b6a02c50ceebab60257691b7b1b/src/models/StyledNativeComponent.js#L21).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 ## unreleased
 
+- Fix an issue where the `injectGlobal` deprecation warning errors for production, by [@benmagyar](https://github.com/BenMagyar) (see [#2015](https://github.com/styled-components/styled-components/pull/2015))
+
 ## [3.4.7] - 2018-09-17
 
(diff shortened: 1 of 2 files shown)


BenMagyar:
Closing the PR since v3.4.8 fixes this!

PR #1954, opened 2018-08-27
add Themed* interfaces for react-native

elnygren:
This addresses https://github.com/styled-components/styled-components-website/issues/216

**Explanation**

Basically, https://www.styled-components.com/docs/api#define-a-theme-interface doesn't work for React Native. The reason is that `native/index.d.ts` is missing interfaces such as `ThemedStyledInterface`, `ThemedReactNativeStyledFunction` and `ThemedStyledComponentsModule`.

**Defining a theme after this PR**

https://www.styled-components.com/docs/api#define-a-theme-interface

```
// styled-components.ts
import * as styledComponents from 'styled-components/native'
import { ThemedStyledComponentsModule } from 'styled-components/native'

import ThemeInterface from './theme';

const {
  default: styled,
  css,
  ThemeProvider
} = styledComponents as ThemedStyledComponentsModule<ThemeInterface>;

export { css, ThemeProvider };
export default styled;
```

Issue #1952, opened 2018-08-27
TypeScript typings: Node, StyleSheet are not found


elnygren:
Where are these coming from? https://github.com/styled-components/styled-components/blob/fcf6f3804c57a14dd7984dfab7bc06ee2edca044/typings/styled-components.d.ts#L187-L190

tsc fails with output:

```
node_modules/styled-components/typings/styled-components.d.ts:188:11 - error TS2304: Cannot find name 'StyleSheet'.

188   sheet?: StyleSheet
              ~~~~~~~~~~

node_modules/styled-components/typings/styled-components.d.ts:189:12 - error TS2304: Cannot find name 'Node'.

189   target?: Node
               ~~~~

node_modules/styled-components/typings/styled-components.d.ts:205:13 - error TS2304: Cannot find name 'StyleSheet'.

205   instance: StyleSheet
                ~~~~~~~~~~
```

elnygren:
fix?

```
patch-package
--- a/node_modules/styled-components/typings/styled-components.d.ts
+++ b/node_modules/styled-components/typings/styled-components.d.ts
@@ -184,9 +184,11 @@ interface StylesheetComponentProps {
   sheet: ServerStyleSheet
 }

+export class StyleSheet {}
+
 interface StyleSheetManagerProps {
   sheet?: StyleSheet
-  target?: Node
+  target?: React.ReactNode
 }

 export class StyleSheetManager extends React.Component<
```

probablyup:
Hey folks, TS typings for styled-components have been moved to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped), so please feel free to move your issue there where you'll receive better community support.

PR #2019, opened 2018-09-19
fix: Define StyleSheet and change Node to React.ReactNode

nicogreenarry:
This is parallel to [a PR in the DefinitelyTyped repo](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/29022). From that PR:
> As referenced [here](https://github.com/styled-components/styled-components/issues/1952), the `StyleSheet` and `Node` types were undefined. This PR defines `StyleSheet`, and replaces `Node` with `React.ReactNode`, following the example found [here](https://github.com/styled-components/styled-components/pull/1954).

TODO:
- [ ] Make sure that the tests still pass: `npm test` and `npm run flow` (for the type checks)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "styled-components",
-  "version": "3.4.9",
+  "version": "3.4.10",
   "description": "Visual primitives for the component age. Use the best bits of ES6 and CSS to style your apps without stress 💅",
   "typings": "typings/styled-components.d.ts",
   "main": "dist/styled-components.cjs.js",
--- a/typings/styled-components.d.ts
+++ b/typings/styled-components.d.ts
@@ -184,9 +184,11 @@ interface StylesheetComponentProps {
   sheet: ServerStyleSheet
 }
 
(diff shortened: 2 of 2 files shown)


probablyup:
I mean I'll take your PR, but the official typings are on DT now so this may never get shipped

probablyup:
@nicogreenarry yup we removed the typings in v4. It's set to go out on October 1 and the beta is available now: https://medium.com/styled-components/announcing-styled-components-v4-better-faster-stronger-3fe1aba1a112

nicogreenarry:
OK cool. Your call whether it makes more sense to just close this PR, then. I'll try using the beta version in my project and see if that works. Thanks!

nicogreenarry:
I'm going to close this - per [this comment in the DT PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/29022#pullrequestreview-156940662), my change isn't correct anyway. I'm going to keep trying to solve the fix on the DT end.

PR #2021, opened 2018-09-20
inline static hoisting

quantizor:
Ironically I think this shrinks the bundle size slightly. This PR inlines a version of `hoist-non-react-statics`, with some optimizations since we don't have to worry about React 15

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,6 @@
   "dependencies": {
     "@emotion/is-prop-valid": "^0.6.5",
     "css-to-react-native": "^2.0.3",
-    "hoist-non-react-statics": "^3.0.0",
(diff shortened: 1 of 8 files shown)


probablyup:
```
npm install styled-components@4.0.0-beta.7-hoist
```

for anyone who wants to test

mridgway:
Was there any motivation to do this other than package size? While inlining it does make your library smaller when bundled, in most applications it will actually make the bundle larger since most applications will use `hoist-non-react-statics` anyway.

probablyup:
We’re inclined it because of a bug

probablyup:
Inlined*

PR #2029, opened 2018-09-24
fix createGlobalStyles + keyframes

### LINKED ISSUE #2027: Can't use keyframes in createGlobalStyle
Version: 4.0.0-beta.8

I ran into this problem when attempting to add an animation to a body background, so I created [this codesandbox reproduction](https://codesandbox.io/s/lzrxkq9kz).

Basically, when `keyframes` is used within a `createGlobalStyle` block, the value of the `animation` property is `[object Object]`

image

I searched around and couldn't find any mentions or reference to this issue, so I apologize if it's been brought up and you're aware of it.

probablyup:
~~I'm not sure if this is the right solution. Ideally we wouldn't run the `css` helper until render time but since we're hashing the completed rules I'm not sure if that can be changed.~~ It wasn't. But the PR has been updated to the correct solution.

Fixes #2027

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -13,6 +13,7 @@
 .*/dist/.*
 .*/lib/.*
 .*/src/test/.*
+.*/src/vendor/.*
 .*/src/.*/test/.*
 .*/example/.*
 .*/integration-test/.*
--- a/src/constructors/createGlobalStyle.js
+++ b/src/constructors/createGlobalStyle.js
@@ -6,6 +6,7 @@ import StyleSheet from '../models/StyleSheet';
 import { StyleSheetConsumer } from '../models/StyleSheetManager';
 import determineTheme from '../utils/determineTheme';
 import { ThemeConsumer, type Theme } from '../models/ThemeProvider';
+// $FlowFixMe
 import hashStr from '../vendor/glamor/hash';
 import css from './css';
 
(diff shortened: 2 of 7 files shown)


src/utils/flatten.js
```diff
@@ -28,7 +29,11 @@ export const objToCss = (obj: Object, prevKey?: string): string => {
  */
 const isFalsish = chunk => chunk === undefined || chunk === null || chunk === false || chunk === '';
 
-export default function flatten(chunk: any, executionContext: ?Object, styleSheet: ?Object): any {
+export default function flatten(
+  chunk: any,
+  executionContext: ?Object,
+  styleSheet: ?Object = StyleSheet.master
```
probablyup: This is the primary change of the PR, assigning default back to `StyleSheet.master`

probablyup:
I’m going to delete it. Drives me friggin crazy. Separate PR though

imbhargav5:
Looks good to me. Perhaps we can add one test case where animation is added within interpolation just to be doubly sure? 
```
${props => props.animate ?  `animation: ${keyframeAnimation1} ease 1s` :  `` }
```

Fer0x:
Offtopic: should we add some warning when someone tries to use keyframes within react-native?

mxstbr:
@Fer0x yeah that seems smart! People have been confused about that. Wanna PR it?

probablyup:
@imbhargav5 that example actually wouldn't work because of the lack of `css` helper haha.

imbhargav5:
Alright then, I think we can merge this if all is good?

PR #2031, opened 2018-09-26
Add warning when using `keyframes` on React Native

Fer0x:
Helper `keyframes` cannot be used on React Native, so we should add a corresponding warning.

https://github.com/styled-components/styled-components/pull/2029#issuecomment-423969660

### DIFF
--- a/src/constructors/keyframes.js
+++ b/src/constructors/keyframes.js
@@ -20,7 +20,9 @@ export default function keyframes(
     typeof navigator !== 'undefined' &&
     navigator.product === 'ReactNative'
   ) {
-    console.warn('Helper `keyframes` cannot be used on React Native');
+    console.warn(
+      '`keyframes` cannot be used on ReactNative, only on the web. To do animation in ReactNative please use Animated.'
+    );
   }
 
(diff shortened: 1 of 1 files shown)


src/constructors/keyframes.js
```diff
@@ -14,6 +14,15 @@ export default function keyframes(
   strings: Styles,
   ...interpolations: Array<Interpolation>
 ): Keyframes {
+  /* Warning if you've used keyframes on React Native */
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    typeof navigator !== 'undefined' &&
+    navigator.product === 'ReactNative'
+  ) {
+    console.warn('Helper `keyframes` cannot be used on React Native');
```
mxstbr: I'd change the wording slightly to something like "`keyframes` cannot be used on ReactNative, only on the web. To do animation in ReactNative please use Animated."

Fer0x:
@mxstbr During production building with tools like webpack, all lines with condition `process.env.NODE_ENV !== 'production'` will be dead code and stripped out from bundle.

PR #2062, opened 2018-10-04
add protocol to open collective link

iamstarkov:
without it some terminals dont understand that it is a link.
When they do it is very convenient to just click on it see what's there.
Perhaps it can drive up the conversion of donations (I hope it does at least a little bit)


iamstarkov:
I think "master" shouldn't be a default branch if de facto it is a "develop" branch

probablyup:
Develop goes away as soon as we launch v4

probablyup:
Thank you!

PR #2071, opened 2018-10-07
fix self-referential behavior in combo selectors

probablyup:
When a component refers to itself with the ampersand placeholder the behavior today is a bit inconsistent and won't work as intended if your component has dynamic styles.

For example:

```jsx
const Comp = styled.div`
  color: ${p => p.color};

  & + & {
    margin-left: 10px;
  }
`;

<div>
  <Comp color="red">Hello</Comp>
  <Comp color="blue">World</Comp>
</div>

// renders
// <div>
//   <div class="sc-a a">Hello</div>
//   <div class="sc-a b">Hello</div>
// </div>
```

The second instance of `Comp` will not get the `margin-left` because a different className is generated due to the different color prop, and today ampersand refers to the current styling context  className.

It effectively looks like this:

```css
// Comp -> .sc-a
// color: red -> .a
// color: blue -> .b

.sc-a {}
.a { color: red }
.a + .a { margin-left: 10px }
.b { color: blue }
.b + .b { margin-left: 10px }
```

Note how the combo is `.a + .a` and `.b + .b`, so only two adjacent components with the exact same styling props would trigger a match. This is a surprising behavior and makes compositions more troublesome.

The solution is to treat the second `&` as a reference back to the component itself (`.sc-a`), not the style of the component.

Then the styles look like this:

```css
// Comp -> .sc-a
// color: red -> .a
// color: blue -> .b

.sc-a {}
.a { color: red }
.a + .sc-a { margin-left: 10px }
.b { color: blue }
.b + .sc-a { margin-left: 10px }
```

And any `Comp` next to a `Comp` will receive the `margin-left`. Note that the reason the first ampersand isn't also switched to the static component class is that will break style isolation.

[before](https://codesandbox.io/s/xpk25v94rz):

& + & -> .b + .b
&[disabled] { & + & } -> .b[disabled] + .b[disabled]

[after](https://codesandbox.io/s/xomz3rj41q):

& + & -> .b + .sc-a
&[disabled] { & + & } -> .b[disabled] + .sc-a[disabled]

fixes #784

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ script:
   - yarn test
   - yarn flow
   - yarn lint
+  - yarn size
 notifications:
   email:
     on_failure: change
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 ## Unreleased
 
+- Fix how ampersand is handled in self-referential selector combinations, e.g. `& + &` (see [#2071](https://github.com/styled-components/styled-components/pull/2071))
+
 ## [v4.0.0-beta.10] - 2018-10-04
 
 - Add support for `as` to be used with `attrs` for better polymorphism, by [@imbhargav5](https://github.com/imbhargav5) (see [#2055](https://github.com/styled-components/styled-components/pull/2055))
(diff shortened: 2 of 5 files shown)


probablyup:
`styled-components@4.0.0-beta.10-2` if anyone wants to help test. Can also hack on the "after" sandbox in OP

imbhargav5:
Perhaps we can add a test for a combination of `&` & `+`  or something just to be sure.

imbhargav5:
@probablyup Just did

probablyup:
@imbhargav5 done

imbhargav5:
Awesome.

probablyup:
@kitten any issues or tests I should add to the suite?

kitten:
@probablyup looks solid to me; the only thing we could do differently is to actually add it as a stylis middleware, to its selector input transform function.

PR #2072, opened 2018-10-08
inline stylis compressed via gcc advanced mode

probablyup:
drops ~0.25kb gzipped

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ script:
   - yarn test
   - yarn flow
   - yarn lint
+  - yarn size
 notifications:
   email:
     on_failure: change
--- a/package.json
+++ b/package.json
@@ -67,7 +67,6 @@
     "memoize-one": "^4.0.0",
     "prop-types": "^15.5.4",
     "react-is": "^16.3.1",
-    "stylis": "^3.5.0",
     "stylis-rule-sheet": "^0.0.10"
   },
   "devDependencies": {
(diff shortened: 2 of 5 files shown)


kitten:
I honestly don't think this is worth it 😅 Maybe you should contribute this compression script to stylis instead. Otherwise we'd need to vendor stylis often and frequently on any patch it receives

probablyup:
> Otherwise we'd need to vendor stylis often and frequently on any patch it receives

That's true, but it has barely changed in the recent months. I was also thinking since it's a critical piece of our infra it's probably a good idea to keep it locked down anyway such as to avoid a bad release mucking things up.

kitten:
Oh, it hasn't changed much? I guess that's to be expected. In that case, if we don't expect it to be patched frequently now, I'm fine with merging this :+1:

kitten:
> inline stylis compressed via gcc advanced mode 

Can we add a script to the repo and put stylis into devDependencies to make future upgrades safely reproducible though? I'm fine if that's done separately or not at all for now, but eventually it should be obvious what we're doing to it and make it fool-proof

probablyup:
@kitten sure. I also noticed Mitchell was doing something similar for emotion: https://github.com/emotion-js/emotion/blob/6cb8d15438b01b8623af42f304d5ea3032332187/packages/stylis/README.md

probablyup:
After further thought I agree with you and will just see if I can contribute it to stylis

PR #2084, opened 2018-10-09
fix createGlobalStyle HMR issue

probablyup:
I don't love this solution but it does work.

### DIFF
--- a/src/constructors/createGlobalStyle.js
+++ b/src/constructors/createGlobalStyle.js
@@ -12,6 +12,11 @@ import css from './css';
 
 import type { Interpolation } from '../types';
 
+// place our cache into shared context so it'll persist between HMRs
+if (IS_BROWSER) {
+  window.scCGSHMRCache = {};
+}
+
 export default function createGlobalStyle(
   strings: Array<string>,
   ...interpolations: Array<Interpolation>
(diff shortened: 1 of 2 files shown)


probablyup:
Oh yeah probably

probablyup:
Actually I'm not sure. There are cases where this could pop up in production like when hot rendering page transitions in next

probablyup:
It's out...

mhaagens:
Looks like `<GlobalStyles />` needs to be wrapped in a `styled.X` for HMR to work.
In a top level app component this doesn't work;
```
  render() {
    return (
      <>
        <h1>Hello World</h1>
        <GlobalStyle />
      </>
    );
  }
```
but this does;
```
  render() {
    return (
      <Container>
        <h1>Hello World</h1>
        <GlobalStyle />
      </Container>
    );
  }
```
So you need to wrap a styled component (or maybe just use somewhere, haven't checked) around the `<GlobalStyles />` in the top level component to get HMR working. Maybe its a React Hot Loader issue, not sure, but that solved it at least.

probablyup:
Uhh, that should definitely not be the case.

mhaagens:
I can share the repo with you if you want to take a look.
Definitely what happens in my app, but I might be doing something stupid heh.

mhaagens:
@probablyup Just wondering if you had a chance to take a look?

probablyup:
No sorry, I'm very swamped at work so I don't expect having time to look at this until the weekend

mhaagens:
Gotcha, no worries! Thanks for helping me out 🙏

PR #2085, opened 2018-10-09
Add 'allowUserMedia' and 'allowPaymentRequest' as valid attributes

asoltys:
These are superseded by the allow attribute but should still be valid as 
legacy attributes on an iframe tag

ref: https://wicg.github.io/feature-policy/#legacy-attributes


probablyup:
@asoltys looks like you need to rebase

asoltys:
@probablyup Oops sorry, should be ok now I think, I rebased master and force pushed

asoltys:
On it! https://github.com/emotion-js/emotion/pull/903

Thanks for the version bump too

PR #2088, opened 2018-10-10
Setup for sourceMap support for dev environment

yiochen:
> This merge request is not ready

I created this merge request to get feedback on my approach. Let me know if I missed anything.

As discussed in [this issue](https://github.com/styled-components/babel-plugin-styled-components/issues/14), the source map support will **only** be available for dev environment for two reasons:

1. CSS sourcemap will greatly increase the bundle size of the app
2. SpeedyTag.insertRule doesn't insert comment, sourceMap needs to live inside a comment

Changes need to be made to both the styled-components repo and the babel-plugin-styled-components repo

1. We need to render the styles into style tag
2. Each style tag can only have one sourceMap, because styles are dynamically generated and added to the style tag, we need to make sure one style tag contains one rule.
3. The only source mapping we can do is to map the begining of a rule to the beginng of the styled-component code. We can't do mapping for each css property.

In this repo (with this merge request), I added support to pass sourceMap into styled component with `withConfig`

```
styled.div.withConfig({
    sourceMap: `/*# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbIk9yaWdpbmFsLmpzIl0sIm5hbWVzIjpbXSwibWFwcGluZ3MiOiJBQUNBIiwiZmlsZSI6ImRpc3QuanMiLCJzb3VyY2VzQ29udGVudCI6WyIvLyBUaGlzIGxpbmUgd2lsbCBiZSBza2lwcGVkXG5JIHdpbGwgc2V0IGNzcyBjbGFzcyBoZXJlXG5JIHdpbGwgc2V0IGNvbG9yIGhlcmVcbiJdfQ== */`
})`
    color: red
`
```

This should generate a style tag that contains rules **only** for this component
```
<style>
/* sc-component-id: sc-a*/
.sc-a {
  color: red;
} /*# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VzIjpbIk9yaWdpbmFsLmpzIl0sIm5hbWVzIjpbXSwibWFwcGluZ3MiOiJBQUNBIiwiZmlsZSI6ImRpc3QuanMiLCJzb3VyY2VzQ29udGVudCI6WyIvLyBUaGlzIGxpbmUgd2lsbCBiZSBza2lwcGVkXG5JIHdpbGwgc2V0IGNzcyBjbGFzcyBoZXJlXG5JIHdpbGwgc2V0IGNvbG9yIGhlcmVcbiJdfQ== */
</style>
```

The second part of the work would be done in babel plugin repo to generate those sourceMap comment.

Because this uses `withConfig`, looks like `injectGlobal` and `keyframe` don't support it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 ## unreleased
 
+## [3.4.9] - 2018-09-18
+
+- Remove the `injectGlobal` warning; it's not actionable since the replacement API is in v4 only, so why say anything?
+
 ## [3.4.8] - 2018-09-17
 
 - Fix the `injectGlobal` warning not being properly guarded for production, by [@probablyup](https://github.com/probablyup)
@@ -566,7 +570,9 @@ _v3.3.1 was skipped due to a bad deploy._
 
 - Fixed compatibility with other react-broadcast-based systems (like `react-router` v4)
 
-[unreleased]: https://github.com/styled-components/styled-components/compare/v3.4.7...master
+[unreleased]: https://github.com/styled-components/styled-components/compare/v3.4.9...master
+[v3.4.9]: https://github.com/styled-components/styled-components/compare/v3.4.8...v3.4.9
+[v3.4.8]: https://github.com/styled-components/styled-components/compare/v3.4.7...v3.4.8
 [v3.4.7]: https://github.com/styled-components/styled-components/compare/v3.4.6...v3.4.7
 [v3.4.6]: https://github.com/styled-components/styled-components/compare/v3.4.5...v3.4.6
 [v3.4.5]: https://github.com/styled-components/styled-components/compare/v3.4.4...v3.4.5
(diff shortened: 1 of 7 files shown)


yiochen:
Looks like the bundle size limit failed the build. Can we increase the limit?

probablyup:
If it fails it’s because code is getting through to production that
shouldn’t be

yiochen:
@probablyup not sure if I follow what you said. In the package.json, we have a `test:size` script that checks the bundle size. I checked before my changes, the bundle size was 11.48kb.
```
./dist/styled-components.cjs.min.js: 11.48KB < maxSize 11.5KB (gzip)
```
My changes increased the size by 0.02kb, which breaks the travis build.
The size limit 11.5kb is in package.json. I wonder if I should increase the size limit for ./dist/styled-components.cjs.min.js to make my travis build pass.

probablyup:
Oh did you make this change against master? You’ll need to rebase in the
morning when we release v4

yiochen:
ah, ok, I will rebase tomorrow and see if it passes CI

PR #2093, opened 2018-10-12
Transient props

### LINKED ISSUE #439: Separate HTML attributes from styling props
## Background
I am trying to style a 3rd party "masked" input. The styles are applied correctly, however, there are props I pass into the "styled" element that are used within the CSS logic of the styled component that I don't want passed down to the 3rd party input. This logs an error saying there are undesired props being put on the input (this is also an issue with the 3rd party component that isn't properly stripping out what should go on the input).

I could see it being an enhancement where we specify what props to terminate at the "styled" part. However, this also could just be looked at as the 3rd party component having an issue not removing those props. I'm fine with it being an issue with the 3rd party developer but just wanted to mention the issue and see if this is happening to other people and if it is worth it to build some extra logic in.

#### Screenshot
screen shot 2017-01-31 at 11 31 51 am

### LINKED ISSUE #2129: Feature Request: Option to pass "as" prop down to wrapped component when using styled(Comp)
First off, thanks again for your hard work on sc4! I'm really enjoying the new features it introduced, the `as` prop in particular.  I got really excited when I read about it first, as we had just introduced a `Base` component to our UI library that would enable the same functionality for all components that used it. We even chose the same prop name.

But as we know the devil's in the details, and so I noticed during migration to v4 that there's an important difference between letting a `Base` component handle the `as` prop and using s-c's own prop: The first styled component to receive the `as` prop uses it directly, and there's no way other than using a different prop name to prevent this. This is not a problem when all components wrapped with `styled()` are styled components themselves, but when they're not (and quite a few of our UI elements aren't), it leads to veeeery unexpected behaviour.

It's basically the same issue that the ReaKit people are facing: https://github.com/reakit/reakit/issues/226

Now, our codebase is already using the `as` prop and we'd strongly prefer not to change it. Not just because it'd be a huge amount of work. It would also be a shame since there should be a consistent API for changing the rendered element/component of both "plain" sc components and our own custom ones. I WANT to use s-c's `as` prop. But due to the issue mentioned above, achieving this is very complicated at the moment.

Consider this innocent piece of code for example:
```jsx
const CustomNavBarTab = styled(NavBarTab)`
  /* ... styles ... */
`;

render(
  <CustomNavBarTab as="a" ref={this.refHandler} ... >
);
```
where NavBarTab is NOT a primitive styled-component. It does take the `as` prop (and ref) and forwards it to a nested styled-component. This used to work with the `Base` component, but breaks with sc's 'as' prop, since `as="a"` effectively changes the above component definition to `const CustomNavBarTab = styled('a')`, disregarding all of `NavBarTab`'s styles & behaviour.

To work around this fully while trying to keep the outer component API the same, something crazy like the below would have to be written:

```jsx
const StyledCustomNavBarTab = styled(React.forwardRef(({innerAs, ...props}, ref) =>
  <NavBarTab {...props} as={innerAs} ref={ref} />
))`
  /* ... styles ... */
`;

const CustomNavBarTab = React.forwardRef(({as, ...otherProps}, ref) => {
  return (
    <StyledCustomNavBarTab
      ref={ref}
      innerAs={as}
      {...otherProps}
    />
  );
});
```

You'll probably agree that this really shouldn't be necessary just to keep refs and the "as" prop working in a consistent way.

As you know from my persistent lobbying for an sc-specific API for filtering props near the component definition, **the goal of all of this is to keep styled-components a "transparent" layer**. I don't want to require component users to know whether a component is a primitive styled-component or not.

I WISH I could just write the component somewhat like this, indicating to styled-components that the `as` prop should be passed down to the wrapped component:
```jsx
const CustomLinkTab = styled(NavBarTab, {ignoreAs: true})`
  /* ... styles ... */
`;
```

I don't care much about the specifics of the API as long as it doesn't require piling on wrapper components just for passing on or tranforming or filtering props.

Is this something you'd consider adding to styled-components? I imagine it would be fairly easy to implement, but the hard part would be settling on an API for it.

### LINKED ISSUE #2131: v4 passes through unknown HTML attribute to the DOM
## Environment
npm packages:
1. react: v16.5.2
2. react-dom: v16.5.2
3. styled-components: v4.0.2

## Reproduction
https://codesandbox.io/s/zn5vrk0vpm

## Expected Behavior
https://codesandbox.io/s/20xlp242pr
There is no unknown-prop warning if I'm using `styled(styled.button).attrs({ customProp })` with styled-components v3.4.10.

## Actual Behavior
There is unknown-prop warning with styled-components v4.

If this is not a bug, and there is something missing. Please let me know, thanks!

probablyup:
This commit introduces a new (currently styled-components-specific) concept called "transient props". When a prop is prefixed with a dollar sign (e.g. `$foo={false}`), it will only be applied to the immediate component and not forwarded to any deeper children or layers of the component stack.

In the case of styled-components, this equates to a styling-only prop that you may consume in your style interpolations but will never end up on the DOM node or in your wrapped component (other props can still be accessed normally). For example:

```JS
const Input = styled.input`
  color: ${props => props.$color};
  background: ${props => props.disabled ? '#DDD' : '#FFF'};
`

<Input $color="#333" /> // => <input />, $color not passed through
<Input $color="#333" disabled /> // => <input disabled />, disabled passed through
```

This would allow us to get rid of the prop-whitelist in the next major version (v5?), which would considerably drop the bundle size of the package.

Fixes #2131 
Fixes #2129 
Fixes #439


probablyup:
This could also potentially land in v4.x but without the existing prop validation being removed. If it catches on we can deprecate prop validation and remove it in v5.

cloud-walker:
It is a JSX concept coming from react? Or just a styled-components JSX concept?

Kovensky:
Just a styled-component's thing; the _only_ props that have any special meaning in JSX are `key` and `ref`.

chengjianhua:
Can I make a request? Why not `_` or `__`? The style of using sign `$` feel likes not be consistent with preferences of most of us (use camelCase almost, a little underscore_case...) 🙈.  Let me remember jQuery or Angular or Vue...

diondiondion:
Is the `as` prop going to become `$as`?

And is there still a chance you'll _also_ implement a static "prop-filtering at component definition time" solution similar to [what's described here](https://github.com/styled-components/styled-components/pull/1682#issuecomment-382900468), for people like me who really don't want to expose styled-component-specific props in their UI components?

(For example like this:
```
styled(Comp).filterProps(['customProp1', 'customProp2'])`
  /* styles */
`;
```
)

On a side-note, the `as` prop could also do with a way of being forwarded to the underlying component instead of applying directly to the one it's attached to.

probablyup:
There are a lot of linter rules around use of _ and personally I just
prefer how $ looks. It’s a bit easier to write imo.

kitten:
:100: Still a big fan of this :heart:

probablyup:
> Is the as prop going to become $as?

We can definitely look into that. That particular change would have to be in v5 if we do land it.

mxstbr:
Just realised a downside to this approach: it makes the abstraction leaky by making styled-components very different from normal React component. That in turn makes it hard to evolve that component without breaking changes to any user of the component.

To elaborate, imagine a Button component with a primary property. At first, your implementation would probably be something like this:

```JS
export const Button = styled.button``;
```

So throughout the whole application you use the Button like so:

```JSX
<Button $primary disabled />
```

Now imagine you get some new requirements and your Button needs some custom behaviour, so you change the implementation to something like this:

```JS
const StyledButton = styled.button``;

export const Button = (props) => (
  <StyledButton>{/* ... */}</StyledButton>
)
```

Now you're stuck in a bad place since you have only got these two options:

1. Continue naming the prop `$primary`, pretending as though the component was a styled one. Unfortunately, it's not, so people who think it is will have a bad time, e.g. when trying to use it in a component selector `${Button}`
2. Rename the prop to `primary` and have to comb through the entire application and however many hundreds of usages to change the prop name

That sucks. :confused:

---

I'm leaning towards the "style-prop" approach (a single blacklisted prop that's always filtered and everything else is passed through), specifically a prop that doesn't have any s-c specific connotation (I like `variant`). It keeps the abstraction intact and makes for much easier evolution of existing components.

probablyup:
> by making styled-components very different from normal React component

It's a choice to be made by the developer. If they need multiple layers of components to receive the prop, they should not mark it as transient and just filter it out at the appropriate level so it doesn't hit the DOM.

mxstbr:
> It's a choice to be made by the developer. If they need multiple layers of components to receive the prop, they should not mark it as transient and just filter it out at the appropriate level so it doesn't hit the DOM.

A big part of what makes styled-components so nice in my mind is that it's "Just React Components™️". I would not want to require users to choose upfront whether they'll ever have more complex requirements—those are always changing, so I think they'd end up always having to wrap each styled component they export, which is not great.

> nesting a prop especially for s-c isn't all that different from giving it a prefix.

That would be the case if the prop was called "sc" or something like that, but not if giving it a generic name like `variant="primary"`. By doing that you don't end up in the bad place I described above, as whether or not the component is a styled component or not is an implementation detail. (every component can have a `variant` prop)

diondiondion:
> That would be the case if the prop was called "sc" or something like that, but not if giving it a generic name like `variant="primary"`.

Sure, every component _can_ have a `variant` prop, but the truth is that normally every component _wouldn't_ have one – unless you made it a convention across your codebase to move _all_ custom component props into a `variant` prop. Since that's not very likely, the variant prop would still feel arbitrary and come with the same confusing downsides as a prefix, just in a more neutral guise. In practice I'd still have to know or think about whether a prop for a given component is attached directly or as part of the variant object.

So I'd still be one of those who'd
> end up always having to wrap each styled component they export, which is not great.

probablyup:
Alright I think we've reached an internal agreement on how to proceed for v5 and are comfortable landing the `$prop` filtering change in v4 without removing existing filtering, right @mxstbr?

christianalfoni:
Just wanted to chime in. I think @mxstbr is correct leaning towards a single prop for custom properties, if I interpret this correctly. The reason is that there are actually two problems:

1. Creating a prop that happens to be a valid HTML attribute, but it was not intended as such. Happens to me all the time. Call me a bad programmer, but things like `width`, `height` etc. easily ends up in the DOM

2. Creating a prop that is not a valid HTML attribute, where React will yell at you

The `$` suggestion solves problem 2 at the cost of adding special conventions on props and issues refactoring, as pointed out above. But it does not solve problem 1.

I think the only way to solve both issues is to pass in custom props explicitly to the component. `variant` does not really make sense to me, as the property is specifically there for styled-components. So I would name it what it is: `styleProps`, the props used to style the component.

```js
<MySpecialDiv styleProps={{
  width: 100,
  height: 100
}} />
``` 

What makes this a good solution is that the component, it being an element or actual component, works exactly the same. Styled components does not interfer with this in any way. You explicitly state what to be available in the styling composition via the `styleProps` property:

```js
export const MySpecialDiv = styled.div`
  width: ${props => props.width + "px"};
  height: ${props => props.height + "px"};
`
```

If you for some reason need the the same property passed as attribute to element or to a component it can be duplicated:

```js
<MyCanvas width={width} height={height} styleProps={{
  width,
  height
}} />
``` 

This also explicitly states that you are both adding width/height to the DOM and you use the values in the styling of the canvas.

Fer0x:
I appreciate this solution like compromise to having whitelist of attributes which is not grouped by html elements. 

However I feel this solution like proposal of private fields in js. It seems too controversial for end users.
For example, if you have already prepared props with html attributes you can’t just use ```<MyStyled {...props} />```. Each time you need to transform props to prefix them with `$` sign. Another example - when you try to pass some props to component - you don't know - is it styled component or just regular react component and you don't have propTypes which can help you to choose right prop. For example, let's imagine that we have a styled input which takes `type` prop to customize input's color and $type to pass it as html attribute:
```jsx
<StyledInput type="primary" $type="password" />
```
It looks very confusing.

I think this feature should be implemented like this
```jsx
<StyledInput type="primary" foo="bar" htmlAttrs={{ type: 'password', 'maxLength': 10 }} />
```
It allows to use the same names of props like html attributes and allows to use spread with attributes.

christianalfoni:
@Fer0x Yeah, that is like flipping it around. Instead of passing explicit style props, you pass explicit HTML attributes. What is very good about this idea as well is that the props passed to style composition is still just props:

```js
export const StyledInput = styled.input`
  color: ${({ type, foo, htmlAttrs }) => {}};
`
```

To me this makes perfect sense 👍

Fer0x:
And for static components it can be used like:

```jsx
const Input = styled.input.attrs({
	htmlAttrs: {
		type: 'password'
	}
})``;
```

Or we can made a new constructor:

```jsx
const Input = styled.input.htmlAttrs({
	type: 'password'
})``;
```

Fer0x:
@styled-components/core what can you say about my examples?

probablyup:
It doesn’t work because if you’re wrapping a custom component, the
underlying component has to understand what htmlAttrs is. If you were to
just spread apply it’s contents against the other top level props, there
could be collisions.

Fer0x:
@probablyup
> It doesn’t work because if you’re wrapping a custom component, the underlying component has to understand what htmlAttrs is. If you were to just spread apply it’s contents against the other top level props, there could be collisions.

Ok, but custom components also needed in knowing how to handle props with $. If I use third party component, like react-input and want to customized it with styled() and pass html prop options. Which variant should I use? Maybe it will be ‘options’. But then, imagine that I have some task to render a table and for each row I need to render input or select depending on some logic. And input is standard styled component but select is custom select mentioned above. I need to pass $value to first but value to second? It’s a very confusing, because on the top both components are styled components, but with very different behavior. I mean a lot of custom components are used the same names of props as html attributes. Mix of “$prop” and “prop” in code can make usage of styled-components very painful.

probablyup:
Actually they don’t. The transient props proposal means $components are not
forwarded below the styles component HOC layer.

christianalfoni:
Though in what way are the `.attrs` affected by this? Sorry, I do not know the internals. But as I see it the `htmlAttrs` property would act as extending the element with `attrs`?

jgoux:
Hello all 👋 ,

maybe we could rely on the existing `attrs` and change its `Factory` behaviour?

Instead of : 
> A function that receives the props that are passed into the component and computes a value, that is then going to be merged into the existing component props.

Why not have : 
> A function that receives the props that are passed into the component and returns new props.

Then we could use it like this : 
```js
const MyComponent = styled.div.attrs(props => {
  const { propToFilter, anotherPropToFilter, propToTransform, ...rest } = props

  // Here we don't include the props to filter and we apply a prop transformation
  const newProps = {
    propToTransform: propToTransform * 2,
    ...props
  }

  return newProps
})`
  ${({ propToFilter, anotherPropsToFilter, transformedProp, ...rest }) => /* We have access to { ...props, ...newProps } ! */}
`
```

So the requirement here is to pass two different sets of props : 
- The result of `attrs` (newProps) to the wrapped component, so we don't have the filtered ones
- The result of merging `props` and `newProps` (`{ ...props, ...newProps }`) to the styled HOC layer, so we have access to everything (filtered, transformed and initial props!)

What do you think? 💡

probablyup:
If that did happen it’d have to be a major release, but I think it would be
confusing for most people. The current behavior is more like dynamic
defaultProps and migrating to something like you described isn’t
codemoddable and would be prone to issues when component folding occurs
like styled(StyledComponent) since the behavior is to merge the attrs.

jgoux:
Hey @probablyup , thanks for the feedback!

> If that did happen it’d have to be a major release, but I think it would be confusing for most people. 

On the contrary, I think it's less confusing than introducing a new naming convention which increase the API surface of the library. Also `attrs` is the current layer where we can touch props between the component's props and the styled HOC, so I thought it was cool to be able to reuse it.

I agree that using `attrs` for this is maybe a big step. We could add an extra argument + a deprecation notice. This would solve the codemod issue I guess : 
```js
// .attrs = (props, merge = true) => {}
const MyComponent = styled.div.attrs((props, false) => {
  const { propToFilter, anotherPropToFilter, propToTransform, ...rest } = props

  // Here we don't include the props to filter and we apply a prop transformation
  const newProps = {
    transformedProp: propToTransform * 2,
    ...rest
  }

  return newProps
})`
  ${({ propToFilter, anotherPropsToFilter, transformedProp, ...rest }) => /* We have access to { ...props, ...newProps } ! */}
`
```

It's not pretty but it does the job until everyone catch up.


Another solution would be to create a new ".attrs" like function with the proposed behaviour (no idea about the name). But then we would again increase the API surface...


> The current behavior is more like dynamic defaultProps

Yes, and I'd like to generalize this behavior from prop initialization to prop transformations. Do you see any case where my proposal wouldn't work for defaultProps?

> component folding occurs like styled(StyledComponent) since the behavior is to merge the attrs

I'm not sure I understand here, if `StyledComponent` props are already filtered, what prevent you from merging with the external `styled()` ?

probablyup:
Unfortunately this doesn't solve the central issue of multiple levels of components consuming the same sets of props... it only supports 2 layers, where in practice there can be more than that. Going to close for the time being and seek further alternatives. For the record, no other proposal has addressed the 3+ layer issue either, other than JSX namespacing which I think the most correct solution.

geelen:
Hey so, I've been pretty quiet on this for a while, mainly because I really want JSX namespace props and thought that if I looked away for long enough they'd be implemented by the time I look again, alas!

I think putting this in terms of "transient" props is actually a really good one–all we're then doing is defining a convention for a prop that can _only_ be passed a single level deep. Using the `$` character is really the only option given that `#` is a syntax error. But this has a couple of nice properties:

* **It's extremely close to the existing implementation.**<br/>This is important because as far as I can see it, dropping the whitelist is going to cause a boatload of runtime warnings, with no good codemod possible. So the migration path is going to be awkward for everyone, let's make it as less awkward as possible.
* **It's dumb as a plank**.<br/>_"Props starting with $ don't get passed through. All other ones do"_. As long as _all_ props are available inside interpolations, it's an easy decision to make.
* **It can be customisable**.<br/> @diondiondion is right that "developers HAVE to manage (i.e. decide whether to pass on or not) their component's props at some point", but for the base case where SC is wrapping a `div` or `button` , adding that info on every _definition_ just makes things clumsy. When you're building up more complex component APIs on the other hand, you might wanna change things:

```js
styled.button` ... ` // equivalent to:
styled.button.rejectProps(
  propName => propName .startsWith('$')
)` ... ` // default

// works great for <button large type="submit">

styled(Comp).rejectProps(
  propName => new Set(['dont','pass','these']).has(propName)
)` ... ` // overridden

// works better for <Component dont pass these/>
```

Plus we could also do some things like allowing anything specified in `attrs` to be added _after_ this has been run, so you could do:

```js
styled.div.attrs({
  $someProp: props => props.$someProp
})` ...` 
```

This would be pretty easily wrapped up in a helper:

```js
styled.div.attrs({
  ...passProp('$someProp', '$someOtherProp')
})` ...` 
```

Now for some more controversial ideas:

* **We should alias `props.x` to match the prop named `$x`**<br/>In other words, `${props => props.foo}` matches both `<Comp foo/>` and `<Comp $foo/>`. This would dramatically simplify the migration path since its only callsites that need updating, not components. This better fits the concept of a "transient prop" rather than an actual SC-specific API. It just means "don't pass this on" rather than actually being a new name.
* **Alternatively, let's do more babel**<br/>We could technically add JSX namespaces to our babel plugin, I believe. Maybe `$propName` is the no-babel version and `style:prop` or some other syntax is the "preferred" one. I hate when other people introduce non-standard syntax to something they don't own so yeah this is probably a bad idea.
* **Maybe $ is our new thing?**<br/>Last year I played around with a much more Sass-y version of SC syntax but I couldn't figure out the parser grammar required to actually implement it:

```js
styled.div`
  color: $color || black;
  $bold? {
    font-weight: 700;
  }
  opacity: $variant = 'subtle' ? 0.8 : 1.0;
`

// compiles to
styled.div`
  color: ${ props => props.$color || props.theme.color || 'black' };
  ${ props => props.$bold || props.theme.bold && css`
    font-weight: 700;
  ` }
  opacity: ${ props => (props.$variant || props.theme.variant) === 'subtle' ? 0.8 : 1.0 };
`
```

Bit wild I know, and totally just an initial thought about a syntax, but `$` would work really well there (since it's the start of the interpolation syntax anyway), so maybe using `$` just for transient props isn't the worst thing to be starting with...

Anyway, just a few thoughts. 👋 everyone!

satya164:
> We should alias props.x to match the prop named $x

Worth noting that it won't be possible to statically type this sort of thing.

> We could technically add JSX namespaces to our babel plugin

Babel plugins can't add new syntax afaik. You'll need to fork Babylon to be able to add a non standard syntax.

diondiondion:
Great, thanks for chiming in @geelen. Still don't like the $ sign based default filtering, but as long as there's customisation via the `rejectProps` function, I'm all for it. 👍

The only case this probably doesn't solve is the `as` prop, it would be great to be able to tell a styled component to ignore & pass it on to the wrapped component. (See  #2129 for details)

Any idea how that could be taken into account?

hallister:
I'm in agreement with @diondiondion's previous comment. I guess I'm lost on why it's clumsy to handle it at the component definition level. Is there *really* a situation where you might want to pass a prop sometimes but other times not? That sounds like an implementation issue.

But as long as customization exists via `rejectProps` (although, I'd argue `filterProps` or `transientProps` makes more semantic sense, as you're not rejecting them, you're just not passing them to the next level) I'll also jump on board. Hacky workarounds that either break syntax highlighting or throw errors for invalid DOM attributes are worse than anything suggested here.

chengjianhua:
Recently I just found that using `data-xxx` as props name solved all my problem. It can pass down any numbers of layers and didn't cause React to emit warnings about non-DOM properties passed to dom elements.

Everything is ok, no extra props, no additional layers, no extra mechanism, leveraging the native attributes of HTML spec. Only one caveat that names are more verbose.

dan-dr:
@chengjianhua it may be valid but they bloat your html

joemaffei:
@dan-dr I really like your solution. It's worth noting that lodash's omit method is fairly heavy (I measured it at 5.98kB today, Bundlephobia says it's 2.5kB). You might want to try a more lightweight implementation: https://gist.github.com/joemaffei/39976001f70d57b43ade0482994f2db2

Jakobo:
If you don't want to monkey patch `styled`, you can also create a HOC for consuming props. This just takes @joemaffei and @dan-dr's great work and moves it to a regular 'ole HOC.

```
const without = (C, omitProps = []) => {
  const omitSingle = (object = {}, key) => {
    if (key === null || key === undefined || !(key in object)) return object;
    const { [key]: deleted, ...otherKeys } = object;
    return otherKeys;
  };

  const omit = (object = {}, keys) => {
    if (!keys) return object;
    if (Array.isArray(keys)) {
      // calling omitMultiple here would result in a second array check
      return keys.reduce((result, key) => {
        if (key in result) {
          return omitSingle(result, key);
        }
        return result;
      }, object);
    }
    return omitSingle(object, keys);
  };

  const WithoutPropsComponent = ({ children, ...props }) => {
    if (typeof C === "string") {
      return (
        <React.Fragment>
          {React.createElement(C, omit(props, omitProps), children)}
        </React.Fragment>
      );
    } else {
      return <C {...omit(props, omitProps)} children={children} />;
    }
  };

  return WithoutPropsComponent;
};
```

Used like this
```
// foo is consumed, and not passed on to MyMisbehavingThing
const MyStyledThing = styled(without(MyMisbehavingThing, 'foo'))`
  color: ${props => (props.foo === "bar" ? "#f00" : "#00f")};
`;
```

[Here's a codesandbox](https://codesandbox.io/s/l50mlwqo1q) that shows the HOC consuming & discarding the specified props.

pumanitro:
For a temporary solution we can do it as follows: 
```
const SomeComponent = styled(({propIdontWant1, propIdontWant2, ...restProps} => <ComponentIWantToUse {...restProps}/>)`
 background-color: ${props => props.propIdontWant1}
`;
```

brandonkal:
Why is this even something Styled Components or any library  should be concerned with? This would be much better for React to handle. Pass everything down and have it filtered at the end. React wouldn't even need to have an arbitrary whitelist of valid DOM attributes. As a developer, you would prefix or namespace attributes you want to reach the DOM:

`<Button primary dom:disabled />`
OR
`<Button primary $disabled />`

This is extremely easy to lint.

Using a dollar prefix as a signal that "this is not a DOM attribute" works well. But it is awkward if it becomes a "this property is implemented as a SC/styletron/etc specific attribute". I'd much rather put a $ sign in front of all props that are not HTML attributes than have it be specific to any library. But at that point, it makes sense to go one step further and only namespace DOM attributes.

mxstbr:
@brandonkai we submitted a RFC to that effect in May of 2017, but it does not look like it will be implemented anytime soon. (https://github.com/facebook/jsx/issues/81)

samtimalsina:
> For a temporary solution we can do it as follows:
> 
> ```
> const SomeComponent = styled(({propIdontWant1, propIdontWant2, ...restProps} => <ComponentIWantToUseOrHTMLElement {...restProps}/>)`
>  background-color: ${props => props.propIdontWant1}
> `;
> ```

This temporary solution worked for us too. There's a small typo here. Better written as:

```javascript
const SomeComponent = styled(({propIdontWant1, propIdontWant2, ...restProps}) => <ComponentIWantToUseOrHTMLElement {...restProps}/>)`
 background-color: ${props => props.propIdontWant1}
`;
```

tommedema:
Has anyone gotten these workarounds to work in a type safe manner with Typescript?

```javascript
// props is `any`
const MyLabel = styled(({display, ...rest}) => <label {...rest}/>)`
  display: ${props => props.display ? 'inline' : 'none'}
`

// Type '{ display: boolean; }' does not satisfy the constraint...
// And: props: never
const MyLabel2 = styled<{ display: boolean }>(({display, ...rest}) => <label {...rest}/>)`
  display: ${props => props.display ? 'inline' : 'none'}
`
```

3sanket3:
@tommedema It should be as below.

```
import React, { LabelHTMLAttributes } from "react";

const MyLabel2 = styled(
  ({
    display,
    ...rest
  }: LabelHTMLAttributes<HTMLLabelElement> & { display: boolean }) => (
    <label {...rest} />
  )
)`
  display: ${props => props.display ? 'inline' : 'none'}
`;
```

I lost my css intellisense though. Folks coding JS, is it the same for you or it's issue with TS only?

mertindervish:
@3sanket3 

For IntelliSense, you can assign the "Transient" component into a variable as below:

```js
import React, { LabelHTMLAttributes } from 'react'

const TransientMyLabel2 = ({
  display,
  ...rest
}: LabelHTMLAttributes<HTMLLabelElement> & { display: boolean }) => (
  <label {...rest} />
)

const MyLabel2 = styled(TransientMyLabel2)`
  display: ${props => (props.display ? 'inline' : 'none')};
`
```

sami616:
Adding `$` to props you don't want passed down seems like a really neat solution to me. Would be very handy for them to be aliased without the `$` when interpolating like @geelen suggested. 😄

dimitropoulos:
couple things I want to add to the conversation:

## [**Styletron has already been doing this for a long time and it works well!**](https://www.styletron.org/react/#props-filtering)

The approach they [use](https://www.styletron.org/react/#props-filtering) is:

> All props starting with $ are filtered out and not passed to the underlying element.

I used styletron in quite extensively and I can say from experience that the approach taken in styletron (to use `$`) was both easy/obvious to work with for very junior and very senior developers alike.  In fact the team never had any problems around this the way I have been having problems with (on a different project) the gotchas that come with styled-component's current implementation.

## Maintaining a whitelist causes problems

It doesn't matter if styled-components maintains that list themselves, or depends on a 3rd party to maintain it.. you're always going to get problems due to ["hey add my library's props too"-ism](https://github.com/styled-components/styled-components/pull/1823), as well as the [ever-evolving HTML attributes spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes) adding common properties like `importance` and `loading` that are likely to be already in use in projects (and thus be a breaking change due to change in functionality when these new attributes are introduced).  Let me say that again: since the browser can change at any time, we can expect that breaking changes will occur such that the same project (if using a props whitelist) will have different functionality even though they're using the same version of the library in question.  See the thread below with Dan Abramov's tweet for ideas on how to prevent this.

#### Current approach
While styled-components _used_ to maintain a [hardcoded regex](https://github.com/styled-components/styled-components/blob/dbf280c48da7e7cddc53695677898f266a4bee6f/src/utils/validAttr.js) with all whitelisted properties, as of @probablyup's [PR from August 2018](https://github.com/styled-components/styled-components/pull/1937) styled-components moved to use [`@emotion/is-prop-valid`](https://www.npmjs.com/package/@emotion/is-prop-valid).  There is a somewhat [systemic problem](https://github.com/emotion-js/emotion/issues/1362) with this, which I will describe below.

#### "Dan says so"
[Dan Abramov warns](https://twitter.com/dan_abramov/status/822526388557250560) not to use whitelists (styled-components is using a whitelist).

> Please don’t use prop whitelists in components! This defeats the whole purpose of the DOM warning you were seeing.

This appeal to authority logical fallacy is brought to you by Dan's good reputation.

### My use case

Recently I was upgrading an older project that used an older version of styled-components (before the move to `@emotion/is-prop-valid`).  _After_ styled-components moved to `@emotion/is-prop-valid` a [PR was merged that added `loading` to the list](https://github.com/emotion-js/emotion/issues/1362) of things passed through in the whitelist.  This was the right thing to do technically speaking, since the DOM spec _is what it is_ and we should support it.  That said, as the commentators in that thread noticed (one of which is @probablyup, the author of this PR) thatL

> this is kinda a breaking change (like any change to this whitelist, but the odds of breaking somebody's code in this case is quite huge in case of the "loading" prop)
> - https://github.com/emotion-js/emotion/issues/1362#issuecomment-495873283, @Andarist 

Indeed, this sent me on a wild hunt to figure out why I kept getting this error because [I was ignorant to this newly-added experimental HTML attribute](https://github.com/whatwg/html/pull/3752).  I had assumed (because I had seen `loading` used before with no problems in exactly the same context) that such a thing wouldn't present a problem.  What's worse is that the debugging for something like this is extremely tedious:

67782849-bfd80600-fa3f-11e9-89a5-084abb19a240

```
Warning: Received `false` for a non-boolean attribute `loading`.

If you want to write it to the DOM, pass a string instead: loading="false" or loading={value.toString()}.

If you used to conditionally omit it with loading={condition && value}, pass loading={condition ? value : undefined} instead.
    in div (created by Context.Consumer)
    in StyledComponent (created by styled.div)
    in styled.div
    in Unknown (created by Poll)
    in div (created by Context.Consumer)
    in StyledComponent (created by styled.div)
    in styled.div (created by Poll)
    in Poll
    in Unknown
    in div (created by Context.Consumer)
    in StyledComponent (created by styled.div)
    in styled.div
    in Unknown (created by Context.Consumer)
    in Route (created by AppRouter)
    in Switch (created by AppRouter)
    in Router (created by HashRouter)
    in HashRouter (created by AppRouter)
    in AppRouter
    in Provider
    in ThemeProvider
```

In true form, there were _many_ places in the app that used `loading`, which made this quite a bear to track down.

To compound the frustration, [`styled-component`'s documentation on this doesn't help my case](https://www.styled-components.com/docs/faqs#why-am-i-getting-html-attribute-warnings) because I was composing another component (i.e. `styled(MyComponent)`) where I wasn't able to pull out the props related to the error.

Once I figured out what was going on, which took me quite some time because I had to understand how the whitelist worked (there's no mention of whitelist in the styled-components codebase since it's just a dependency now and `is-valid-prop` isn't something obvious if you don't already know about it as we might expect someone encountering this problem might), I was able to fix it by changing `loading` to `isLoading`.

-----

I wanted to highlight my use case because I think it's a good example of why using a whitelist can lead to:
- constantly breaking changes (in a library outside of styled-components, no less)
- extraordinarily tricky to debug for the average user of styled-components
- non-obvious based on the same code working well in the past between upgrades (from what I can tell this wasn't mentioned in any styled-components release notes).

vjsingh:
@probablyup Any updates on this? Why is this closed?

probablyup:
@vjsingh It was added in 5.1

vjsingh:
@probablyup cool - thanks!

PR #2105, opened 2018-10-16
Add error to keyframes when interpolated into raw string

mxstbr:
This will now throw an error, prompting the user to use the css\`\`
helper as is required in v4 (but wasn't in v3):

```
const rotate = keyframes``

const styles = `
  animation-name: ${rotate};
`

```

### DIFF
--- a/src/utils/errors.md
+++ b/src/utils/errors.md
@@ -63,4 +63,4 @@ _This error was replaced with a dev-time warning, it will be deleted for v4 fina
 
 ## 12
 
-It seems you are interpolating a keyframe declaration (%s) into a vanilla string. This was supported in styled-components v3, but is not longer supported in v4. Please wrap your string in the css\`\` helper (see https://www.styled-components.com/docs/api#css).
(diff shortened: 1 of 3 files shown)


src/utils/errors.md
```diff
@@ -60,3 +60,7 @@ Cannot find a StyleSheet instance. Usually this happens if there are multiple co
 ## 11
 
 _This error was replaced with a dev-time warning, it will be deleted for v4 final._ [createGlobalStyle] received children which will not be rendered. Please use the component without passing children elements.
+
+## 12
+
+It seems you are interpolating a keyframe declaration (%s) into a vanilla string. This was supported in styled-components v3, but is not longer supported in v4. Please wrap your string in the css\`\` helper (see https://www.styled-components.com/docs/api#css).
```
probablyup: ~into a vanilla string~ into an untagged string template.

probablyup: Let's also just add a quick sentence on why:

The `css` helper is needed here because it is what causes the animation code to be injected into the stylesheet.

PR #2107, opened 2018-10-16
Add suppressMultiMountWarnings prop to disable warning on multiple cgs mount

### LINKED ISSUE #2082: Add prop to createGlobalStyle created-component to allow multiple mounts / disable multi-mount warning
The new function createGlobalStyle is awesome, and I recognize that mounting/unmounting/mounting multiple times *could* be in issue for some people (hence why I understand the warning in development), but I would like to propose a way to suppress the warning on a per-component basis.

A simple example use case where it *doesn't matter* if we mount/unmount/mount is when we are specifically adjusting **font-face** declarations (which need to be at the outer scope of a stylesheet).

The public api could be something simple like (disregard lack of proper sanitization, this is just an example):

```
  const GlobalFontStyler = createGlobalStyle`
    ${({fontFaceRelatedStuff})=> fontFaceRelatedStuff}
  `


<GlobalFontStyler 
  suppressMultiMountWarnings={true}
  fontFaceRelatedStuff={this.props.fontFaceRelatedStuff}
/>

```

## npmPackages:
 - babel-plugin-styled-components: ^1.3.0 => 1.3.0 
 - styled-components: 4.0.0-beta.10 => 4.0.0-beta.10 

## Steps to reproduce

1. create a component with createGlobalStyle
2. mount, unmount, mount again

## Actual Behavior

createGlobalStyle behaves properly, but warns in development regarding mounting multiple times, which is unnecessary/not-beneficial in some cases.

imbhargav5:
This PR fixes #2082 by adding a warning to disable warning when `suppressMultiMountWarnings` prop is added. @probablyup Adding styles for all the instances of the component seems to be a bit hard to do right now. Is the warning enough or should we make the components add multiple styles for the CGS component?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 ## Unreleased
 
+- Add suppressMultiMountWarnings prop to disable warning on multiple cgs mount, by [@imbhargav5](https://github.com/imbhargav5) (see [#2107](https://github.com/styled-components/styled-components/pull/2107))
+
 ## [v4.0.0] - 2018-10-15
 
 This is a rollup of the highlights of beta 0-11 for convenience. See the [migration guide](https://www.styled-components.com/docs/faqs#what-do-i-need-to-do-to-migrate-to-v4) for easy updating steps and the [beta announcement blog](https://medium.com/styled-components/announcing-styled-components-v4-better-faster-stronger-3fe1aba1a112) for our summary of v4's changes, thought process, etc.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 ## Unreleased
 
-- Add suppressMultiMountWarnings prop to disable warning on multiple cgs mount, by [@imbhargav5](https://github.com/imbhargav5) (see [#2107](https://github.com/styled-components/styled-components/pull/2107))
+- Add suppressMultiMountWarning prop to disable warning on multiple cgs mount, by [@imbhargav5](https://github.com/imbhargav5) (see [#2107](https://github.com/styled-components/styled-components/pull/2107))
 
 ## [v4.0.0] - 2018-10-15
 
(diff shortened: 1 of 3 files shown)


src/constructors/createGlobalStyle.js
```diff
@@ -59,7 +59,8 @@ export default function createGlobalStyle(
       if (
         process.env.NODE_ENV !== 'production' &&
         IS_BROWSER &&
-        window.scCGSHMRCache[this.state.styledComponentId] > 1
+        window.scCGSHMRCache[this.state.styledComponentId] > 1 &&
+        !this.props.suppressMultiMountWarnings
```
probablyup: Let's make this "suppressMultiMountWarning" instead. And add a `propType` and `defaultProps` definition for it.

imbhargav5: Done.

imbhargav5:
I think, given that this will be an uncommon scenario and like you said, an escape hatch, adding it as `props` should be good enough. Leaving it unchanged. Let me know if you would like to put it on the class.

probablyup:
Maybe add a changelog before merging

imbhargav5:
Ah! That's right.

imbhargav5:
Alright then. Just added changelog. Please approve it again and I will merge.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this
 
 ## Unreleased
 
+- Add suppressMultiMountWarnings prop to disable warning on multiple cgs mount, by [@imbhargav5](https://github.com/imbhargav5) (see [#2107](https://github.com/styled-components/styled-components/pull/2107))
```
probablyup: It’s without the s 😆

imbhargav5: LOL

wmertens:
I'm not happy with this API. It causes a long string to be embedded in the runtime JS, creating objects with it etc, even though it is to quench a dev-time warning.

How about instead adding an option to the cgs creation, or maybe a a construct like

```js
const Foo = createGlobalStyle`/* stuff */`
if (!__PROD__) Foo.suppressMMWarn = true
```

? That way it can be removed from production builds and there is no need to provide the prop on every instance.

PR #2108, opened 2018-10-16
Failing test for ampersand selector

quentez:
Since beta 11 of Styled Components `4.0.0`, we've been experiencing a regression where some selectors that use the ampersand operator will not be properly converted.

Input:

```
.foo > & {}
```

Expected output:

```
.foo > .x {}
```

Actual output:

```
.x .foo > .x {}
```

The culprit appears to be this regular expression: https://github.com/styled-components/styled-components/blob/12cde61a9bb6dbbd167cbb2b3244c1354b32d566/src/utils/stringifyRules.js#L7

Unfortunately, I don't think I have enough context to fix this without breaking other things that depend on it.

Issue #2095, opened 2018-10-13
V4 media queries prioritized higher than extended styles


ksabry:
In v4, `MoreStyles` in this code will have a background-color of red or green, instead of the expected blue

```javascript
const Styles = styled.div`
  @media (min-width: 1000px) {
    background-color: red;
  }
  @media (max-width: 1000px) {
    background-color: green;
  }
`;

const MoreStyles = styled(Styles)`
  background-color: blue;
`;

```

V4 version with unexpected behavior: https://codesandbox.io/s/vq6k94rpz0
V3 version with expected behavior: https://codesandbox.io/s/0z48zxvn
(exactly the same code)

From what I can tell this is due to v4 adding styles to an existing class (of lower priority than the media query) rather than to a new class like v3 does:

**V3**:
![V3](https://i.imgur.com/gkLjxEM.png)

**V4**:
![V4](https://i.imgur.com/vkfyVFY.png)

Only work around I've found is something clunky like this

```javascript
const MoreStyles = styled(Styles)`
  @media screen {
    background-color: blue;
  }
`;
```

If this is the intended behavior what is the correct way to achieve this functionality? Also if it is intended it should be probably be mentioned in https://www.styled-components.com/docs/faqs#what-do-i-need-to-do-to-migrate-to-v4

probablyup:
The way v4 currently works is styles are folded and flattened together in the case of nested `styled(StyledComponent)`, so the background blue is essentially getting grouped with the other static rules (height and width) and the media queries are appended after since they aren't part of the same style context.

Your v3 example works because before the components were not folded and each had its own generated style class. The blue background class comes after the first class due to runtime insertion order and then due to how CSS places higher precedence on equal-weight styles lower in the sheet, the blue is applied.

In terms of fixes, there are three I can think of:

1) Use a precedence boost:

  ```
  && {
    background-color: blue;
  }
  ```

  This seems like the better solution to me because it makes it more obvious that you intend for this particular style to be applied over other siblings.

2) Wrap it in a media query like you did; this forces the style to be broken off the main clump since media queries can not be nested inside a class definition and thus are inserted after the main set, becoming implicitly higher precedence

3) Use something like `!important`

ksabry:
Thanks for the response, the precedence boost seems like the best solution for what I want.

Just to be clear, I take it that the current behavior is intended and won't be changed back, as doing so would probably require some messy logic around conditionally folding each attribute. If this is the case feel free to close the issue

[...]

PR #2109, opened 2018-10-17
reimplement self-reference replacement as a stylis plugin

probablyup:
this fixes the failing test case and another edge case

Fixes #2108 
Fixes the second issue noted toward the end of #2095

### DIFF
--- a/src/test/styles.test.js
+++ b/src/test/styles.test.js
@@ -55,6 +55,10 @@ describe('with styles', () => {
         & + & {
           margin-bottom: 4px;
         }
+
+        & > & {
+          margin-top: 4px;
+        }
       }
 
       & + & {
@@ -93,6 +97,7 @@ describe('with styles', () => {
       .b.b.b{ border:1px solid red; }
       .b[disabled]{ color:red; }
       .b[disabled] + .sc-a[disabled]{ margin-bottom:4px; }
+      .b[disabled] > .sc-a[disabled]{ margin-top:4px; }
       .b + .sc-a{ margin-left:4px; }
       .b + .sc-a ~ .sc-a{ background:black; }
       .b ~ .sc-a{ margin-right:4px; }
@@ -103,6 +108,7 @@ describe('with styles', () => {
       .c.c.c{ border:1px solid red; }
       .c[disabled]{ color:red; }
       .c[disabled] + .sc-a[disabled]{ margin-bottom:4px; }
+      .c[disabled] > .sc-a[disabled]{ margin-top:4px; }
       .c + .sc-a{ margin-left:4px; }
       .c + .sc-a ~ .sc-a{ background:black; }
       .c ~ .sc-a{ margin-right:4px; }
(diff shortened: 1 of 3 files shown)


codeaid:
Just adding a link to the test case I created to demonstrate the issue in #2095:
https://stackblitz.com/edit/react-9lm5xr

src/test/styles.test.js
```diff
@@ -55,6 +55,10 @@ describe('with styles', () => {
         & + & {
           margin-bottom: 4px;
         }
+
+        & > & {
```
kitten: After looking at the tests, do we have snapshots for these three combinators in particular?

probablyup: Other than the ones in that group?

kitten: sorry, my bad. I meant stylis level tests but I think we can live without them

probablyup:
@codeaid fixed https://stackblitz.com/edit/react-ktxdtq

you can install via `styled-components@dev` for the moment until we cut a patch release in the next day or two

codeaid:
> @codeaid fixed https://stackblitz.com/edit/react-ktxdtq

Legend! Just tested it in StackBlitz and it's indeed fixed. That's some quick turnaround!

src/utils/stringifyRules.js
```diff
@@ -43,25 +43,63 @@ const parseRulesPlugin = _insertRulePlugin(rule => {
...
+ *
+ * & + & {
+ *   color: red;
+ * }
+ *
+ * The second ampersand should be a reference to the static component class. stylis
+ * has no knowledge of static class so we have to intelligently replace the base selector.
+ */
+const selfReferenceReplacementPlugin = (context, _, selectors) => {
+  if (context === 2 && selectors[0].lastIndexOf(_selector) > 0) {
```
toolmantim: We've had to revert to 4.0.0 in buildkite/site#234 because this line seems to have broken vendor prefix of webfonts? The arguments when `selectors[0]` was undefined is:

```
{ context: 2,
  selectors: [],
  _:
   'font-family:BK-Circular;src:url(/_next/static/assets/assets/fonts/lineto-circular-book-0415b07a.woff2) format("woff2");font-weight:normal;font-style:normal;font-display:swap;' }
```

I wonder if the vendor prefixing `context` of 2 is being confused with the plugin `context` of 2?

probablyup: If you could provide a code sample I could use in a unit test, that'd be great. In the meantime, I made this PR that should work around the specific error you mentioned: https://github.com/styled-components/styled-components/pull/2114

toolmantim: Thanks @probablyup 🙏🏼 That fixed it.

I've tried for an hour to create the smallest possible test case (both browser rendered and server rendered), but failed to recreate it I'm sorry. Crummy bug report I know 😔

Here's the commands to recreate it exactly though, if you needed:

```bash
git clone https://github.com/buildkite/site.git
cd site && git checkout da15e8cd
yarn && yarn start
```

PR #2133, opened 2018-10-21
Filter attrs with @emotion/is-prop-valid

Fer0x:
Just quick fix of #2131


src/models/StyledComponent.js
```diff
@@ -100,26 +100,28 @@ class StyledComponent extends Component<*> {
     const elementToBeCreated = this.props.as || this.attrs.as || target;
     const isTargetTag = isTag(elementToBeCreated);
 
-    const propsForElement: Object = { ...this.attrs };
+    const propsForElement: Object = {};
+    const computedProps: Object = { ...this.attrs, ...this.props };
```
probablyup: Is there a way to avoid the second object allocation? This is a hot code path so if we can eliminate it, that'd be good.

Fer0x: @probablyup Generally we can write merged props to `propsForElement` and filter them in following loop.

Fer0x: @probablyup Generally we can write merged props to `propsForElement` and filter them in following loop.

Fer0x: @probablyup Generally we can write merged props to `propsForElement` and filter them in following loop.

Issue #2136, opened 2018-10-22
`Unable to find node on an unmounted component` in Sandbox


mkascel:
Trying to get Sandbox up and running so I can work on a PR, but the app is throwing on load, it appears to be an issue with `findDOMNode` in `classNameUseCheckInjector`:

## System:
 - OS: macOS High Sierra 10.13.6
 - CPU: x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
 - Memory: 166.41 MB / 16.00 GB
 - Shell: 3.2.57 - /bin/bash
## Binaries:
 - Node: 8.11.4 - ~/.nvm/versions/node/v8.11.4/bin/node
 - Yarn: 1.10.1 - /usr/local/bin/yarn
 - npm: 5.6.0 - ~/.nvm/versions/node/v8.11.4/bin/npm
## npmPackages:
 - babel-plugin-styled-components: ^1.5.1 => 1.5.1 

## Steps to reproduce
- Fork styled-components repo
- `cd sandbox/`
- `yarn install`
- `yarn start`
- Go to http://localhost:3000

## Expected Behavior
- Site loads

## Actual Behavior
- favicon shown in tab, flash of background, then app unmounts with error
```
react-dom.development.js?054f:55 Uncaught Error: Unable to find node on an unmounted component.
    at invariant (react-dom.development.js?054f:55)
    at findCurrentFiberUsingSlowPath (react-dom.development.js?054f:4158)
    at findCurrentHostFiber (react-dom.development.js?054f:4266)
    at findHostInstance (react-dom.development.js?054f:17676)
    at Object.findDOMNode (react-dom.development.js?054f:18145)
    at StyledComponent.componentDidMount (classNameUseCheckInjector.js?0b11:22)
    at commitLifeCycles (react-dom.development.js?cada:14362)
    at commitAllLifeCycles (react-dom.development.js?cada:15463)
    at HTMLUnknownElement.callCallback (react-dom.development.js?cada:100)
    at Object.invokeGuardedCallbackDev (react-dom.development.js?cada:138)
invariant @ react-dom.development.js?054f:55
findCurrentFiberUsingSlowPath @ react-dom.development.js?054f:4158
findCurrentHostFiber @ react-dom.development.js?054f:4266
findHostInstance @ react-dom.development.js?054f:17676
findDOMNode @ react-dom.development.js?054f:18145
componentDidMount @ classNameUseCheckInjector.js?0b11:22
[...]

PR #2142, opened 2018-10-22
Fix resolving react-dom in sandbox

Fer0x:
Fix https://github.com/styled-components/styled-components/issues/2136

### DIFF
--- a/sandbox/server/webpack.config.js
+++ b/sandbox/server/webpack.config.js
@@ -11,6 +11,7 @@ const {
   serverApp,
   clientApp,
   styledComponentsSrc,
+  reactDOM,
 } = SANDBOX_PATHS;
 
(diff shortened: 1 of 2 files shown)


probablyup:
Why is it getting duplicated?

probablyup:
AppVeyor doesn't like this for some reason...

Fer0x:
@probablyup 

> AppVeyor failed to assign job to 'gce' build environment after 3 attempt(s). The envronment is either unavailable or there are no free workers, however alternate build environment was not configured by the system administrator. Please try again later.

I don't think that's problem with PR

probablyup:
Going to merge this to fix it. I think we should probably make sandbox be a yarn workspace though so we can just use the devDep specified at the top level.

PR #2150, opened 2018-10-24
Fix import rules hoisting for ssr

microcood:
Unlike for browser, for ssr ```@import``` rules are not hoisted, what results with unnecessary latency on page load (e.g. font flickering on page reload even with cached fonts).


kitten:
@mxstbr Yea, I think we probably should. Since this adds the import exception to SSR it's more of a patch, which we should shop before any refactors 👌

probablyup:
@kitten is this approved?

probablyup:
Superseded by #2344

Issue #2137, opened 2018-10-22
Component references in object-notation styles


mkascel:
Per the conversation with @mxstbr [in spectrum](https://spectrum.chat/thread/776b2d0b-cb81-42de-a8f1-7ca56b2afb48), I've gotten nested selectors and media queries working with the new object-notation, but not Component references like `${Text}`.  

Here's my Sandbox for an example (check out the `ObjectTest` file): https://codesandbox.io/s/xz9q6ovvlo

## System:
 - OS: macOS High Sierra 10.13.6
 - CPU: x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
 - Memory: 166.41 MB / 16.00 GB
 - Shell: 3.2.57 - /bin/bash
## Binaries:
 - Node: 8.11.4 - ~/.nvm/versions/node/v8.11.4/bin/node
 - Yarn: 1.10.1 - /usr/local/bin/yarn
 - npm: 5.6.0 - ~/.nvm/versions/node/v8.11.4/bin/npm
## npmPackages:
 - babel-plugin-styled-components: ^1.5.1 => 1.5.1 

## Expected Behavior
- `<Text>` component has red text

## Actual Behavior
- `<Text>` component has purple text

probablyup:
Unfortunately I don't think this will ever work due to how JS object literal keys work.

probablyup:
Here is a workaround: https://codesandbox.io/s/yv8ojjr5wj, but it relies on internals. I suppose we could add a util to make the pattern a little neater:

```js
import styled, { getSelector } from 'styled-components';

const Thing = styled.div({
  color: 'red',
});

const OtherThing = styled.div({
  [getSelector(Thing)]: {
    color: 'blue',
  },
});
```

diegohaz:
@probablyup `getSelector` would be really helpful. I was actually looking for it when I asked [on spectrum](https://spectrum.chat/?thread=ed35297e-0b75-43ec-9682-d37a5cecda92).

It turned out that I created a small library so I don't rely on styled-components' internals on my projects: https://github.com/diegohaz/styled-selector

thysultan:
Maybe a ToString method on the component could implicitly extract this when used as an object key.

```
function a () {}
a.toString = () => { return this.styledComponentId }
var obj = {[a]: ''}
```

diegohaz:
@thysultan That's how Emotion does.

I guess it would also work in template strings out of the box.

ericberens:
[...]

PR #2151, opened 2018-10-25
add getSelector util and export it from base

timurista:
This should solve #2137
Basically adding a getSelector util so you get '.ac-Sdsd' if you need to access this inside an object for doing object notationy things.

For the related conversation about adding a new util method:
https://spectrum.chat/thread/ed35297e-0b75-43ec-9682-d37a5cecda92


src/utils/getSelector.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+import isStyledComponent from './isStyledComponent';
+
+function getSelector(chunk: any) {
+  return isStyledComponent(chunk) ? `.${chunk.styledComponentId}` : chunk.toString();
```
timurista: basically copy from flatten method, but exposed as standalone function

probablyup: Let's keep it specifically for outputting the styled-components selector and do a dev-time warning if you try to use the util with a non-styled-component.

probablyup:
If you could also add a changelog entry for your work, that'd be great!

timurista:
Thanks @probablyup looks like that commit fixed the issue!

PR #2166, opened 2018-10-28
Avoid subscribing static components

mxstbr:
I'm not sure if this breaks anything, but if it doesn't it might prevent some unnecessary re-renders?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


probablyup:
This makes sense in theory. Breaks a ton of tests though...

mxstbr:
What the hell are those errors too... 😱

mxstbr:
@kitten why exactly do we do `this.styleSheet = styleSheet` over and over and over every render?

probablyup:
I think it can change in SSR scenarios from first construction to final emit?

mxstbr:
Should probably also do on native btw?

probablyup:
We don't have a concept of "staticness" in native at the moment. I looked 😛

Issue #2159, opened 2018-10-25
useless className warning


B-Reif:
I'm using styled-components to style a `rc-tooltip`, which takes the property `overlayClassName` instead of `className` in order to style the tooltip's overlay. My code contains a very basic wrapper to map the prop:

```
const TooltipStyleWrapper = props => {
	const { className, ...rest } = props;
	return <Tooltip overlayClassName={className} {...rest} />;
};
```

When this component mounts, styled-components issues the className warning:
> It looks like you've wrapped styled() around your React component (TooltipStyleWrapper), but the className prop is not being passed down to a child. No styles will be rendered unless className is composed within your React component.

In truth, I _am_ passing down `className`, but by a different name. I did a brief search for this issue and saw PR #2156 for adding another prop to suppress this warning.

I don't think it's appropriate for styled-components to pry into my components to see how I'm using the provided `className`, and I _really_ don't think it's appropriate that I should have to add another obtuse prop to suppress this warning. The above code already exists as a bridge between `sc` and `rc-tooltip`, and I very much don't want to go to even further lengths to build this bridge. If a warning is going to exist, it should be smart enough only to warn me when something is _actually wrong_.

Fer0x:
This warning is not about that you are using className prop name in your code. It search in nesting children to sure that you have a real DOM node which uses passed classnames.

Fer0x:
There is only one problem with this warning when asynchronous rendering is used. I have some ideas and will try to make some solution to handle it.

B-Reif:
The className here is ultimately rendered in a node that's set up with a portal, so it's both asynchronous (drawn later, when someone mouses over the tooltip) and detached (rendered elsewhere than the DOM tree).

If it's only searching the nested DOM nodes, it won't find content rendered in a portal.

probablyup:
You'll be able to opt out of this warning for edge cases like you've described: https://github.com/styled-components/styled-components/pull/2156

Probably will cut a release tomorrow.

B-Reif:
I'm really not happy with the need to add another prop just to suppress this warning, and I don't really view deferred or relocated rendering as an 'edge case'. I should be able to render the className later, or somewhere else, or not at all. Once the className is handed off,  it's not styled-component's business how, if, or when I choose to use it.

probablyup:
> Once the className is handed off, it's not styled-components's business how, if, or when I choose to use it.

[...]

PR #2168, opened 2018-10-29
Make classnames warnings warn once

Fer0x:
Follow https://github.com/styled-components/styled-components/issues/2159#issuecomment-433746372

### DIFF
--- a/src/utils/classNameUsageCheckInjector.js
+++ b/src/utils/classNameUsageCheckInjector.js
@@ -2,6 +2,8 @@
 import ReactDOM from 'react-dom';
 import getComponentName from './getComponentName';
 
+const didWarnAboutClassNameUsage = new Set();
+
 export default (target: Object) => {
   let elementClassName = '';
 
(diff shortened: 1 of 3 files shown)


Fer0x:
I did it with Set/className in the some way it did in React 
https://github.com/facebook/react/blob/6938dcaacbffb901df27782b7821836961a5b68d/packages/react-reconciler/src/ReactFiberClassComponent.js#L748

src/utils/classNameUsageCheckInjector.js
```diff
@@ -13,13 +15,18 @@ export default (target: Object) => {
       targetCDM.call(this);
     }
 
+    const componentName = getComponentName(this.props.forwardedClass.target);
```
probablyup: I don't think this is going to be high fidelity enough because `getComponentName` can resolve to "Component" if `displayName` isn't set and warnings won't be emitted.

What we can do instead is add `this.type` (which is the unique constructor class/function I think?) to the Set and that way it'll still track each new component individually in case they share the same name.

Fer0x: @probablyup As I mentioned above, React has the same solution (and has "Component" too in some cases). I have tried `this.type` but it was `undefined` in tests.

probablyup: How about `this.constructor`?

probablyup: The React code is fine for putting a name in a warning, but not good enough for tracking which components have received warnings.

Fer0x: @probablyup Ok, I have found how to obtain type: `this. _reactInternalFiber.type`. Are you admit to use it?

probablyup: `this.constructor` doesn't work? I tried it out in a sandbox...

Fer0x:
@probablyup `this._reactInternalFiber.type` is always identical instance of `models/StyledComponent`. So, I have found another solution - to use `this.props.forwardedClass.target`.

PR #2172, opened 2018-10-30
Add StyledProvider to bring in customisations for default behaviours

### LINKED ISSUE #2167: Add StyledProvider
There's been a decent amount of discussion and feature requests around customizing default behaviors of styled components, so let's make it happen.

Here are some of the individual requests:

- scoped selectors https://github.com/styled-components/styled-components/issues/1789
- ability to disable vendor prefixing https://github.com/styled-components/styled-components/issues/719
- disabling speedy mode https://github.com/styled-components/styled-components/issues/2038

It's been proposed by @kitten I believe that we could add another context provider/consumer combo to facilitate configuration overrides that do not need to be performed every time a styled component is created.

imbhargav5:
Attemps to fix #2167 

- [x] Add scope prop 
- [ ] Should scope cascade? 🤔 @probablyup 
- [ ] Disable vendor prefixing
- [ ] Disable speedy mode

### DIFF
--- a/src/models/ComponentStyle.js
+++ b/src/models/ComponentStyle.js
@@ -46,7 +46,7 @@ export default class ComponentStyle {
      * Hashes it, wraps the whole chunk in a .hash1234 {}
      * Returns the hash to be injected on render()
      * */
-  generateAndInjectStyles(executionContext: Object, styleSheet: StyleSheet) {
+  generateAndInjectStyles(executionContext: Object, styleSheet: StyleSheet, scope: string) {
     const { isStatic, componentId, lastClassName } = this;
     if (
       IS_BROWSER &&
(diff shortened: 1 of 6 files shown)


imbhargav5:
@probablyup Is this PR the right approach for the issue? I am thinking we will need a context for the scope behaviour and to opt out of vendor prefixing. Speeding mode, would be a global flag somewhere. How do you think we should go about it? Also if there are multiple styled providers nested , should the scopes add up ? 
```jsx
<StyledProvider scope="prefix1">
<StyledProvider scope="prefix2">
  <StyledDiv/>
</StyledProvider>
</StyledProvider>
```
My opinion is that, the scope prefix should just be "prefix2" because more scopes will make it a little bit of a mess. What do you think?

imbhargav5:
@mxstbr @probablyup @kitten Thoughts?

imbhargav5:
I think we can have DISABLE_SPEEDY, and DISABLE_VENDOR_PREFIXES be configurable using StyleSheetManager and scope via StyledProvider. StyledProvider, in my mind, is a context provider for a sub tree where you can give scope. I can't think of any other way at the moment. 🤔

probablyup:
Let's park this for v5. Given the performance issues we're seeing with context consumers, I want to give the React team some time to fix the issue and perhaps do the hooks transition first so our tree is overall more lightweight

imbhargav5:
Yes. I was thinking so too.

PR #1171, opened 2017-09-18
Support for non-keyed .attrs factory function

oliverlaz:
This PR enhances the ``.attrs`` function support for a top level factory function which receives the component props as an argument.

Example:
```javascript
const Input = styled.input.attrs(props => ({
  placeholder: props.placeholder
}))`
  color: red;
`
```

This change is very handy when integrating a styled-component within [``redux-form``](https://redux-form.com) [``Field``](https://redux-form.com/7.0.4/docs/api/field.md/) element where this code:
```javascript
const Input = styled.input.attrs({
  onChange: props => props.input.onChange,
  onBlur: props => props.input.onBlur,
  onFocus: props => props.input.onFocus,
  ...
})`
  color: red;
`
```

can be easily changed to:
```javascript
const Input = styled.input.attrs(props => ({
   ...props.input // ensure all redux-form props and handlers are passed to the wrapped input component
})`
  color: red;
`
```

PR #2176, opened 2018-10-31
Support for .attrs factory function (props => attrs)

oliverlaz:
A follow up PR of the discussion in #1171. The changes here are based on `v4.0.3`.

I made the smoke tests and test suite pass on my side so I'm opening this PR for an early review.
In the next few days, I will try this build in one of my apps (just to make sure that I didn't break something else).

ping @probablyup, @Fer0x


src/models/StyledComponent.js
```diff
@@ -203,21 +186,27 @@ export default function createStyledComponent(target: Target, options: Object, r
...
 
   // fold the underlying StyledComponent attrs up (implicit extend)
   const finalAttrs =
     // $FlowFixMe
-    isTargetStyledComp && target.attrs ? { ...target.attrs, ...attrs } : attrs;
+    isTargetStyledComp && target.attrs
+      ? (context: Context): Object => ({
+          // $FlowFixMe <- target.attrs
+          ...processAttrs(target.attrs, context),
+          ...processAttrs(attrs, context),
```
Fer0x: Maybe creating of this function here and in constructor can be extracted as `createAttrsResolver` and use typings.

Fer0x:
@oliverlaz Awesome work, I'm glad you found time to finish it.

src/models/StyledComponent.js
oliverlaz: Sure, I will update this PR later.

src/utils/attrsResolver.js
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * Creates new attributes resolver function based on given attrs configuration.
+ *
+ * When the resolver function is invoked, it will return a plain object
+ * with keys set as attribute names and their respective evaluated values.
+ */
+export default function createAttrsResolver(...attrs: Array<?Attrs>): AttrsResolver {
+  const nonNullAttrs: Array<Attrs> = attrs.filter(Boolean);
```
probablyup: This is too opinionated. We should only be ignoring `undefined`.

probablyup:
Thanks so much for this @oliverlaz! I ended up [implementing this a different way](#2200) that's a bit less code and helps support some more future use cases, but credited you in the changelog 🙇

oliverlaz:
Hey @probablyup, thank you! Looking forward to see #2200 landed on master :)

PR #2089, opened 2018-10-10
add SC_DISABLE_SPEEDY environment attribute

probablyup:
Workaround for #2038

Setting this attribute to anything truthy will disable "speedy" mode. Note that you'll want to use
  something like `webpack.DefinePlugin` to write the variable since most bundlers don't write any environment variables other than `NODE_ENV` by default:

  ```js
  // webpack.config.js

  {
    plugins: [
      new webpack.DefinePlugin({
        'process.env.SC_DISABLE_SPEEDY': true,
      })
    ]
  }
  ```

Issue #2038, opened 2018-09-29
Yandex crawler doesn't see styles in production mode


them4g:
I have a problem with [Yandex Webvisor](https://metrica.yandex.com/promo/webvisor/). In production mode styled-components injects styles to CSSOM. Because of this yandex bot can't render page properly (it renders page as plain html without styles).

When I disabled speedy mode in production it helps. And I'm wondering if we can make this mode to be optional (e.g. check for FORCE_DISABLE_SPEEDY enviroment variable is set to true). I'll glad if you suggest better solution.

probablyup:
An environment variable could work. It's been discussed before to give an option to disable speedy mode... I'm not opposed. @mxstbr @kitten @imbhargav5?

dashed:
I'm also interested in a way to control `DISABLE_SPEEDY` thru an env variable for another usecase.

mxstbr:
What's your use case @dashed?

----

Are you sure that injecting the styles via slow mode will show up on Yandex @them4g, have you tried and verified that? 

I'm not necessarily opposed to making this a configuration option, but I'd rather keep them to a minimum.

probablyup:
If anyone knows a dev over at Yandex they should probably also just add support for CSSOM styles since the other CSS-in-JS libraries use the same set of APIs. I know some analytics tools have upgraded to detect them.

probablyup:
If someone at Yandex sees this, this issue requires your crawler to make use of the CSSOM APIs: https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet

In prodution, we use `insertRule` which causes the CSS to only be accessible via `document.styleSheets[0].cssRules` etc

dashed:
> What's your use case @dashed?

@mxstbr This would sound stupid; but a segregated environment to allow styles to be editable using DevTools.

devrelm:
@mxstbr My company has another use-case: [percy.io](https://percy.io). I'd already mentioned this [on another issue](https://github.com/styled-components/styled-components/issues/1915#issuecomment-413645595).

@probablyup's PR (#2089) is appreciated, but we'd rather not have to set up a separate build that sets that environment variable. Could an additional option be available where we could set a global variable to disable speedy? I would happily provide the PR.

[...]

PR #2185, opened 2018-11-03
add SC_DISABLE_SPEEDY runtime override

devrelm:
This is, if nothing else, a temporary fix until #2167 is resolved.

Currently, using Percy doesn't work for my company. We are using `styled-components` within a `create-react-app` app, which doesn't allow its users to override `NODE_ENV` for builds.

This PR allows a global `SC_DISABLE_SPEEDY` variable to be set before `styled-components` is imported. For my company, this would look something like this:

```html
<script>SC_DISABLE_SPEEDY=/percy/.test(location.host)</script>
<script src="/main.js></script>
```

#### Related

#2038
#2089

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 ## Unreleased
 
 - Performance optimization for fully static (no function interpolation) styled-components by avoiding using `ThemeConsumer` since it isn't necessary, by [@mxstbr](https://github.com/mxstbr) (see [#2166](https://github.com/styled-components/styled-components/pull/2166))
+- Allow disabling "speedy" mode via global `SC_DISABLE_SPEEDY` variable, by [@devrelm](https://github.com/devrelm) (see [#2185](https://github.com/styled-components/styled-components/pull/2185))
 
 ## [v4.0.3] - 2018-10-30
 
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,6 +1,7 @@
 // @flow
 
 declare var __DEV__: ?string;
+declare var SC_DISABLE_SPEEDY: ?boolean;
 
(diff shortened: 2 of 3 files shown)


probablyup:
Needs at least one test

devrelm:
@probablyup updated with tests and changelog entry.

devrelm:
Updated tests to reflect changes from #2188

brianlenz:
@probablyup, any chance this PR could be backported to a v3 release? We need this exact behavior in order to get prerendering working (e.g. prerender.io). The problem right now is that we need to do prerendering on a production build, which results in component styles not being included in the DOM due to the use of CSSOM `insertRule`. We would set up a rule very similar to the host example here, only we'd set `SC_DISABLE_SPEEDY` based on the user agent (e.g. when it's for prerender). I realize you're on v4 releases now, but we've not upgraded yet, so a v3 release would be great. If you're not planning to do any further v3 releases, we'll start exploring the v4 upgrade path in anticipation of the next v4 release that includes this PR.

Thanks!

mxstbr:
We probably won't backport this change, so I think the best course of action would be for you to upgrade to v4.

If that's not possible for whatever reason, we'd probably merge a PR backporting this to the last v3 release!

brianlenz:
Hi @mxstbr, we are attempting to upgrade to styled-components 4.1, and we're running into a bit of dependency hell trying to get all of our code and dependencies to play nice after updating (we are using TypeScript, btw). It would be immensely helpful to us if you could backport this to a v3 release so we can make use of this functionality and save the upgrade to styled-components 4 for a future day. We are currently running styled-components 3.3.3, but I'm guessing upgrading to 3.4 should be much simpler for us than upgrading to 4.1.

Thanks for considering it!

mxstbr:
Hey, feel free to create a `v3` branch off the latest commit that's in v3 and submitting a PR against it that integrates this feature, we'd probably publish a new release if it works as intended!

brianlenz:
Thanks, we'll do that!

PR #2187, opened 2018-11-04
move react-is to peerDeps

probablyup:


packages/styled-components/package.json
```diff
@@ -63,20 +63,21 @@
...
     "stylis": "^3.5.0",
     "stylis-rule-sheet": "^0.0.10",
     "supports-color": "^5.5.0"
   },
   "peerDependencies": {
-    "react": ">= 16.8.0",
-    "react-dom": ">= 16.8.0"
+    "react": ">= 16.8",
+    "react-dom": ">= 16.8",
+    "react-is": ">= 16.8"
```
aretecode: This is helpful, I've had to use `resolutions` to update these dependencies in consuming packages. Thanks for changing it 👍

probablyup:
Correct, but most people have it installed already. It’s important to move
to peer deps for deduping.

mxstbr:
I strongly disagree. I have never once installed `react-is` manually and literally have no clue what that package even does. I do not want to require users of `styled-components` to also have to install a dependency without ever actually using it themselves.

If people need to dedupe they should just use the yarn `resolutions` feature which handles exactly that.

```JS
{
  "resolutions": {
    "react-is": "whatever"
  }
}
```

That will make sure to only install a single version of it across your monorepo.

kilpatty:
I noticed this was merged into the beta for v5, has the intention on this changed, and if so is there another issue to read up on the reasoning here?

Thanks!

omBratteng:
I was wondering the same as @kilpatty.
I see it was added in ff771d5e668907b51e7fbe4b69e9d55e69781274 by @kitten. Was this intentional?

PR #2196, opened 2018-11-09
Added test showing that styled-calls yield non-functions

gertsonderby:
I ran into some issues wrt. testing that seemed to be caused by styled components not being JS function type, somehow. Test replication was very easy - a simple styled component is of type `'object'`! A quick look at the constructors used did not yield an immediately obvious source, so I submit it here.

### DIFF
--- a/src/test/basic.test.js
+++ b/src/test/basic.test.js
@@ -357,5 +357,21 @@ describe('basic', () => {
 
       expect(console.warn).not.toHaveBeenCalled();
     });
+
+    it('creates component with type function for DOM elements', () => {
+      const Comp = styled.div`
+        color: red;
+      `;
+
+      expect(typeof Comp).toEqual('function');
+    });
+
+    it('creates component with type function for wrapped components', () => {
+      const Comp = styled(({ className }) => <div className={className} />)`
+        color: red;
+      `;
+
+      expect(typeof Comp).toEqual('function');
+    });
   });
 });


kachkaev:
@gertsonderby the reason why SC are objects is because they are wrapped into `React.forwardRef`. Why seeing them as functions it necessary for you?

joeegan:
One reason would be react-redux connect compatibility.

kachkaev:
Could you please provide an example that does not work for you now?

joeB2C2:
Slightly more details here: https://spectrum.chat/thread/9b5cc71e-100e-4380-a976-b40166332896

mxstbr:
As @kachkaev correctly stated, styled components are not functions due to `React.forwardRef` being used, which returns an object. If that bothers you for whatever reason, please submit an issue with React itself and discuss whether `React.forwardRef` can use a function under the hood!

> One reason would be react-redux connect compatibility.

`react-redux` has to add support for `React.forwardRef` components, there's nothing we can do from our side to help this.

joeB2C2:
OK cool, I'll try and get round to raising a PR for react-redux

mxstbr:
It seems like they recently did, see https://github.com/reduxjs/react-redux/issues/914

Might just have to upgrade!

joeB2C2:
Excellent, thanks for looking into it

PR #2199, opened 2018-11-10
Fix styled components causing update problem

### LINKED ISSUE #2195: Removing PureComponents breaks inputs
Hi, I did a bisect to find the problem we were having with styled components v4 and surveyjs. I ended up with narrowing the problem to this commit:
a7ab4704fbe0de7450ff0ff634e5024cad24fd31

I understand that PureComponents might cause some problem for you but you have to have some strategy not to mess with props.

The last version of v3 is fine if that helps.

olafura:
In certain situations styled components react component seems to be
updating when it shouldn't this is a regression from v3 which does don't
have this behaviour. Since PureComponent is too new this patch implements
the basic behaviour of it. This fixes the error we are getting which is in
our big app a deeply nested SurveyJs inputs aren't being updated because
before we can update them we have rendered a new version.

Fixes #2195

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "@emotion/is-prop-valid": "^0.6.8",
     "babel-plugin-styled-components": ">= 1",
     "css-to-react-native": "^2.2.2",
+    "fbjs": "^0.8.17",
     "memoize-one": "^4.0.0",
     "prop-types": "^15.5.4",
     "react-is": "^16.6.0",
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -2,6 +2,8 @@
 
 import validAttr from '@emotion/is-prop-valid';
 import React, { createElement, Component } from 'react';
+// $FlowFixMe
+import shallowEqual from 'fbjs/lib/shallowEqual';
 import ComponentStyle from './ComponentStyle';
 import createWarnTooManyClasses from '../utils/createWarnTooManyClasses';
 import determineTheme from '../utils/determineTheme';
@@ -66,6 +68,12 @@ class StyledComponent extends Component<*> {
     if (process.env.NODE_ENV !== 'production' && IS_BROWSER) {
       classNameUsageCheckInjector(this);
     }
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    return (
+      !shallowEqual(this.props, nextProps) || !shallowEqual(this.state, nextState)
+    );
   }
 
   render() {
--- a/src/models/StyledNativeComponent.js
+++ b/src/models/StyledNativeComponent.js
@@ -1,5 +1,7 @@
 // @flow
 import React, { createElement, Component } from 'react';
+// $FlowFixMe
+import shallowEqual from 'fbjs/lib/shallowEqual';
 import determineTheme from '../utils/determineTheme';
 import { EMPTY_OBJECT } from '../utils/empties';
 import generateDisplayName from '../utils/generateDisplayName';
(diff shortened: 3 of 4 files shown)


olafura:
I've tried making a simple create react app that simulates this, and diffs those files to try to identify the problem but I haven't found anything yet. But this does fix it.

olafura:
I did some digging and it is, ran out of memory when trying to use jsondiffpatch in the browser, with circularjson because of circular dependencies.

Thankfully I could stringify the results of the different props and use my JsonDiffEx elixir library to do the diff and it's "forwardedClass" that is the culprit.

See why you need a big application for it to happen :grin:
```
...excerpt the rest in children diff with the same nested structure
"forwardedClass":["~children~props~children~props~children~_owner~stateNode~_reactInternalFiber~return~return~return~return~return~return~return~return~return~stateNode~props~children~0~_owner~stateNode~_reactInternalFiber~return~return~return~return~return~return~return~return~return~return~return~return~return~return~return~stateNode~_reactInternalFiber~return~return~firstEffect~stateNode~__reactInternalInstance$stnc3qi6n3g~return~stateNode~props~creator~refs~root~__reactInternalInstance$stnc3qi6n3g~return~return~stateNode~__reactEventHandlers$stnc3qi6n3g~children~_owner~return~stateNode~0~0~__reactInternalInstance$stnc3qi6n3g~return~return~stateNode~__reactEventHandlers$stnc3qi6n3g~children~0~_owner~alternate~return~return~stateNode~__reactInternalInstance$stnc3qi6n3g~return~stateNode~_reactInternalFiber~return~return~stateNode~__reactInternalInstance$stnc3qi6n3g~return~return~return~stateNode~__reactEventHandlers$stnc3qi6n3g~children~_owner~return~return~return~return~return~type",
   "~children~props~children~props~children~_owner~stateNode~_reactInternalFiber~return~return~return~return~return~return~return~return~return~stateNode~props~children~0~_owner~stateNode~_reactInternalFiber~return~return~return~return~return~return~return~return~return~return~return~return~return~return~return~stateNode~_reactInternalFiber~return~return~firstEffect~stateNode~__reactInternalInstance$stnc3qi6n3g~return~stateNode~props~creator~refs~root~__reactInternalInstance$stnc3qi6n3g~return~return~stateNode~__reactEventHandlers$stnc3qi6n3g~children~_owner~return~stateNode~0~0~__reactInternalInstance$stnc3qi6n3g~return~return~stateNode~__reactEventHandlers$stnc3qi6n3g~children~0~_owner~alternate~return~return~stateNode~__reactInternalInstance$stnc3qi6n3g~return~stateNode~_reactInternalFiber~return~return~stateNode~__reactInternalInstance$stnc3qi6n3g~return~return~return~stateNode~__reactEventHandlers$stnc3qi6n3g~children~_owner~return~return~return~return~stateNode~props~forwardedClass"]
```

olafura:
forwardedClass seems to be a way of aggregation classes, but I don't know the codebase. I'm getting the string from forwardedClass from the toString because of the circular json library.

The problem with tree walking is that you have to limit scope in languages that allow for references in the element you are walking. So usually you clone the object and prune things you don't need. So React seems to have _owner on children which is causing the library to do more work than it needs to when walking the tree.

probablyup:
https://github.com/benjamine/jsondiffpatch/issues/152

jsondiffpatch is not compatible with react fiber trees because there are many cyclic references and a fiber tree cannot be marshaled into JSON. I'm not really sure what your use case is, but it's not something we can support.

olafura:
@probablyup there is a bug in your code this simple patch fixes it

probablyup:
It's not a bug, it's intentionally written that way because we don't know what's beneath that may be using legacy context.

olafura:
@probablyup it stops unnecessary updates

PR #2200, opened 2018-11-10
allow attrs to be called with a function

probablyup:
attrs is now implemented as an array internally to make it
easier to do merging etc and handle different types of things
passed in

Closes #1171 

This is an alternate implementation of https://github.com/styled-components/styled-components/pull/2176 that will allow more & easier customization in the future (and is smaller.)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,15 +1,14 @@
 {
   "parser": "babel-eslint",
-  "extends": ["airbnb", "prettier", "prettier/flowtype"],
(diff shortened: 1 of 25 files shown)


imbhargav5:
This is so amazing. :D Also, love the "probably a bad idea" test cases.

probablyup:
Haha I can't take credit for those tests!

probablyup:
mmk this is ready to go

imbhargav5:
:fire:

PR #2210, opened 2018-11-13
 ensure theme is accessible to function-form attrs 

### LINKED ISSUE #2208: new attrs as a function doesn't pass theme to the function provided
## Environment
 - React-native IOS
## System:
 - CPU: x64 Intel(R) Core(TM) i5-5350U CPU @ 1.80GHz
 - Memory: 43.83 MB / 8.00 GB
 - Shell: 5.6.2 - /usr/local/bin/zsh
## Binaries:
 - Node: 11.0.0 - /usr/local/bin/node
 - Yarn: 1.12.1 - /usr/local/bin/yarn
 - npm: 6.4.1 - /usr/local/bin/npm
## npmPackages:
 - styled-components: ^4.1.1 => 4.1.1 

## Reproduction
1. create a global theme and provide it using <ThemeProvider>. 
2. create a styled component with attrs, where the function expects a theme to be in the props. (eg, `styled.div.attrs(props => ({ backgroundColor: props.theme.bgcolor}))```)

## Expected Behavior
theme is provided to the props of the attrs function

## Actual Behavior
theme is not provided and the app crashes with  "cannot read property 'bgcolor' of undefined".

probablyup:
fixes #2208

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 ## Unreleased
 
+- Fix function-form attrs to receive the full execution context (including theme) (see [#2210](https://github.com/styled-components/styled-components/pull/2210))
+
 ## [v4.1.1] - 2018-11-12
 
 - Put back the try/catch guard around a part of the flattener that sometimes receives undetectable SFCs (fixes an errand hard error in an edge case)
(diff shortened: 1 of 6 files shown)


probablyup:
Do we need to do that explicitly? I was thinking forwardRef would automatically resolve any set `defaultProps` and pass them along as normal...

mxstbr:
Not sure, might be worth adding a test for!

antoinerousseau:
I feel like this removes the fact that before the props were forwarded as HTML attributes, and not anymore, or is it completely unrelated?

Example:

```tsx
const Infos = styled.div`
  position: ${(props: {fixed: boolean}) => (props.fixed ? 'fixed' : 'absolute')};
  right: 0;
`

// in render:

// `test` attribute is in DOM with:
<div test="ok" /> 

// `test` attribute is not in DOM with:
<Infos fixed test="ok" /> 
```

PR #2223, opened 2018-11-17
Allowing `forwardedRef` prop to drill through `styled(Wrapper)`

therealparmesh:
Let's say you have a `<FancyButton />` component with a `forwardedRef` prop like the following:

```jsx
const Button = styled.button`
  font-size: 16px;
  color: blue;
`;

const FancyButton = { forwardedRef, className, text, onClick } => (
  <Button ref={forwardedRef} className={className} onClick={onClick}>{text}</Button>;
);
```

Now let's say a consumer of `<FancyButton />` does something like the following:
```jsx
const StyledFancyButton = styled(FancyButton)`
  color: green;
`

<StyledFancyButton forwardedRef={someRef} text="Foo" onClick={() => {}} />
```

**In this scenario, doing `styled(WrappableComponent)` makes `forwardedRef` `null` down the chain.**

My proposal is simple - to allow drilling of the `forwardedRef` prop if `ref` is not defined.

This prop name is somewhat encouraged by [the React refs docs](https://reactjs.org/docs/forwarding-refs.html) - take a look at the code example right above the section titled "Displaying a custom name in DevTools".

I'm looking forward to hearing some feedback about this issue and this PR.

Keep up the great work on this amazing library!

### DIFF
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -147,8 +147,13 @@ class StyledComponent extends Component<*> {
       }
 
       if (key === 'forwardedClass' || key === 'as') continue;
-      else if (key === 'forwardedRef') propsForElement.ref = computedProps[key];
-      else if (!isTargetTag || validAttr(key)) {
+      else if (key === 'forwardedRef') {
+        if (computedProps.isRefExternal) {
+          propsForElement[key] = computedProps[key];
+        } else {
+          propsForElement.ref = computedProps[key];
+        }
+      } else if (!isTargetTag || validAttr(key)) {
         // Don't pass through non HTML tags through to HTML elements
         propsForElement[key] = computedProps[key];
       }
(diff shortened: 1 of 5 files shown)


probablyup:
Can you add some tests?

therealparmesh:
Sure thing!

therealparmesh:
Added test and entry. Thank you for taking the time to review this!

src/models/StyledComponent.js
```diff
@@ -275,7 +280,12 @@ export default function createStyledComponent(target: Target, options: Object, r
    * instead of extending ParentComponent to create _another_ interim class
    */
   const WrappedStyledComponent = React.forwardRef((props, ref) => (
-    <ParentComponent {...props} forwardedClass={WrappedStyledComponent} forwardedRef={ref} />
+    <ParentComponent
+      {...props}
+      forwardedClass={WrappedStyledComponent}
+      forwardedRef={ref !== null ? ref : props.forwardedRef}
```
probablyup: Thinking about this some more, I believe we can just start implementing an early pass at https://github.com/styled-components/styled-components/pull/2093 by making `forwardedRef` -> `$forwardedRef` and omitting it during prop computation. That way no new prop has to be added and it's just an implementation detail.

What do you think?

therealparmesh: Ah I see. I like the idea of "transient props" better than this manual props drilling logic. This PR is a more clever (not in the good way) solution.

therealparmesh: I'm gonna go ahead and close this. See if I can't help out with the other effort instead. Thanks!

PR #2225, opened 2018-11-18
isTag() additionally check for casing of the first char

asvetliakov:
It's very common to write tests like this:

```jsx
jest.mock("../SomeComponent", () => "MockedComponent");

it("test", () => {
  expect(testRenderer(<Component someParam="a" />)).toMatchSnapshot();
});
```

where ```Component``` may be:
```jsx
import SomeComponent from "./SomeComponent";
export default styled(SomeComponent)` /* stuff */`;
```

In this case ```SomeComponent``` becomes string-typed internally similar to dom-like components and only name with passed props is being rendered is snapshots.

Without this fix, the ```styled-components``` thinks as ```SomeComponent``` is like dom-like component and applies props filtering to it, resulting in filtered out custom props (```someParam``` in example above) from test snapshots.

### DIFF
--- a/src/utils/generateDisplayName.js
+++ b/src/utils/generateDisplayName.js
@@ -4,5 +4,5 @@ import isTag from './isTag';
 import type { Target } from '../types';
 
 export default function generateDisplayName(target: Target): string {
-  return isTag(target) ? `styled.${target}` : `Styled(${getComponentName(target)})`;
+  return isTag(target) ? `styled.${getComponentName(target)}` : `Styled(${getComponentName(target)})`;
 }
--- a/src/utils/generateDisplayName.js
+++ b/src/utils/generateDisplayName.js
@@ -4,5 +4,6 @@ import isTag from './isTag';
 import type { Target } from '../types';
 
 export default function generateDisplayName(target: Target): string {
-  return isTag(target) ? `styled.${getComponentName(target)}` : `Styled(${getComponentName(target)})`;
+  // $FlowFixMe
+  return isTag(target) ? `styled.${target}` : `Styled(${getComponentName(target)})`;
 }
--- a/src/utils/getComponentName.js
+++ b/src/utils/getComponentName.js
@@ -1,6 +1,6 @@
 // @flow
 import type { ComponentType } from 'react';
 
-export default function getComponentName(target: ComponentType<*>): string {
-  return target.displayName || target.name || 'Component';
+export default function getComponentName(target: ComponentType<*> | string): string {
+  return typeof target === "string" ? target : target.displayName || target.name || 'Component';
 }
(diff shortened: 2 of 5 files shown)


src/utils/isTag.js
```diff
@@ -2,5 +2,5 @@
 import type { Target } from '../types';
 
 export default function isTag(target: Target) /* : %checks */ {
-  return typeof target === 'string';
+  return typeof target === 'string' && target.charAt(0) === target.charAt(0).toLowerCase();
```
probablyup: let's wrap this in a `NODE_ENV !== 'production'` check

asvetliakov: I think it's not necessary. Built-in tag components are always lower-cased by [react spec](https://reactjs.org/docs/jsx-in-depth.html#user-defined-components-must-be-capitalized)

probablyup: Yes, I just don't want to include the code in production since it will never happen :)

src/utils/generateDisplayName.js
```diff
@@ -4,5 +4,5 @@ import isTag from './isTag';
 import type { Target } from '../types';
 
 export default function generateDisplayName(target: Target): string {
-  return isTag(target) ? `styled.${target}` : `Styled(${getComponentName(target)})`;
+  return isTag(target) ? `styled.${getComponentName(target)}` : `Styled(${getComponentName(target)})`;
```
probablyup: why was this change added? it's not desired

asvetliakov: Flow can't coerce correctly to string type for some reason

probablyup: Ah ok, just put a `// $FlowFixMe`

asvetliakov: done

src/utils/getComponentName.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 import type { ComponentType } from 'react';
 
-export default function getComponentName(target: ComponentType<*>): string {
-  return target.displayName || target.name || 'Component';
+export default function getComponentName(target: ComponentType<*> | string): string {
+  return typeof target === "string" ? target : target.displayName || target.name || 'Component';
```
probablyup: this can be reverted as well

probablyup:
Finally, just need a test and a changelog entry :) thanks for your work!

src/utils/getComponentName.js
asvetliakov: I don't think so, without it the generated display name will be ```Styled(Component)``` with mocked components because string doesn't have either displayName || name

probablyup: It won't be because you're accounting for the uppercase string name in `isTag`. It'll end up as `styled.YourMockedName`

asvetliakov: No, it won't, because ```isTag``` will return false for such components (check for first lower case letter will fail)

asvetliakov:
@probablyup Ping

PR #2234, opened 2018-11-21
Update tips-and-tricks.md

milasevicius:


mxstbr:
`ref` works now thanks to React.forwardRef, as the text above the codeblock states correctly! No more need for `innerRef` :+1:

Thanks for taking the time to submit this PR, appreciated!

PR #2239, opened 2018-11-23
 inject static classnames from folded components

### LINKED ISSUE #2232: v4 styles for child not applied
If you put a styled() over a styled() you cannot style the child from the parent anymore.

With version 3.4.9 and/or 2.0.1 this still worked, but with version 4.x.x not anymore.

Hint: If you replace the "<StyledD" with "<D", everything works as expected, because you remove one styled-instance like that.

### System:
 - OS: macOS High Sierra 10.13.5
 - CPU: (4) x64 Intel(R) Core(TM) i7-5557U CPU @ 3.10GHz
 - Memory: 191.48 MB / 16.00 GB
 - Shell: 3.2.57 - /bin/bash
### Binaries:
 - Node: 8.11.3 - ~/n/bin/node
 - npm: 5.6.0 - ~/n/bin/npm
 - Watchman: 4.7.0 - /usr/local/bin/watchman
### npmPackages:
 - babel-plugin-styled-components: ~1.8.0 => 1.8.0
 - styled-components: ~4.0.3 => 4.0.3

## Reproduction
https://codesandbox.io/s/nnnpyx5jrl

## Steps to reproduce
Just run the example.

## Expected Behavior
It must be possible to style the child as well. The border of the children must be green.

## Actual Behavior
It is not possible to style the child. The border of the children is brown (which is wrong).

probablyup:
fixes #2232

this allows for interim components that have been folded to still
be targeted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 - Remove `createGlobalStyle` multimount warning; Concurrent and Strict modes intentionally render the same component multiple times, which causes this warning to be triggered always even when usage is correct in the application (see [#2216](https://github.com/styled-components/styled-components/pull/2216))
 
+- Folded components are now targettable via component selector as in v3 if you used the old `.extend` API (see [#2239](https://github.com/styled-components/styled-components/pull/2239))
+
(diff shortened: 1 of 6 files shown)


delijah:
Can confirm fix of #2232

PR #2244, opened 2018-11-26
Don't generate a UID name from the macro for the import

mxstbr:
This is unnecessary, as the import name is already guaranteed to be unique. By not rewriting the import name, we can make the Babel plugin css prop work with the macro!

Before:

```JS
import styled from 'styled-components/macro';

// Before this was turned into
import _styled from 'styled-components';

// Now it's turned into
```

/cc @satya164

### DIFF
--- a/src/macro/index.js
+++ b/src/macro/index.js
@@ -16,6 +16,7 @@ function styledComponentsMacro({ references, state, babel: { types: t }, config 
 
   // references looks like this
   // { default: [path, path], css: [path], ... }
+  let customImportName;
   Object.keys(references).forEach(refName => {
     if (!allowedImports.includes(refName)) {
       throw new MacroError(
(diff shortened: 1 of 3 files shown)


mxstbr:
Hold up on merging this, @lucleray correctly pointed out that this breaks if you do

```
import myStyled from 'styled-components';
import styled from 'another-package';
```

Gotta think about this a bit more, I guess we have to pass through the unique name to the Babel plugin somehow?

mxstbr:
New plan is implemented in latest commit:

- Still generate unique name for import
- Pass generated name through to Babel plugin via Babel state so that the Babel plugin can basically do `const importName = state.customImportName || generateUIDName()`

Just gotta update the Babel plugin PR now

mxstbr:
This is ready for a review now that the Babel plugin PR is updated! It all works locally perfectly:

screenshot 2018-11-26 at 17 10 15

probablyup:
I think the issue is /macro isn't included in this list in the babel plugin: https://github.com/styled-components/babel-plugin-styled-components/blob/master/src/utils/detectors.js#L1

mxstbr:
Well, that would be an even easier solution if it works...

lucleray:
@probablyup
The `styled-components/macro` import is replaced by `styled-components` before it goes through the babel-plugin.
(The babel-macro plugin removes `styled-components/macro` , and the macro.js files adds `styled-components` back)

probablyup:
Ah ok, let's ship this then.

Issue #2218, opened 2018-11-15
Warning: Unknown event handler property


khrizt:
It looks like in some cases the props passed to a nested wrapped styled component end up in the final html tag when in v3 using extend did not.

I'm not sure if this is a bug or I'm using something incorrectly but as seeing that in the v3 this warning did not appear I chose to file a bug report instead of asking in stack overflow.

## System:
 - OS: Linux 4.15 Ubuntu 18.04.1 LTS (Bionic Beaver)
 - CPU: x64 Intel(R) Core(TM) i5-6500 CPU @ 3.20GHz
 - Memory: 483.48 MB / 15.53 GB
 - Container: Yes
 - Shell: 4.4.19 - /bin/bash
## Binaries:
 - Node: 9.10.0 - /usr/local/bin/node
 - Yarn: 1.12.3 - /usr/bin/yarn
 - npm: 5.6.0 - /usr/local/bin/npm
## npmPackages:
 - styled-components: ^4.0.0 => 4.1.1

## Reproduction

https://codesandbox.io/s/mq48yr83vy

## Steps to reproduce

Just access to this codesandbox at take a look at the console (I can't see it in the condesandbox console, but I can in the browser console).

## Expected Behavior

Not produce a warning.

## Actual Behavior

Produce a warning. In the v3 I had this as an extend and I never saw this warning.

juanmagalhaes:
I'm having this issue too. I'm using a custom HOC that I created to enhance the component with a prop `onOutsideClick` and basically I pass all the props back to the wrapped component like this:

```
      render() {
        return (
          <div ref={this.setWrapperRef} style={wrapperStyle}>
            <WrappedComponent {...this.props} />
          </div>
        );
      }

```

I know that if I just filter this property out of `this.props` it will stop passing `onOutsideClick` to the wrapped component (Which can be a styled component) and the warning will go away, but it is annoying to have to do that. What if I really wanted to have a property called `onWhatever` would this warning be raised too?

probablyup:
Thanks, this does seem to be a bug.

mbrowne:
It seems to happen with any prop beginning with `on` regardless of case, e.g. we're seeing it with a prop named `onlyOneColumn`.

mxstbr:
Do you reckon this is a bug in `is-valid-prop` @probablyup?

probablyup:
Probably

kokilaw:
[...]

PR #2251, opened 2018-11-27
Update @emotion/is-prop-valid

Andarist:
fixes https://github.com/styled-components/styled-components/issues/2218
related https://github.com/emotion-js/emotion/issues/1042

This fails atm only on this test - https://github.com/styled-components/styled-components/blob/3056e3110dd7b64ecbe2dc3609e338cb7d73da91/src/utils/test/validAttr.test.js#L414-L419 . Is this something that you feel strongly about supporting? IMHO its not a real life use case (even if technically its correct) - `data-` & `aria-` are written with lowercase by convention.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
   },
   "homepage": "https://styled-components.com",
   "dependencies": {
-    "@emotion/is-prop-valid": "^0.6.8",
+    "@emotion/is-prop-valid": "^0.7.3",
     "@emotion/unitless": "^0.7.0",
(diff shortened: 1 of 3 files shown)


probablyup:
Yes it should continue to work

mxstbr:
That's a bug in is-prop-valid

Andarist:
Tests are passing, but the size check is broken - this is not fault of this PR though.

probablyup:
Thanks!

Issue #2246, opened 2018-11-26
Default props behavior changed between v3 and v4


TimonVS:
The behavior of how `defaultProps` are handled has changed between s-c v3 and v4, I'm not exactly sure if this change was intentional but it was not apparent in the changelog, so that's why I'm filing it :).

Anyways, in v3 I could have a `Text` component with some default props such as the font. I could then create a `Heading` component which is based on `Text` but may provide some different default props. In v3, the following example works fine and both the `Heading` and `Text` components always have a default `font` prop. In v4 however, the `defaultProps` of the `Heading` overwrites all the default props of `Text`. If I don't define any `defaultProps` for `Heading` it seems to work fine.

```
const Text = styled.span`
  font-family: ${({ font }) => {
    if (!font) console.error('FONT PROP NOT FOUND!')
    return font
  }};
`

Text.defaultProps = {
  font: 'sans-serif'
}

const Heading = styled(Text)`
  display: block;
  font-size: 38px;
`

Heading.defaultProps = {}
```

## Environment

`npx envinfo --system --binaries --npmPackages styled-components,babel-plugin-styled-components --markdown --clipboard`

N/A, see reproduction.

## Reproduction

https://www.webpackbin.com/bins/-KeeZCr0xKfutOfOujxN
https://codesandbox.io/s/rkmNRByE4

The exact same demo in two different versions of Styled Components:

s-c v3, `Heading` and `Text` are both using `sans-serif`: https://codesandbox.io/s/6wmwk9l8kk
s-c v4, the `font-family` for `Heading` is undefined because the `font` prop is undefined: https://codesandbox.io/s/9jlw0zq5yo

## Expected Behavior

The `Heading` component should inherit the `defaultProps` from the `Text` component like it did in v3. I think in v3 the `defaultProps` would be merged?

![image](https://user-images.githubusercontent.com/876666/49029693-83ec4f00-f1a5-11e8-88ef-0d902e58c475.png)

## Actual Behavior

The `defaultProps` of `Heading` overwrites the `defaultProps` of the `Text` component even though `Heading` is based on the `Text` component.

![image](https://user-images.githubusercontent.com/876666/49029715-923a6b00-f1a5-11e8-885b-4b7e83b136d1.png)

probablyup:
So I think what's happening here is we don't attempt to fold the defaultProps together. In v3 each styled component layer was actually rendered separately, whereas in v4 there is only 1 component rendered no matter how many `styled` override layers there might be.

[...]

PR #2258, opened 2018-12-01
WIP: fix attrs handling

oomathias:
Fix muted context, introduced in `4.1.2` by #2210 

`context` was created here

https://github.com/styled-components/styled-components/blob/a41dfd0537f96b2a8572409382faedcdc936849e/src/models/StyledComponent.js#L175

passed here

https://github.com/styled-components/styled-components/blob/a41dfd0537f96b2a8572409382faedcdc936849e/src/models/StyledComponent.js#L189

and muted here

https://github.com/styled-components/styled-components/blob/a41dfd0537f96b2a8572409382faedcdc936849e/src/models/StyledComponent.js#L213

which generates side effects described and discovered while testing regarding this issue https://github.com/styled-components/styled-components/issues/2246#issuecomment-443435142

`attrs` wasn't overridable because props where created and passed between components in a reverse order.

Maybe we should add some test to prevent this in the future.

### DIFF
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -193,7 +193,7 @@ class StyledComponent extends Component<*> {
       /* eslint-disable guard-for-in */
       // $FlowFixMe
       for (key in resolvedAttrDef) {
-        if(this.attrs[key]) continue
+        if (this.attrs[key]) continue;
         attr = resolvedAttrDef[key];
 
         if (!attrDefWasFn) {
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -186,13 +186,14 @@ class StyledComponent extends Component<*> {
 
       if (isFunction(resolvedAttrDef)) {
         // $FlowFixMe
-        resolvedAttrDef = resolvedAttrDef({ ...props, theme });
+        resolvedAttrDef = resolvedAttrDef(context);
         attrDefWasFn = true;
       }
 
       /* eslint-disable guard-for-in */
       // $FlowFixMe
       for (key in resolvedAttrDef) {
+        if(this.attrs[key]) continue
         attr = resolvedAttrDef[key];
 
         if (!attrDefWasFn) {
@@ -258,7 +259,7 @@ export default function createStyledComponent(target: Target, options: Object, r
   const finalAttrs =
     // $FlowFixMe
     isTargetStyledComp && target.attrs
-      ? Array.prototype.concat(target.attrs, attrs).filter(Boolean)
+      ? Array.prototype.concat(attrs, target.attrs).filter(Boolean)
       : attrs;
 
   const componentStyle = new ComponentStyle(
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -186,7 +186,7 @@ class StyledComponent extends Component<*> {
 
       if (isFunction(resolvedAttrDef)) {
         // $FlowFixMe
-        resolvedAttrDef = resolvedAttrDef(context);
+        resolvedAttrDef = resolvedAttrDef({ ...props, theme });
         attrDefWasFn = true;
       }
 
--- a/src/test/overriding.test.js
+++ b/src/test/overriding.test.js
@@ -35,31 +35,31 @@ describe('extending', () => {
   });
 
   describe('inheritance', () => {
-
     const setupParent = () => {
       const colors = {
         primary: 'red',
         secondary: 'blue',
         tertiary: 'green',
-      }
+        quaternary: 'yellow',
+      };
       const Parent = styled.h1`
         position: relative;
         color: ${props => colors[props.color]};
       `;
       return Parent;
-    }
+    };
 
     const addDefaultProps = (Parent, Child, Grandson) => {
       Parent.defaultProps = {
         color: 'primary',
-      }
+      };
       Child.defaultProps = {
         color: 'secondary',
-      }
+      };
       Grandson.defaultProps = {
         color: 'tertiary',
-      }
-    }
+      };
+    };
 
     it('should override parents defaultProps', () => {
       const Parent = setupParent();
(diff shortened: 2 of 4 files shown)


oomathias:
cc @probablyup since you are the author of the changes

oomathias:
I added two tests, only the dynamic one was failing.
It's just a side effect of the muted context. My knowledge of the `styled-components`' core is limited, and it's clearly not the best test case to prevent the regression.

src/models/StyledComponent.js
```diff
@@ -186,7 +186,7 @@ class StyledComponent extends Component<*> {
 
       if (isFunction(resolvedAttrDef)) {
         // $FlowFixMe
-        resolvedAttrDef = resolvedAttrDef(context);
+        resolvedAttrDef = resolvedAttrDef({ ...props, theme });
```
probablyup: I don't think this is. Each successive call of attrs should receive the full execution context, including any change made by a previous `attrs` invocation.

oomathias: The problem is that the context seems to be constructed from the Child to the Parent, which leads to "unexpected" results. see: https://github.com/styled-components/styled-components/issues/2246#issuecomment-443435142

oomathias: I see what you mean but with your code:
```js
const Parent = styled.div.attrs(props => ({
  color: 'second',
}))``
const Child = styled(Parent).attrs(props => ({
  color: props.color + 'ary',
}))``
```
`<Child />` `color` was actually `secondary`, but for me, it should work the other way around.
static `attrs` are overridden by a Child, I don't see why it shouldn't be the case with dynamic `attrs`.

oomathias:
Ok, second try.

I reversed the way attrs are handled. Maybe I'm wrong, but I think it's better to start from the styled child to the parent.

```js
const Parent = styled.div.attrs(props => ({
  color: 'red',
}))``
const Child = styled(Parent).attrs(props => ({
  color: props.color || 'blue',
}))``
<Child color='yellow' />
```

I'm starting from `Child` to `Parent` attrs. If Child colors attrs is resolved, from my understanding, it overrides any parents color attrs, so we don't even need to evaluate them (unless we want to support chaining override). If that's actually what we want, deprecating `.attrs({ color: props.color || 'red' }) is maybe a bad idea, because we have access to the color key and can prevent the resolution of attrs we don't need to resolve.

Can someone from the core team confirm how attrs want to be handled?

src/test/overriding.test.js
```diff
@@ -76,12 +76,54 @@ describe('extending', () => {
...
+
     describe('when overriding with another component', () => {
       it('should override parents defaultProps', () => {
         const Parent = setupParent();
-        const Child = styled(Parent).attrs({as: 'h2'})``;
-        const Grandson = styled(Child).attrs({as: 'h3'})``;
-        console.log
+        const Child = styled(Parent).attrs({ as: 'h2' })``;
+        const Grandson = styled(Child).attrs({ as: 'h3' })``;
+        console.log;
```
probablyup: can you remove these console logs

probablyup:
The topmost attrs should always take precedence. The stack should unwind bottom-up, with the final wrapping layer being the ultimate source of truth.

oomathias:
If the final wrapping layer is the source of truth, I don't get why it should unwind the other way.

If you take this code:
```js
const Parent = styled.div.attrs(props => ({
  color: props.color + 'C'
}))``
const Child = styled(Parent).attrs(props => ({
  color: props.color + 'B'
}))``
<Child color='A' />
```
With the last version `4.1.2` the result is `ACB` [4.1.2 demo](https://codesandbox.io/s/8jr8rmm4j)
With `4.1.1` the result is `AB` [4.1.1 demo](https://codesandbox.io/s/l2516y59z) 

When you are defining `.attrs` I think it should work like the React tree and not like the styled-components tree. I would actually expect the result to be `AB` (or `ABC` if we support nesting override).

Actually this code doesn't work with `4.1.2` and I think this is widely used:
```js
const Parent = styled.div.attrs(props => ({
  color: props.color || 'C'
}))``
const Child = styled(Parent).attrs(props => ({
  color: props.color || 'B'
}))``
<Child />
```
In `4.1.2` color is `props.color` or `C` and would never be `B`. And I would expect color to be `props.color` or `B` since the last layer is the source of truth.

What I'm doing here is unwind it top-down, with the first layer being the source of truth. In my first example, the result is `AB` (I don't see a way to support `ABC` with the new `attrs` format).

Can you confirm for each of my example which result would you expect?

src/test/overriding.test.js
oomathias: sure

oomathias:
Well, in that case, we shouldn't pass any context between attrs.

probablyup:
Why? There's no harm in building it up sequentially.

oomathias:
Did you read my two example? Please give me what you expect for the results. Maybe we just don't agree on the results.

probablyup:
I guess my issue here is one of what is more used.

Based on components I've seen in the wild, it's more common to see multiple levels of attrs that deal with different props (for example component A cares about "color", and component B cares about "disabled".) For this pattern to work, we do need to sequentially built up context.

Your examples deal with multiple levels of attrs that modify or are derived from the same prop. This pattern I haven't seen very much... so I'm not sure realistically how useful it is to change all the existing behavior to support it. I'm also not really understanding what your actual use case is for the pattern... the examples are contrived by design to showcase the functionality.

oomathias:
I didn't see components in the wild where `B` is using `disabled` from `A.attrs()`. B for me is getting it from `<B disabled />` or `<C disabled />` or `B.attrs()`.
```js
const Button = styled(SomeComponent).attrs(props => ({
  color: props.disabled ? 'red' : 'blue'
}))``
```
I would expect `<Button disabled />` to be `red`, that's not always the case in `4.1.2` depending on what was done in `SomeComponent`.

mxstbr:
> I just think that it might be very confusing that attrs from Parent have priority over props.

`.attrs` always have priority over props, otherwise why have it at all, right? It should have priority over props always, if you just want to set a default then use `defaultProps`!

oomathias:
Ok. I'm closing, I think the way dynamic `attrs` is merged should be documented. The props are passed to the Parent and attrs is constructed bottom-up (so in reverse compared to the "React tree").
```js
const Parent = styled.div.attrs({
  color:  'red'
})``
const Child = styled(Parent).attrs(props => ({
  color: props.color
}))``
<Child /> // Child is always red the way attrs works.
```

This would reflect more what's happening starting from `4.1.2` (which was a breaking change)
```js
const Child = styled(Parent).attrs(attrs => ({
  color: attrs.color
}))``
```
Naming it `props` in the documentation is confusing, at least for me, since I expected it to be a always a `prop` or `attrs` from a child. And I can think of powerful utils if the way attrs is folded stays like it is now (hence adding a test and documentation).

I agree that `defaultProps` is the way to go, but it isn't possible in `v4` until we agree on my other PR.

oomathias:
Ok. I'm closing, I think the way dynamic `attrs` is merged should be documented. The props are passed to the Parent and attrs is constructed bottom-up (so in reverse compared to the "React tree").
```js
const Parent = styled.div.attrs({
  color:  'red'
})``
const Child = styled(Parent).attrs(props => ({
  color: props.color
}))``
<Child /> // Child is always red the way attrs works.
```

This would reflect more what's happening starting from `4.1.2` (which was a breaking change)
```js
const Child = styled(Parent).attrs(attrs => ({
  color: attrs.color
}))``
```
Naming it `props` in the documentation is confusing, at least for me, since I expected it to be a always a `prop` or `attrs` from a child. And I can think of powerful utils if the way attrs is folded stays like it is now (hence adding a test and documentation).

I agree that `defaultProps` is the way to go, but it isn't possible in `v4` until we agree on my other PR.

PR #2260, opened 2018-12-03
add defaultProps folding

### LINKED ISSUE #2246: Default props behavior changed between v3 and v4

oomathias:
Fix #2246 

Attempt to fold defaultProps.

```js
const Text = styled.p`
  font-family: ${props => props.font};
`
Text.defaultProps = {
  font: 'sans-serif'
}

const Heading = styled(Text)``
Heading.defaultProps = {
  as: 'h2'
}
```

`Heading` is rendered with `font: 'sans-serif'`

@probablyup you mentionned that `defaultProps.style` should have a special case. Which one?

I didn't make the changes for Native, I'm waiting for some feedback.

### DIFF
--- a/packages/styled-components/src/models/StyledComponent.js
+++ b/packages/styled-components/src/models/StyledComponent.js
@@ -298,6 +298,31 @@ export default function createStyledComponent(target: Target, options: Object, r
   // $FlowFixMe
   WrappedStyledComponent.target = isTargetStyledComp ? target.target : target;
 
+  // fold defaultProps
+  // $FlowFixMe
+  WrappedStyledComponent._defaultProps = EMPTY_OBJECT;
+  // $FlowFixMe
+  Object.defineProperty(WrappedStyledComponent, 'defaultProps', {
+    enumerable: true,
+    get() {
+      return this._defaultProps;
+    },
+    set(defaultProps) {
+      this._defaultProps = {
+        ...this._defaultProps,
+        ...defaultProps,
+        ...(this._defaultProps.style || defaultProps.style
+          ? {
+              style: {
+                ...(this._defaultProps.style || EMPTY_OBJECT),
+                ...(defaultProps.style || EMPTY_OBJECT),
+              },
+            }
+          : EMPTY_OBJECT),
+      };
+    },
+  });
+
   // $FlowFixMe
   WrappedStyledComponent.withComponent = function withComponent(tag: Target) {
     const { componentId: previousComponentId, ...optionsToCopy } = options;
--- a/packages/styled-components/src/models/StyledNativeComponent.js
+++ b/packages/styled-components/src/models/StyledNativeComponent.js
@@ -222,6 +222,32 @@ export default (InlineStyle: Function) => {
       ? // $FlowFixMe
         target.target
       : target;
+
+    // fold defaultProps
+    // $FlowFixMe
+    WrappedStyledNativeComponent._defaultProps = EMPTY_OBJECT;
+    // $FlowFixMe
+    Object.defineProperty(WrappedStyledNativeComponent, 'defaultProps', {
+      enumerable: true,
+      get() {
+        return this._defaultProps;
+      },
+      set(defaultProps) {
+        this._defaultProps = {
+          ...this._defaultProps,
+          ...defaultProps,
+          ...(this._defaultProps.style || defaultProps.style
+            ? {
+                style: {
+                  ...(this._defaultProps.style || EMPTY_OBJECT),
+                  ...(defaultProps.style || EMPTY_OBJECT),
+                },
+              }
+            : EMPTY_OBJECT),
+        };
+      },
+    });
+
     // $FlowFixMe
     WrappedStyledNativeComponent.withComponent = function withComponent(tag: Target) {
       const { displayName: _, componentId: __, ...optionsToCopy } = options;
--- a/packages/styled-components/src/native/test/native.test.js
+++ b/packages/styled-components/src/native/test/native.test.js
@@ -297,6 +297,74 @@ describe('native', () => {
     });
   });
 
+  describe('defaultProps' , () => {
+    it('should extends parents defaultProps', () => {
+      const Parent = styled.Text({}, props => ({
+        color: props.color,
+        backgroundColor: props.backgroundColor,
+        borderColor: props.borderColor
+      }));
+      Parent.defaultProps = {
+        color: 'red',
+      };
+      const Child = styled(Parent)``;
+      Child.defaultProps = {
+        backgroundColor: 'blue',
+      };
+      const Grandson = styled(Child)``;
+      Grandson.defaultProps = {
+        borderColor: 'green',
+      };
+
+      const parentWrapper = TestRenderer.create(<Parent />);
+      const childWrapper = TestRenderer.create(<Child />);
+      const grandsonWrapper = TestRenderer.create(<Grandson />);
+
+      const parentText = parentWrapper.root.findByType('Text');
+      const childText = childWrapper.root.findByType('Text');
+      const grandsonText = grandsonWrapper.root.findByType('Text');
+
+      expect(parentText.props.style).toMatchObject([{
+        color: 'red'
+      }]);
+      expect(childText.props.style).toMatchObject([{
+        color: 'red',
+        backgroundColor: 'blue'
+      }]);
+      expect(grandsonText.props.style).toMatchObject([{
+        color: 'red',
+        backgroundColor: 'blue',
+        borderColor: 'green'
+      }]);
+    });
+
+    it('should extends parents default style', () => {
+      const Parent = styled.Text``;
+      Parent.defaultProps = {
+        style: { color: 'blue' },
+      };
+
+      const Child = styled(Parent)``;
+      Child.defaultProps = {
+        style: { backgroundColor: 'red' },
+      };
+
+      const parentWrapper = TestRenderer.create(<Parent />);
+      const childWrapper = TestRenderer.create(<Child />);
+
+      const parentText = parentWrapper.root.findByType('Text');
+      const childText = childWrapper.root.findByType('Text');
+
+      expect(parentText.props.style).toMatchObject([{}, {
+        color: 'blue',
+      }]);
+      expect(childText.props.style).toMatchObject([{}, {
+        color: 'blue',
+        backgroundColor: 'red',
+      }]);
+    });
+  })
+
   describe('expanded API', () => {
     it('should attach a displayName', () => {
       View.displayName = 'View';
--- a/packages/styled-components/src/test/defaultProps.test.js
+++ b/packages/styled-components/src/test/defaultProps.test.js
@@ -1,0 +1,89 @@
+import { expectCSSMatches, resetStyled } from './utils';
+
+// @flow
+import React from 'react';
+import TestRenderer from 'react-test-renderer';
+
+let styled;
+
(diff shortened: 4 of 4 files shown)


oomathias:
Make sense, style is merged now.

src/test/defaultProps.test.js
```diff
@@ -52,5 +52,26 @@ describe('extending', () => {
         .f{ position:relative; color:red; background-color:blue; border-color: green; }
       `);
     });
+
+    it.only('should merge parents style', () => {
```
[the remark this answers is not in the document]
oomathias: already done ;)

src/models/StyledComponent.js
```diff
@@ -298,6 +323,17 @@ export default function createStyledComponent(target: Target, options: Object, r
...
 
+  // $FlowFixMe
+  const targetDefaultProps = target.targetDefaultProps || {};
+  // $FlowFixMe
+  const defaultProps = target.defaultProps || {};
+  // $FlowFixMe
+  WrappedStyledComponent.targetDefaultProps = {
+    ...targetDefaultProps,
+    ...defaultProps,
+    ...mergeStyleProp(defaultProps.style, targetDefaultProps.style),
```
mxstbr: I don't think there's a need for this separate util, let's just make this

```suggestion
...({ ...defaultProps.style, ...targetDefaultProps.style })
```

oomathias: This wouldn't work. It's missing the `style` key, we can't spread `undefined` and we can't send `style: undefined` or `style: {}`. I change the util for mergeProp, so we can use it for other special case. I could use reduce, or lodash.get equivalent but I think this way is faster.

mxstbr: Ah I see, okay let's just keep i this way then! Good catch

oomathias:
Ok, I'm waiting the sign-off but I still need to do the same inside `StyledNativeComponent`.

src/models/StyledComponent.js
```diff
@@ -298,6 +315,20 @@ export default function createStyledComponent(target: Target, options: Object, r
...
+    ...restTargetDefaultProps
+    // $FlowFixMe
+  } = target.targetDefaultProps || {};
+  // $FlowFixMe
+  const { style: styleDefaultProp, ...restDefaultProps } = target.defaultProps || {};
+  // $FlowFixMe
+  WrappedStyledComponent.targetDefaultProps = {
+    ...restTargetDefaultProps,
+    ...restDefaultProps,
+    ...mergeProp('style', targetStyleDefaultProp, styleDefaultProp),
```
probablyup: Can simplify to

```js
const targetDefaultProps = target.defaultProps || EMPTY_OBJECT;

WrappedStyledComponent.targetDefaultProps = {
    ...targetDefaultProps,
    ...mergeProp('style', target.targetDefaultProps.style, targetDefaultProps.style),
};
```

I'm not even sure we need the concept of `targetDefaultProps`. Just set `defaultProps` on the forwardRef right?

oomathias: Hmm yeah, I might be able to optimize I will look at it tomorrow. And I see a case where I might have done something wrong.

```js
const Text = styled.p`
  font-family: ${props => props.font};
`
Text.defaultProps = {
  font: 'sans-serif'
}
```
Here `Text` is `WrappedStyledComponent` so `Text.defaultProps` would actually override it, that's why I used `targetDefaultProps`.

oomathias:
Ok so as I expected theming wasn't fully supported, I pushed a fix. I have the feeling that it starts to be ugly.

I also found something that might not be wanted: 
```js
const Parent = styled.div``;
Parent.defaultProps = {
  style: { color: 'blue' },
};
<Parent style={{ border: 'none' }} />
// Parent has color: blue and border: none
```
Why? Because `defaultProps` is applied by React on `forwardRef`, and I can't know which style key was from a `prop` or `defaultProp`. I could do a diff of style's keys and reconstruct style (we have defaultProps.style and props.style+defaultProps.style).
It's kinda the same case as `determineTheme` except here I'm dealing with objects I have to merge.

Let me know what you think.

src/models/StyledComponent.js
```diff
@@ -275,9 +278,29 @@ export default function createStyledComponent(target: Target, options: Object, r
...
    * instead of extending ParentComponent to create _another_ interim class
    */
-  const WrappedStyledComponent = React.forwardRef((props, ref) => (
-    <ParentComponent {...props} forwardedComponent={WrappedStyledComponent} forwardedRef={ref} />
-  ));
+  const WrappedStyledComponent = React.forwardRef((props = EMPTY_OBJECT, ref) => {
+    // $FlowFixMe
+    const defaultProps = WrappedStyledComponent.defaultProps || EMPTY_OBJECT;
+    // $FlowFixMe
+    WrappedStyledComponent.defaultProps = {
```
oomathias: I could just merge `theme` here because it's only used here https://github.com/styled-components/styled-components/blob/d45ef9d0edb70a5b3bcbf74273f7acd5e8efdf9a/src/models/StyledComponent.js#L125
But since it's named `defaultProps`, i'm sending a proper defaultProps to `forwardedComponent`.

oomathias:
I found a cleaner way, the new POC. https://github.com/styled-components/styled-components/pull/2260/files

oomathias:
@mxstbr @probablyup Could you review this so we can progress on this PR?

_I'm developing complex components and I really need this._

lucarge:
Hi everyone 👋 Any updates on this issue? Looks like the newer @oomathias implementation is pending a review, and at my company this issue is a major blocker to upgrade from v3.

Any chance to see this merged soon? 🙏

horyd:
Looking forward to having this change in - I've successfully upgraded our app using @oomathias branch of 4.1.2 with this feature and it's working well

horyd:
Does anything else need to be completed for this feature to be added?

oomathias:
I rebased/squashed on master, since the architecture of the project changed.
Also added support for React Native with some tests related.

I'm waiting for feedback from the core team.

packages/styled-components/src/models/StyledComponent.js
```diff
@@ -298,6 +298,31 @@ export default function createStyledComponent(target: Target, options: Object, r
   // $FlowFixMe
   WrappedStyledComponent.target = isTargetStyledComp ? target.target : target;
 
+  // fold defaultProps
+  // $FlowFixMe
+  WrappedStyledComponent._defaultProps = EMPTY_OBJECT;
+  // $FlowFixMe
+  Object.defineProperty(WrappedStyledComponent, 'defaultProps', {
```
probablyup: Why do we need to use a setter here? Doing the merge once at construction time should be sufficient I think...

oomathias: Hey! Thanks for reviewing. 
That's what I did with my first implementation, but I had to merge twice
```js
const A = styled.div``
A.defaultProps = {}

const B = styled(A)`` // A's defaultProps at construction time
B.defaultProps = {}

<B />  // B's defaultProps at render time
```

I would also have to modify this: https://github.com/styled-components/styled-components/blob/e6f1b85726512cde163dd5fcdd7fb0f2439d4a09/packages/styled-components/src/models/StyledComponent.js#L125

I can totally code it without a setter, but I thought it was more complex to read.

probablyup: > I would also have to modify this

Why? If you do the merge once at construction time (during `createStyledComponent`), the new defaultProps will be correct for when `determineTheme` is used in that component. Just set defaultProps on `WrappedStyledComponent`.

oomathias: Yeah sorry the problem was elsewhere, here are the problems I was facing: 

1) I don't think it's possible to fold directly on `WrappedStyledComponent.defaultProps`.
  ```js
  // Folding defaultProps' target to WrappedStyledComponent.defaultProps at construction time
  const A = styled(X)``
  A.defaultProps = { color: 'red' }
  // WrappedStyledComponent.defaultProps is now  { color: 'red' } and we lost everything.
  // WrappedStyledComponent.defaultProps is A.defaultProps
  ```
Easily fixable by using another static property to preserve the context (that I'm going to call _defaultProps for the example)

2) Like I wrote there https://github.com/styled-components/styled-components/pull/2260#discussion_r267719350 I was merging at construction time and at render time. 

3) The main difficulty appears at render time, https://github.com/styled-components/styled-components/blob/e6f1b85726512cde163dd5fcdd7fb0f2439d4a09/packages/styled-components/src/models/StyledComponent.js#L279

`defaultProps` are already applied by React, we have access to them (but are not merged yet), _defaultProps (merged except the last defaultProps), props (is a mix of defaultProps and props), and we still need to merge defaultProps (including style) and reapply them. 

Example:
```js
const A = styled.div``
A.defaultProps = { style: { color: 'red' } }
<A /> // expected: color: red

const B = styled(A)``
B.defaultProps = { style: { backgroundColor: 'red' } }
// Case B1
<B /> // expected: color: red and backgroundColor: red
// Case B2
<B style={{ color: blue }} />  // expected: color: blue
````
```js
// At render time
B1 => B1._defaultProps == { style: { color: 'red' } }
   => B1.defaultProps == { style: { backgroundColor: 'red' } }
   => B1.props == { style: { backgroundColor: 'red' } }
result expected => { style: { backgroundColor: 'red', color: 'red' } }

B2 => B2._defaultProps == { style: { color: 'red' } }
   => B2.defaultProps == { style: { backgroundColor: 'red' } }
   => B2.props == { style: { color: 'blue' } }
result expected => { style: { color: 'blue' } }
```

By using a setter and merging defaultProps before the rendering, everything works as expected. Maybe I missed something...

probablyup:
Going to take another look at this over the weekend, sorry for the delay.

probablyup:
Superseded by https://github.com/styled-components/styled-components/pull/2597

PR #2265, opened 2018-12-04
Fix hoisting of new react features and statics between forwardRef

eps1lon:
- fixes hoisting of `contextType` and `getDerivedStateFromError`
- fixes hoisting of `defaultProps`, `displayName` and `propTypes` between two `React.forwardRef` components

`hoist-non-react-statics` was previously inlined in #2021 with bundle size as the main argument. However maintenance cost was not considered which is required for new react features and fixes to the existing functionality. 

Bundle size was also only considered in isolation it seems. However `hoist-non-react-statics` is a very popular package (included in `react-router`, `react-redux` and virtually any library with HOCs these days). This means that as soon as `styled-components` is used in combination with those packages the original PR would add bundle size because the bundle would basically contain 2 versions.

The actual difference in bundle size might be close to zero now with [mridgway/hoist-non-react-statics e89c7a6168edc19eeadb2d149e600b888e8b0446](https://github.com/mridgway/hoist-non-react-statics/commit/e89c7a6168edc19eeadb2d149e600b888e8b0446). Maybe even in favor of the dependency since computed property keys add transpilation overhead.

Maybe @probablyup could shed some light on how he measured bundle size difference so that we can see how big the difference would be now.


probablyup:
It wasn’t just bundle size, there were bugs when multiple versions of the
library made their way into bundles.

eps1lon:
How would that be possible? If your library requires a specific  version (e.g. 3.x) then the layout of node_modules should ensure that you won't use 2.x in the code of `styled-components`. 

There are issues if components decorated as styled-components are later consumed by HOCs that use outdated versions of hnrs since 2.x does not support forwardRef. Those wouldn't be fixed by inlining however.

Do you have a reproduction for that?

mxstbr:
See here: https://github.com/gatsbyjs/gatsby/issues/8224

eps1lon:
@mxstbr It looked like the root cause of that issue was that an (WRT to semver) incompatible version of `hoist-non-react-statics` (hnrs) was resolved by the build setup. 
```
├─ styled-components
│  └─ hoist-non-react-statics@3.1.0
├─ hoist-non-react-statics@2.5.5
```
But the version actually used within `styled-components` was still somehow `2.5.5`. If this was the case then the build setup should be fixed. Otherwise it would imply that you need to inline every dependency so that the build setup of other users does not use a different semver range.

Or did you inline it as a temporary hotfix? Considering that you are now behind on features I don't think that was a good idea.

mxstbr:
Yes, exactly. Duplicate versions, so we inlined it :wink:

@probablyup made that call, so I'll leave it up to him to decide whether we want to un-do it!

eps1lon:
Sounds like this should be resolved in the build setup of the consuming packages. Possible workarounds:

- [webpack alias](https://webpack.js.org/configuration/resolve/#resolve-alias)
- [babel-plugin-module-resolver](https://www.npmjs.com/package/babel-plugin-module-resolver)
- [yarn resolutions](https://yarnpkg.com/lang/en/docs/selective-version-resolutions/)

All of which are IMO easier accessible than patching code of a dependency (in this case `styled-components`).

probablyup:
I’d prefer to continue inlining the hoist. The bug we inlined it for was extremely difficult to identify and I’m not confident it has been resolved since. It’s easy enough to add a new method etc. now and then as React releases things.

eps1lon:
>  The bug we inlined it for was extremely difficult to identify and I’m not confident it has been resolved since. 

The bug was not with `styled-components`.

> It’s easy enough to add a new method etc. now and then as React releases things.

It's obviously not which is evident by the release of this bug and no action for over <s>a month</s> three weeks (are probably more fair due to holidays).

If you still insist on this "just because" I would suggest setting up a proper fork under the styled-components org and pull it in as a github dependency. The current approach is clearly not working mainly because there is not automated process that ensures feature parity between your fork and the original repo.

probablyup:
In the time you've spent trying to convince me that we should do a fork the update could have been finished. Please accept the decision and move on.

PR #2281, opened 2018-12-11
make use of already destructured values to get the styledComponentId constant

manpenaloza:
this PR aims to simplify code to generate the `styledComponentId` constant. it basically does 2 main things:
1. make usage of already destructured/(defaulted) `options` properties
2. remove some duplicate from the `else` block in that condition. as far as I can see this is not needed as the `componentId` already has a default value at that (run)time.

### DIFF
--- a/src/models/StyledComponent.js
+++ b/src/models/StyledComponent.js
@@ -239,6 +239,7 @@ class StyledComponent extends Component<*> {
 }
 
 export default function createStyledComponent(target: Target, options: Object, rules: RuleSet) {
+  
   const isTargetStyledComp = isStyledComponent(target);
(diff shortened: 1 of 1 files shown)


manpenaloza:
ok, this code change leads different class names compared to the current implementation that does not use the destructured property values. theoretically: shouldn't the value of `styledComponentId` be the same considering the modifications of this pr?

mxstbr:
🤔 Not sure to be completely honest!

probablyup:
I think this change is ok. Needs a rebase against master now and a changelog entry but the change is sound.

kitten:
Probably out of date; closing due to inactivity

PR #2285, opened 2018-12-13
Use mobile first techniques in media templates examples

jamesgorrie:
Given the example here:
https://zellwk.com/blog/how-to-write-mobile-first-css/

Should we not be promoting best practice by giving mobile first media queries?


mxstbr:
I think we should remove this section entirely, it's not very fitting with the rest of the docs as we don't recommend any abstractions anywhere apart from this one—and we don't even use it ourselves.

Let's just get rid of it!

jamesgorrie:
Do you mean `Media templates` or the whole tips and tricks?
As this whole file feels like suggestions on abstractions?

Out of interest - What technique do you use?

PR #2316, opened 2019-01-07
React native errors

olehreznichenko:
I spent a few hours when first time trying react-native with styled-component, it was an adaptation web version, and I don't know about these cases like not working keyframes and global styles. What I see an only error like `keyframes` is not a function I guess. I was not helpful. Maybe I wrong make installation or wrong import module. Hope this friendly errors will help for next-gen react-native devs :).
Chooses `throw new Error`, not the `console.warn`, because stack traces will help to find where exactly you import `keyframes` or `createGlobalStyle`
Also added peerDependencies react-native in `native/package.json`, I dont know will it work with instalation.

### DIFF
--- a/native/package.json
+++ b/native/package.json
@@ -3,5 +3,6 @@
   "private": true,
   "main": "./dist/styled-components.native.cjs.js",
   "jsnext:main": "./dist/styled-components.native.esm.js",
-  "module": "./dist/styled-components.native.esm.js"
+  "module": "./dist/styled-components.native.esm.js",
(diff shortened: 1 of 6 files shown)


probablyup:
> Also added peerDependencies react-native in native/package.json, I dont know will it work with instalation.

That part won't do anything, as that package.json is only used as an entry for module resolution, not for installation. You could add it to `optionalPeerDependencies` instead in the main package.

kitten:
No reaction to an older comment; closing due to staleness

PR #2321, opened 2019-01-14
Make the GlobalStyleComponent call the base constructor with props.

jaydenseric:
This PR ensures the `GlobalStyleComponent` created by `createGlobalStyle` calls the base constructor with `props`.

This is a React best practice for class components:

> Class components should always call the base constructor with `props`.
> — https://reactjs.org/docs/state-and-lifecycle.html#adding-local-state-to-a-class

Not doing so can cause issues with `this.props` being undefined when rendered by libraries that are not very fault tolerant.

React's `renderToString` accounts for it here: https://github.com/facebook/react/blob/v16.7.0/packages/react-dom/src/server/ReactPartialRenderer.js#L532

Apollo's `getDataFromTree` used to account for it here: https://github.com/apollographql/react-apollo/blob/v2.2.4/src/getDataFromTree.ts#L80

Fixes https://github.com/jaydenseric/graphql-react/issues/17 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 ## Unreleased
 
-- Make the `GlobalStyleComponent` call the base constructor with `props`.
+- Make the `GlobalStyleComponent` created by `createGlobalStyle` call the base constructor with `props`.
 
 ## [v4.1.2] - 2018-11-28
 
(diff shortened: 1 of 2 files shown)


jaydenseric:
Flow types gave me hell, both in setting up VSCode to make this PR, and in trying to learn how to add types for `props`. Will the way I have done it allow people to use the component with no props at all?

probablyup:
> Flow types gave me hell, both in setting up VSCode to make this PR, and in trying to learn how to add types for props. Will the way I have done it allow people to use the component with no props at all?

We don't export our flow types, so don't worry about that. They'll actually be removed from the lib in general probably for v5.

Issue #2227, opened 2018-11-18
Importing fonts inside global styles cases re-importing fonts on every re-render


SilencerWeb:
## Environment

```
## System:
 - OS: macOS 10.14.1
 - CPU: (4) x64 Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
 - Memory: 330.83 MB / 8.00 GB
 - Shell: 5.3 - /bin/zsh
## Binaries:
 - Node: 11.0.0 - /usr/local/bin/node
 - Yarn: 1.10.1 - /usr/local/bin/yarn
 - npm: 6.4.1 - /usr/local/bin/npm
 - Watchman: 4.9.0 - /usr/local/bin/watchman
## npmPackages:
 - styled-components: ^4.1.1 => 4.1.1
```

## Reproduction
https://github.com/SilencerWeb/eslint-config-website

## Steps to reproduce
Create file .env with this line - `NODE_PATH="src/"`, then do the following:
```
git clone https://github.com/SilencerWeb/eslint-config-website
git checkout 5043b2e272b6f
npm i
npm run start
```

## Expected Behavior
On re-rendering the elements, components should update without blinking.

## Actual Behavior
On re-rendering the elements, components blink.

I'm sure it is because of new global styles API that was presented in 4th version of styled-components. I had the same issue with my app on electron and react but I thought it was next.js issue, not styled-components, I had to downgrade to 3th version and use old global styles API. In the current case, I had to move all my fonts to public folder - https://github.com/SilencerWeb/eslint-config-website/commit/7cd1722a816669b1a6d2642475cb11e4c7a77721. Please, check in Chrome, Opera or Safari - in these browser this bug exists for sure.

probablyup:
This is only in dev right? In production you definitely should not see blinking. Dev tends to be rougher since we're using the old text node way of injecting styles, which causes a bigger restyle/relayout.

SilencerWeb:
@probablyup unfortunately, in production too :(

danbopes:
I'm having similar issues with this. When my page initially loads, I don't mind a small flicker as the fonts load in. The problem is, when I create another element on the page (In my case, I was adding a dropdown with some styled-components), it caused a re-render of the globalStyles which then caused a flicker as my fonts attempted to download again.

probablyup:
I'll try and look into this sometime during the week.

egoarka:
[...]

PR #2327, opened 2019-01-15
Seperate global style

doniyor2109:
It should solve #2227

I have created new separate GlobalStyleSheet so that it could insert global styles into separate stylesheet.

GlobalStyleSheet is same as usual StyleSheet but it contains its own `GLOBAL_SC_ATTR` attribute


doniyor2109:
@mxstbr That is better solution. Thanks for the feedback. Then I will close the PR and try to use your solution.

mxstbr:
Amazing, I'd love to see that! :100: Looking forward

PR #2333, opened 2019-01-17
#2080 forgot to replace \n to '' in className

manglobe:
https://github.com/styled-components/styled-components/pull/2080

PR #2338, opened 2019-01-18
add test for #2333

manglobe:
Here's a new problem:
At first, I think it's an easy test, just copy the #2080's test and rewrite several words.
But it not works?!
`styled.div` used for two times will break the test. 
I can't understand why, can anyone help me?
And why `renderIntoDocument(components)` return null while components created by styled-components?

Here's my local test result , if I copy the same code, it failed


probablyup:
@manglobe can you rebase this one?

PR #2344, opened 2019-01-21
Move extra rules to separate style tag

doniyor2109:
PR for #2227.

Created extra rules for `@import` and `@font-face` css rules so that each of them is inserted into separate style tag.


probablyup:
I'm open to this, but I think the work should be moved over into #2357 since it's a total rewrite of the stylesheet implementation

.gitignore
```diff
@@ -2,4 +2,5 @@ node_modules
 dist
 .DS_Store
 *.log
-package-lock.json
\ No newline at end of file
+package-lock.json
+.idea
```
d7my11: Shouldn't we keep `.gitignore` for project-related files/directories only and move others (ide-related, system-related..) to our own global gitignore file?

doniyor2109: I did not know about global gitignore. Thanks for feedback.

doniyor2109:
@probablyup I could not migrate these changes to new rewrite. I did not understand how new rewrite works in general and that is why it was difficult for me

probablyup:
@kitten what's your take on this PR?

kitten:
@probablyup we do have to implement this for import at-rules in the v5 sheet, but I'm not sure whether we still want to add the extra complexity for font-face. It really depends on how complex the implementation will turn out.

The problem where all assets reload if cache headers are missing, seems to be wide ranging in Chrome and includes images and all other assets, so it might not make sense to extract everything as it's something that should be solved on the user's end.

So I was thinking of just reordering rules since that means we can continue to use a single style tag in v5. We do however then need to add a docs entry on this issue

IljaDaderko:
I'm on v5 rc2 at the moment, and font-face issue appears on chrome and safari (mac-os) only during first re-render.

kitten:
Closing due to age, inactivity, missing reproduction, or lack of details.

Issue #2342, opened 2019-01-21
Use hooks


mxstbr:
We should use React hooks once they land in a stable release! Here are a couple first ideas:

- [ ] Rework theming system to be based on `useContext` to avoid DevTools bloat with `ThemeConsumer` components
- [ ] Implement `useTheme` hook (see #2340)
  - [ ] ~~Maybe remove `withTheme` HOC in favor of the hook?~~
- [ ] Rework StyleSheet infrastructure to use hooks

Any other ideas for how we could leverage hooks?

probablyup:
Can you add this to the v5 rollup ticket?

probablyup:
> maybe remove withTheme HOC in favor of the hook?

I'm 👎 on this because there are use cases where you need the theme in bound functions and that flow becomes awkward without dependency injection.

TrySound:
Hooks work fine with flow.

probablyup:
We could probably introduce a `useStyles` hook that injects immediately once called and returns a className. And we could internally do a `useEffect` thing to automatically clean it up on unmount of the component in the stylesheet.

probablyup:
Something like:

```js
const className = useStyle`
  color: ${theme.colors.red};
`;
```

Is that possible, using a TTL as a hook?

TrySound:
I thought styled components are against classes. How are you gonna implement SSR with this?

imbhargav5:
@probablyup Exactly what I had in mind.

TrySound:
Emotion introduced jsx pragma to solve SSR problem.

probablyup:
> I thought styled components are against classes.

What do you mean? Like React classes? Ideally we'd like to use lighter representations in the library for things, but we have no overall objection to them.

mxstbr:
> We could probably introduce a useStyle hook that injects immediately once called and returns a className.

I don't really like this, that would fit better in a separate / complementary library based on the same (new) core StyleSheet infra to share code. Maybe `styled-hook`?

probablyup:
I guess so? Would be pretty easy to add and just another way to do things.

mxstbr:
Yeah but it does not fit well into the "never see another class again"-thing that the entire styled-components API has right now. I would much rather this (seemingly lower-level) hook live in another package, especially once we land the monorepo refactor (#2326) and the new StyleSheet infra (which will also be a separate package) it should be easy.

probablyup:
[...]

PR #2348, opened 2019-01-24
refactor: use StyleSheet.master as default value in StyleSheetContext

alexandernanberg:
I think it makes sense to use the `StyleSheet.master` as the default value in the `StyleSheetContext`. Removes some unnecessary (I think) logic in `createGlobalStyle` and `StyledComponent`. This will also make the Hooks migration (#2342) a bit simpler.

I thought this would be a pretty straight forward change – but a lot of the tests broke and I'm not 100% sure why. Could it be the `master` getter in `StyleSheet.js` that sets the `master` variable right away instead of when it's needed?

### DIFF
--- a/packages/styled-components/src/constructors/createGlobalStyle.js
+++ b/packages/styled-components/src/constructors/createGlobalStyle.js
@@ -81,8 +81,8 @@ export default function createGlobalStyle(
 
       return (
         <StyleSheetConsumer>
-          {(styleSheet?: StyleSheet) => {
-            this.styleSheet = styleSheet || StyleSheet.master;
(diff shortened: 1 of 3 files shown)


mxstbr:
The `StyleSheetManager` is only used when using the `ThemeProvider` iirc, which we do not require and lots of components (and our tests) do not expect the app to be wrapped in anything.

Every styled components need to be able to be rendered stand-alone, which is why we have to default in them instead of the manager, if that makes sense?

alexandernanberg:
Ah I think I get it! Closing since I don't think there is any way to change this.

PR #2349, opened 2019-01-24
[WIP/Exploration] Hooks

alexandernanberg:
Some early exploration of react hooks 

Related #2342


alexandernanberg:
Absolutely – once I get the sandbox working with these changes 😬 Right now I get this error `Hooks can only be called inside the body of a function component.` when trying to SSR, will investigate further!

mxstbr:
Maybe they aren't supported in SSR?

alexandernanberg:
I'm pretty certain they are 100% supported in SSR, could be a bug in their end or something weird we're doing... hooks works fine in `ThemeProvider` but not in `StyleSheetProvider` for some reason

gaearon:
You probably have two Reacts in node modules.

alexandernanberg:
Ah I finally figured out how to fix the sandbox error! I think there are some issues when symlinking the react dependency (like with lerna), but in this case it was possible to solve with a webpack alias.

alexandernanberg:
@mxstbr I was able to get some screenshots now 😏 

Before (the entire Hero component won't even fit on my screen 😅 )
image

After 😍 
image

mxstbr:
Amazing 😍😍😍

Jessidhia:
If the `StyledComponent` class is itself ported to being a `forwardRef` body component, this could become even shallower :fire:

probablyup:
@alexandernanberg excellent work! have you had a chance to look at the test failures?

alexandernanberg:
@probablyup Thanks! Yeah but I've been unable to come up with a solution... there seems to be an issue with `jest.resetModules()` that causes react to throw the `Hooks can only be called inside the body of a function component.` error. 

I tried with Jest's `moduleNameMapper` as suggested in [this comment](https://github.com/facebook/react/issues/13991#issuecomment-463486871) but that didn't work either 🤔

Jessidhia:
I did solve it in my PR (#2390) but it's not very pretty. It's something to do with how `react` communicates with its reconciler.

alexandernanberg:
Ah nice, copied that fix here as well!

alexandernanberg:
I'm going to close this since #2390 supersedes it and has a better implementation 🎉

PR #2356, opened 2019-01-28
[WIP] feat: upgrade babel to v7

alexandernanberg:
Also upgrades jest to v24 and react-native to 0.58 since they have better babel ^7 compatibility.

Still WIP but I haven't noticed any bundlesize regression yet (which was AFAIK the reason we haven't upgraded yet). All test but the macro ones are passing atm, will need to take a look at them later.


probablyup:
Supeceded by #2509

Issue #2120, opened 2018-10-18
Investigate ServerStyleSheet construction performance


mxstbr:
It seems like `new ServerStyleSheet()` is our biggest SSR perf bottleneck according to the SSR benchmarks added in #2119, which in turn makes it seem like `StyleSheet.clone()` is the actual bottleneck.

Since as of v4 everything is in the component-context, I'm not 100% certain why we need to clone anything anymore on the server. It seems as though creating a cache of styles per each request should be a cheap operation since nothing should be shared anymore between requests?

Going to defer to @kitten's expertise since he wrote most of that—can we maybe speed up or get rid of the cloning entirely?

mxstbr:
I just tried to "stupidly" remove cloning (by making the `clone()` method only `return new StyleSheet()`) and it broke quite a few tests, but I can't quite tell why. Many of the test failures look like this:

```diff
- .sc-a {} .c{color: red;}</style>"
+ .c{color:red;}</style>"
```

Why is the `.sc-a {}` missing?

kitten:
@mxstbr The `.sc-a {}` is injected as part of the deferred injection. It's also what determines the eventual order.

### Current StyleSheet — deferred injection

Basically our current model keeps **order of first injection** per component. When new rules are then injected they're appended to first rule the component injected. However, we care about **order of creation** for components instead, so we want to inject a placeholder as soon as the component is created. Now, because we don't want to immediately inject rules we have deferred injection to "fake insert" as first rules. That's when `.sc-a {}` is also injected.

When we clone a `StyleSheet` (typically `master` to a `ServerStyleSheet`) the `master` has all the deferred rules and the "child" has all the rest. Since we want to preserve the order we've established we copy all of that information over by "cloning".

So what you're seeing is missing "placeholder" rules and component injection order. You should in that change-set see React-render-order instead.

That's 100% @geelen's genius system, which has just been refactored in v3 (?) :100:

### Proposed system — injection by indices

(Hey, this sounds like `z-index`?! Indeed we'll keep track of the order of rules with a number)

To also capture what my proposal is, we can capture the **creation order** with a simple counter. If we then get to `render` we already have a `StyleSheet` from context with fallback to `master` (completely universal, yay). In `render` we can then instruct the `StyleSheet` to inject the component's rendered rules and also give it the **creation index**.

The `StyleSheet` keeps a list of all _creation indices_ and the number of rules that have been inserted for them. (This means that all these numbers added up are equal to the number of inserted rules)

[...]

PR #2357, opened 2019-01-28
[WIP] styled-sheet

mxstbr:
### Summary

This is a complete rewrite and modernisation of the core StyleSheet to speed up SSR performance and be much smaller bundle size-wise.

### Background

In styled-components v3 and below we had to clone styles (`injectGlobal`, `keyframes`) from the global "master" StyleSheet to the request-specific StyleSheet (`new ServerStyleSheet`) for every server-side rendering request. That is an expensive operation and a big bottleneck for SSR performance.

With the new `createGlobalStyle` component and lazy injection of `keyframes` in v4 all styles injection happens at render-time, i.e. on every request. That means we do not have to clone any styles anymore! If we rewrite our StyleSheet to avoid cloning we should be able to get a big speedup in server-side rendering performance, as well as removing a lot of complexity and thus reducing the bundle size!

<details>
  <summary>Click for a more detailed technical explanation of the proposed changes</summary>

<br />

> This was taken from @philpl's comment in #2120.

## Current StyleSheet — deferred injection

Our current model keeps **order of first injection** per component. When new rules are then injected they're appended to the first rule the component injected. However, we care about **order of creation** for components instead, so we want to inject a placeholder as soon as the component is created. Now, because we don't want to immediately inject rules we have deferred injection to "fake insert" as first rules. That's when `.sc-a {}` is also injected.

When we clone a `StyleSheet` (typically `master` to a `ServerStyleSheet`) the `master` has all the deferred rules and the "child" has all the rest. Since we want to preserve the order we've established we copy all of that information over by "cloning".

## Proposed system — injection by indices

We can capture the **creation order** of components with a simple counter. If we then get to `render` we already have a `StyleSheet` from context with fallback to `master` (completely universal, yay). In `render` we can then instruct the `StyleSheet` to inject the component's rendered rules and also give it the **creation index**.

The `StyleSheet` keeps a list of all _creation indices_ and the number of rules that have been inserted for them. (This means that all these numbers added up are equal to the number of inserted rules)

We can then take the **creation index** and walk the complete list of _creation indices_ until either the next index is greater or the current one is equal. All of the _number of rules_ in the list up until the matching index can then be added up. What we get is the **injection index**, which can be passed to `sheet.insertRule` :tada:

### Babel-less SSR

Another problem we tried to solve is Babel-less SSR. How can we ensure consistent classes, ordering, and unique names **without** the Babel plugin?

The problem we need to solve is that there has to be a _“single consistent moment”_ between the result of SSR and the rehydration where the client renders the _same result_. What we're trying to achieve is consistency between the two. We can't use our “best guess displayName” since it's ordered by creation, which might be different on the server and the client.

Essentially we can keep another counter for the **render order**. Since in SSR the rendered tree is enforced to be the same on the client we can count upwards for rendered `StyledComponent`s. The first time we encounter a component _while rendering_ we can give it a number to use as its `componentId` instead of a “best guess“ name.

### Concretely

- Each component has an [order index] that determines the injection order for a group of CSS rules ([rule group])
- Each component injects any number of CSS rules into its [rule group]
- Hence the [rule groups][rule group] must stay ordered and grouped
- [rule group]s can also grow when more CSS rules are injected into it

#### Component Creation

This is the moment when the user creates a `StyledComponent`.

- An [order index] is being assigned to the component
- The [order index] is created by incrementing a global counter

#### Rule injection

This is when a `StyledComponent` is being rendered

- An amount of CSS rules have been processed and returned by stylis
- The component calls `RuleSheet#insertRules` using CSS rules and its [order index]
- The index of the underlying `CSSStyleSheet#insertRule` needs to be determined
- This can be determined by summing up the sizes of all [rule groups][rule group] for which their [order index] `<=` the passed [order index]

When an optimisation on the above operation is made we thus need to take into account:

- How do we determine the [order index] relative to all others efficiently?
- How do we sum up the number of rules left of the current [rule group] efficiently?

> To be answered

#### SSR

To restore the correct ordering the output CSS needs to be rehydrated. This is basically a process
of marshalling (during SSR) and unmarshalling (on the client).

The [order index] of all components must be restored. This can be done by having a mapping
of [order index] to a [component ID]. This [component ID] can be replaced with the [render order index]
which means that the "final" [order index] can only be determined during rendering.

Subsequently when a component is unknown its new [order index] must be larger than the largest [order index]
that SSR has generated.

Apart from this process everything remains the same.

#### Summary

1. SSR maps each [component ID] to an [order index] (Rehydration retrieves this information)
2. If the SSR result is "spread out" due to streaming, all rules are migrated to a new `style` tag in the `head`
3. An incremental [order index] is assigned to each created `StyledComponent`
4. During rendering we use the [component ID] or capture the [render order index] as a fallback
5. If the [component ID] matches to a rehydrated [order index] it replaces the component's own [order index] (see 3.)
6. CSS rules are generated
7. The [order index] is used to insert the new rules into the [rule group] (See [Rule injection](#rule-injection))

As an optimisation the CSS rules' hash (which is also their `className`) can be looked up in a `Map` to avoid
duplicate injection.

#### Glossary

- <a name="order-index"><strong>order index:</strong></a> This is a number that assigns a _priority_ order to a [rule group]. Rules in such a group are meant to be injected in ascending order of their index.
- <a name="rule-group"><strong>rule group:</strong></a> This is a group of CSS rules that can never be separated and can grow as components append new rules to it.
- <a name="component-id"><strong>component ID:</strong></a> This is a unique identifier for a component that is important to match information to it between SSR and rehydration.
- <a name="render-order-index"><strong>render order index:</strong></a> This is a number that signifies in which order a given component rendered to all other tracked components, which can be used to replace [component ID]s for SSR.

[order index]: #order-index
[render order index]: #render-order-index
[rule group]: #rule-group
[component ID]: #component-id

</details>

<br />

### TODO

- [ ] Step 1: Build the new StyleSheet in a new `styled-sheet` package
  - [x] Create new package in the monorepo
  - [x] Build first implementation as test bed
  - [ ] Figure out API
  - [ ] Move to classes-as-order-markers instead of comments-as-order-markers
  - [ ] Add tons of unit tests
- [ ] Step 2: Integrate the new `styled-sheet` into the core library
  - [ ] Rip out old StyleSheet
  - [ ] Add `orderIndex` to `StyledComponent.js` that counts up for each created component
  - [ ] Add same `orderIndex` to `keyframes` and `createGlobalStyles`
  - [ ] Output the `orderIndex` in the `styledComponentId` comment during SSR
  - [ ] Implement SSR rehydration based on the `orderIndex` in the comment
  - [ ] Replace `styledComponentId` with `renderIndex`
- [ ] Step 3: Publish alpha release
  - [ ] Test in production apps (Spectrum)
- [ ] Step 4: Publish public beta release
  - [ ] Announce and get people to submit bugs
- [ ] Step 5: Release styled-components v5
- [ ] Step 6: Implement rehydration from `<link>` tags for "static" style extraction

**If you want to get involved and help out with this please comment below!** We can use all the hands we can get, especially to write tests and to figure out the API.

*Ref issue #2120*


packages/styled-components/src/models/ServerStyleSheet.js
```diff
@@ -1,96 +1,57 @@
...
-          html += tag.toHTML(streamAttr);
-        }
-
-        /* force our StyleSheets to emit entirely new tags */
-        instance.sealAllTags();
-
-        /* prepend style html to chunk */
+        const html = sheet.toHTML();
+        // Force Sheet to "forget" about previously emitted CSS
+        sheet.reset();
```
mxstbr: Since there should be a new `sheet` on every request this seems like unnecessary work, shouldn't the entire sheet be garbage collected once the request is done?

kitten: This is the code inbetween chunks on the SSR stream. We need to ensure that previously emitted CSS is not emitted again. Since there's no "useful" information in a GroupedTag apart from just CSS, the Stylesheet can just throw it away and create a new one

packages/styled-sheet/src/Sheet.js
```diff
@@ -0,0 +1,96 @@
...
+import { RuleGroupTag, DefaultTag, VirtualTag } from './tags';
+import { makeCssMarker, wrapInStyleTag } from './utils';
+import GroupRegistry from './GroupRegistry';
+import rehydrate from './rehydrate';
+
+// Ensure we only rehydrate once
+let SHOULD_REHYDRATE = !IS_BROWSER;
+
+const makeRuleGroupTag = (target?: HTMLElement, forceServer?: boolean): GroupedTag => {
+  const tag = forceServer ? new DefaultTag(target) : new VirtualTag();
```
mxstbr: This condition is inverse, right? It should be

```suggestion
  const tag = !forceServer ? new DefaultTag(target) : new VirtualTag();
```

mxstbr:
I just added my first (working) test for the Sheet, yay! https://github.com/styled-components/styled-components/blob/084e07e4302b54ec21e61a38368499e588ce6d98/packages/styled-sheet/test/Sheet.test.js#L4-L13

Can you elaborate on the reason one has to call `GroupRegistry.registerRuleGroup('name')` to get a group index back? That seems like a weird API. Maybe something like this would be clearer?!

```JS
const group = new Group('name');
sheet.inject(group, 'key', []);
```

mxstbr:
For posterity so we do not forget: @kitten pointed that out that we could move to classes-as-order-markers instead of comments-as-order-markers, which would make it possible to rehydrate from a `<link>` tag. That would be useful for static pre-rendering a la Gatsby and react-static.

probablyup:
@mxstbr do a little benchmark around usage of `new` - it's a known perf impactor :/

kitten:
@probablyup we'll revert the decision around Group and go back to the GroupRegistry singleton. That being said there's less allocations now, even including this change which only happens while the StyledComponent and ComponentStyle are being created :) during updates there's shouldn't ever be an allocation in styled-sheet

PR #2365, opened 2019-01-31
Filter suppressClassNameWarning to not to pass down to wrapped components

### LINKED ISSUE #2346: Do not pass suppressClassNameWarning to wrapped components
### System:
 - OS: macOS 10.14
 - CPU: (4) x64 Intel(R) Core(TM) i5-7267U CPU @ 3.10GHz
 - Memory: 75.45 MB / 8.00 GB
 - Shell: 3.2.57 - /bin/bash
### Binaries:
 - Node: 8.15.0 - /usr/local/opt/node@8/bin/node
 - npm: 6.4.1 - /usr/local/opt/node@8/bin/npm
### npmPackages:
 - babel-plugin-styled-components: ^1.10.0 => 1.10.0

## Summary
* suppressClassNameWarning produces React warning
suppressClassNameWarning should either only be using lowercase letters instead of camelCase, or it should be filtered from being applied to the DOM. Otherwise, React complains.

(see https://github.com/styled-components/styled-components/pull/2156/files/fe66a4f9aae8f4736c8c6bc08df64e56d892844d)

### Steps to reproduce
* add suppressClassNameWarning as a property or attribute

### Expected Behavior
* styled-components warning is suppressed

### Actual Behavior
* styled-components warning is suppressed but then React gives warning:

Warning: React does not recognize the `suppressClassNameWarning` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `suppressclassnamewarning` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

taneba:
fixes #2346 

Once I tried the solution that @leantide [mentioned in the related issue](https://github.com/styled-components/styled-components/issues/2346#issuecomment-459407874), but it didn't work (I think it is because React likely freezes props). 
What I did Instead is to filter suppressClassNameWarning when passing props to child component.

### DIFF
--- a/packages/styled-components/src/models/StyledComponent.js
+++ b/packages/styled-components/src/models/StyledComponent.js
@@ -145,9 +145,10 @@ class StyledComponent extends Component<*> {
         this.warnInnerRef(displayName);
       }
 
-      if (key === 'forwardedComponent' || key === 'as') continue;
+      if (key === 'forwardedComponent' || key === 'as' || key === 'suppressClassNameWarning')
+        {continue;}
       else if (key === 'forwardedRef') propsForElement.ref = computedProps[key];
-      else if ((!isTargetTag && key !== 'suppressClassNameWarning') || validAttr(key)) {
+      else if (!isTargetTag || validAttr(key)) {
         // Don't pass through non HTML tags through to HTML elements
         propsForElement[key] = computedProps[key];
       }
--- a/packages/styled-components/src/models/StyledComponent.js
+++ b/packages/styled-components/src/models/StyledComponent.js
@@ -147,7 +147,7 @@ class StyledComponent extends Component<*> {
 
       if (key === 'forwardedComponent' || key === 'as') continue;
       else if (key === 'forwardedRef') propsForElement.ref = computedProps[key];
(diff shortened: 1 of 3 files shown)


taneba:
Updated! Thanks @mxstbr for taking a look!

leantide:
Is there a next step on this?

taneba:
I'm not sure. 
@mxstbr Is there anything left to do?

packages/styled-components/src/models/StyledComponent.js
```diff
@@ -147,7 +147,7 @@ class StyledComponent extends Component<*> {
 
       if (key === 'forwardedComponent' || key === 'as') continue;
       else if (key === 'forwardedRef') propsForElement.ref = computedProps[key];
-      else if (!isTargetTag || validAttr(key)) {
+      else if ((!isTargetTag && key !== 'suppressClassNameWarning') || validAttr(key)) {
```
probablyup: can you move this up to L148 where `forwardedComponent` and `as` are? It can just be skipped along with those known props

taneba: good point, done!

PR #2389, opened 2019-02-16
Add svg marker tag to domElements.js

ZvezdocheTag:
`TypeError: styled_components__WEBPACK_IMPORTED_MODULE_8__.default.marker is not a function`

https://developer.mozilla.org/en-US/docs/Web/SVG/Element/marker


probablyup:
Can you add a quick changelog entry please?

ZvezdocheTag:
Sure

PR #2390, opened 2019-02-17
Reimplement StyledComponent using hooks (continued off #2349)

Jessidhia:
This continues from #2349, but I think very little of it survived.

Is the dream of shallow trees here? 🔥 

image

---

This also included a bunch of infrastructure stuff because the flow version is so old it didn't have types for hooks, and I couldn't copy&paste the definitions either because it used new flow builtins. The infrastructure itself probably needs a revamp.


packages/styled-components/src/constructors/createGlobalStyle.js
```diff
@@ -24,98 +23,57 @@ export default function createGlobalStyle(
...
-    styleSheet: Object;
+  function GlobalStyleComponent(props: GlobalStyleComponentPropsType) {
+    const styleSheet = useStyleSheet();
+    const theme = useContext(ThemeContext);
 
-    static globalStyle = style;
+    if (process.env.NODE_ENV !== 'production' && React.Children.count(props.children)) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        `The global style component ${styledComponentId} was given child JSX. createGlobalStyle does not render children.`
```
Jessidhia: TBH the right thing to do here is to just delete `createGlobalStyle` (breaking this API again...) and add a `useGlobalStyle` instead.

packages/styled-components/src/models/StyledComponent.js
```diff
@@ -40,212 +47,240 @@ function generateId(_ComponentStyle: Function, _displayName: string, parentCompo
...
+  );
 
-            attr = attr(context);
+  // NOTE: this has to be called unconditionally due to the rules of hooks
+  // it will just do nothing if it's not an in-browser development build
+  // NOTE2: there is no (supported) way to know if the wrapped component actually can
+  // receive refs -- just passing refs will trigger warnings on any function component child :(
+  // -- this also means we can't keep doing this check unless StyledComponent is itself a class.
+  //
+  // const refToForward = useCheckClassNameUsage(
```
Jessidhia: This is essentially not implementable with hooks, unless we make a dummy intermediate component **just** for calling `findDOMNode` in it, which sounds... icky. This warning was icky to begin with, as it triggers `StrictMode` warnings.

packages/styled-components/src/test/constants.test.js
```diff
@@ -12,6 +9,8 @@ describe('constants', () => {
 
   describe('SC_ATTR', () => {
     function renderAndExpect(expectedAttr) {
+      const React = require('react');
+      const TestRenderer = require('react-test-renderer');
```
Jessidhia: The `jest.resetModules()` call would break things without re-requiring the packages as the `react` would still hold a reference to the dispatcher that was imported at the top level.

packages/styled-components/src/utils/useCheckClassNameUsage.js
```diff
@@ -0,0 +1,78 @@
...
+const didWarnAboutClassNameUsage = new Set();
+function useDevelopmentCheckClassNameUsage(
+  ref: Ref<*>,
+  target: Target,
+  generatedClassName: string,
+  suppressClassNameWarning: boolean
+) {
+  const ownRefDefault: { current: any } = useRef(null);
+  const ownRef = suppressClassNameWarning ? null : ownRefDefault;
+  const refToUse = useCombinedRefs(ref, ownRef);
```
Jessidhia: As I said earlier, I tried implementing this by using refs to the child component, but this would end up passing refs even to pure function components.

alexandernanberg:
Very nice work! A question though – will the display name in the devtools show the components name or just `styled.whatever`?

Eg

```js
const Title = styled.h1``
``` 

Would that display as `Title` or `styled.h1`?

Jessidhia:
It should work the same (that is something the babel plugin does, not the component). That said, I didn't actually test it. I'll check tomorrow 😴

Jessidhia:
Here it is, running on my real work code 😎 

image

probablyup:
TODO: make a v5 branch prior to merging

probablyup:
Just a note - I have not forgotten about this! Working to get #2413 #2456 #2427 in to cut a final 4.x minor before starting the new 5.x develop branch

packages/styled-components/src/models/StyledComponent.js
wmertens: plus, it has false positives, like for react-tooltip

probablyup: Yeah just delete it

packages/styled-components/src/constructors/createGlobalStyle.js
probablyup: You can't currently use hooks in class components, so i'd rather not.

kitten:
I'm merging this into the `canary` branch, which will be used to hold all major v5 changes and will also soon be published under the `@canary` tag on npm once it's ready.

mxstbr:
Thank you so much for working on this @Jessidhia!!!

PR #2392, opened 2019-02-17
Add support for shouldForwardProp

stevesims:
Adds support for use of `shouldForwardProp` config method, which can be used to explicitly determine which props should be forwarded on.

If this config is not provided then existing functionality is preserved (i.e. all props passed to components, or for HTML elements check using `validAttr`).

Inspired by the same feature in emotion.

This feature enables the ability to filter props that would otherwise get passed through to other components, which in turn may get passed through to the DOM.

### DIFF
--- a/packages/styled-components/src/models/StyledComponent.js
+++ b/packages/styled-components/src/models/StyledComponent.js
@@ -113,6 +113,7 @@ class StyledComponent extends Component<*> {
       defaultProps,
       displayName,
       foldedComponentIds,
+      shouldForwardProp,
       styledComponentId,
       target,
     } = this.props.forwardedComponent;
@@ -138,6 +139,16 @@ class StyledComponent extends Component<*> {
     const propsForElement = {};
     const computedProps = { ...this.attrs, ...this.props };
 
+    // use explicit `shouldForwardProp` function if one is set, otherwise
+    // if no explicit `shouldForwardProp` set then
+    //   if we're on a tag, use `validAttr`
+    //     i.e. Don't pass through non HTML tags through to HTML elements
+    // else (we're on a component - pass throug all by default)
+    //   () => true
+    const shouldReallyForwardProp = shouldForwardProp || (
+      isTargetTag ? validAttr : () => true
+    );
+
     let key;
     // eslint-disable-next-line guard-for-in
     for (key in computedProps) {
@@ -147,8 +158,7 @@ class StyledComponent extends Component<*> {
 
       if (key === 'forwardedComponent' || key === 'as') continue;
       else if (key === 'forwardedRef') propsForElement.ref = computedProps[key];
-      else if (!isTargetTag || validAttr(key)) {
-        // Don't pass through non HTML tags through to HTML elements
+      else if (shouldReallyForwardProp(key)) {
         propsForElement[key] = computedProps[key];
       }
     }
@@ -238,6 +248,20 @@ class StyledComponent extends Component<*> {
   }
 }
 
+function getShouldForwardProp(options, target, isTargetStyledComp) {
+  // $FlowFixMe
+  if (isTargetStyledComp && target.shouldForwardProp) {
+    if (options.shouldForwardProp) {
+      // compose nested shouldForwardProp calls
+      // $FlowFixMe
+      return prop => target.shouldForwardProp(prop) && options.shouldForwardProp(prop);
+    }
+    return target.shouldForwardProp;
+  }
+
+  return options.shouldForwardProp;
+}
+
 export default function createStyledComponent(target: Target, options: Object, rules: RuleSet) {
   const isTargetStyledComp = isStyledComponent(target);
   const isClass = !isTag(target);
@@ -284,6 +308,8 @@ export default function createStyledComponent(target: Target, options: Object, r
   // $FlowFixMe
   WrappedStyledComponent.componentStyle = componentStyle;
   WrappedStyledComponent.displayName = displayName;
+  // $FlowFixMe
+  WrappedStyledComponent.shouldForwardProp = getShouldForwardProp(options, target, isTargetStyledComp);
 
   // $FlowFixMe
   WrappedStyledComponent.foldedComponentIds = isTargetStyledComp
@@ -331,6 +357,7 @@ export default function createStyledComponent(target: Target, options: Object, r
       componentStyle: true,
       displayName: true,
       foldedComponentIds: true,
+      shouldForwardProp: true,
       styledComponentId: true,
       target: true,
       withComponent: true,
--- a/packages/styled-components/src/test/__snapshots__/props.test.js.snap
+++ b/packages/styled-components/src/test/__snapshots__/props.test.js.snap
@@ -1,0 +1,14 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`props shouldForwardProp should inherit shouldForwardProp for wrapped styled components 1`] = `
+Array [
+  <div
+    className="sc-a c"
+    id="test-1"
+  />,
+  <div
+    className="sc-a sc-b d"
+    id="test-2"
+  />,
+]
+`;
--- a/packages/styled-components/src/test/props.test.js
+++ b/packages/styled-components/src/test/props.test.js
@@ -1,5 +1,5 @@
 // @flow
-import React from 'react';
+import React, { Fragment } from 'react';
 import TestRenderer from 'react-test-renderer';
 
 import { resetStyled, expectCSSMatches } from './utils';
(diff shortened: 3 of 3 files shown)


stevesims:
We have an example where this would be useful in our component library.

We're router agnostic, so our `Link` component, which accepts several props to affect styling, styles an `a` tag, and in-use one tends to use `as={RouterLink}`.  This works, but one ends up with the styling props rendered into the DOM.

The FAQs suggest using an interim component to remove props, however that technique is not compatible with the use of `as` (as the provided component replaces the version that filters out props) and thus cannot be router agnostic.

Emotion solves this problem by allowing a `shouldForwardProp` config function to be provided when defining a styled component.

With `shouldForwardProp` you'd define the component as something like this:
```jsx
const Link = styled('a').withConfig({
  shouldForwardProp: prop => !['fancy', 'colorful'].includes(prop)
})`[styles go here]`
```

and then to use it:
```jsx
<Link as={RouterLink} to="destination" fancy="yes" colorful>Link to destination</Link>
```

penx:
Fixes #439

stevesims:
Hey @probablyup any chance you could take a look at this?

Thanks in advance

mohsinulhaq:
Hi, any update on this? This is a very basic feature from emotion missing in styled-components.

massile:
Would love to see this !

probablyup:
As I've said a bunch of times, I'm 👎 on it

parkroolucas:
@probablyup I'm sorry if I'm OOTL here, but what is the suggestion you propose as an alternative to this?

probablyup:
The only solution that makes sense and works for multiple levels and types of HOC is [namespacing in JSX](https://github.com/facebook/jsx/issues/81). We can achieve this via babel plugin if someone wants to take a crack at it.

penx:
@probablyup please could you link to where this has been discussed previously for those of us not as familiar with all the discussions here? There are 160+ comments on  #439 and I don't see anything from you on why this isn't desired.

penx:
for context, we recently moved from emotion to styled components and this is the one thing we miss. I'm not sure what the downside is to adding this to the API, even if another solution could be found in the longer term

probablyup:
https://github.com/styled-components/styled-components/pull/2093 and the constellation of linked tickets in the OP

penx:
I can see a lot of discussion around having a convention for prop names/prefixes, but I can't see any argument against also having a filter function as an option in the config. As above, no-one has mentioned shouldForwardProp outside of this PR and #439 so it's hard to find any objections to it!

If you're aware of any specific posts/discussion against this that I've missed please let me know.

The benefits are:

- alignment with emotion API
- something that for many use cases/users (including us) will work now

diondiondion:
I'd also be grateful to see some practical examples of use cases where `shouldForwardProp` (or a similar solution) doesn't work (or "doesn't make sense"). Just trying to understand the thinking behind this, as I can't imagine JSX namespaces or prop naming conventions to be a user/developer-friendly solution at all. They remind me of the previously common `innerRef` pattern that devs don't have to resort to anymore now that `React.forwardRef` exists.

probablyup:
It immediately falls apart any time two HOCs could consume the same prop.

ancestorak:
@probablyup That's an intrinsic problem with HOCs, though. I don't think it's relevant for the kinds of problems this PR is trying to solve.

stevesims:
I must admit that I'm struggling to imagine an example of using two HOCs that both want to consume the same prop.

My usual expectation when using an HOC is that it will consume the props it is interested in, and not pass those props through.  So if I have a component that's wrapped in 2 HOCs I'd typically expect filtering to happen.

Besides which, as an HOC author, you're free to write your HOC as you see fit and document it accordingly - if you wanted to leave all props intact then you could set `shouldForwardProp: () => true`, or if you wanted the default styled-components behaviour you'd just not set that config at all.

probablyup:
@stevesims the semantic-ui + styled-components example has come up about a dozen times across all the issues

probablyup:
> That's an intrinsic problem with HOCs, though. I don't think it's relevant for the kinds of problems this PR is trying to solve.

Correct and I'd rather fix the problem at the source and introduce an overall pattern that works for all cases rather than implementing layers of hacky patches that don't work combinatorially.

diondiondion:
The `as` prop just represents another side of the same issue to me. For it (or similar props), styled-components needs an API to specify that a component should ignore & pass it on to the wrapped component.

With this & `shouldForwardProp`, the vast majority of prop-passing woes in styled-components would be solved.

penx:
As per 
https://github.com/styled-components/styled-components/issues/439#issuecomment-470590713

I don't think namespaced props would be a desirable resolution to the problem for us. In fact, for our use case, I'd consider it more hacky that shouldForwardProp as we'd be putting implementation details in to our prop names (though I'd need to more fully understand the proposed solution and how styled-components would use it).

For us, where we're providing a component library that has very few dependencies other than styled-components, we haven't encountered any scenarios where we're not in control of the prop names used by any HOCs that we're using (most HOCs we use are ones that we've written ourselves are part of our component library). It's hard to imagine a scenario where we'd encounter this.

Most of all, I'm in favour of aligning with the emotion API where there isn't a clear performance/maintenance impact and the API extension has clear use cases. I don't see how conflicting HOC prop names means that a shouldForwardProp API extension is hacky.

brabeji:
Ironically, "solving" problems with HOCs:

https://github.com/Jephuff/namespace-hoc#usage-to-namespace-a-portion-of-a-group-of-hoc

The abstraction just leaks.

ancestorak:
@probablyup You've closed this PR in favor of a proposal in an upstream project that a) probably won't come to fruition b) is not necessarily better as it comes with its own downsides. This seems premature to me. Please consider that while the problem prop conflicts is pretty rare, the issue of props being passed through to DOM is a common pain point for styled-components users.

penx:
A couple more use cases where I think an `sc` namespace would be a worse solution:

- a developer has a built and distributed a component using another library such as emotion and wants to port or convert it to styled-components, but keep the same API. They don't want to introduce `sc:`  prefix as this is an implementation detail, and they don't want their users to have to refactor their code if they choose to move to the styled-components version

- a developer has a component library built with styled-components. Their current API is a mix of unprefixed and `sc:` prefixed props. They want to port or convert it to another library, but this means either they can't keep their existing API or have to implement their own, clashing, `sc` namespace, negating the reason for having namespacing in the first place

penx:
I'd also be concerned that an `sc:` namespace conflicts with the desires of ISTF to allow styles to be distributed without a runtime, @kitten?

parkroolucas:
Read through #2093 and checked all the ideas that are there.

Currently React throws a whole bunch of (non-critical) errors in my app because custom props are being passed to the DOM level. I've grit my teeth and stuck with it for a while, because I expected an elegant solution to be implemented for what I saw (and still see) to be a relatively small issue with a trivial fix (emotion-style `shouldForwardProps` or `rejectProps` or similar)

Now that the `$` prefixing pattern has been introduced, do I have to go ahead and change all of my components to use this pattern? I would **really, really** much rather be able to perform an easy fix on the implementation details of my hundreds of styled components, as opposed to having to change the prop names on the >1000 times I am using these components across my projects.

diondiondion:
> Now that the $ prefixing pattern has been introduced

I don't think this PR was actually merged, it has been closed unresolved.

So, no, you don't have to go and change your props, but there's also still no official, non-hacky way to solve this.

Styled-components is clearly stalled in a limbo of "the perfect solution is the enemy of the good" when it comes to this issue. It's sad.

madiodio:
Namespacing props really isn't a scalable nor a reliable solution here when you're working with an UI library components or design systems which primarily use `styled-components`. So like many have mentioned here, this seems like a trivial feature to add that would solve a lot of pain of dealing with props which is the main selling of CSS in JS lib like SC and there is quite not a single valid explanation about the rejection of the solution presented in this PR. That's a bit sad.

kristojorg:
The issue seems tied to the nature of HOCs as a whole. What if we ditch them altogether in favor of composition:

```js
const Styler = style`
    color: red
`
const Button = () => (
    <Styler>
        <button>click</button>
    </Styler>
);
```

The `Styler` component clones it's children and passes down a classname, that's it. You now control exactly what props and what components are actually being rendered. No more issues with `as` props or whatever. 

Any thoughts on this pattern?

gcosgreave-us:
Is there any movement on this? It would be really good to be able to make a styled component and just pass it the custom props that are needed. Seems like an issue with styled components atm 🤷‍♂️

mikearche:
Guess I'll just have 1000 warnings then

Martinnord:
I'm wondering the same as @gcosgreave-us and prob +100 styled-components users. This should indeed be a basic feature that the lib provides. Especially when using styled-system, your DOM gets flooded with props.

Thanks.

penx:
reopened as #3006

PR #2413, opened 2019-02-24
Try to handle React's streaming renderer yielding at inopportune times

Dsan10s:
Added a test for interleaveWithNodeStream.  The niche case where the
style tag gets interleaved into a textarea causes the style tag to be
rendered as text in the text area.

Hello styled-components maintainers, thank you for this useful CSS-in-JS library!  It's been super useful for my projects so far

I have come across a very odd issue that I think is worth addressing

Very rarely, the chunks generated by react renderToNodeStream  will have one chunk end with the opening tag and contents of a textarea, and the next chunk will start with the closing tag of the textarea.  Combined with styled-components interleaveWithNodeStream, prepending the <style> tags to the chunk causes an issue specifically with textareas in this case.  For any other elements, this has not caused an issue, since a style tag as a child of another tag typically just inserts the style tag into the DOM.  However, textareas are an exception, as any children of a textarea will be interpreted as a string, and rendered as the initial textArea text, rather than being rendered as HTML node into the DOM.

I've created the following test to reproduce the issue, let me know if I can help further or provide further information.  Many thanks!


mxstbr:
Wow good catch, thanks for the reproduction @Dsan10s! I will defer to @probablyup on this one

Dsan10s:
Sounds good, thanks @mxstbr !

Dsan10s:
Hi guys, any updates on this?  Our app still has this issue :(

Let me know if there's any other way I can help out!

@mxstbr @probablyup

probablyup:
@mxstbr I don't know if this is necessarily the _right_ fix but I can't think of a better solution since you can't really introspect the rendered stream well

PR #2422, opened 2019-03-03
Readd production guard on flatten function error

mariel9999999:
This production guard was removed in https://github.com/styled-components/styled-components/commit/9a2ad9db4d50752e32ee4292758968c0c4de7d22#diff-1626f8a40e7d0e93acaa5ca5370598a4.

I'm not sure why that was the case but this is causing quite a bad performance hit in our medium/large sized application. Things seem to slow down quite a bit because of the new operator usage.

Is it possible to avoid the use of the new operator here in development as well? It causes a very noticeable delay when components are mounting (And this seems to be only noticeable in large sized applications that have a lot of dynamic styled components). This was not an issue in v3.


mariel9999999:
Closing this as I was unable to reproduce. Created an issue instead: https://github.com/styled-components/styled-components/issues/2424

PR #2427, opened 2019-03-05
Fix macro with ts

### LINKED ISSUE #2378: Babel macro conflicts w/ TS
#### System:
 - OS: macOS 10.14.1
 - CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
 - Memory: 163.83 MB / 16.00 GB
 - Shell: 5.0.2 - /usr/local/bin/bash
#### Binaries:
 - Node: 11.9.0 - /usr/local/bin/node
 - npm: 6.5.0 - /usr/local/bin/npm
#### npmPackages:

## Reproduction

Create a new CRA-based app w/ TypeScript. Add `@types/styled-components` & `styled-components`, and update App.tsx to look like the following:

```jsx
import React, { Component } from 'react';
import styled, { DefaultTheme } from 'styled-components/macro';
import logo from './logo.svg';
import './App.css';

const Link = styled.a``;

class App extends Component<{ theme: DefaultTheme }> {
  render() {
    return (
      <div className="App">
        <header className="App-header">
          <img src={logo} className="App-logo" alt="logo" />
          <p>
            Edit <code>src/App.tsx</code> and save to reload.
          </p>
          <Link
            className="App-link"
            href="https://reactjs.org"
            target="_blank"
            rel="noopener noreferrer"
          >
            Learn React
          </Link>
        </header>
      </div>
    );
  }
}

export default App;
```

Note that we're importing a type, `DefaultTheme`, from `styled-components/macro`, but otherwise, nothing special here.

## Steps to reproduce

Run `npm run build`.

## Expected Behavior

The build completes successfully.

## Actual Behavior

Build errors with the below:

```
Creating an optimized production build...
Failed to compile.

./src/components/views/EmployeeDetail/Charts/CompensationHistory.tsx
MacroError: Invalid import: DefaultTheme. You can only import default, css, keyframes, createGlobalStyle, isStyledComponent, ThemeConsumer, ThemeContext, ThemeProvider, withTheme, ServerStyleSheet, StyleSheetManager, __DO_NOT_USE_OR_YOU_WILL_BE_HAUNTED_BY_SPOOKY_GHOSTS from 'styled-components/macro'.
```

The error appears somewhat related to [this change](https://github.com/kentcdodds/babel-plugin-macros/pull/96) in babel-plugin-macros but it's also directly caused by the validation done by the macro [here](https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/macro/index.js#L17-L23).

My current theory is the macros PR changed the order the macros were being run, so the styled-components macro is being run against an AST that still contains imported types. The macro then finds an identifier it's not expecting (in this case, `DefaultTheme`) and bombs out.

Couple potential solutions (happy to PR any of them):

1. Remove the identifier validation completely.
2. Warn & skip if invalid identifier is found.
3. Silently skip if invalid identifier is found
4. Enable providing a list of "expected identifiers" that the macro should ignore.
5. Add TS types to "expected identifiers"  and skip.

mAAdhaTTah:
Fixes #2378. That issue will mistakenly closed, as the issue lives in the macro, not in the types. I figured providing a solution would help clarify the issue.

I also added a test for multiple imports, as I didn't expect to see them split in the test I wrote, so I wanted to confirm and test that behavior before I added my change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,10 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 - Compatibility with react-native-web 0.11 (see [#2453](https://github.com/styled-components/styled-components/issues/2453)).
 
 - Add stack trace to .attrs warning (see [#2486](https://github.com/styled-components/styled-components/issues/2486)).
+
 - Fix functions as values for object literals. (see [2489](https://github.com/styled-components/styled-components/pull/2489))
+
+- Remove validation in Babel macro, by [@mAAdhaTTah](http://github.com/mAAdhaTTah) (see [#2427](https://github.com/styled-components/styled-components/pull/2427))
 
 ## [v4.2.0] - 2019-03-23
 
--- a/packages/styled-components/src/macro/index.js
+++ b/packages/styled-components/src/macro/index.js
@@ -1,10 +1,7 @@
 // @flow
-import { createMacro, MacroError } from 'babel-plugin-macros';
+import { createMacro } from 'babel-plugin-macros';
 import babelPlugin from 'babel-plugin-styled-components';
 import { addDefault, addNamed } from '@babel/helper-module-imports';
-import * as styled from '..';
-
-const allowedImports: Array<string> = Object.keys(styled).filter(helper => helper !== '__esModule');
 
 function styledComponentsMacro({ references, state, babel: { types: t }, config = {} }) {
   const program = state.file.path;
@@ -14,14 +11,6 @@ function styledComponentsMacro({ references, state, babel: { types: t }, config 
   // { default: [path, path], css: [path], ... }
   let customImportName;
   Object.keys(references).forEach(refName => {
-    if (!allowedImports.includes(refName)) {
-      throw new MacroError(
-        `Invalid import: ${refName}. You can only import ${allowedImports.join(
-          ', '
-        )} from 'styled-components/macro'.`
-      );
-    }
-
     // generate new identifier
     let id;
     if (refName === 'default') {
--- a/packages/styled-components/src/macro/test/__snapshots__/macro.test.js.snap
+++ b/packages/styled-components/src/macro/test/__snapshots__/macro.test.js.snap
@@ -190,3 +190,24 @@ _styled.div.withConfig({
 })(['background:', ';'], p => p.error ? 'red' : 'green');
 "
 `;
+
+exports[`macro should work with types alongside import: should work with types alongside import 1`] = `
+"
+import styled, { DefaultTheme } from '../../macro'
+
+      ↓ ↓ ↓ ↓ ↓ ↓
+
+import { DefaultTheme as _DefaultTheme } from 'styled-components';
+import _styled from 'styled-components';
+"
+`;
+
+exports[`macro should work with types: should work with types 1`] = `
+"
+import { DefaultTheme } from '../../macro'
+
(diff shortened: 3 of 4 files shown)


mAAdhaTTah:
Bumping this, as it's preventing us from upgrading to the latest version of react-scripts (it bumps the version of babel-macro, which surfaced this issue). Happy to do whatever is needed to land this.

Thanks!

Igorbek:
I don't really get how it relates to TypeScript. Does it?

mAAdhaTTah:
The test is the best explanation:

```
import styled, { DefaultTheme } from '../../macro'
```

This will throw a `MacroError` because of the `includes` check I removed. This makes it impossible to import TypeScript types from `styled-components/macro` because those types don't actually exist in the package.

mAAdhaTTah:
The issue linked above (#2378) has more context, if that helps.

lucleray:
@mAAdhaTTah Have you tried importing `DefaultTheme` from `styled-components` instead of `styled-components/macro` ?

IMO it doesn't really makes sense to import a type from a macro (maybe @kentcdodds has more insights on this?).

mAAdhaTTah:
@lucleray I discussed with @kentcdodds on the babel-macro repo; his thought was you should actually [whitelist the TS types](https://github.com/kentcdodds/babel-plugin-macros/pull/96#issuecomment-462508628). I figure that's probably a step too far, since you don't maintain the TS types, so I just removed the validation.

> Have you tried importing `DefaultTheme` from `styled-components` instead of `styled-components/macro`?

This would work, but then requires a separate import for SC imports vs types. Also means I can't add a lint rule that says "only import from `styled-components/macro`", and I have to remember when I should be importing from the macro vs the root package.

It's minor overhead, but for what benefit? The macro doesn't need the imports to be valid in order to do its thing, so I'm not sure why we should keep it and push that off to the consumers.

lucleray:
@mAAdhaTTah You don't want your application to break if the order in which typescript and babel-macros are run in CRA change and the best way to do that is to import the types from `styles-components`.

If you start importing the types from the macro and typescript is run before babel-macro, you'll get an error.

IMO it still makes sense to show an error here, because it's not completely equivalent to import from the macro and from sc.

mAAdhaTTah:
> If you start importing the types from the macro and typescript is run before babel-macro, you'll get an error.

This is only true with the check in place.

> IMO it still makes sense to show an error here, because it's not completely equivalent to import from the macro and from sc.

Why not? Validating whether the imports are valid doesn't seem like the responsibility of the babel plugin so much as the bundler (whether that's webpack/rollup/etc. or the browser), and given that the types for the macro are [re-exports of the root types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/styled-components/macro.d.ts), they should be entirely equivalent, except for the fact that the macro throws when encountering type imports.

mAAdhaTTah:
Sorry to be a pain, but bumping this again.

I think I'd like to suggest thinking about it from this perspective: Setting aside TypeScript at all here, what benefit does validating the imports in the macro provide to either the consumers or maintainers of `styled-components`?

mAAdhaTTah:
@probablyup Last poke, I swear, but we've got 2 approvals, can we get this merged & released? Thanks!

probablyup:
@mAAdhaTTah yup just need you to add a changelog entry

mAAdhaTTah:
@probablyup Done!

probablyup:
Could you possibly make a second PR against canary when you get a chance?

PR #2428, opened 2019-03-05
Reduce GC pressure by avoiding to create new object types

### LINKED ISSUE #2424: Flatten function performance issue v4
Having recently tried to upgrade our styled-components from v3 to v4 in a relatively large react app - we've ran into a strange performance issue that seems to be tied to the flatten function. When commenting out this part of flatten, the issue seems to resolve:

https://github.com/styled-components/styled-components/blob/9b81695d35074974c42b943a82456c9ae4b42500/packages/styled-components/src/utils/flatten.js#L59-L72

Here are screenshots of performance snapshots run with chrome dev tools when performing the same action in our app on v3 vs. v4:

**v3**
screen shot 2019-03-04 at 11 06 48 am

**v4**
screen shot 2019-03-04 at 10 59 45 am

Strangely, we have not been able to reproduce this in a clean slate react app.

## System:
 - CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
 - Memory: 2.29 GB / 16.00 GB
 - Shell: 5.3 - /bin/zsh
## Binaries:
 - Node: 8.14.0 - /usr/local/bin/node
 - npm: 6.5.0 - /usr/local/bin/npm
## npmPackages:
 - babel-plugin-styled-components: ^1.4.0 => 1.10.0

## Reproduction

Unable to reproduce with a basic react app created via `create-react-app`. 

## Expected Behavior
Performance is as good if not better than v3.

## Actual Behavior
Performance is sluggish, significantly slower than v3.

alansouzati:
Closes #2424 

If you inspect the issue, you will see there is a major GC collector happening. `flatten` function is called a lot of times and every time we interpolate some style that code block is executed.

I'm avoiding to do `new chunk` and, instead, replaced with a new util called `isStatelessFunction`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,12 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 ## Unreleased
 
+- Fix memory leak issue when using component selector styles. (see [#2424](https://github.com/styled-components/styled-components/issues/2424)).
+
 - Make the `GlobalStyleComponent` created by `createGlobalStyle` call the base constructor with `props` (see [#2321](https://github.com/styled-components/styled-components/pull/2321)).
 
 - Move to Mono repository structure with lerna [@imbhargav5](https://github.com/imbhargav5) (see [#2326](https://github.com/styled-components/styled-components/pull/2326))
+
 
 ## [v4.1.3] - 2018-12-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,11 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 ## Unreleased
 
-- Fix memory leak issue when using component selector styles. (see [#2424](https://github.com/styled-components/styled-components/issues/2424)).
+- Reduced GC pressure when using component selector styles. (see [#2424](https://github.com/styled-components/styled-components/issues/2424)).
 
 - Make the `GlobalStyleComponent` created by `createGlobalStyle` call the base constructor with `props` (see [#2321](https://github.com/styled-components/styled-components/pull/2321)).
 
 - Move to Mono repository structure with lerna [@imbhargav5](https://github.com/imbhargav5) (see [#2326](https://github.com/styled-components/styled-components/pull/2326))
-
 
 ## [v4.1.3] - 2018-12-17
 
--- a/packages/styled-components/src/utils/flatten.js
+++ b/packages/styled-components/src/utils/flatten.js
@@ -2,6 +2,7 @@
 import { isElement } from 'react-is';
 import getComponentName from './getComponentName';
 import isFunction from './isFunction';
+import isStatelessFunction from './isStatelessFunction';
 import isPlainObject from './isPlainObject';
 import isStyledComponent from './isStyledComponent';
 import Keyframes from '../models/Keyframes';
@@ -55,12 +56,13 @@ export default function flatten(chunk: any, executionContext: ?Object, styleShee
 
   /* Either execute or defer the function */
   if (isFunction(chunk)) {
-    if (executionContext) {
+    if (isStatelessFunction(chunk) && executionContext) {
       let shouldThrow = false;
 
+      const result = chunk(executionContext);
+
       try {
-        // eslint-disable-next-line new-cap
-        if (isElement(new chunk(executionContext))) {
+        if (isElement(result)) {
           shouldThrow = true;
         }
       } catch (e) {
@@ -71,7 +73,7 @@ export default function flatten(chunk: any, executionContext: ?Object, styleShee
         throw new StyledError(13, getComponentName(chunk));
       }
 
-      return flatten(chunk(executionContext), executionContext, styleSheet);
+      return flatten(result, executionContext, styleSheet);
     } else return chunk;
   }
 
--- a/packages/styled-components/src/utils/flatten.js
+++ b/packages/styled-components/src/utils/flatten.js
@@ -57,19 +57,9 @@ export default function flatten(chunk: any, executionContext: ?Object, styleShee
   /* Either execute or defer the function */
   if (isFunction(chunk)) {
     if (isStatelessFunction(chunk) && executionContext) {
-      let shouldThrow = false;
-
       const result = chunk(executionContext);
 
-      try {
-        if (isElement(result)) {
-          shouldThrow = true;
-        }
-      } catch (e) {
-        /* */
-      }
-
-      if (shouldThrow) {
+      if (isElement(result)) {
         throw new StyledError(13, getComponentName(chunk));
       }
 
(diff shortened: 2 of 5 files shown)


probablyup:
@alansouzati can you add a changelog entry for this as well?

probablyup:
Published under `styled-components@4.1.4-alpha.6` for testing

packages/styled-components/src/utils/flatten.js
```diff
@@ -55,12 +56,13 @@ export default function flatten(chunk: any, executionContext: ?Object, styleShee
 
   /* Either execute or defer the function */
   if (isFunction(chunk)) {
-    if (executionContext) {
+    if (isStatelessFunction(chunk) && executionContext) {
```
tbranyen: Since classes used to go into this code path and be checked if they are styled components or not, is this a breaking change?

alansouzati:
@tbranyen pointed out that we may need to move the check for a full compatibility with the previous implementation. I'm taking a pass at that.

alansouzati:
yeah we just validated it, it is not a breaking change. not further changes required it seems.

packages/styled-components/src/utils/flatten.js
```diff
@@ -55,12 +56,13 @@ export default function flatten(chunk: any, executionContext: ?Object, styleShee
...
       let shouldThrow = false;
 
+      const result = chunk(executionContext);
+
       try {
```
alansouzati: I think we can remove the try catch too!!

packages/styled-components/src/utils/flatten.js
```diff
@@ -55,23 +56,14 @@ export default function flatten(chunk: any, executionContext: ?Object, styleShee
...
-        // eslint-disable-next-line new-cap
-        if (isElement(new chunk(executionContext))) {
-          shouldThrow = true;
-        }
-      } catch (e) {
-        /* */
-      }
-
-      if (shouldThrow) {
+      if (isElement(result)) {
```
kitten: This doesn't quite seem safe. We're calling a function that might be a component and are doing so to throw a proper warning if it is;

I don't think I even agree with the previous error to be honest. A return value of a component might not always be an element, so this doesn't seem safe.

It's also unsafe to simply call this because components might contain hooks now, so we'd still need the try-catch: https://runkit.com/kitten/5c7f0daf31404d0011b3a130

@probablyup @mxstbr Are we sure we really need this error in production? Seems a little overzealous. Maybe an error in development or even just a `console.error` would do the trick?

@alansouzati btw, can you change the changelog entry to say it's reducing memory usage when running component selector styles? It seems a little extreme to call it a memory leak, since I don't think this is leaking, it simply seems to be allocating a lot of memory that the GC is slow to clean up 😅 Still bad, but I'd hate to confuse people tbh

alansouzati: sure I can re-word the change log. I will wait to do that until you decide if you want the check or not. Then I will update the PR in one go.

alansouzati: The goal of this PR was to maintain the same features while fixing an important issue that is affecting performance.

alansouzati: Is there an easy way to tell if something is a styled component or not ? the check could be changed to validate that instead.

probablyup: Yeah we have a util for it

alansouzati: Cool does it make sense to use that instead? Unfortunately I don’t have a lot of the context for this.

alansouzati:
any updates here? This is a rather pressing issue.

packages/styled-components/src/utils/flatten.js
probablyup: We check if it's a styled component at https://github.com/styled-components/styled-components/pull/2428/files#diff-ce0d3892db66efa4deef4063f44d3022R53 so not necessary

alansouzati: I'm confused. can we remove the check or not?

probablyup: I'm just responding to your earlier question, interpolated styled components are handled earlier in `flatten` so you don't have to worry about them in the try/catch

probablyup: To @kitten's point, I think you're right and we can just do a dev-time warning & swallow the error in prod.

alansouzati: is it ok to do that in a separate PR?

alansouzati: I'm happy to do it in this PR if you can point me to a reference on how you are envision this.

CHANGELOG.md
```diff
@@ -6,10 +6,13 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this
 
 ## Unreleased
 
+- Fix memory leak issue when using component selector styles. (see [#2424](https://github.com/styled-components/styled-components/issues/2424)).
```
tbranyen: I noticed the comment saying to update this and I agree this isn't a leak, but helps prevent unnecessary GC cycles.

mxstbr:
I agree with @kitten and @probablyup, two changes needed here:

- [ ] Move warning to be dev-only like our other warnings by checking `process.env.NODE_ENV`
- [ ] Rework changelog entry to say "reduce GC pressure" or something to that effect rather than memory leak

alansouzati:
PR updated with the requested changes.

packages/styled-components/src/utils/flatten.js
probablyup: The `isStyledComponent` check happens higher in the function, so shouldn't be.

packages/styled-components/src/utils/flatten.js
```diff
@@ -55,23 +55,19 @@ export default function flatten(chunk: any, executionContext: ?Object, styleShee
...
+        );
       }
 
-      if (shouldThrow) {
-        throw new StyledError(13, getComponentName(chunk));
-      }
-
-      return flatten(chunk(executionContext), executionContext, styleSheet);
+      return flatten(result, executionContext, styleSheet);
     } else return chunk;
```
probablyup: The else branch here should also have the warning, no? If `isStatelessFunction(chunk)` returns false that implies a class, which is not a valid thing to interpolate.

probablyup: ping @alansouzati

probablyup:
I'd love to get the PR in this week and cut a minor release if you have time to address my last comment there!

packages/styled-components/src/utils/flatten.js
alansouzati: sorry I've missed this. I think the warning should be only in one place. probably by moving things around. let me take a careful pass at this.

alansouzati: I don't think we should add a warning in the else.

`isStatelessFunction(chunk) && executionContext`

this is deferring the execution since `executionContext` is empty.

alansouzati: @mxstbr please advise here. I don't have the full context of this flatten operation so I can definitely be missing something.

probablyup: Ok I see! Yes this is perfect, thanks @alansouzati!

PR #2430, opened 2019-03-05
Delete unused src folder.

alansouzati:
I believe the `src` folder needs to be removed now that you moved to monorepo.


mxstbr:
All old versions of styled-components link to that file when an error happens in production, which is why we left that file where it is. We need to revert this PR.

probablyup:
Ahhhh crap

mxstbr:
No worries, reverted in #2443

alansouzati:
i see. it might be a good idea adding a markdown comment to avoid this to be deleted in the future

alansouzati:
never mind. just saw that you actually added the comment now

PR #2431, opened 2019-03-05
ADD tests for multiple styleFns & styleObjs

ndelangen:
Styled-components supports this right?

```js
styled.div(
  ({ color }) => ({ color }),
  ({ color }) => ({ borderColor: color })
);
```

&

```js
styled.div(
  { color: 'blue' },
  { borderColor: 'blue' }
);
```

I couldn't find any tests for this behavior.

### DIFF
--- a/packages/styled-components/src/test/basic.test.js
+++ b/packages/styled-components/src/test/basic.test.js
@@ -127,6 +127,33 @@ describe('basic', () => {
     expectCSSMatches('.b { color:blue; }');
   });
 
+  it('should allow you to pass in multiple functions returning a style objects', () => {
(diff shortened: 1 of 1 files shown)


ndelangen:
Btw I was trying out this behavior in the playgrounds on the website and it didn't. When trying it in https://codesandbox.io/s/6yv9jo2jr it did work.

probablyup:
I don't know if we've ever explicitly supported it, but it's cool that it works? What do you think @mxstbr / @kitten, do we want to canonicalize this?

probablyup:
Closing since we don't want to encourage this. Thanks though!

PR #2436, opened 2019-03-07
Refactored code to use more named exports

### LINKED ISSUE #2418: [FEATURE REQUEST] Export utils, contexts, and constructors
I understand that the react hooks migration will take sometime. It will most likely be part of the next [major version](https://github.com/styled-components/styled-components/issues/1934) of styled-components.

At the same time, I'm facing [performance issues](https://github.com/styled-components/styled-components/issues/2215) with styled-components. This issue is very pressing, and I'm confident that moving to react hooks will get rid of the problem.

So I decided to take a stab at creating my own version of [StyledComponent](https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/models/StyledComponent.js) that replaces `*Consumer` with `useContext(...)`. The basic idea is to do something like:

```
import styled from "styled-components-no-context";

const Box = styled.div``; // no consumer will be added because we are using useContext
```

Initially I was very confident that this solution would work, until I've looked into `node_modules/styled-components/dist`. I was really hoping the `esm` distribution would maintain the module exports so that I could just replace `../utils/determineTheme` with `styled-components/lib/utils/determineTheme`.

Instead, you are controlling what gets exported or not. The StyleSheet module is inside a `__DO_NOT_USE_OR_YOU_WILL_BE_HAUNTED_BY_SPOOKY_GHOSTS` variable.

I'm a very superstitious guy, so I don't want to use that variable. 😆 

Jokes apart, there would be quite a few things I would need to manually copy over: all utils, the constructors, and contexts. You don't expose `StyleSheetContext` so there is no way for me to do `useContext(StyleSheetContext)`.

I'm a big fan of styled-components. At the same time, I don't want to sit and wait for a v5, not even talking about other breaking changes that will come that will prevent me to migrate immediately once it is out.

I decided to take a look at emotion and see how they structure their dist content. I really liked the architecture there where everything lives in its own package, for example:

```
import { createContext, createElement, forwardRef, Component, Fragment } from 'react';
import createCache from '@emotion/cache';
import { getRegisteredStyles, insertStyles } from '@emotion/utils';
import { serializeStyles } from '@emotion/serialize';
import { StyleSheet } from '@emotion/sheet';
import css from '@emotion/css';
export { default as css } from '@emotion/css';
```

Both [styled-components and emotion](https://www.youtube.com/watch?v=AqFqduEEw7A&t=19s) are suffering from the context issue. I feel it will take some time until you both move away from it. In the meantime, as it stands today, emotion offers a better architecture to allow consumers to migrate earlier.

Please don't get me wrong, I want to keep using styled-components, but I'm afraid I will be out of options if I'm not able to fix the performance issues in my organization.

alansouzati:
Closes #2418 by adding utils, models, and constructors to the secretInternals.


packages/styled-components/src/constructors/test/keyframes.test.js
```diff
@@ -36,7 +37,7 @@ describe('keyframes', () => {
         opacity: 1;
       }
     `.getName()
-    ).toMatchInlineSnapshot('"bcCCNc"');
```
alansouzati: I believe this test was wrong before. We are using the `'../../test/utils'` and it should mock the `generateAlphabeticName`

packages/styled-components/src/constructors/test/keyframes.test.js
```diff
@@ -2,11 +2,12 @@
 import React from 'react';
 import TestRenderer from 'react-test-renderer';
 
-import css from '../css';
-import keyframes from '../keyframes';
-import Keyframes from '../../models/Keyframes';
 import { expectCSSMatches, getCSS, resetStyled } from '../../test/utils';
```
alansouzati: the order definitely matters here. And I feel the mock call should always come first.

alansouzati:
https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html

packages/styled-components/src/hoc/withTheme.js
```diff
@@ -1,11 +1,9 @@
 // @flow
 import React, { type ComponentType } from 'react';
-import { ThemeConsumer, type Theme } from '../models/ThemeProvider';
-import determineTheme from '../utils/determineTheme';
-import getComponentName from '../utils/getComponentName';
-import hoistStatics from '../utils/hoist';
```
alansouzati: this is an example of the default export is bad. the name of the function is misleading since it is hoist non statics... but you can give any name since it is the default export...

probablyup:
I'm fine with your internal refactorings, but as I said before I don't want to expose actual internals. If you wanted to pare this down to just exposing the `StylesheetContext` as well, I'd accept that.

mxstbr:
Yeah we definitely do not want to add all our utils and stuff to the external API, that would make any change to them a breaking one 😱😱😱😱😱

alansouzati:
Let me start by saying that I understand where you are coming from. I was not expecting that you would approve this PR right away, I just wanted to get the ball rolling and chat about this. Exposing just the `StylesheetContext` seems like a good compromise for now. I will update the PR accordingly.

Although, let me explain what I'm trying to achieve in more details in hopes that I would convince you to expose the utils and constructors too.

For context, I've observed that throughout my app I have very repetitive constructs, and I never use `children` to style my components. So, imagine this structure:

```javascript
const Box = styled.div`...`;
const Text = styled.span`...`;

// for example, I never do:
const Test = styled.span`
  ${props => props.children && '...'}
`;

<Box>
  <Box>
      <h1>VM1</h1>
  </Box>
  <Text>online</Text>
</Box>
```

This is a simplified example of a listitem in a list grid. I've been experiencing performance issues with this. I have a feeling that PR #2428 will make things better.

But while learning more about the root causes for the issue, I've noticed that `flatten` function is called a LOT of times. Most of times the function return is the same styles. So, that got me thinking, I must be able to memoize [this](https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/models/ComponentStyle.js#L58).

The problem is that if I memoize based on `executionContext` children will be there and it is always different.

As a result, I want to create my custom `lean` styled that makes opinionated assumptions like removing children from `executionContext` for memoization. This change does not make sense to be in styled-components since other folks can be using it. I've tested locally and memoizing significantly improves the update cycles for my list grid.

We could make a flag `styled('div', { memoize: true })` but this feels too much, I'm not sure though.

I think a good start is to make things easier for creating custom `styled` components like you have StyledComponent and StyledNativeComponent. This was my intent with this PR. If we expose all secret internals under `__DO_NOT_USE_OR_YOU_WILL_BE_HAUNTED_BY_SPOOKY_GHOSTS`, I understand that I'm not expecting documentation or compatibility support. At the same time, creating my custom lean styled will be much easier since I will not have to replicate most of styled components code.

alansouzati:
PR updated to remove the exports 👍

mxstbr:
I actually think that change could make sense for styled-components generally. In two years (or however long its been) I have never seen anybody adjust the styling based on children, and I have seen a lot of styled-components code!

Can you open a new issue so we can discuss that?

alansouzati:
https://github.com/styled-components/styled-components/issues/2440

.eslintignore
```diff
@@ -0,0 +1,11 @@
+node_modules
```
alansouzati: this was to fix a precommit issue. precommit is running `eslint . --fix` but as we moved to monorepo, this is failing since `.eslintrc` got moved.

packages/styled-components/package.json
```diff
@@ -100,7 +100,7 @@
   "bundlesize": [
     {
       "path": "./dist/styled-components.min.js",
-      "maxSize": "16kB"
+      "maxSize": "18kB"
```
probablyup: Can we drop this back down now?

packages/styled-components/src/hoc/withTheme.js
probablyup: I don’t really understand the beef with default exports. The point is the implementation is opaque and the end developer doesn’t need to know how the internals are structured to use the thing.

probablyup:
Would you mind creating a separate PR just exposing the StyleSheetContext? I'm not sure I agree with all the internal default exporting changes TBH, but I'd like to get that change out in the next minor release.

probablyup:
Opened #2456 as a more minimal implementation of the desired external change. I'd prefer not to merge this one as it would cause massive thrash on the hooks refactor PR

PR #2444, opened 2019-03-11
Revert "Delete unused src folder."

probablyup:
Reverts styled-components/styled-components#2430


mxstbr:
Already reverted in #2443!

PR #2446, opened 2019-03-13
Fix compatibility with react-native-web v0.11.0

msand:
V0.11.0 released yesterday has deprecated the default export, thus we need to add `* as` as explained by @necolas
https://github.com/necolas/react-native-web/issues/1287#issuecomment-472550053


packages/styled-components/src/native/index.js
```diff
@@ -1,7 +1,207 @@
 // @flow
 
 /* eslint-disable import/no-unresolved */
-import * as reactNative from 'react-native';
+import {
+  AccessibilityInfo,
+  ActivityIndicator,
+  ReactNativeART,
```
necolas: ART is gone from RNW@0.11 and several other modules are already deprecated in RN or soon will be

packages/styled-components/src/native/index.js
```diff
@@ -1,7 +1,207 @@
 // @flow
 
 /* eslint-disable import/no-unresolved */
-import * as reactNative from 'react-native';
+import {
```
necolas: Only a few of these modules are components that can be styled. Perhaps a limited subset could be imported instead? It might also be worth reassessing the "alias" API altogether. Leaving it up to consumers to do `styled(View)` instead seems like it would play better with lazy eval and load, as well as avoiding issues related to future component deprecations in React Native needing to be fixed in this library as well as any app code depending on it.

msand: I limited it to the ones which are aliased now

msand:
@necolas Using require seems to keep both the CI happy, and seems to work in the minimal repro I made: https://github.com/msand/react-native-web-styled-components-bug/commit/17c95270eb186be33ea3f81f3060d63036bec488

So this seems to be the simplest patch to fix the issue, and keep the imports lazy:
```
-import reactNative from 'react-native';
+const reactNative = require('react-native');
```

I'm still getting some strange behavior with v0.11 in a larger app, seems none of the styles for react-native-web are applied. But I'll keep investigating and try to isolate it, or stay with 0.10 for now.

packages/styled-components/src/native/index.js
kilbot: I just ran into this problem, thanks for working on a fix!

probablyup:
Is this a semver-major change?

msand:
It might affect some build configurations, although I think it should work fine for most users. At least safer with a semver major, as there will almost certainly be some projects with very lax version constraints.

necolas:
I don't think this is a good solution. This patch loads the legacy CJS entry point from RNW, which means you pull in the *entire* library no matter how little of it is being used in app code.

probablyup:
Alright then please make a followup PR against canary.

packages/styled-components/src/native/index.js
EvanBacon: Using `import * as reactNative from react-native` will always include every react native web module. This is something we really want to avoid.

PR #2458, opened 2019-03-23
Improve example app

### LINKED ISSUE #1358: More complex Example app

maxxgl:
Expanding the functionality of the example app to provide more visual testing.

Fixes #1358.

Most of the test components are quite simple, and rely on the `background` property to make it easy to identify at a glance if there any errors. All components should have at least one `mediumseagreen` element, signifying that the basic feature of the component is functioning.

New example app: 
example-test-js-example-page-should-match-screenshot-1-snap

Caveats: 
Due to the test setup, there is an issue that all of the `styled` or `React` references must be in the index file for linting. There are some code indentation liberties taken to make the code print correctly as well.


probablyup:
Ooh, this actually would be great to pass through something like https://percy.io/ to help prevent regressions between builds. @mxstbr @kitten

kitten:
That pretty awesome! @probablyup Do we still have the screenshot regression tests on the website? Could just reuse that code for now

maxxgl:
@kitten @probablyup I was finally able to get this to pass the tests.

The current Travis [snapshot test](https://github.com/maxxgl/styled-components/blob/1358-improve-example-app/packages/styled-components/integration-test/example.test.js) uses the `google-chrome-beta` binary as set in the Travis config (which I didn't realize at first). Which means that there are potentially quite a few problems when people test this locally, due to either the browser version being used by puppeeter, or the host system's browser quirks (especially font-related).

Honestly, after getting a little frustrated trying to make this pass tests, I'm not sure I would recommend merging this considering the current snapshot strategy.

probablyup:
What we could do instead is set up storybook in the example app and use https://www.chromaticqa.com/ to do the automated visual regression testing. Then we could drop using the chrome binary entirely.

probablyup:
I'm going to co-opt this and implement chromatic :) will make sure you retain authorships on the commits though 👍

probablyup:
Closing in favor of #2589 which rolls up these changes with different end tooling

PR #2483, opened 2019-04-06
fix: ensure nonce can be found and applied

### LINKED ISSUE #2363: Client-side nonce support
I am trying to craft an intelligent Content-Security-Policy while using styled-component. The suggested solution to this seems to be to [use SSR to set the __webpack_nonce__ variable](https://github.com/styled-components/styled-components/issues/887), ensuring that webpack inlines in in all created style-tags. However, I cant find a solution to this problem when you are not using SSR.

Without knowing the internals of styled-components, it seems to me that it should be possible to specify a nonce in a non-ssr environemt. Something like:

```
import { setNonce } from "styled-components"
setNonce(myNonce)
```

This would allow users to craft their own solutions that is independent from both SSR and webpack. In my use case it would be trivial to inject a unique nonce with every page request, even though SSR is not feasible.

mattcosta7:
fix #2363 

When applying a nonce in a client side only environment, prior to this change strict scoping in modules limits the ability to access __webpack_nonce__ if it doesn't exist on the window or global object

This change checks global and window for the presence of __webpack_nonce__ before checking the variable itself to alleviate that scoping issue, and only uses the first one that exists.

### DIFF
--- a/packages/styled-components/src/utils/nonce.js
+++ b/packages/styled-components/src/utils/nonce.js
@@ -1,6 +1,13 @@
 // @flow
-/* eslint-disable camelcase, no-undef */
+/* eslint-disable camelcase, no-undef, no-nested-ternary */
 
 declare var __webpack_nonce__: string;
 
-export default () => (typeof __webpack_nonce__ !== 'undefined' ? __webpack_nonce__ : null);
+export default (global => () => {
+  if (typeof global.__webpack_nonce__ !== 'undefined') {
+    return global.__webpack_nonce__;
+  } else if (typeof __webpack_nonce__ !== 'undefined') {
+    return __webpack_nonce__;
+  }
+  return null;
+})(typeof global !== 'undefined' ? global : typeof window !== 'undefined' ? window : {});


probablyup:
Isn’t webpack nonce a magic string? I don’t think it’s a real variable?

mattcosta7:
Maybe I'm misunderstanding how to use it in a client side environment, which is very possible too?

In a client side environment (ie a cra application served via cdn), we can inject a nonce into the head at serve time like 

```
<script nonce="qsdasdsa">
   __webpack_nonce__ = "qsdasdsa"
</script>
```

without this change, this gets ignore by styled-components, and generated styles do not have the nonce added to the style tags.  Debugging showed that the variables were available in the global scope, but accessing them inside the styled-components returned undefined. (rather than the null or string) it should return.

ixtli:
`__webpack_nonce__` is documented here https://webpack.js.org/guides/csp/. It's the only way to make style injection work with very strict CSPs, which very important for applications that need to pass rigorous security evaluations.

(The tl;dr: is that a compliant browser will reject styles that dont have the `nonce` property set to match the html documents CSP header in order to keep post-load malicious scripts from injecting whatever they want. It's hard to circumvent because assuming its random every time the malicious script would have to guess the string.)

mattcosta7:
@probablyup I looked into this more, and I think I understand now what you were getting at. 

The issue was that I was attempting to use this in a dev hmr setting as well (to make sure it was working), and __webpack_nonce__ doesn't seem to actually work in that environment, but in a build it does.  I'll try to update the documentation there a little later on, to make it more easy to understand. 

In this respect, it appears that my changes aren't needed, and this is actually not an issue with the implementation here

probablyup:
Alright, going to close. Feel free to reopen if something changes.

Issue #2414, opened 2019-02-24
Feature request: Functions as values for object literals


mrmckeb:
Hopefully this isn't something that's been raised before, I couldn't find a related issue.

I love the new object support in `styled-components`, and while I know that this isn't something that aligned with the project's original goals, I wonder if this could be improved to support functions as values? There may also be some good technical reasons why this shouldn't happen, so if that's the case, please share those too.

As an example:
```js
var A = styled.a(({ big }) => ({
  fontSize: big ? 24 : 16,
}));
```

Would become:
```js
var A = styled.a({
  fontSize: ({ big }) => big ? 24 : 16,
});
```

probablyup:
Hmm I don't see why not, I'd accept a PR for it. You'd need to make the change in `flatten` to pass executionContext into `objToCss` and invoke the function with it (but only if confirmed that it's a function and not a SFC.)

mrmckeb:
If no one else picks this up, I'll take a look in the coming weeks :)

d7my11:
I didn't see this issue, but this is done as part of #2489 which fixes #2468 and achieve this enhancement

mrmckeb:
Ah, amazing @d7my11, I had this open in a tab as I planned to pick it up this weekend - so thanks for letting me know ;)

mrmckeb:
Hi @probablyup, my understanding is that this is released now - and we can close this issue?

mrmckeb:
Actually, I've tested this today @probablyup and @d7my11 - this works, but there are two issues I've spotted so far:
1. The TypeScript definitions don't support this.
2. Returning a number creates invalid CSS.
```
const Spinner = styled.div({
  width: () => 40 * 2,
});

// Sets width of '80' not '80px'

If anyone can point me in the right direction, I'd be happy to help with one or both of those.
```

d7my11:
> Returning a number creates invalid CSS.

Why `px` and not other unit?
because you should handle the unit yourself based on the use case. This is the same behaviour in template literals as well:

```js
const Spinner = styled.div`
  width: ${() => 40 * 2};
`;

 /* will result (width: 80;) without the unit */ 
```

> The TypeScript definitions don't support this.

will look into it

mrmckeb:
@d7my11 I don't disagree, but this is not the behaviour that you see with an object.

This is also the standard behaviour whenever you set any style in JavaScript. Unit-less values default to pixels (where appropriate).

```tsx
const Spinner = styled.div({
[...]

PR #2489, opened 2019-04-09
Fix objects with functions - replace (objToCss) with (objToCssArray)

### LINKED ISSUE #2468: Functions inside of style objects, using multiple selectors
## Environment

> ### System:
>  - OS: macOS High Sierra 10.13.6
>  - CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
>  - Memory: 86.22 MB / 16.00 GB
>  - Shell: 5.6 - /usr/local/bin/zsh
> ### Binaries:
>  - Node: 10.15.0 - /usr/local/bin/node
>  - Yarn: 1.5.1 - ~/.local/share/npm/bin/yarn
>  - npm: 6.8.0 - ~/.local/share/npm/bin/npm
>  - Watchman: 4.9.0 - /usr/local/bin/watchman

## Reproduction

See [CodeSandbox][repro]

[repro]: https://codesandbox.io/s/817wl8l08l

## Steps to reproduce

- Create an object style using comma-separated tags, and a property implemented as a function. See `BrokenStyle`, below:

```js
const WorkingStyle = createGlobalStyle`
  h1, h2, h3, h4, h5, h6 {
    font-family: ${({theme}) => theme.fonts.heading};
  }
`;

const BrokenStyle = createGlobalStyle({
  'h1, h2, h3, h4, h5, h6': {
    fontFamily: ({theme}) => theme.fonts.heading,
  },
});
```

## Expected Behavior

- The two implementations `WorkingStyle` and `BrokenStyle` are equivalent.

## Actual Behavior

- When using `BrokenStyle`, it seems as though `styled-components` stringifies my function and subsequently attempts to parse that as CSS. This new syntax is _amazing_, but not quite what I was expecting. Check out the output from the [above CodeSandbox][repro] (newlines added for clarity here):

```
h1 font-family:function fontFamily(_ref2),
h2 font-family:function fontFamily(_ref2),
h3 font-family:function fontFamily(_ref2),
h4 font-family:function fontFamily(_ref2),
h5 font-family:function fontFamily(_ref2),
h6 font-family:function fontFamily(_ref2){
  -webkit-var:theme = _ref2.theme;
  -moz-var:theme = _ref2.theme;
  -ms-var:theme = _ref2.theme;
  var:theme = _ref2.theme;
  return:theme.fonts.heading;
}
```

## Submitter note

If you'd like help with a fix, I'd love to tackle it; this is right up my alley. I'm not looking to enable my use case _per se_, I'd just like to prevent the next person from going through the same 20 minutes I did 😄 

I could see this being resolved in a few ways:

- Update `styled-components` so my sample works as I had expected
- Add a small note to the docs: "don't do this!" (Check out the implementation of `WorkaroundStyle` [in the CodeSandbox][repro]; technically this performs better, but there's no impact anyways since it usually runs once at page load time.)
- Implement a warning or safeguard to prevent this edge case (probably with an accompanying docs note)
- Something else?

d7my11:
Fixes: #2468 
Addresses: #2414
***

The issue was in `objToCss` method that didn't handle methods for objects.
- [x] Replace `objToCss` with `objToArray` which handles methods as well. 
- [x] Cover `createGlobalStyle` and `styled` with more test cases.


mrmckeb:
Out of interest, are there any performance metrics around a) objects/functions over template literals, and b)  methods as properties?

d7my11:
@mrmckeb I'm not 100% sure actually, but here is my understanding:
using template literal will be converted by babel to arrays if they have embedded expressions and get `.concat()`ed at the end.

Using objects gets converted by styled-components and it gets slower with nested objects because you have to go through every key for each object.

So, based on my assumptions, using template literal is faster specially if the object contains nested objects.

I would like to hear from somebody who is fully aware of the project to verify or correct me if i'm wrong.

mrmckeb:
Good information, thanks @d7my11 - and great work.

probablyup:
I'd also love if you could do this against `canary` instead or open a second PR for it.

d7my11:
@probablyup is there any minor releases for bugs before merging `canary` into `master`? if so, will open a second PR for this. If not, will change the base branch.

probablyup:
Yeah 4.x isn’t closed out yet

d7my11:
Will open another PR then

probablyup:
If you could build it on top of @kitten's latest PR that'd be great as I think he made some changes to code near what you're changing

d7my11:
@probablyup opened another PR for `canary` #2542

probablyup:
@d7my11 can you get this rebased against master? Looks like a test file just needs a little fix

d7my11:
@probablyup done

packages/styled-components/src/utils/flatten.js
```diff
@@ -14,19 +14,32 @@ import addUnitIfNeeded from './addUnitIfNeeded';
...
-    .filter(key => !isFalsish(obj[key]))
-    .map(key => {
-      if (isPlainObject(obj[key])) return objToCss(obj[key], key);
-      return `${hyphenate(key)}: ${addUnitIfNeeded(key, obj[key])};`;
+
+const objToCssArray = (obj: Object, prevKey?: string): Array<string | Function> => {
+  const rules = []
+  const keys = Object.keys(obj)
+
+  keys
```
probablyup: this would be cleaner as a reduce I think

d7my11: true, but usually it's slower than `.forEach` because the result has different type than the items. i did a quick benchmark for that [here](http://jsben.ch/A6Ad9)

probablyup: Fair enough!

probablyup:
Last thing, a changelog entry would be lovely

d7my11:
@probablyup thanks for reviewing

PR #2491, opened 2019-04-10
added details-menu tag helper

njenia:


probablyup:
Not a real HTML element

Issue #2254, opened 2018-11-28
styled-components does not inject styles in head


JrmKrb:
Hello,

As said in the title, I cannot make styled-components work in my project. I am using React with Ruby via the `webpacker` gem if that's of any help. The CSS is not included in my `<head>`, however it is creating on-fly CSS classes and gives those classes to my element `h2` in the below example, but those classes do not have any CSS linked to them.

I am using React 16.6.0 and styled-components 4.1.1.

### System:
 - OS: Linux 4.4 Ubuntu 16.04.5 LTS (Xenial Xerus)
 - CPU: (4) x64 Intel(R) Core(TM) i5-6300HQ CPU @ 2.30GHz
 - Memory: 784.58 MB / 7.63 GB
 - Container: Yes
 - Shell: 4.3.48 - /bin/bash
### Binaries:
 - Node: 8.4.0 - /usr/local/heroku/bin/node
 - Yarn: 1.12.3 - /usr/bin/yarn
 - npm: 6.1.0 - ~/project/.bin/npm
### npmPackages:
 - styled-components: ^4.1.1 => 4.1.1 

## Reproduction
 I do not know how to reproduce this issue, it might be linked to my setup. I spent quite some time on it and still don't know where it's coming from. I basically just `yarn add styled-components`, added this snippet in my file:

```javascript
import styled from "styled-components"

const Title = styled.h2`
  font-size: 5.5em;
  text-align: center;
  color: blue;
```

and used `<Title> My Title </Title>` instead of my former `<h2> My Title </h2>` tag

## Expected Behavior
   My title should get the CSS I defined.

## Actual Behavior
   My title is inside a `<h2>` tag with styled-components classes `sc-bdVaJa csRtfh` but those classes do not have any CSS. No CSS is injected in my head tag. The only injected tag seems to be

```html
<style data-styled="" data-styled-version="4.1.1">
/* sc-component-id: sc-bdVaJa */
</style>
```

mxstbr:
- Do the styles show up, i.e. does the `h2` have a font-size of 5.5em and a color of blue?
- Does RoR/webpacker server-side render the React components? 
- Which browser are you using?

JrmKrb:
Hello, thanks for the quick answer!

- The style does not show up, my h2 elements has the classes generated by `styled-components` but there is no CSS for those classes and thus for my element

- I do not server-render my components

[...]

PR #2498, opened 2019-04-12
fix bug where style tag has been removed from the DOM

ryanirilli:
Due to how [rehydrate](https://github.com/styled-components/styled-components/blob/c3eedbeb79621acd668e15ed2d63107528ed81c3/packages/styled-components/src/models/StyleTags.js#L410) works, there is a scenario where the `Tag.styleTag` will be an element that has been removed from the DOM causing styles not to be rendered. 

This is happening for us because there is another javascript bundle being loaded on our page after our style tag has been created with the same version of styled-components. 

This ensures the style tag be added back to the DOM before trying to insertRules to it. There may be a better approach here but this is the simplest fix as far as I can tell.

 I did not see existing tests for `insertRules` or I would have added one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 ## Unreleased
 
+- Ensure style tag is present in the head when calling insertRules on the browser tag. (see [#2498](https://github.com/styled-components/styled-components/pull/2498))
+
 ## [v4.2.0] - 2019-03-23
 
 - Reduced GC pressure when using component selector styles. (see [#2424](https://github.com/styled-components/styled-components/issues/2424)).
--- a/packages/styled-components/src/models/StyleTags.js
+++ b/packages/styled-components/src/models/StyleTags.js
@@ -255,6 +255,13 @@ const makeBrowserTag = (el: HTMLStyleElement, getImportRuleTag: ?() => Tag<any>)
     const importRules = [];
     const cssRulesSize = cssRules.length;
 
+    if (IS_BROWSER && !el.parentElement) {
+      const { head } = document;
+      if (head) {
+        head.appendChild(el);
+      }
+    }
+
     for (let i = 0; i < cssRulesSize; i += 1) {
       const rule = cssRules[i];
       let mayHaveImport = extractImport;


matthewverde:
An alternative to this would be to change how the list of elements is grabbed for the rehydration step, or how elements are removed from the DOM during the rehydration step. It seems like removing an element from the DOM when it's parent is the target element (aka the head) should never be done.

probablyup:
I believe there’s another PR up for this also

ryanirilli:
@probablyup do you think this is a reasonable change?

ryanirilli:
Thanks @probablyup, I added a changelog entry.

ryanirilli:
Anything else we need to do to get this merged?

kitten:
@ryanirilli The sheet implementation for canary (v5) will change shortly and Alters how this works.

There's a small issue here in that the original target is not respected. Either way though, thanks for opening this 👍 I'm afraid though that I have some more concerns around merging this 😅

I'm unsure of whether we'll merge this, since it might have a slight impact on performance and predictability. Inserting rules is quite a frequent operation and adding code to basically defeat other libraries is not quite in-scope, in my opinion, but a hack.

It does come up when other versions of SC clash, but that's quite a common issue for which we've got some workarounds, presuming that all duplicate versions of SC are new enough.

I'm general, I don't believe that every library working with the DOM should defend against other libraries aggressively altering elements 😅 we'll have a think and will leave se more comments on this PR.

I'll make sure that we port this over to canary, if we do go ahead with this 👍

stonebk:
> It does come up when other versions of SC clash, but that's quite a common issue for which we've got some workarounds, presuming that all duplicate versions of SC are new enough.

@kitten what are the workarounds? This is currently a huge headache for us since it completely breaks style insertion during development. Our current workaround is a forked implementation of styled-components with this fix implemented.

> I'm general, I don't believe that every library working with the DOM should defend against other libraries aggressively altering elements 😅 we'll have a think and will leave se more comments on this PR.

@ryanirilli can probably speak to this better than I can, but this isn't defending against other libraries, it's defending styled-components against itself. We have two dependencies on the page, both using styled-components on the same version -- styled-components should be aware enough not to clobber it's own style tag.

kitten:
Currently `styled-components` already defends against multiple versions being on one page; It [checks for the correct version attribute](https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/models/StyleSheet.js#L83) so in that case you're bundling two identical versions of `styled-components`?

Unfortunately we can't just readd the style tag and assume that everything is going to continue working correctly, since the removal of them indicates that one version of `styled-components` has rehydrated the styles of the other, which will probably lead to other subtle bugs.

The `canary` version of `styled-components` will defend against this by setting the `SC_ATTR` to `"active"` so that another version doesn't "consume" its styles.

ryanirilli:
@kitten the issue we face is that we have two apps (Webpack bundles) loading on the same page that are using the same version of styled components, thus, causing styled-components to go down the rehydrate path and removes the first style tag from the DOM at the [cleanup step](https://github.com/styled-components/styled-components/blob/c3eedbeb79621acd668e15ed2d63107528ed81c3/packages/styled-components/src/models/StyleTags.js#L422). 

However, the style tag node still exists and styles are being injected into it which led us to this fix. The logic here is, if you're going to keep injecting styles into this tag, it _needs_ to be in the DOM. 

I agree, it feels like a hack, but it is reasonable to make the check or at least throw an error here. See #2254 for other people experiencing this issue. 

Perhaps a better fix would be to mark the style tag as "seen" in `makeStyleTag` with a data attribute and remove that from the list [here](https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/models/StyleSheet.js#L83)

In any case, we just would like to figure out how we can support this and defend against this reasonably. Open to suggestions!

probablyup:
@ryanirilli you should use the CDN version of styled components in this case and set your webpack bundles to use the window alias instead

ryanirilli:
> @ryanirilli you should use the CDN version of styled components in this case and set your webpack bundles to use the window alias instead

We do actually use SSR though, is there a viable option that can support that as well?

kitten:
Ah, there's another solution to this though 👍 you can provide two different SC_ATTR constants for both apps which will cause them to use different attributes for their style tags https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/constants.js#L4

probablyup:
Another option is to pin each app to a different patch version of the library, but @kitten's recommendation is better since you then aren't potentially losing out on a bugfix

ryanirilli:
If Thank you @kitten @probablyup this is a great idea, I didn't realize we could define those vars via define plugin. We can definitely try that!

probablyup:
Either option means you're loading duplicate copies of the library though, which is unfortunate. If you use the CDN version for your client side bundles and NPM version for SSR it'll all work fine together (just make sure they're all the same version of s-c.)

ryanirilli:
100% agree, we are are in discussion about a long term solution to shared dependencies like this but we're looking for a short term solution at the moment.

kitten:
@ryanirilli thanks again for bringing this problem to us though! It's been mentioned a lot of times but I'll think about how we can make this less painful in version 5 👍

I'll close this issue for now, but please feel free to reach out with any questions you have here or on Twitter.

By the way, maybe a long term solution would be a shared DLL bundle via we packs DLLPlugin?

ryanirilli:
Thank you! and yes DLL plugin is top of the list at the moment but we haven't explored it fully yet. Thanks again all for your support.

ryanirilli:
@kitten It was discovered that this solution is in the docs [here](https://www.styled-components.com/docs/advanced#avoiding-conflicts-with-thirdparty-styles-and-scripts). 

Perhaps elevating that last paragraph could help. I realize it _is_ uncommon but since people are encountering it, the solution may be more discoverable in another section of the site.

brendonco:
> Ah, there's another solution to this though 👍 you can provide two different SC_ATTR constants for both apps which will cause them to use different attributes for their style tags https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/constants.js#L4

How do I set SC_ATTR if I'm not using create react cli?

PR #2501, opened 2019-04-15
Make possible to initialize SC_ATTR via REACT_APP_SC_ATTR env variable

oliverlaz:
When working with two or more different styled-components instances on a single page (or application), `SC_ATTR` environment variable could be used for distinguishing between those instances.

However, in create-react-app based applications, environment variables are filtered out by matching them with a well-known pattern (`REACT_APP_`) -> [see here](https://github.com/facebook/create-react-app/blob/bffc296a2c8967d46bed40d3120d82f2752496dc/packages/react-scripts/config/env.js#L73).

This PR enables usage of `REACT_APP_SC_ATTR` environment variable as a fallback and its easier configuration via a `.env` file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ All notable changes to this project will be documented in this file. If a contri
 _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/)._
 
 ## Unreleased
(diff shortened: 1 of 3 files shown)


probablyup:
Can you add this for the other env variable too? `SC_DISABLE_SPEEDY`

oliverlaz:
Sure! PR is updated.

probablyup:
@oliverlaz can you look into the CI failure?

oliverlaz:
Hey @probablyup, Somehow this DISABLE_SPEEDY flag messes up the integration test.
At first, I just thought it is a Travis issue but later I managed to reproduce it on my side as well (integration test suite requires a built version of styled-components).
I will ping you back once I find the culprit.

PR #2509, opened 2019-04-18
Clean up dependencies and configs in canary branch

kitten:
> **NOTE:** This is a PR against `canary` and will probably break / change some things, which is acceptable, since `canary` is not published yet.


.sail.yml
```diff
@@ -0,0 +1,70 @@
...
+      - GOOGLE_CHROME_BINARY: "/usr/bin/chromium-browser"
+      - PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: "true"
+  testWeb:
+    image: philpl/alpine-chrome-node:latest
+    command:
+      - yarn
+    args:
+      - test:web
+  testNative:
+    image: philpl/alpine-chrome-node:latest
```
probablyup: Can we use a public image? Just thinking if something happens it'll be really hard to fix without your context

kitten: Nope; I'll eventually move the Docker image somewhere else and this literally just includes `USER root` but there's no Docker image that works perfectly out of the box for us :disappointed:

kitten: Removed the integration tests for now which enables us to use the default Docker image for node. They don't seem to be running anyway.

Also removed native/primitives for now. They don't seem to work due to a react-native change and our rather weird Jest config 😅

Issue #1934, opened 2018-08-21
5.0 Roadmap


probablyup:
Test it out:

```shell
npm install styled-components@beta react@^16.8 react-dom@^16.8 react-is@^16.8
```

_5.0 milestone_

- [x] ~~rewrite stylesheet + tag classes for the needs of today (#2522)~~
- [x] ~~move react-is to peer dependencies (https://github.com/styled-components/styled-components/pull/2187)~~
- [x] ~~use react v16.7+ hooks to simplify the context consumer architecture (#2390)~~
- [x] ~~babel 7 migration (maybe... there were bundle size regressions before) (#2509)~~
- [x] ~~internal rewrite to use hooks (#2390)~~
- [x] ~~bump minimum versions of react, react-native to hooks-compatible versions~~
- [x] ~~remove code paths deprecated in v4 (https://github.com/styled-components/styled-components/pull/2604)~~
- [x] ~~fix cGS style clobbering issues (https://github.com/styled-components/styled-components/pull/2824)~~

_5.1 candidate_

- [ ] stylis -> sweetsour (depending on when it's ready for use and testing)
- [ ] figure out CSP / nonce support (#2363)
- [ ] solution to drop the prop whitelist (https://github.com/styled-components/styled-components/pull/2093)
- [ ] source maps for styles (https://github.com/styled-components/styled-components/issues/827)
- [ ] scoping of styles via StyleSheetManager (https://github.com/styled-components/styled-components/issues/1789)

_didn't do_

- [ ] ~~use React.warn and React.error if 16.9 comes out before v5 (https://github.com/styled-components/styled-components/issues/2584)~~ React removed them

follow along on the ~canary~ [v5 branch](https://github.com/styled-components/styled-components/tree/v5)

_evolving_

Andarist:
👋 From what I understand you are approaching v4 release and I would like to make a proposal for it. 
While

Andarist:
👋 From what I understand you are approaching v4 release and I would like to make a proposal for it. 

While prop whitelist is obviously non-ideal solution and has its costs (2kb+) the approach is used by more than 1 library, it would be great if we could just use the same module for it so the cost would be one time thing. 

I'd like to propose using [is-prop-valid](https://github.com/emotion-js/emotion/tree/master/next-packages/is-prop-valid). It's obviously used by emotion, but also by popmotion since recently and I know that if we could all agree on a single module @diegohaz would love to dedupe his copied `validAttr` from reakit.

[...]

PR #2514, opened 2019-04-20
Feature: toggling vendor styles on and off

aretecode:
- https://github.com/styled-components/styled-components/issues/719 (_toggling vendor prefixing on/off_)
- https://github.com/styled-components/styled-components/issues/1934 (_5.0 Roadmap_)

- This is implemented using `process.env`, though we could implement it many other ways, since stylis is instantiated immediately, this is the most straight forward & minimal way to achieve the feature, and will have no performance impact if a bundler defines the env & dead code elimination takes place. Please suggest if you have a preferred approach to achieving an implementation of this feature. It could also be moved to [constants](https://github.com/styled-components/styled-components/blob/canary/packages/styled-components/src/constants.js) file.

- A test suite for `stringifyRules` is created
- Two test cases are added, both of them use `jest.resetModules()` to ensure the `stringifyRules` file is evaluated with the updated process.env, only for this test case
1. To check the default functionality (having prefixes)
2. With the `STYLIS_SHOULD_PREFIX` env set, expect that there are no prefixes
- teardown/afterAll ensures the process.env is reset

### DIFF
--- a/packages/styled-components/src/utils/stringifyRules.js
+++ b/packages/styled-components/src/utils/stringifyRules.js
@@ -19,7 +19,7 @@ const stylis = new Stylis({
   global: false,
   cascade: true,
   keyframe: false,
-  prefix: true,
+  prefix: process.env.STYLIS_SHOULD_PREFIX === undefined,
   compress: false,
   semicolon: false, // NOTE: This means "autocomplete missing semicolons"
 });
--- a/packages/styled-components/src/utils/test/stringifyRules.test.js
+++ b/packages/styled-components/src/utils/test/stringifyRules.test.js
@@ -1,0 +1,41 @@
+// @flow
+
+/**
+ * @todo could do the same for `compress`
+ */
+describe('stringifyRules', () => {
+  afterAll(() => {
+    delete process.env.STYLIS_SHOULD_PREFIX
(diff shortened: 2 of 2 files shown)


aretecode:
to test integration, I've also published this to provide a codesandbox
https://codesandbox.io/s/ww11zzzmkw

Screen Shot 2019-04-20 at 3 46 18 PM

probablyup:
This actually requires a much different implementation... something we can perhaps do in v5 instead. Basically we need to initialize stylis dynamically for `StylesheetContext` and if you manually compose that provider then we can pass stylis-specific options there.

probablyup:
Superceded by #2565

aretecode:
@probablyup thanks for making it available! :) Going the dynamic route should make it easier to use, even if it's a little slower.

For ease of navigation for those who are interested in the lines changed as a diff:
- https://github.com/styled-components/styled-components/pull/2565/files#diff-086b9ead88eff464ee20bc94958af1a0R14
- https://github.com/styled-components/styled-components/pull/2565/files#diff-339db951123215c19cca8f07372c38ceR240

probablyup:
The speed difference will be imperceptible, no worries there

PR #2524, opened 2019-05-01
Optimise ComponentStyle name hashes

kitten:


packages/styled-components/src/utils/hasher.js
```diff
@@ -2,36 +2,15 @@
...
-      (a ^= c.charCodeAt(d) & 255),
-        (a = 1540483477 * (a & 65535) + (((1540483477 * (a >>> 16)) & 65535) << 16));
-  }
-  a ^= a >>> 13;
-  a = 1540483477 * (a & 65535) + (((1540483477 * (a >>> 16)) & 65535) << 16);
-  return (a ^ (a >>> 15)) >>> 0;
-}
+export const hash = (x: string): number => {
+  /* prettier-ignore */
+  for (var h = 5381 | 0, i = 0, l = x.length | 0; i < l; i++)
```
probablyup: what's this one?

kitten: That's djb2 which has acceptably low collision rates even compared with murmurhash, is also not a cryptographic hasher, and quite short https://gist.github.com/MohamedTaha98/ccdf734f13299efb73ff0b12f7ce429f

probablyup:
@kitten looking at the tests it seems like the placeholder class is receiving css as well where it should be an empty selector

probablyup:
Ok double injection is fixed but it's worth noting we're doing something a little different in the new sheet implementation here that will lead to a lot of test churn:

We're now attaching the static styles to the stable className instead of putting them into an additional className. Previously the stable className did not contain styles and was only used for targeting purposes. I think this is actually fine and what we ultimately want, but just wanted to call it out.

kitten:
I've tested this against a couple of large apps and it seems 100% safe to merge, so I'll go ahead and merge this into `feat/styled-sheet`

PR #2542, opened 2019-05-12
replace objectToCss with objectWithCssArray

d7my11:
Original PR #2489 
Fixes `objectToCss` didn't handle objects with methods.


probablyup:
@d7my11 if you merge from canary the tests should be happy

PR #2558, opened 2019-05-24
Fix theme detection in react native when no ThemeProvider is used

### LINKED ISSUE #2367: Theme prop and defaultProp on a styled component is not working on react native
## System:
 - OS: macOS 10.14.2
 - CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
 - Memory: 289.74 MB / 16.00 GB
 - Shell: 3.2.57 - /bin/bash
## Binaries:
 - Node: 8.15.0 - ~/.nvm/versions/node/v8.15.0/bin/node
 - npm: 6.4.1 - ~/.nvm/versions/node/v8.15.0/bin/npm

## Reproduction

I want to define theme props for a styled component, using the React Native version, but theme is always set as an *empty object*, even when I define a defaultProps or a prop to the component. 

*Bug possibly only on native, I tested on the browser and it works:* https://codesandbox.io/s/y2ko222zqv

I analysed the code and noticed a difference between native and browser versions on method `StyledNativeComponent.render` versus `StyledComponent.renderInner`

On browser it considers the `props.theme`:

```    
    if (componentStyle.isStatic) {
      generatedClassName = this.generateAndInjectStyles(EMPTY_OBJECT, this.props);
    } else if (theme !== undefined) {
      generatedClassName = this.generateAndInjectStyles(determineTheme(this.props, theme, defaultProps), this.props);
    } else {
      generatedClassName = this.generateAndInjectStyles(this.props.theme || EMPTY_OBJECT, this.props);
    }
```

while on Native it doesnt:

```       
       if (theme !== undefined) {
          var themeProp = determineTheme(_this2.props, theme, defaultProps);
          generatedStyles = _this2.generateAndInjectStyles(themeProp, _this2.props);
        } else {
          generatedStyles = _this2.generateAndInjectStyles(theme || EMPTY_OBJECT, _this2.props);
        }
```

Possibly very simple fix, I cant right now do a merge request and run tests, late here, I thought I was crazy since the behaviour wasnt working :)

## Steps to reproduce
- create a styled component for a native element like Text, using a `prop.theme` to style it, you can define defaultProps as well
- do not use a theme provider in the app
- use the styled component passing a theme object as a prop

Same as I did on sandbox, but import the native version and use any 

## Expected Behavior

- when you use a prop as theme
-- expect the styled component to use the theme props
- when you do not inform any theme prop
-- expect the styled component to use the defaultProps

## Actual Behavior
- when you use a prop as theme
-- sets an empty object as theme
- when you do not inform any theme prop
-- sets an empty object as theme

JelteF:
This fixes #2367, the previous code would result in the always being set to EMPTY_OBJECT. This brings it in line with the same code in [StyledComponent.js](https://github.com/styled-components/styled-components/blob/772280915f3ce342181d55b6b0e1f6fc5050b67f/packages/styled-components/src/models/StyledComponent.js#L130-L133).

### DIFF
--- a/packages/styled-components/src/models/StyledNativeComponent.js
+++ b/packages/styled-components/src/models/StyledNativeComponent.js
@@ -79,7 +79,7 @@ class StyledNativeComponent extends Component<*, *> {
             const themeProp = determineTheme(this.props, theme, defaultProps);
             generatedStyles = this.generateAndInjectStyles(themeProp, this.props);
           } else {
-            generatedStyles = this.generateAndInjectStyles(theme || EMPTY_OBJECT, this.props);
+            generatedStyles = this.generateAndInjectStyles(this.props.theme || EMPTY_OBJECT, this.props);
           }
 
           const propsForElement = {


JelteF:
For any other people running into this, I've published this patch to npm under: https://www.npmjs.com/package/@stream-io/styled-components

packages/styled-components/src/models/StyledNativeComponent.js
```diff
@@ -79,7 +79,7 @@ class StyledNativeComponent extends Component<*, *> {
             const themeProp = determineTheme(this.props, theme, defaultProps);
             generatedStyles = this.generateAndInjectStyles(themeProp, this.props);
           } else {
-            generatedStyles = this.generateAndInjectStyles(theme || EMPTY_OBJECT, this.props);
+            generatedStyles = this.generateAndInjectStyles(this.props.theme || EMPTY_OBJECT, this.props);
```
probablyup: This doesn't actually fix anything

packages/styled-components/src/models/StyledNativeComponent.js
```diff
@@ -79,7 +79,7 @@ class StyledNativeComponent extends Component<*, *> {
             const themeProp = determineTheme(this.props, theme, defaultProps);
```
probablyup: I'm wondering if this line should be pulled up above L78 and then use `themeProp` in L82 where you changed it to `this.props.theme`.

JelteF: I tried that first, but then I also had to change determineTheme, to support undefined as a theme. And for completeness you also need to use it like that in the respective place in `StyledComponent.js`. It does allow you to remove the if else statement in both places. I wanted to keep this PR small, to make it easy to get it reviewed and this bug fixed. So I went for the simple change that is the same as in `StyledComponent.js`

probablyup: Well we never want the theme value to be undefined, it should always be at least an empty object.

probablyup:
Closing in favor of #2567

PR #2563, opened 2019-05-29
remove className usage checker & suppression prop

### LINKED ISSUE #2154: React.StrictMode throws warning because of `findDOMNode` use
```
Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of StyledComponent which is inside StrictMode. Instead, add a ref directly to the element you want to reference.
```

I am aware that it is only run in development-mode, but it clutters console output when trying to find errors.

@Fer0x Could we change this here somehow? (by not using findDOMNode)

I don't think it is a problem, not even sure if there is a proper solution. Keeping it for later reference.

_Originally posted by @optunetobi in https://github.com/styled-components/styled-components/pull/2073/review_comment/create_

probablyup:
It ends up creating a ton of dev noise in nextjs due to usage of React loadable and breaks things like React strict mode for which there is no current corollary.

Fixes #2154


packages/styled-components/src/models/StyledComponent.js
```diff
@@ -74,7 +72,7 @@ class StyledComponent extends Component<*> {
           // eslint-disable-next-line no-console
           console.warn(
             `Functions as object-form attrs({}) keys are now deprecated and will be removed in a future version of styled-components. Switch to the new attrs(props => ({})) syntax instead for easier and more powerful composition. The attrs key in question is "${key}" on component "${displayName}".`,
-            `\n ${(new Error()).stack}`
```
mxstbr: Is there a reason why it was this way, maybe browser compat? Is this safe to change?

probablyup: idk, prettier removed it so I assume a safe transform

PR #2565, opened 2019-05-29
allow dynamic creation & configuration of stylis per stylesheet

### LINKED ISSUE #719: Consider toggling vendor prefixing on/off (stylis v3)

probablyup:
Closes #719 #844 
Enables #2514 

If this works the way I think it will, the stylis stringifier is now dynamically created with the new stylesheet and can be replaced at runtime via use of `StylesheetManager` as a React context provider.


packages/styled-components/src/utils/stylis.js
```diff
@@ -0,0 +1,98 @@
...
+  const stylis = new Stylis({
+    global: false,
+    cascade: true,
+    keyframe: false,
+    prefix: true,
+    compress: false,
+    semicolon: false, // NOTE: This means "autocomplete missing semicolons"
+
+    // user-given overrides
+    ...options,
```
mxstbr: We don't want users to be able to override the other options, right? E.g. if they set `global: true` or `cascade: false` that's bad, no?

probablyup: Messing with stylis in general is kind of "here be dragons" territory... definitely a pro-level functionality and anyone getting this deep would need to read up on what the options actually do.

probablyup:
> Can we add a test to make sure RTL support is possible with this patch?

Definitely. I want to get all the canary tests passing first before merging this so I have some assurance I'm not accidentally breaking anything.

mxstbr:
> I want to get all the canary tests passing first before merging this so I have some assurance I'm not accidentally breaking anything.

Good call.

packages/styled-components/src/models/StyleSheetManager.js
```diff
@@ -1,13 +1,15 @@
...
 import React, { useContext, useMemo, type Node, type Context } from 'react';
 import PropTypes from 'prop-types';
 import StyleSheet from '../sheet';
 import StyledError from '../utils/error';
+import createStylisInstance from '../utils/stylis';
 
 type Props = {
   children?: Node,
   sheet?: StyleSheet,
+  stylisMiddlewares?: Array<Function>,
```
kitten: Small nit but unfortunately the plural is just "middleware"; shall we call this "stylisPlugins" instead?

packages/styled-components/src/sheet/Sheet.js
```diff
@@ -13,6 +14,8 @@ let SHOULD_REHYDRATE = IS_BROWSER;
 class StyleSheet implements Sheet {
   isServer: boolean;
 
+  stringifier: Function;
```
kitten: It might make sense to type this

packages/styled-components/src/utils/stylis.js
```diff
@@ -0,0 +1,98 @@
...
+   * has no knowledge of static class so we have to intelligently replace the base selector.
+   */
+  const selfReferenceReplacementPlugin = (context, _, selectors) => {
+    if (context === 2 && selectors.length && selectors[0].lastIndexOf(_selector) > 0) {
+      // eslint-disable-next-line no-param-reassign
+      selectors[0] = selectors[0].replace(_selectorRegexp, selfReferenceReplacer);
+    }
+  };
+
+  stylis.use([...middlewares, selfReferenceReplacementPlugin, parseRulesPlugin, returnRulesPlugin]);
```
kitten: Same naming thing here

packages/styled-components/src/utils/stylis.js
```diff
@@ -0,0 +1,100 @@
+import Stylis from 'stylis/stylis.min';
```
StevenLangbroek: *feel free to ignore, reading up on what's going on*

This feels like it should be taken care of by the bundler, but I'm probably just missing some backstory here. Is there a reason this references the minified version?

probablyup: https://github.com/thysultan/stylis.js/blob/4561e9bc830fccf1cb0e9e9838488b4d1d5cebf5/package.json#L3

packages/styled-components/src/utils/stylis.js
```diff
@@ -0,0 +1,100 @@
...
+
+  // Wrap `insertRulePlugin to build a list of rules,
+  // and then make our own plugin to return the rules. This
+  // makes it easier to hook into the existing SSR architecture
+
+  let parsingRules = [];
+
+  // eslint-disable-next-line consistent-return
+  const returnRulesPlugin = context => {
+    if (context === -2) {
```
StevenLangbroek: What does `-2` stand for here? Could this maybe made clearer with an explicit mapping / var naming?

probablyup: Honestly, I'm not sure. The docs for stylis are kind of bad so it was trial and error to get it working. This isn't new code, just moved from a different file.

kitten: It's documented here https://github.com/thysultan/stylis.js#plugins

We might want to extract this to some named constants. `-2` is the last step in the plugin lifecycle, so it allows us to return a different result from the stylis function itself after it's done processing all CSS rules.

packages/styled-components/src/utils/stylis.js
kitten: `stylis` commits the minified JS file, so I think we switched over at some point since it yielded better results, i.e. was more minified than what another minifier compressing styled-components + stylis might achieve. But that might've changed to be fair.

packages/styled-components/src/models/ComponentStyle.js
```diff
@@ -81,7 +80,7 @@ export default class ComponentStyle {
       const name = generateName(dynamicHash >>> 0);
 
       if (!styleSheet.hasNameForId(componentId, name)) {
-        const cssFormatted = stringifyRules(css, `.${name}`, undefined, componentId);
+        const cssFormatted = styleSheet.stringifier(css, `.${name}`, undefined, componentId);
```
kitten: So, this might be a nit, but it'd seem like `stringifier` couples the new Sheet implementation to Stylis technically, or at least to some kind of `stylis`-like stringifier even if it's refactored, which makes the Sheet stuff on its own less reusable for others, once we make it a separate package.

Does it make sense to have an intermediate object that holds the sheet and the stringifier separately in the `StyleSheetManager` context?

probablyup: It would end up being a much bigger change... but also the stringification process is pretty closely linked with stylesheet injection so I don't think the coupling is actually a bad thing.

kitten: I mean, it's not a bad thing, but it will make it hard to extract `styled-sheet` without coupling it with stylis. We can always make this change to enable such an extraction separately though.

probablyup:
@kitten @mxstbr ok this should be ready to go. added tests for stylis options and plugins

packages/styled-components/src/models/StyleSheetManager.js
```diff
@@ -16,25 +17,32 @@ export const StyleSheetConsumer = StyleSheetContext.Consumer;
...
-    }
-  }, [sheet, target]);
+  return useMemo(
+    () => {
+      if (sheet) {
+        return sheet;
+      } else if (target) {
+        return new StyleSheet(false, target);
+      } else {
+        return masterSheet;
```
probablyup: when using `StyleSheetManager` if `sheet` isn't passed it should just fall back to masterSheet

probablyup:
Also ran benchmarks and there's no change 👍

packages/styled-components/src/utils/stylis.js
StevenLangbroek: Interesting. I imagine this happens often enough that looking this up on an object will incur a perf hit?

StevenLangbroek: Maybe just a comment linking to that section here would be enough?

PR #2569, opened 2019-05-29
Fix tests for canary changes

probablyup:
Feel free to hop on this PR and push up stuff @kitten and @mxstbr


probablyup:
@kitten how does this work? https://github.com/styled-components/styled-components/blob/0164a8d8122ffe5eaee7894851202231add1eac6/packages/styled-components/src/sheet/Rehydration.js#L25

I was trying to fix the SSR tests and was seeing some weird output in the style tags

packages/styled-components/src/models/StyledComponent.js
```diff
@@ -253,10 +240,11 @@ function useStyledComponentImpl<Config: {}, Instance>(
       props.className, // eslint-disable-line react/prop-types
       styledComponentId,
       attrs.className,
-      generatedClassName
+      generatedClassName !== styledComponentId ? generatedClassName : null
```
probablyup: for a static style use case the generated name seems to just be the static component id, so without this the className is added twice

packages/styled-components/src/models/test/StyleSheetManager.test.js
```diff
@@ -202,51 +203,6 @@ describe('StyleSheetManager', () => {
     div.parentElement.removeChild(div);
   });
 
-  describe('ssr', () => {
-    it('should extract CSS outside the nested StyleSheetManager', () => {
```
probablyup: don't really understand what this test is trying to accomplish since refs don't work in an SSR use case

packages/styled-components/src/test/styles.test.js
```diff
@@ -291,28 +291,4 @@ describe('with styles', () => {
       expect(el.getAttribute('nonce')).toBe('foo');
     });
   });
-
-  it('should handle deferredInject and inject correctly', () => {
```
probablyup: AFAIK this code path no longer exists, so just removing the test

kitten:
@probablyup Regarding how the SSR output markers work;

```js
`${SC_ATTR}.g${group}[id="${id}"]`
```

This is a group marker.
It contains the group's number (group ordering) and the component ID. It then also adds the group's names. It is inserted _after_ a group/component's rules.

Compare this to what we had in v4:

```js
`/* sc-component-id: ${id} */`
```

So in v4 we would inject a single comment _before_ a group/component's rules. The names however were stored separately on the `data-styled` tag, where as now in v5 they're stored inside this group marker.

This allows us to use the `CSSStyleSheet` (i.e. speedy) API to rehydrate rules, which is much faster than parsing and traversing the rules manually with custom logic. We also save ourselves the need for a "stylis splitter" in this case.

Having the names non-globally also allows us to have more precise caching and being able to remove a group's names when the group is deleted, which is nice for global styles.

The `group` number in the new marker is also added so that we can rehydrate the ID to group number mapping.

probablyup:
@kitten would you mind taking a crack at those rehydration and ssr tests? Feeling lost even with that explainer TBH

kitten:
@probablyup Done! :tada:

packages/styled-components/src/constructors/test/createGlobalStyle.test.js
```diff
@@ -230,26 +234,15 @@ describe(`createGlobalStyle`, () => {
 
     render(<Comp />);
     const el = document.querySelector('[data-test-el]');
-    expect(getCSS(document).trim()).toMatchInlineSnapshot(`
-      "/* sc-component-id:sc-global-3005254895 */
```
kitten: all these snapshots basically look at the inserted CSS in non-speedy mode. In canary we never insert any comment markers anymore, so those are gone. The rest of the CSS remains identical

packages/styled-components/src/constructors/test/createGlobalStyle.test.js
```diff
@@ -332,7 +317,7 @@ describe(`createGlobalStyle`, () => {
     );
 
     expect(console.warn.mock.calls[0][0]).toMatchInlineSnapshot(
-      `"The global style component sc-global-2176982909 was given child JSX. createGlobalStyle does not render children."`
+      `"The global style component sc-global-a was given child JSX. createGlobalStyle does not render children."`
```
kitten: Canary switched `GlobalStyle` over to use the same name generator, hence we don't see hashes here or in other snapshots anymore

packages/styled-components/src/constructors/test/keyframes.test.js
```diff
@@ -7,12 +7,15 @@ import keyframes from '../keyframes';
 import Keyframes from '../../models/Keyframes';
 import { expectCSSMatches, getCSS, resetStyled } from '../../test/utils';
 
-/**
- * Setup
- */
+// Disable isStaticRules optimisation since we're not
+// testing for ComponentStyle specifics here
+jest.mock('../../utils/isStaticRules', () => () => false);
```
kitten: In some files I'm disabling this so we minimise changes to the snapshot. I've only done it for files that would've had a lot of changes otherwise and that don't care about the `isStatic` optimisation

packages/styled-components/src/models/test/StyleSheetManager.test.js
kitten: yep, looks kind of useless. Might've tested CSS insertion into different targets, but it doesn't look like it does this correctly anyway

packages/styled-components/src/test/constants.test.js
```diff
@@ -83,17 +83,17 @@ describe('constants', () => {
...
     it('should be true in test NODE_ENV', () => {
       process.env.NODE_ENV = 'test';
-      renderAndExpect(true, '.b { color:blue; }');
+      renderAndExpect(true, '.sc-a { color:blue; }');
     });
 
     it('should be true in development NODE_ENV', () => {
       process.env.NODE_ENV = 'development';
-      renderAndExpect(true, '.b { color:blue; }');
+      renderAndExpect(true, '.sc-a { color:blue; }');
```
kitten: This is actually a neat change! Every change here is a case where `isStatic` is enabled and effective

packages/styled-components/src/test/ssr.test.js
```diff
@@ -304,6 +235,14 @@ describe('ssr', () => {
       color: green;
     `;
 
+    // These create a long chunk of (hopefully) uninterrupted HTML
+    const elements = new Array(100).fill(0).map((_, i) => (
+      <div key={i}>*************************</div>
+    ));
+
+    // This is the result of the above
+    const expectedElements = '<div>*************************</div>'.repeat(100);
```
kitten: The snapshot before was just too damn big

probablyup:
Woot! Let's get canary green 💚

PR #2570, opened 2019-05-29
chromatic visual regression testing

probablyup:
Builds on #2458 by setting up a storybook for each example and then using chromatic to automatically snapshot and provide visual regression testing


mxstbr:
@kitten switched us to sail.ci on the canary branch, maybe worth targeting that for this patch?

probablyup:
Closing in favor of #2589

Issue #2545, opened 2019-05-13
"as" prop is breaking rebass component usage


mikedloss:
## Environment
`npx envinfo --system --binaries --npmPackages styled-components,babel-plugin-styled-components --markdown --clipboard`
```
## System:
 - OS: macOS High Sierra 10.13.6
 - CPU: (8) x64 Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
 - Memory: 138.20 MB / 16.00 GB
 - Shell: 5.3 - /bin/zsh
## Binaries:
 - Node: 8.11.2 - /usr/local/bin/node
 - Yarn: 1.13.0 - /usr/local/bin/yarn
 - npm: 6.7.0 - /usr/local/bin/npm
 - Watchman: 4.9.0 - /usr/local/bin/watchman
```

## Reproduction
https://codesandbox.io/s/56pvp6y5x

## Steps to reproduce
I created 2 components directly viewed in the codesandbox link - 1) a custom "passthrough" component that is wrapped by a `styled` tag, and 2) a Rebass component wrapped by a `styled` tag. In the codesandbox link, I have a few examples. The issue comes up when the `as` prop is being used.

## Expected Behavior
Both the passthrough and rebass components should be styled when being passed styling props (according to rebass/styled-system documentation), regardless of there being an `as` prop being passed to either component.

## Actual Behavior
The `as` prop is causing the 1st component to not have any styling, OR render the passthrough component at all.

probablyup:
Please use the new `forwardedAs` prop introduced in 4.3.x 😄 

https://www.styled-components.com/docs/api#forwardedas-prop

probablyup:
So when you wrap a custom component (non primitive, aka "div" or non-styled-component) in our HOC, e.g. `styled(CustomComponent)` if you use the `"as"` prop, that actually totally replaces the underlying component being rendered similar to the `.withComponent` API.

Since rebass components are styled-components, wrapping a rebass `Box` in `styled()` is totally fine, however in the case of your `BoxOne` component which introduces an interim function component, it's no longer a styled-component and you need to use the new `forwardedAs` prop to send your intended `"as"` down to the lower component layer.

Here's your sandbox updated: 
https://codesandbox.io/s/styled-components-rebass-bug-18fnp?fontsize=14

probablyup:
[...]

PR #2573, opened 2019-05-30
add forwardedAs prop

probablyup:
This works around the most common issue with #439 and #2545  around integrating the `styled()` HOC with other components that also accept an "as" prop

### DIFF
--- a/packages/styled-components/src/test/props.test.js
+++ b/packages/styled-components/src/test/props.test.js
@@ -1,4 +1,5 @@
 // @flow
+/* eslint-disable react/prop-types */
 import React from 'react';
 import TestRenderer from 'react-test-renderer';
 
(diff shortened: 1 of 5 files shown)


probablyup:
Just a followup after an internal convo: they're orthogonal issues so `shouldForwardProp` would be a separate implementation that doesn't fix the primary issue in the OP.

packages/styled-components/src/test/props.test.js
```diff
@@ -38,4 +41,18 @@ describe('props', () => {
...
   });
+
+  it('should forward the "as" prop if "innerAs" is used', () => {
+    const Comp = ({ as: Component = 'div', ...props }) => <Component {...props} />;
+
+    const Comp2 = styled(Comp)`
+      background: red;
+    `;
+
+    expect(TestRenderer.create(<Comp2 innerAs="button" />).toJSON()).toMatchInlineSnapshot(`
```
StevenLangbroek: Does this test really confirm that it fixes the bug. The issue was never that "as" didn't get applied, it's that styling was dropped from the originally `styled` component...

probablyup: I added an additional assertion to the test to make sure the className contains the correct style.

PR #2581, opened 2019-06-01
add seal fn back to ServerStyleSheet for backward compat

probablyup:


packages/styled-components/src/models/ServerStyleSheet.js
```diff
@@ -110,4 +101,8 @@ export default class ServerStyleSheet {
 
     return readableStream.pipe(transformer);
   }
+
+  seal() {
+    this.sealed = true;
```
probablyup: doesn't really change anything as far as I can see, but just here for API backward compat

probablyup:
@kitten any comments on this one?

packages/styled-components/src/models/ServerStyleSheet.js
```diff
@@ -66,7 +58,7 @@ export default class ServerStyleSheet {
     }
 
     return <style {...props} />;
-  }
+  };
 
   interleaveWithNodeStream(input: any) {
```
kitten: I guess we do want to readd setting seal to true. But we can get away with setting it right at the top before the stream and only throwing an error here I think

probablyup: @kitten want to push onto my PR? According to the tests sealed is being set correctly in an error scenario so not 100% what you want

probablyup:
@kitten ok I refactored ServerStyleSheet a tad to allow what we need while still enforcing the error scenarios against the public API

PR #2582, opened 2019-06-01
refactoring & allow disabling CSSOM injection via StyleSheetManager

probablyup:


packages/styled-components/src/constants.js
```diff
@@ -5,7 +5,8 @@ declare var __VERSION__: string;
 
 export const SC_ATTR = (typeof process !== 'undefined' && process.env.SC_ATTR) || 'data-styled';
 export const SC_ATTR_ACTIVE = 'active';
-export const SC_VERSION_ATTR = 'data-styled-version';
+export const SC_ATTR_MODE = 'data-styled-injection-mode';
```
probablyup: I think this will be useful going forward for debugging purposes too :)

packages/styled-components/src/sheet/Sheet.js
```diff
@@ -22,6 +22,8 @@ class StyleSheet implements Sheet {
 
   tag: void | GroupedTag;
 
+  useCSSOM: boolean;
```
kitten: Lets maybe make this part of the constructor and refactor it to take a prams object?

Are we also calling this CSSOM? We could continue to call it speedy injection or use "useLegacyInjection" or sth

packages/styled-components/src/sheet/dom.js
```diff
@@ -20,15 +20,16 @@ const findLastStyleTag = (target: HTMLElement): void | HTMLStyleElement => {
...
   const head = ((document.head: any): HTMLElement);
   const parent = target || head;
   const style = document.createElement('style');
   const prevStyle = findLastStyleTag(parent);
   const nextSibling = prevStyle !== undefined ? prevStyle.nextSibling : null;
 
   style.setAttribute(SC_ATTR, SC_ATTR_ACTIVE);
-  style.setAttribute(SC_VERSION_ATTR, SC_VERSION);
+  style.setAttribute(SC_ATTR_VERSION, SC_VERSION);
+  style.setAttribute(SC_ATTR_MODE, useCSSOM ? 'cssom' : 'text');
```
kitten: Do we really need this tag in production?

probablyup: Nope, I'll remove it. Just something I thought might be useful down the line if more injection modes end up happening for debugging purposes.

packages/styled-components/src/sheet/Sheet.js
probablyup: I definitely think calling it CSSOM is more obvious... "speedy" is cute but an unofficial name and not helpful if people are trying to google for our usage of it you know?

probablyup: > Lets maybe make this part of the constructor and refactor it to take a prams object?

It'd break backward compat. We can look into that for v6 though.

kitten: Will it though? The Stylesheet was a secret internal and I'd expect it to break anyway since most people will use it for the old static reset method.

I do think it's reasonable to break backwards compatibility here. Jest- styled-components will need to be changed anyway

probablyup: It's not really a secret internal, we allow people to make their own sheet to supply to `StyleSheetManager` in v3+.

probablyup: Ok nevermind it's actually in that spooky ghosts object so I guess it is. Alrighty.

packages/styled-components/src/models/StyleSheetManager.js
```diff
@@ -17,40 +18,56 @@ export const StyleSheetConsumer = StyleSheetContext.Consumer;
...
-  if (props.stylisOptions || props.stylusPlugins) {
-    sheet.stringifier = createStylisInstance(props.stylisOptions, props.stylusPlugins);
-  }
+      return sheet;
+    },
+    [
+      props.disableCSSOMInjection,
+      props.sheet,
+      props.stylisOptions,
+      props.stylusPlugins,
```
kitten: That's a typo here and the above and this are likely to cause issues here if people forget to keep their references equal. Given the severity of some people accidentally making this mistake, should we just make these constant on mount?

probablyup:
@kitten refactor complete :-)

packages/styled-components/src/models/StyleSheetManager.js
probablyup: Thanks for catching this 🤦‍♂

probablyup: @kitten done in the latest commit 👍

PR #2585, opened 2019-06-02
use shallowequal for equivalency checking instead

probablyup:
fixes `stylisPlugins` triggering the warning unnecessarily

### DIFF
--- a/packages/styled-components/src/models/StyleSheetManager.js
+++ b/packages/styled-components/src/models/StyleSheetManager.js
@@ -1,6 +1,7 @@
 // @flow
 import React, { useContext, useMemo, useState, type Node, type Context } from 'react';
 import PropTypes from 'prop-types';
+import shallowequal from 'shallowequal';
 import StyleSheet from '../sheet';
 import createStylisInstance from '../utils/stylis';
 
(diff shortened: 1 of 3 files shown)


packages/styled-components/src/models/StyleSheetManager.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 import React, { useContext, useMemo, useState, type Node, type Context } from 'react';
 import PropTypes from 'prop-types';
+import shallowequal from 'shallowequal';
```
kitten: Remains to be seen whether this is correctly minified away I suppose

yarn.lock
```diff
@@ -1922,7 +1922,32 @@ babel-code-frame@^6.26.0:
     esutils "^2.0.2"
     js-tokens "^3.0.2"
 
-babel-core@^6.26.0, babel-core@^7.0.0-bridge.0:
+babel-core@^6.26.0:
```
kitten: Seems like the `yarn.lock` is broken. Have you maybe forgotten to run `yarn` in the root? 😅

packages/styled-components/src/models/StyleSheetManager.js
probablyup: I checked and it gets DC'd properly due to the NODE_ENV wrapper 👍

PR #2586, opened 2019-06-02
prevent component folding when interim HOCs are present

probablyup:


packages/styled-components/src/test/basic.test.js
```diff
@@ -338,9 +372,7 @@ describe('basic', () => {
       expect(Named2.styledComponentId).toBe('Name-bar');
     });
 
-    // this no longer is possible in React 16.6 because
-    // of the deprecation of findDOMNode; need to find an alternative
-    it.skip('should work in StrictMode without warnings', () => {
+    it('should work in StrictMode without warnings', () => {
```
probablyup: this works now because of removing the className checker, so unskipped it

PR #2588, opened 2019-06-02
peer dependency & globals adjustments for bundle size

probablyup:
Brings our bundle size down to ~14.2 kB


probablyup:
@kitten any comments on this one?

mxstbr:
If I understand this correctly, this will mean people have to `npm install --save prop-types styled-components` to make styled-components work? That doesn't seem very ideal given that prop-types for styled components are basically useless, I'd much rather remove them?

probablyup:
@mxstbr like remove them entirely? wouldn't that be sad for dev feedback?

mxstbr:
I have never seen a proptype error from a styled component. Our factory functions make it pretty much impossible to mess them up?!

kitten:
I'd say our Prop Types aren't really that useful anymore but if we want to keep them we can also move the package back to dependencies and exclude it from the bundle manually in our config

probablyup:
That works, will make the change

PR #2589, opened 2019-06-02
 implement visual regression testing via storybook + chromatic

probablyup:
Something's up with the integration, reached out to their Spectrum community and waiting to hear back.


packages/styled-components/src/utils/error.js
```diff
@@ -1,24 +1,7 @@
...
-
-      module.exports = md.split(/^#/gm).slice(1).reduce((errors, str) => {
-        const [, code, message] = str.split(/^.*?(\\d+)\\s*\\n/)
-        errors[code] = message
-
-        return errors;
-      }, {});
-    `
-    : {};
+const ERRORS = process.env.NODE_ENV !== 'production' ? errorMap : {};
```
probablyup: Double checked that this is DCE'ing correctly.

probablyup:
For whatever reason `babel-plugin-preval` stopped working with the react native tests. 🤷‍♂ 

I replaced it with a general script and tried to weave it in to all the typical places you'd want it to be automatically run.

PR #2598, opened 2019-06-05
update flow, implement defaultProps folding for v5

probablyup:
Copied from the v4 PR:

> I specifically chose the merge-anything library because of its default handling for non plain-object types (replace, not merge.) It's about 500b gzipped so not too bad of an addition and I'm not sure we could do much better hand-rolling it.


.flowconfig
```diff
@@ -30,9 +30,6 @@
 .*/integration-test/.*
 .*/sandbox/.*
 
-[libs]
```
probablyup: flow gets run from the top anyway, so might as well just keep the types at the root of the repo

PR #2612, opened 2019-06-09
allow usage of css tagged template inside style objects

### LINKED ISSUE #2561: Creating keyframes using styled objects
Hello, thank you for maintaining Styled Components.

I would like to echo the sentiment shared in this [Spectrum thread](https://spectrum.chat/styled-components/help/how-can-i-use-a-custom-animation-with-the-style-object-syntax~91d624ea-a5a3-4d0e-9314-4da860059112): creating custom animations using styled objects. This is something I would like to use in styled-components.

Here is an example using styled objects with `keyframes`, but requiring an interpolated string when consuming the animation: https://codesandbox.io/s/animation-objects-b6h6h

```jsx
const animation = keyframes({
  from: {
    opacity: 0
  },
  to: {
    opacity: 1
  }
})

// Keyframes must be consumed as an interpolated string 
const Component = styled.div`
  animation: ${animation} 3s;
`
```

Please let me know if there's anything I can do to assist.

probablyup:
fixes #2561

### DIFF
--- a/packages/styled-components/src/utils/flatten.js
+++ b/packages/styled-components/src/utils/flatten.js
@@ -9,40 +9,62 @@ import Keyframes from '../models/Keyframes';
 import hyphenate from './hyphenateStyleName';
 import addUnitIfNeeded from './addUnitIfNeeded';
 
+type ObjectToCSSArrayProps = {
+  chunk: Object,
+  executionContext?: Object,
+  prevKey?: string,
+  styleSheet?: Object,
+};
+
 /**
  * It's falsish not falsy because 0 is allowed.
  */
-const isFalsish = chunk => chunk === undefined || chunk === null || chunk === false || chunk === '';
+function isFalsish(chunk: any): boolean {
+  return chunk === undefined || chunk === null || chunk === false || chunk === '';
+}
 
+function objToCssArray(args: ObjectToCSSArrayProps): Array<string | Function> {
+  const rules = [];
+  const keys = Object.keys(args.chunk);
 
-const objToCssArray = (obj: Object, prevKey?: string): Array<string | Function> => {
-  const rules = []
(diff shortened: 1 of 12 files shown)


packages/styled-components/src/utils/flatten.js
```diff
@@ -9,40 +9,62 @@ import Keyframes from '../models/Keyframes';
...
-  keys
-    .forEach(key => {
-      if (!isFalsish(obj[key])) {
-        if (isPlainObject(obj[key])) {
-          rules.push(...objToCssArray(obj[key], key))
+    if (!isFalsish(value)) {
+      if (Array.isArray(value) /* likely usage of `css` tagged template literal */) {
+        /* eslint-disable no-use-before-define */
+        return rules.push(
+          `${hyphenate(key)}: ${flatten(value, args.executionContext, args.styleSheet).join('')};`
```
d7my11: not sure about `.join` part actually.
After this change, this should be acceptable and if it's, then i think this will cause a bug.

```js
styled({
  h1: [{
        color: $(props) => props.color
    }]
})
```

we call `flattern` and will end up in `objToCssArray` with an object `{ color: $(props) => props.color }` which will return

```js
['color:', func, ...]
```
and with `.join` will cause the previous bug #2489

d7my11: actually tested it against your branch 
```diff
+   "h1: color:function color(props) {
+           return props.color;
+         };;",
```

also, one more thing to consider, we still need `objectToCssArray` to returns an array for `isStatic` flag.

my suggesting is to keep `objectToCssArray` context-free and split `flatten` handling of the array which is going to be context-free as well, and reuse it in both.

mario-jerkovic:
Is this still WIP and do you need help to finish this PR?

PR #2614, opened 2019-06-11
Throw invariant violation when ref'd theme not found in props

### LINKED ISSUE #2605: [Request] Optional invariant violation when `theme` not found in props
## Background

First off, great job with the library! We use `styled-components` for our internal component library and as a result, a lot of end users of our library don't know the implementation details of our components (and nor should they). The problem, however, is that because almost all components use the `theme` prop, not including the `ThemeProvider` at the top level of the application results in errors that appear like "Cannot find `colors.primary` of undefined". This isn't super clear to users who aren't familiar with how our library is structured (or that it even uses `styled-components` under the hood) and results in a lot of confused Slack messages.

## Request / Idea

Similar to how `react-router-dom` handles it (i.e. `Error: Invariant failed: You should not use <Route> outside a <Router>`), it would be great if there was a way to enable invariant violations when the `theme` prop is not found. Ideally, the error message itself would be customizable since in our case, we don't use the `ThemeProvider` component directly but as part of another component.

If there is any interest in adding a feature like this, I could get a PR going. Lmk!

haldunanil:
closes #2605


packages/styled-components/src/models/StyledComponent.js
```diff
@@ -10,12 +10,14 @@ import React, {
   type AbstractComponent,
   type Ref,
 } from 'react';
+import invariant from 'tiny-invariant';
```
probablyup: Can you just use console.warn please

packages/styled-components/src/utils/hasThemeReferences.js
```diff
@@ -0,0 +1,31 @@
+// @flow
+import { Parser } from 'acorn';
```
probablyup: There has to be a lighter way of doing this, like calling `.toString()` on the interpolation function and looking for `.theme`. We definitely won't want to pull in all of acorn for development

probablyup:
created #2655 as a much more minimal version of this PR

PR #2615, opened 2019-06-11
added a failing test for an 'as' prop bug

shane935:
This is a failing test for a bug I noticed the other day while using the 'as' prop to pass styles down to a react-router link component.  

When you use 'as' to restyle a non-html component it passes down non-standard props to the underlying dom element. Which results in the following react error:
> React does not recognize the `fullWidth` prop on a DOM element.

here is a simplified code sandbox for the issue: https://codesandbox.io/s/flamboyant-bsnns-bsnns

As I am writing this I am realizing there probably isn't much that can be done without potentially breaking custom components that pass non-standard props down multiple levels.  The only solution I can think of is adding an option to strip non-standard props from custom components to the `styled` function.

```
styled(Link, {stripNonStandardProps: true})
```

Happy to give solving this a shot if someone can point me in the right direction.

### DIFF
--- a/packages/styled-components/src/test/expanded-api.test.js
+++ b/packages/styled-components/src/test/expanded-api.test.js
@@ -142,7 +142,7 @@ describe('expanded api', () => {
       expect(TestRenderer.create(<Comp as={Override} />).toJSON()).toMatchSnapshot();
     });
 
-    it.only('does not pass down non standard props to custom component', () => {
+    it('does not pass down non standard props to custom component', () => {
       const Override = props => <figure {...props} />;
       const Comp = styled.div`
         color: red;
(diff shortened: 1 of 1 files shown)


probablyup:
Not a bug, intended behavior.

shane935:
Fair enough, do you have a suggested approach to remove the React error?

probablyup:
I’d either ignore it or restructure your components such that the
styled-component is the lowest level so it is the primitive and the prop
filtering takes place.

Any time we wrap a custom component we always pass all props.

Agnis-Scale3c:
is this going to be addressed any time soon? can't wrap my head around how can an error be intended behaviour?

probablyup:
There isn't a bug here, it's intended behavior. When the rendering target is a custom component and not a primitive HTML element we pass all props because there's no way to determine if a prop is only used for styling or not at this time.

zarabotaet:
Can you delete props which used in styled component even they potentially may used in nested component?

PR #2621, opened 2019-06-15
fix inconsistency between client & server that breaks rehydration

probablyup:


packages/styled-components/src/test/__snapshots__/ssr.test.js.snap
```diff
@@ -1,30 +1,30 @@
...
 
-exports[`ssr should extract both global and local CSS 1`] = `"<h1 class=\\"sc-b c\\">Hello SSR!</h1>"`;
+exports[`ssr should extract both global and local CSS 1`] = `"<h1 class=\\"sc-b\\">Hello SSR!</h1>"`;
 
 exports[`ssr should extract both global and local CSS 2`] = `
 "<style data-styled data-styled-version=\\"JEST_MOCK_VERSION\\">body{background:papayawhip;}
 data-styled.g1[id=\\"sc-global-a\\"]{content:\\"sc-global-a,\\"}
-.c{color:red;}
-data-styled.g2[id=\\"sc-b\\"]{content:\\"c,\\"}
+.sc-b{color:red;}
```
probablyup: This behavior is now consistent with how things render client-side. All static styles go on the base class and dynamic classes are only made for function interpolations 👍

PR #2622, opened 2019-06-15
lower the threshold for the "too many classes" warning

probablyup:
If your component is dynamic enough to have 20+ styling permutations
it's probably time to inline some stuff or the stylesheet will
be massive.


probablyup:
@mxstbr well it's not about speed necessarily, it's just a good practice not to emit an unnecessarily large amount of CSS

probablyup:
@kitten what's a good number then? 200 seems excessively high

probablyup:
@kitten pushed it up to 50, let's merge and see if people complain?

PR #2623, opened 2019-06-16
fix HMR

probablyup:


probablyup:
All the tests pass so I don’t think it will change the hashes unless
there’s an actual style change?

I tested it in my nextjs setup and it works. Also ran the benchmarks and
they seemed unaffected, so I think we’re ok.

PR #2625, opened 2019-06-16
Add useStyle hook

migueloller:
Initial implementation of `useStyle` hook as discussed with @probablyup [here](https://twitter.com/probablyup/status/1140030878069985281) and [here](https://github.com/styled-components/styled-components/issues/2342#issuecomment-456121567).

There might be some Babel plugin work yet to be done. I'm mostly looking for feedback to get this to a mergeable state.

### DIFF
--- a/packages/styled-components/src/models/StyledComponent.js
+++ b/packages/styled-components/src/models/StyledComponent.js
@@ -6,6 +6,7 @@ import React, {
   createElement,
   useContext,
   useDebugValue,
+  useRef,
   type AbstractComponent,
   type Ref,
 } from 'react';
@@ -23,15 +24,16 @@ import hasher from '../utils/hasher';
 import { ThemeContext } from './ThemeProvider';
 import { useStyleSheet } from './StyleSheetManager';
 import { EMPTY_ARRAY, EMPTY_OBJECT } from '../utils/empties';
+import css from '../constructors/css';
 
-import type { Attrs, RuleSet, Target } from '../types';
+import type { Attrs, RuleSet, Target, Styles, Interpolation } from '../types';
 
 /* global $Call */
 
 const identifiers = {};
 
 /* We depend on components having unique IDs */
-function generateId(displayName: string, parentComponentId: string) {
+function generateId(displayName?: string, parentComponentId?: string) {
   const name = typeof displayName !== 'string' ? 'sc' : escape(displayName);
   // Ensure that no displayName can lead to duplicate componentIds
   identifiers[name] = (identifiers[name] || 0) + 1;
@@ -175,6 +177,21 @@ function useStyledComponentImpl<Config: {}, Instance>(
   propsForElement.ref = refToForward;
 
   return createElement(elementToBeCreated, propsForElement);
+}
+
+export function useStyle(styles: Styles, ...interpolations: Array<Interpolation>): string {
+  // We have to force interpolations to be functions because there's no way of knowing wether
+  // a prop is being interpolated. And if it is a prop but not a function, `generateAndInjectStyles`
+  // will fail to insert the new rule.
+  const rules = css(
+    isFunction(styles) ? styles : () => styles,
+    ...interpolations.map(i => (isFunction(i) ? i : (): Interpolation => i))
+  );
+  const id = useRef(generateId());
+  const componentStyle = new ComponentStyle(rules, id.current);
+  const theme = useContext(ThemeContext) || EMPTY_OBJECT;
+
+  return useInjectedStyle(componentStyle, false, { theme });
 }
 
 export default function createStyledComponent(
--- a/packages/styled-components/src/constructors/styled.js
+++ b/packages/styled-components/src/constructors/styled.js
@@ -1,6 +1,6 @@
 // @flow
 import constructWithOptions from './constructWithOptions';
-import StyledComponent from '../models/StyledComponent';
+import StyledComponent, { useStyle } from '../models/StyledComponent';
 import domElements from '../utils/domElements';
 
 import type { Target } from '../types';
@@ -13,3 +13,4 @@ domElements.forEach(domElement => {
 });
 
 export default styled;
+export { useStyle };
--- a/packages/styled-components/src/index.js
+++ b/packages/styled-components/src/index.js
@@ -1,5 +1,5 @@
 // @flow
-import styled from './constructors/styled';
+import styled, { useStyle } from './constructors/styled';
 
 export * from './base';
-export { styled as default };
+export { styled as default, useStyle };


probablyup:
@mxstbr I wouldn't say that's a core tenet, but we've tried hard to abstract it away. In this case though I think the concept is solid and it provides a good migration path for emotion users for instance that are using emotion's `css` export for example.

packages/styled-components/src/models/StyledComponent.js
```diff
@@ -177,6 +179,21 @@ function useStyledComponentImpl<Config: {}, Instance>(
...
+  // will fail to insert the new rule.
+  const rules = css(
+    isFunction(styles) ? styles : () => styles,
+    ...interpolations.map(i => (isFunction(i) ? i : (): Interpolation => i))
+  );
+  const id = useRef(generateId());
+  const componentStyle = new ComponentStyle(rules, id.current);
+  const theme = useContext(ThemeContext) || EMPTY_OBJECT;
+
+  return useInjectedStyle(componentStyle, false, { theme });
```
probablyup: Let's export `useInjectedStyle` and then isolate `useStyle` into its own file. Maybe it's time for us to have a `hooks` directory 🤔

mxstbr:
I'm really not a fan of including this in the core lib, it doesn't really fit. Providing an easy migration path is a good idea though...

Let's have it as a separate package, `use-styled-hook` for example, which uses our new StyleSheet under the hood? That way, migration is easy but we are also not encouraging this behaviour by default!

probablyup:
@mxstbr It needs ComponentStyle though which is an internal

migueloller:
I don't mind implementing this as a separate package. I initially made a PR agains't this repo in @probablyup's guidance.

I can see the bug change in the public API exposing a class for the first time. Although one could also say that it's already exposed to any component that gets styled (i.e., the `className` prop).

If the hook were to be implemented outside of the main repository, there might need some work to be done to expose all necessary internals. Like @probablyup mentioned, the hook currently relies on `ComponentStyle`.

packages/styled-components/src/models/StyledComponent.js
migueloller: While working on this I expected there to be a `hooks` folder. If it's something y'all are open to then I would gladly make one and centralize hooks there. Thoughts?

mxstbr:
Yeah I think I would rather make the core (StyleSheet + whatever else is necessary) modular enough to where something like this hook can live externally. At that point, styled-components exposes the "blessed" API that hides classes, and anybody else can build further APIs on top of the same core to avoid bundle size bloat!

Does that make sense? What exactly would this change need exposed?

packages/styled-components/src/models/StyledComponent.js
probablyup: Definitely open to that, maybe a separate PR?

probablyup:
@mxstbr We'd need to expose the `useInjectedStyle` hook and also ComponentStyle. We could put them under `__PRIVATE__`.

migueloller:
Yeah, the way it's implemented now it's very tied to how Styled Component works internally and some work would have to be done to expose low-level APIs that make sense by themselves.

mxstbr:
I think that's preferable to including this in core—if we can expose sensible low-level APIs that one can use to build something like this useStyle hook, that means anybody can invent new APIs and we can trial new features in an external package before we include them in core! 😍 

Do you want to start planning that @migueloller? I think you might have the best idea of what you'd need, since you are looking at everything with fresh eyes.

probablyup:
Alright @migueloller so I guess this PR can just be reduced to exposing `ComponentStyle` inside `__PRIVATE__`

ryota-murakami:
I was reading this issue from out side of context though, I totally agree @mxstbr's opinion.
But I don't know how much difficult is it internal code level.
I'm hoping we'll getting great direction about this 🤗

probablyup:
@migueloller created https://github.com/styled-components/use-styled-hook and added you as an admin :)

migueloller:
@probablyup, sweet! Thanks for doing that.

By the way, are we renaming the hook to `useStyled`? Or will the package be called `use-styled-hook` but the hook will be a named export called `useStyle`? Or should I rename the repository/package to `use-style-hook`?

probablyup:
We'd only export default the one hook I assume, keep the lib neat and tidy. You can call a default export whatever you like so in the docs I'd say we call it `useStyle` but keep the name of the lib be `use-styled-hook` since "styled" mentally maps a little more cleanly to "styled-components". Open to other naming though if people think it'll be confusing.

alexandernanberg:
Minor nit, why not naming the package `use-styled`? Would be cleaner imo

yordis:
@mxstbr I had been using `styled-components` for such a long time, and I would agree 100% with you that allowing classes is not what I would support.

Before joining my current employer, I had the opportunity to do a lot of green-field projects where I didn't have to worry about classes since everything was in `styled-components.`

I didn't need it at all.

But today,

The only reason I will have to go with `jss` or `material-UI/styles` (is pretty much the setup I will end up with anyway) is because of [makeStyles](https://material-ui.com/styles/api/#makestyles-styles-options-hook) hook.

We are trying to eliminate any Webpack CSS loader, and embrace CSS-in-JS.

But we can't remove the loader if we don't have something like this feature.

Since for those few cases where we are forced to create classes we need this, and we don't want to overwrite styles by creating CSS specificity.

I know that you would prefer to create another package, but I am worried about two things:

1. Who will own this package?

I would propose to use `@styled-components/[my name]` if that is possible.

And keep the package under Styled Components organization umbrella.

2. Do we want to end up with tiny packages for everything?

We don't have to embrace classes since would go against this package, but I think that this is just a matter of education.

With proper documentation and leadership, I don't see any issue.

This feature will allow styled-components to be introduced in legacy code easier than before since these use cases will be present in some cases.

Adding more packages put a lot of burden in consumers where it is pretty low if we keep in the styled-components package.

Just imagine if we have styled-components-native and styled-components.

Probably the best decision you made for the ecosystem was to keep them both together.

I would re-evaluate this decision, I couldn't agree more with you, this goes against the idea of styled-components, and we shouldn't embrace it.

But I see the other side of the coin where legacy codes and the perfect scenario doesn't exist.

And I would prefer to take ownership of educating people and show them how they could embrace the technology in those cases than creating the mini-package chaos.

wmertens:
While I agree that s-c mostly disappears `className`, devs still need to be aware of it, and still need to pass `className` to the DOM when they are wrapping a component with `styled()`.

Furthermore, this is a tiny addition and tree shaking removes it completely. So it would be a bonus for people that want/need it.

Tangentially though, I am struggling to find an example where creating a unique className with something like [useId](https://www.npmjs.com/package/react-use-id-hook) and using regular styled components would not suffice @yordis. Can you elaborate?

yordis:
Hey folks, any updates on this?

probablyup:
Going to close this since it's been decided it won't be in core. @migueloller feel free to open a followup PR to expose the relevant internal so this can be built in userland.

PR #2633, opened 2019-06-20
fix css props when using babel macro with babel-plugin-styled-components 1.10.1

jamesknelson:
Currently, if you follow the guide to using styled-components with Create React App, `css` props will not work. This is due to a [change](https://github.com/styled-components/babel-plugin-styled-components/commit/e3829d28f2b460e097f1499f6091a52b667ef1b4#diff-1fdf421c05c1140f6d71444ea2b27638L13) introduced in babel-plugin-styled-components 1.10.1.

This fix resolves the issue by switching the macro to use `@babel/traverse(path)` instead of `path.traverse()`. For some reason, this makes the `Program` visitor work, whereas it doesn't with `path.traverse()`.

See https://github.com/styled-components/babel-plugin-styled-components/issues/226 for details


probablyup:
@jamesknelson can you add a test please?

jamesknelson:
I can't get the tests running on my machine, so can't do so at the moment. I get this when trying to run `yarn test:web` or `jest`

```
 ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import styled from '../styled';
SyntaxError: Unexpected identifier "import"
```

Looks like the tests are not compiling modules to commonjs for some reason. Any ideas?

probablyup:
@jamesknelson hmm weird, a `yarn` from the root always works for me

jamesknelson:
I'm not sure what's going on - I've run `yarn` in the root, tried a bunch of things with a fresh clone, and can't get tests to work locally no matter what I try.

On the other hand, there is a [macro test](https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/macro/test/macro.test.js) already, and it passes on CI.

probablyup:
I’ll add one to your branch

probablyup:
Released as 4.3.2, thanks so much for tackling this!

PR #2636, opened 2019-06-20
Stop creating invalid CSS when using @-ms-viewport along with createGlobalStyle

cameron-martin:
Currently when using:

```
createGlobalCss`
@-ms-viewport { width: device-width; }
`
```

The following CSS gets injected:

```
@-ms-viewport{{ width: device-width;}}
```

This is invalid CSS and causes JSDOM to print a long error message.

This is due to a bug in stylis (https://github.com/thysultan/stylis.js/pull/148) and this PR only contains a failing test and can be fixed by updating the stylis package once it's fixed there.

### DIFF
--- a/packages/styled-components/src/utils/test/stringifyRules.test.js
+++ b/packages/styled-components/src/utils/test/stringifyRules.test.js
@@ -1,0 +1,7 @@
+import stringifyRules from '../stringifyRules';
+
+it('stringifys @-ms-viewport correctly', () => {
+  const result = stringifyRules(['@-ms-viewport { width: device-width; }'], '');
+
+  expect(result).toEqual(['@-ms-viewport{width: device-width;}']);
+});


probablyup:
Sure, please mark it `.skip` as we don't keep failing tests active in the repo unless we can actually fix them. I'd also make this against canary instead.

cameron-martin:
My intention was to keep this PR open and update it with the new version of stylis once it is released to fix the failing test. However, I can mark it with skip so it can get merged sooner if you prefer.

probablyup:
TBH you might be waiting a while :(

probablyup:
@cameron-martin idk if you're still around or using s-c but in the current `main` branch we're using stylis v4

cameron-martin:
Oh cool and this is fixed in stylis v4. I'll rebase the PR tomorrow and see if it passes (it's real late here atm).

kitten:
If it's fixed in `stylis@4` then I'll go ahead and close this since we already upgraded and this isn't a test we'll need in an upstream repo, so to speak.

PR #2640, opened 2019-06-21
[rfc] use @emotion/stylis

probablyup:
Drops the bundle size by a fair amount (12.1kB vs 13.62kB)... the downside is you
[can't customize stylis options anymore](https://www.npmjs.com/package/@emotion/stylis) but still can supply plugins.

I think it's an acceptable compromise for the bundle size win?

### DIFF
--- a/packages/styled-components/src/utils/stylis.js
+++ b/packages/styled-components/src/utils/stylis.js
@@ -1,6 +1,6 @@
 import Stylis from '@emotion/stylis';
 import _insertRulePlugin from 'stylis-rule-sheet';
-import { EMPTY_ARRAY } from './empties';
+import { EMPTY_ARRAY, EMPTY_OBJECT } from './empties';
 
 const COMMENT_REGEX = /^\s*\/\/.*$/gm;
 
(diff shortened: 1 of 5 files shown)


packages/styled-components/src/utils/stylis.js
```diff
@@ -11,21 +11,8 @@ export type Stringifier = (
...
-  options: Object = EMPTY_OBJECT,
-  plugins: Array<Function> = EMPTY_ARRAY
-) {
-  const stylis = new Stylis({
-    global: false,
-    cascade: true,
-    keyframe: false,
-    prefix: true,
-    compress: false,
-    semicolon: false, // NOTE: This means "autocomplete missing semicolons"
```
probablyup: the only change from this current set of options is `semicolon` is turned on

probablyup:
AFAIK the only setting I've heard someone want to customize was "prefix", and that actually is still customizable in this custom stylis build... instead of the `stylisOptions` prop we started out with for the beta we could instead switch it to something like `disableVendorPrefixes` for that one use case.

probablyup:
Updated to use the new prop for `StyleSheetManager`, bundle size is now 12.18kB still a nice win!

mxstbr:
Did we remove `stylisOptions`?

probablyup:
Yup!

packages/styled-components/src/utils/stylis.js
kitten: I think that's technically a breaking change that might affect plenty of apps without reasonable errors/warnings in place 😅

probablyup:
It’s just ASI for CSS right?

thysultan:
Yes, please don't directly expose a `stylisOptions` to the end user that directly maps 1 to 1 with current stylis options in order to make it easier to upgrade to the next major version(breaking change) of stylis.

probablyup:
@thysultan Good to know, thanks!

zeitamin:
is `disableVendorPrefixes` supposed to work in rc3?

packages/styled-components/src/utils/stylis.js
kricsleo: I've counted with this problem when updating from 4.x to 5.x just because of this config change.  'Not autocomplete missing semicolons' makes some styles can't be compiled right as before. Though I can add semicolons manually when I'm writing new code, but I'm not sure how many old codes are effected by this change. Can we change the default value of `semicolon` to `true`? Though It might cause a little growing in bundle size, but we can get a more smoothing migration experience from 4.x to 5.x.

PR #2652, opened 2019-06-27
remove the concept of foldedComponentIds

probablyup:
since we literally concat the styles together and make an entirely
new component, we don't have to care about what the layers were
that came before and in v5 it actually causes issues to keep that
information around because the base layer could be mounted later
than the overridden component and cause specificity issues

### DIFF
--- a/packages/styled-components/src/test/expanded-api.test.js
+++ b/packages/styled-components/src/test/expanded-api.test.js
@@ -152,6 +152,8 @@ describe('expanded api', () => {
         color: red;
       `;
 
+      Comp.displayName = 'SomethingFun';
+
       const Comp2 = styled(Comp)`
         color: green;
       `;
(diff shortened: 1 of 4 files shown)


packages/styled-components/src/test/expanded-api.test.js
```diff
@@ -160,14 +162,31 @@ describe('expanded api', () => {
...
+                  `);
+
+      expect(TestRenderer.create(<Comp3 as="span" />).toJSON()).toMatchInlineSnapshot(`
+                        <span
+                          className="sc-c f"
+                        />
+                  `);
+
+      expect(getCSS(document)).toMatchInlineSnapshot(
+        `".d{background:blue;color:red;}.e{background:blue;color:red;color:green;}.f{background:blue;color:red;color:green;text-align:center;}"`
```
probablyup: verifying here that the styles are still concatenating correctly for each layer

probablyup:
Me neither 🙈

PR #2655, opened 2019-06-28
 add lightweight dev warning when theme is consumed but not provided

probablyup:
alternate fix for #2605

### DIFF
--- a/packages/styled-components/src/models/StyledComponent.js
+++ b/packages/styled-components/src/models/StyledComponent.js
@@ -134,7 +134,7 @@ function useStyledComponentImpl<Config: {}, Instance>(
   );
 
   if (process.env.NODE_ENV !== 'production' && forwardedComponent.usesTheme && !theme) {
-    console.warn(
+    console.error(
       `Component ${
         // $FlowFixMe
         forwardedComponent.displayName
--- a/packages/styled-components/src/models/StyledComponent.js
+++ b/packages/styled-components/src/models/StyledComponent.js
@@ -27,6 +27,7 @@ import type { Attrs, RuleSet, Target } from '../types';
 
 /* global $Call */
 
+const THEME_PROP_R = /\.theme[.[]/;
 const identifiers = {};
 
 /* We depend on components having unique IDs */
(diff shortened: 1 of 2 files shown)


haldunanil:
@probablyup Curious why you wanted to go with `console.warn` instead of throwing. Wouldn't it be easier, especially when you get a stacktrace on your screen, to see exactly what caused the error?

probablyup:
Throwing is a bit aggressive IMO, could be swayed to make it a console error though. Also hopefully React will release 16.9 or a 16.8 patch soon that ships their warn and error utils that actually log the component stack for you :)

haldunanil:
Ok that'll be dope! In that case `console.error` sounds better since it's not really a warning at that point as you're actually trying to reference something that the consumer component sees as being `undefined`, which will impede the way the component works as opposed to a warning which to me says "hey, this isn't great but carry on", which is not the case here-- you really should address asap.

haldunanil:
Also would you mind if I updated my PR with your suggestions so I can have that coveted `Contributor` badge next to my name 😝?

probablyup:
@haldunanil sure, this one is just for v5 anyway so you can do the corresponding version for v4

haldunanil:
Cool, will open a new PR for that later today

packages/styled-components/src/models/StyledComponent.js
```diff
@@ -261,6 +277,10 @@ export default function createStyledComponent(
   });
 
   if (process.env.NODE_ENV !== 'production') {
```
kitten: So this is probably just a mildly controversial comment but I think we've previously been a bit heavy on warnings.

Yes, they may catch potential issues early, but at some point we'll have to think of the likelihood of them occurring.

Do we really think that this is going to provide value over a stack trace that is already clearly pointing to the same interpolation that could be causing this issue when a theme was missing?

haldunanil:
@probablyup new PR for v4.0: https://github.com/styled-components/styled-components/pull/2657

packages/styled-components/src/models/StyledComponent.js
probablyup: There will only be a stack trace if they access a deeper property like `theme.colors.red`. We backfill `theme` with an empty object so `theme.color` wouldn't trigger a hard error, it'd just be undefined in the styles so the console error definitely would help catch bugs with more subtle usage.

packages/styled-components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "styled-components",
-  "version": "5.0.0-beta.5",
+  "version": "5.0.0-beta.6",
```
kitten: Accidental bump in a PR?

probablyup: Just needs a rebase - it's a stray commit but one that's already in the canary branch 🤷‍♂

PR #2656, opened 2019-06-28
fix component selectors + css prop usage

probablyup:
This PR reverts the change to remove `foldedComponentIds` and reintroduces some v4 behavior to not inject styles into the static component classes.

Without these changes components targeted via component selector that have been wrapped in any way (css prop, styled() hoc, etc.) will not be found for style application which is a regression from v4.

In order to reintroduce `foldedComponentIds` we need to go back to the previous behavior in v4 of not injecting styles into the component stable ssr className so that it remains available for placeholder / targeting usage only.

### DIFF
--- a/packages/styled-components/src/models/ComponentStyle.js
+++ b/packages/styled-components/src/models/ComponentStyle.js
@@ -49,7 +49,7 @@ export default class ComponentStyle {
       }
 
       const cssStatic = flatten(this.rules, executionContext, styleSheet).join('');
-      const name = generateName(phash(this.baseHash, cssStatic) >>> 0);
+      const name = generateName(phash(this.baseHash, cssStatic.length) >>> 0);
 
       if (!styleSheet.hasNameForId(componentId, name)) {
         const cssStaticFormatted = styleSheet.options.stringifier(
--- a/packages/styled-components/src/models/ComponentStyle.js
+++ b/packages/styled-components/src/models/ComponentStyle.js
@@ -17,12 +17,15 @@ export default class ComponentStyle {
 
   componentId: string;
 
+  hasInjected: boolean;
+
   isStatic: boolean;
 
   baseHash: number;
 
   constructor(rules: RuleSet, componentId: string) {
     this.rules = rules;
+    this.hasInjected = false;
     this.isStatic = process.env.NODE_ENV === 'production' && isStaticRules(rules);
     this.componentId = componentId;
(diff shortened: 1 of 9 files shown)


probablyup:
blah incomplete fix, need to test some more -_-

probablyup:
@kitten ok this is good to go, tested in my SSR setup with component selectors etc and it's working as expected

probablyup:
Going to rebase to clean up the tree

probablyup:
Available for test at `5.0.0-beta.6-ej2`

packages/styled-components/src/models/ComponentStyle.js
```diff
@@ -41,19 +44,27 @@ export default class ComponentStyle {
...
 
     if (this.isStatic) {
-      if (!styleSheet.hasNameForId(componentId, componentId)) {
-        const cssStatic = flatten(this.rules, executionContext, styleSheet).join('');
+      if (this.staticRulesId && styleSheet.hasNameForId(componentId, this.staticRulesId)) {
+        return this.staticRulesId;
+      }
+
+      const cssStatic = flatten(this.rules, executionContext, styleSheet).join('');
+      const name = generateName(phash(this.baseHash, cssStatic) >>> 0);
```
kitten: I like how this still preserves the phash optimisation 👍 but given that in prod we know that this can never change and this path of isStatic will never be taken, isn't it safe just to generate a name from the base hash itself?

I know this isn't going to provide much entropy so we could hash the length of cssStatic itself as well. I'm just concerned that this might regress on some of the performance gains we've made

probablyup: like `phash(this.baseHash, 1)`? or just basehash by itself?

kitten: Like ‘phash(this.baseHash, cssStatic.length)’

probablyup:
@kitten tried it and looks good 👍

probablyup:
@kitten rebased and dropped the last commit. Doesn't materially affect bundle size and I'd rather be safe than sorry on hashing stuff

Burdigalien:
Hello, 

I still encounter HASH confusions on beta8. 
Components rendered only on the client side, after a click for example, have a hash collisions to other components on the page, rendered on SSR.

probablyup:
@Burdigalien please open a new issue with a reproduction, thanks!

PR #2663, opened 2019-07-02
show the multi instance warning for all envs

probablyup:
Going to add a section to FAQ about deduping the library within
environments like jest. Not sure why this was so tightly restricted
before because it's a totally valid warning even for browser-like
SSR use cases.


probablyup:
ping @kitten and/or @mxstbr

PR #2672, opened 2019-07-10
Import React Native components using asterisk syntax to support case …

armancharan:
…where React Native has no default export [https://www.notion.so/armancharan/React-Native-Styled-Components-Error-ec66cbc1980e424da3987bd60eb103a9]


armancharan:
Overlooked the fact that asterisk syntax would import all exports by default.

PR #2702, opened 2019-08-07
Update error message with clickable link.

unixchad:


probablyup:
Thanks for the fast turnaround!

PR #2706, opened 2019-08-07
Get rid of the ThemeProvider single child context restriction

### LINKED ISSUE #1325: Remove React.Children.only from theme provider
It there a reason to have [`React.Children.only`](https://github.com/styled-components/styled-components/blob/master/src/models/ThemeProvider.js#L109) in the theme provider? It seems unnecessary and it makes certain layouts not possible. Why not simply return `this.props.children` from render?

vkrol:
We can get rid of that restriction because ThemeProvider uses stable React Context API and children cannot be a top-level element.
Fixes #1325.
The documentation will be updated too.


vkrol:
Sure, https://github.com/styled-components/styled-components/pull/2708.

PR #2708, opened 2019-08-08
Get rid of the ThemeProvider single child context restriction (canary)

vkrol:
This is the follow-up to https://github.com/styled-components/styled-components/pull/2706.


vkrol:
@probablyup No problem!

PR #2721, opened 2019-08-18
Use ownerDocument instead of global document

yamachig:
It seems that on Microsoft Edge, you cannot appendChild an Element from different Document.
This causes a problem, for example, when you are using StyleSheetManager with a new window created by window.open, like this code:
```jsx
import React from 'react';
import { createPortal } from 'react-dom';
import styled, {StyleSheetManager} from 'styled-components';

export const NewWindow = props => {
  const [newWindowBody, setNewWindowBody] = React.useState(null);
  const unloaded = React.useRef(false);

  React.useEffect(() => {
    if (!unloaded.current && !newWindowBody) {
      const w = window.open("about:blank");

      w.addEventListener('load', () => {
        setNewWindowBody(w.document.body);
      });

      w.addEventListener('unload', () => {
        unloaded.current = true;
        setNewWindowBody(null);
      });

      setNewWindowBody(w.document.body);
    }
  });

  return newWindowBody && createPortal(
    <StyleSheetManager target={newWindowBody}>
      {props.children}
    </StyleSheetManager>,
    newWindowBody,
  );
}

const Button = styled.button`
  border: none;
  background: palevioletred;
  color: white;
`;

export const TestNewWindow = () => {
  return (
    <NewWindow>
      <Button>
        Hello from new window!
      </Button>
    </NewWindow>
  )
}
```
This code throws an error on Edge, but not on Chrome or Firefox. (I tested with Microsoft Edge 44.18362.267.0.)

I replaced the global document variables with dynamically created Document from ownerDocument so that the code works properly on Edge.

### DIFF
--- a/packages/styled-components/src/models/StyleTags.js
+++ b/packages/styled-components/src/models/StyleTags.js
@@ -58,7 +58,11 @@ const addUpUntilIndex = (sizes: number[], index: number): number => {
 
 /* create a new style tag after lastEl */
 const makeStyleTag = (target: ?HTMLElement, tagEl: ?Node, insertBefore: ?boolean) => {
-  const el = document.createElement('style');
+  let targetDocument = document;
+  if(target) targetDocument = target.ownerDocument;
+  else if(tagEl) targetDocument = tagEl.ownerDocument;
+
+  const el = targetDocument.createElement('style');
   el.setAttribute(SC_ATTR, '');
   el.setAttribute(SC_VERSION_ATTR, __VERSION__);
 
@@ -68,7 +72,7 @@ const makeStyleTag = (target: ?HTMLElement, tagEl: ?Node, insertBefore: ?boolean
   }
 
   /* Work around insertRule quirk in EdgeHTML */
-  el.appendChild(document.createTextNode(''));
+  el.appendChild(targetDocument.createTextNode(''));
 
   if (target && !tagEl) {
     /* Append to target when no previous element was passed */
@@ -226,7 +230,7 @@ const makeSpeedyTag = (el: HTMLStyleElement, getImportRuleTag: ?() => Tag<any>):
   };
 };
 
-const makeTextNode = id => document.createTextNode(makeTextMarker(id));
+const makeTextNode = (targetDocument, id) => targetDocument.createTextNode(makeTextMarker(id));
 
 const makeBrowserTag = (el: HTMLStyleElement, getImportRuleTag: ?() => Tag<any>): Tag<Text> => {
   const names = (Object.create(null): Object);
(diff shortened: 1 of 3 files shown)


probablyup:
@yamachig could you add a changelog entry and if possible also submit the same PR against canary?

Thanks!

yamachig:
@probablyup thank you for reviewing! OK, I [added a changelog entry](https://github.com/styled-components/styled-components/pull/2721/commits/f5559c1b9e6ff082d8b109f55a5296e6aaa455c6) and submitted the same PR #2726 against canary.

packages/styled-components/src/models/StyleTags.js
```diff
@@ -58,7 +58,11 @@ const addUpUntilIndex = (sizes: number[], index: number): number => {
 
 /* create a new style tag after lastEl */
 const makeStyleTag = (target: ?HTMLElement, tagEl: ?Node, insertBefore: ?boolean) => {
-  const el = document.createElement('style');
+  let targetDocument = document;
+  if(target) targetDocument = target.ownerDocument;
+  else if(tagEl) targetDocument = tagEl.ownerDocument;
```
Jessidhia: I'd write this as:

`const targetDocument = target ? target.ownerDocument : tagEl ? tagEl.ownerDocument : document`

but that's just my anti-mutability bias 🤷‍♀

packages/styled-components/src/models/StyleTags.js
```diff
@@ -58,7 +58,11 @@ const addUpUntilIndex = (sizes: number[], index: number): number => {
...
+
+  const el = targetDocument.createElement('style');
```
Jessidhia: Interesting that this causes no security issues; apparently any `<style>` tags' associated `StyleSheet` are considered same-origin. https://html.spec.whatwg.org/multipage/semantics.html#the-style-element:concept-css-style-sheet-origin-clean-flag

This can still break for CSP reasons, though, but it'd break regardless of whether we use CSSOM or child text nodes.

Jessidhia:
Looks like the travis config / vm is way broken.

packages/styled-components/src/models/StyleTags.js
yamachig: @Jessidhia thank you for reviewing! I agree with anti-mutability, though, the eslint rule (no-nested-ternary) didn't let me do so :)

probablyup:
@yamachig rebase when you get a chance

yamachig:
@probablyup OK, I rebased. Is that correct?

probablyup:
Yeah looks like CI is passing now, good.

PR #2735, opened 2019-09-06
Update .travis.yml

Jessidhia:
Trusty has been EOLed even in Travis, and the GPG keys in it were old enough that things were failing to install.

Try to update it to the new stuff.

Looks like the `bundlesize` command is printing some errors, and there's an unfixed lint warning on master that is being ignored as it's a `warn` not `error`, but otherwise seems fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,12 @@
 language: node_js
-dist: trusty
-sudo: required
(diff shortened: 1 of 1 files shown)


probablyup:
Thanks!

PR #2737, opened 2019-09-06
[canary] attrs should take precedence over props

probablyup:


rudin:
What's the reasoning behind this? I use attrs like this a lot:
const Content = styled(Text).attrs(() => ({textStyle: 'body'}))
And when using this component, I could eventually override this padding prop, like so:
<Content textStyle="caption">

Would this change mean that this is no longer possible?!

probablyup:
@rudin need a more complete example so I can answer your question. `attrs` overrides props, but doesn't replace them.

rudin:
Hi Evan,
Here's a quick example:
https://codesandbox.io/s/crazy-flower-9bwdm
It's using styled-components 4.4. I'm noticing (see uncommented line 22) that 'textStyle' as defined in render does not override 'textStyle' as specified in attrs.
I made the changes as seen on lines 23-25: I'm destructuring 'textStyle' from props and set a default: 'body'. This is nice.

I have to admit that I was expecting the uncommented line to work 🤪Anyway, is there anything going to change in the code as being written in this codesandbox?

probablyup:
@rudin I'm still not entirely sure I understand your question but I'll just say this is now working as it's actually explained in the docs. `attrs` is meant to take in props and adapt/override/add new ones based on the return value.

PR #2745, opened 2019-09-10
DISABLE_SPEEDY can be set to "false" in test environment

bredy:
- allow to control DISABLE_SPEEDY through global variable and environment variable even in case we would like to set it to false


kitten:
Closing due to staleness; also this wouldn't be evaluated statically by all minifiers

PR #2758, opened 2019-09-19
Yarn add

victorhermes:
I added the option to install with yarn in the README.md


probablyup:
Thanks but we chose to keep the instructions in the repo to the lowest common denominator, which is npm at this time.

PR #2761, opened 2019-09-23
fix: add unit test for createGlobalStyle when style tag is detached from DOM and then remove the rules

shomukai:
The same fix for canary:
https://github.com/styled-components/styled-components/pull/2707

In canary, the sheet instance is saved in constructor, so this.sheet always exists when trying to delete CSS rules.
So no code change needed, just add the unit test.


newying61:
@probablyup 
Np, thank you. :)

Issue #2340, opened 2019-01-21
Implement 'useTheme' using Hooks API


markoradak:
Hey there! 👋

I was wondering if there are plans on implementing Hooks API for specific parts regarding context providers? This would help us to use something like `useTheme()` outside of context of styled components, where we're currently using either `withTheme` HOC, or a render props pattern.

The implementation could probably be as simple as following:
```js
const useTheme = (...path) => path.join(".").split(".")
  .reduce((a, c) => (a && a[c] ? a[c] : null), useContext(ThemeContext));
```
and it would allow us to do something like:
```jsx
<Component background={useTheme('colors.blue')} />
```

I guess this has already been thought of, but since I couldn't find any issue related to it, I wanted to get your opinion on it.

I'd be more than happy to do a PR including something like this, maybe under something such as `styled-components/hooks` for now, until 16.8 (hopefully 🤓).

mxstbr:
We want to switch our entire theming setup to use hooks to avoid the DevTools bloat that comes with the `ThemeConsumer` components!

Great idea about the `useTheme` hook, had not thought of that. I will open a general "Let's use hooks!" issue with this included, give me two minutes.

SeanRoberts:
For anyone looking for a dead simple hook while we wait for something official:

```
import { useContext } from 'react';
import { ThemeContext } from 'styled-components';

export default function useStyledTheme() {
  const theme = useContext(ThemeContext);
  return theme || {};
}
```

hon2a:
```javascript
import { useContext } from 'react'
import { ThemeContext } from 'styled-components'
import { get, identity } from 'lodash/fp'

export const useTheme = path => (path ? get(path) : identity)(useContext(ThemeContext) ?? {})
```

wmertens:
I'm reopening this, since #2343 does not include this.

I would like to see this as part of styled-components v5 and not a helper library.

I prefer the simplest version `const useTheme = () => useContext(ThemeContext)`:
* if you want to get some item at some path, you can use plain js or get or wrap the hook
* passing a theme to merge (`const currentTheme = useTheme({color: 'red'})`) is not possible, because it would require wrapping the component with a provider

lcswillems:
Hooks are more and more used and it would be nice to include a `useTheme` hook in the library. I had to create my own this is not convenient.

SeanRoberts proposed a working example:
```
import { useContext } from 'react';
import { ThemeContext } from 'styled-components';

export default function useStyledTheme() {
[...]

PR #2765, opened 2019-09-24
Added useTheme hook

semoal:
Added a simple hook to recover theme context easily

### DIFF
--- a/packages/styled-components/src/base.js
+++ b/packages/styled-components/src/base.js
@@ -18,6 +18,9 @@ import ThemeProvider, { ThemeContext, ThemeConsumer } from './models/ThemeProvid
 /* Import Higher Order Components */
 import withTheme from './hoc/withTheme';
 
+/* Import hooks */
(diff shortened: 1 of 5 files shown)


kachkaev:
Related: https://github.com/styled-components/styled-components/issues/2340

probablyup:
Seems fine to me, though perhaps unnecessary since we expose `ThemeContext` already

lcswillems:
I don't think it is unnecessary. It makes import easier.

For example, in Next.js, there is a `useRouter` hook. They could have not included it in Next.js and just let people import `RouterContext` but it makes more annoying to import it.

A lot of libraries add this little hook to make context easier to import. I think it would make sense to do it in styled components.

Thank you for all your work!

lcswillems:
I don't think it is unnecessary since it makes imports easier.

Thank you for all your great work!

mxstbr:
I think this is a perfectly fine API addition, let's :shipit:!

PR #2771, opened 2019-09-28
[POC] Implement rule hoisting for import/font-face rules

kitten:
> **NOTE:** This is only proof-of-concept quality code

This implements rule hoisting by adding two new abstractions to the `styled-sheet` implementation:

- `WindowedTag`: A wrapper around any Tag that creates a linked list of Tags that all operate on the same underlying tag, with the same interface
- `HoistingTag`: A wrapper around `GroupedTag` that creates two `GroupedTag`s with one `WindowedTag` each internally that are used to separate hoisted from non-hoisted rules

This should work, but I havent tested it yet. It highlights that some things need to be cleaned up. The output logic is pretty ugly now and needs to be put onto the `GroupedTag` itself (?). Not quite sure, but maybe this is fine as it is.

The interfaces worked well when there were three layers, but now that there are five, it seems like an unnecessary addition that leads to indirection. Maybe the code structure can be improved, but that may also be out-of-scope for this PR.

Edit: The SSR output needs to avoid the usual markers for hoisted rules.

### DIFF
--- a/packages/styled-components/src/sheet/HoistedTag.js
+++ b/packages/styled-components/src/sheet/HoistedTag.js
@@ -4,19 +4,19 @@ import type { HoistedTag, GroupedTag, Tag } from './types';
 import { DefaultGroupedTag } from './GroupedTag';
 import { WindowedTag } from './WindowedTag';
 
-const hoistedRuleRe = /\s*@(font-face|import)/;
+const hoistedRuleRe = /\s*@import/;
 
 export class DefaultHoistedTag implements HoistedTag {
   hoistedTag: GroupedTag;
 
   normalTag: GroupedTag;
 
-  length: number;
+  groups: number;
 
   constructor(tag: Tag) {
     this.hoistedTag = new DefaultGroupedTag(new WindowedTag(tag));
     this.normalTag = new DefaultGroupedTag(new WindowedTag(tag));
-    this.length = 0;
+    this.groups = 0;
   }
 
   insertRules(group: number, rules: string[]): void {
@@ -34,17 +34,18 @@ export class DefaultHoistedTag implements HoistedTag {
 
     this.normalTag.insertRules(group, normalRules);
     this.hoistedTag.insertRules(group, hoistedRules);
-    this.length = this.normalTag.length;
+    // We just use the normalTag groups
+    // since the normalTag and hoistedTag have the same groups
+    this.groups = this.normalTag.groups;
   }
 
   clearGroup(group: number) {
     this.normalTag.clearGroup(group);
     this.hoistedTag.clearGroup(group);
-    this.length = this.normalTag.length;
+    this.groups = this.normalTag.groups;
   }
 
-  // TODO: Needs to be replaced to keep hoisted ordering
-  getGroup(group: number) {
-    return this.hoistedTag.getGroup(group) + this.normalTag.getGroup(group);
+  getHoistedAndNormalGroups(group: number) {
+    return { hoisted: this.hoistedTag.getGroup(group), normal: this.normalTag.getGroup(group) };
   }
 }
--- a/packages/styled-components/src/sheet/HoistedTag.js
+++ b/packages/styled-components/src/sheet/HoistedTag.js
@@ -1,0 +1,50 @@
+// @flow
+
+import type { HoistedTag, GroupedTag, Tag } from './types';
+import { DefaultGroupedTag } from './GroupedTag';
+import { WindowedTag } from './WindowedTag';
+
+const hoistedRuleRe = /\s*@(font-face|import)/;
+
+export class DefaultHoistedTag implements HoistedTag {
+  hoistedTag: GroupedTag;
+
+  normalTag: GroupedTag;
+
+  length: number;
+
+  constructor(tag: Tag) {
+    this.hoistedTag = new DefaultGroupedTag(new WindowedTag(tag));
+    this.normalTag = new DefaultGroupedTag(new WindowedTag(tag));
+    this.length = 0;
+  }
+
+  insertRules(group: number, rules: string[]): void {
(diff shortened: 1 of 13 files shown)


packages/styled-components/src/sheet/HoistedTag.js
```diff
@@ -0,0 +1,50 @@
...
+    this.length = this.normalTag.length;
+  }
+
+  clearGroup(group: number) {
+    this.normalTag.clearGroup(group);
+    this.hoistedTag.clearGroup(group);
+    this.length = this.normalTag.length;
+  }
+
+  // TODO: Needs to be replaced to keep hoisted ordering
```
jnielson94: @kitten  - Can you expound a little on what this means?

packages/styled-components/src/sheet/HoistedTag.js
```diff
@@ -0,0 +1,50 @@
...
+      if (hoistedRuleRe.test(rule)) {
+        hoistedRules.push(rule);
+      } else {
+        normalRules.push(rule);
+      }
+    }
+
+    this.normalTag.insertRules(group, normalRules);
+    this.hoistedTag.insertRules(group, hoistedRules);
+    this.length = this.normalTag.length;
```
jnielson94: @kitten - It seems like the `length` for a HoistedTag should account for the hoisted rules? I didn't notice length being used outside of testing and Rehydration, but Rehydration is digging into the hoistedTag to groupedTags to get lengths, so maybe it doesn't matter?

kitten: This number actually signifies the amount of groups; it’s just called length for convenience’s sake. Should definitely rename it or add a comment though :)

packages/styled-components/src/sheet/HoistedTag.js
kitten: This command isn’t yet in use. Instead the Rehydration sheet actually just calls into the hoisted and normal tags directly which is bad.

Basically, the output logic should be as close to every tag as possible. To the point where calling toString on the sheet’s HoistedTag directly gives you the right CSS output.

This will require a couple of changes however:

- the registered names need to either be removed from the SSR output or moved to the tags. It may be possible to generate them on the fly, but that may require even more changes and comes at the cost of sacrificing perf that we fought for 😅
- all hoisted rules must come before normal rules (that’s already done, but hacky as mentioned before)
- the global rules either need special “marker rules” or the marker rules need to be changed so that they don’t interfere with global rules. We should still however be using marker rules if possible. Parsing and splitting the CSS manually, which is what we did before, was pretty slow

packages/styled-components/src/sheet/HoistedTag.js
jnielson94: Ah, that makes a lot more sense. A comment seems like it would make the most sense, since it makes more sense to me to have the property name be consistent among the various `Tags` 😄

packages/styled-components/src/sheet/HoistedTag.js
```diff
@@ -0,0 +1,50 @@
+// @flow
+
+import type { HoistedTag, GroupedTag, Tag } from './types';
+import { DefaultGroupedTag } from './GroupedTag';
+import { WindowedTag } from './WindowedTag';
+
+const hoistedRuleRe = /\s*@(font-face|import)/;
```
jnielson94: As I've been testing this, I've learned that `@imports` and `@font-face` can't be intermingled - an `@import` after an `@font-face` is ignored, and so far as I've found in [the specs](https://www.w3.org/TR/css3-cascade/#at-import) a sheet has to have 
```
@charset rule (if it exists)
@import rules
anything else
```

So, @kitten if we change to only hoisting `@import` statements, I would think we could do a custom marker for them (related to https://github.com/styled-components/styled-components/pull/2771/files#r330312750 ) that could be an `@import` using some combination of media conditions that are impossible to meet (not sure what yet), with the import string containing the marker? Does that sound reasonable?

jnielson94: So @kitten - Basically we only _need_ to hoist `@import` statements, but so far as I can find there isn't a good "marker" that would work to identify what group they're with. So, I'm going to go ahead and take out `font-face` from this regex (and adjust things), try taking out the markers in the hoisted rules, and see what I can figure out for rehydration (in the other thread).

packages/styled-components/src/sheet/HoistedTag.js
jnielson94: @kitten - I gave this a shot in my PR #2775 
I went with making sure the hoisted rules are output first by diverging the HoistedTag from the GroupedTag and then I ended up ignoring the hoisted rules on Rehydration (possibly a bad thing, but I didn't see any negatives in the sandbox, tests, or my app I was testing in locally). I'd appreciate a review of that PR when you get a chance, since I think I hit everything except the "code could probably be cleaned up" notion.

jnielson94:
@kitten Thanks for merging in the PR I had for this! If you or @probablyup have any other TODOs that I can take a look at I'd be glad to, right now I'm not sure what else needs to be done for this to be ready though 😄

jnielson94:
@kitten & @probablyup - Just wanted to ping on this issue again since without it v5 breaks font loading using `@import` statements entirely. Anything that can be done to improve this PR feel free to let me know :)

jnielson94:
@kitten  & @probablyup  - Wanted to throw another ping out on this PR to see if there's anything I can do to move it along :) It'd be nice to get this into v5, otherwise I guess since it's a major version a note can be added to the upgrading guide about `@import` statements not being supported?

probablyup:
This one’s in Phil’s camp, thanks Jordan!

eatrocks:
😞 Disappointed that the issue was closed without resolution nor explanation.

PR #2775, opened 2019-09-30
Import Rule Hoisting Testing

jnielson94:
Working on testing out the POC PR (#2771). Putting up the PR early so that @kitten or @probablyup knows it exists as I work on adding to it (and clean it up).

In the POC PR, Phil added some TODOs and comments that I'll address in this as well if he doesn't get to it before I do 😄 

Edit: I think I covered all the things in #2771 - If there's tests that could be added feel free to mention them and I can do it 😄

### DIFF
--- a/packages/styled-components/src/sheet/GroupedTag.js
+++ b/packages/styled-components/src/sheet/GroupedTag.js
@@ -52,9 +52,9 @@ export class DefaultGroupedTag implements GroupedTag {
 
   clearGroup(group: number): void {
     if (group < this.groups) {
-      const numRules = this.groupSizes[group];
+      const groupSize = this.groupSizes[group];
       const startIndex = this.indexOfGroup(group);
-      const endIndex = startIndex + numRules;
+      const endIndex = startIndex + groupSize;
 
       this.groupSizes[group] = 0;
 
@@ -70,9 +70,9 @@ export class DefaultGroupedTag implements GroupedTag {
       return css;
     }
 
-    const numRules = this.groupSizes[group];
+    const groupSize = this.groupSizes[group];
     const startIndex = this.indexOfGroup(group);
-    const endIndex = startIndex + numRules;
+    const endIndex = startIndex + groupSize;
 
     for (let i = startIndex; i < endIndex; i++) {
       css += `${this.tag.getRule(i)}\n`;
--- a/packages/styled-components/src/sheet/GroupedTag.js
+++ b/packages/styled-components/src/sheet/GroupedTag.js
@@ -8,13 +8,13 @@ const BASE_SIZE = 1 << 8;
 export class DefaultGroupedTag implements GroupedTag {
   groupSizes: Uint32Array;
 
-  length: number;
+  groups: number;
 
   tag: Tag;
 
   constructor(tag: Tag) {
     this.groupSizes = new Uint32Array(BASE_SIZE);
-    this.length = BASE_SIZE;
+    this.groups = BASE_SIZE;
     this.tag = tag;
   }
 
@@ -35,7 +35,7 @@ export class DefaultGroupedTag implements GroupedTag {
 
       this.groupSizes = new Uint32Array(newSize);
       this.groupSizes.set(oldBuffer);
-      this.length = newSize;
+      this.groups = newSize;
 
       for (let i = oldSize; i < newSize; i++) {
         this.groupSizes[i] = 0;
@@ -51,10 +51,10 @@ export class DefaultGroupedTag implements GroupedTag {
   }
 
   clearGroup(group: number): void {
-    if (group < this.length) {
-      const length = this.groupSizes[group];
+    if (group < this.groups) {
+      const numRules = this.groupSizes[group];
       const startIndex = this.indexOfGroup(group);
-      const endIndex = startIndex + length;
+      const endIndex = startIndex + numRules;
 
       this.groupSizes[group] = 0;
 
@@ -66,13 +66,13 @@ export class DefaultGroupedTag implements GroupedTag {
 
   getGroup(group: number): string {
     let css = '';
-    if (group >= this.length || this.groupSizes[group] === 0) {
+    if (group >= this.groups || this.groupSizes[group] === 0) {
       return css;
     }
 
-    const length = this.groupSizes[group];
+    const numRules = this.groupSizes[group];
     const startIndex = this.indexOfGroup(group);
-    const endIndex = startIndex + length;
+    const endIndex = startIndex + numRules;
 
     for (let i = startIndex; i < endIndex; i++) {
       css += `${this.tag.getRule(i)}\n`;
--- a/packages/styled-components/src/sheet/HoistedTag.js
+++ b/packages/styled-components/src/sheet/HoistedTag.js
@@ -11,6 +11,9 @@ export class DefaultHoistedTag implements HoistedTag {
 
   normalTag: GroupedTag;
 
+  /*
+  The number of Groups in the tag
+  */
   length: number;
 
   constructor(tag: Tag) {
@@ -34,6 +37,8 @@ export class DefaultHoistedTag implements HoistedTag {
 
     this.normalTag.insertRules(group, normalRules);
     this.hoistedTag.insertRules(group, hoistedRules);
+    // We just use the normalTag length
+    // since the normalTag and hoistedTag have the same groups
     this.length = this.normalTag.length;
   }
 
--- a/packages/styled-components/src/sheet/HoistedTag.js
+++ b/packages/styled-components/src/sheet/HoistedTag.js
@@ -11,9 +11,6 @@ export class DefaultHoistedTag implements HoistedTag {
 
   normalTag: GroupedTag;
 
-  /*
-  The number of Groups in the tag
-  */
   groups: number;
 
   constructor(tag: Tag) {
--- a/packages/styled-components/src/sheet/HoistedTag.js
+++ b/packages/styled-components/src/sheet/HoistedTag.js
@@ -4,7 +4,7 @@ import type { HoistedTag, GroupedTag, Tag } from './types';
 import { DefaultGroupedTag } from './GroupedTag';
 import { WindowedTag } from './WindowedTag';
 
-const hoistedRuleRe = /\s*@(font-face|import)/;
+const hoistedRuleRe = /\s*@import/;
 
 export class DefaultHoistedTag implements HoistedTag {
   hoistedTag: GroupedTag;
@@ -14,12 +14,12 @@ export class DefaultHoistedTag implements HoistedTag {
   /*
   The number of Groups in the tag
   */
-  length: number;
+  groups: number;
 
   constructor(tag: Tag) {
     this.hoistedTag = new DefaultGroupedTag(new WindowedTag(tag));
     this.normalTag = new DefaultGroupedTag(new WindowedTag(tag));
-    this.length = 0;
+    this.groups = 0;
   }
 
   insertRules(group: number, rules: string[]): void {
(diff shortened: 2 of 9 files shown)


packages/styled-components/src/sheet/test/WindowedTag.test.js
```diff
@@ -0,0 +1,192 @@
...
+describe('WindowedTag acts like a tag', () => {
+  let virtualTag;
+  let tag;
+  beforeEach(() => {
+    virtualTag = new VirtualTag();
+    tag = new WindowedTag(virtualTag);
+  });
+
+  /*
+   * Copy/Pasted from Tag.test.js since it requires different setup than the tags tested there
```
jnielson94: I'd love to not duplicate these tests in this file (and with the GroupedTag tests), but I wanted to make sure that the `WindowedTag` passes the same tests we're using for the `Tag`(s), but I couldn't think of a good way to utilize the setup in Tag.test.js :(

packages/styled-components/src/sheet/test/WindowedTag.test.js
```diff
@@ -0,0 +1,192 @@
...
+    expect(windowedTag2.insertRule(0, '.a-second {}')).toBe(true);
+    expect(windowedTag1.insertRule(2, '.c {}')).toBe(true);
+    expect(windowedTag2.insertRule(2, '.c-second {}')).toBe(true);
+
+    expect(windowedTag1.getRule(0)).toBe('.a {}');
+    expect(windowedTag1.getRule(1)).toBe('.b {}');
+    expect(windowedTag1.getRule(2)).toBe('.c {}');
+    expect(windowedTag1.length).toBe(3);
+
+    // The current implementation allows windows to "bleed" if you request an offset that lands in another window
```
jnielson94: @kitten - I noticed while testing that the "windows" can "bleed" into other windows, but I don't think that's actually a bad thing? We could guard against it, but I don't know that it will be something we need to worry about in actual usage?

kitten: Yea, I figured given the current constraints this shouldn’t ever happen anyway 👍

packages/styled-components/src/sheet/test/GroupedTag.test.js
```diff
@@ -2,78 +2,82 @@
 
 import { VirtualTag } from '../Tag';
 import { DefaultGroupedTag } from '../GroupedTag';
+import { WindowedTag } from '../WindowedTag';
 
-let tag;
-let groupedTag;
+const describeGroupedTag = makeTagAndGroupedTag => {
```
jnielson94: I'm not super familiar with Flow typing, so I wasn't sure if it was needed to try and add types for this function `makeTagAndGroupedTag`. If y'all want them I can try to figure it out though :)

packages/styled-components/src/sheet/test/HoistedTag.test.js
```diff
@@ -0,0 +1,121 @@
...
+
+import { VirtualTag } from '../Tag';
+import { DefaultHoistedTag } from '../HoistedTag';
+
+describe('Works like a groupedTag if there are no hoisted rules', () => {
+  /* 
+  This set of tests is mostly copy/pasted from GroupedTag.test.js,
+  but since we need to dig into the normal/hoisted tags
+  it didn't make sense to try and reduce the duplication.
+   */
```
jnielson94: Not sure if this comment is super needed, but I wanted to make sure that I called this copy/paste and adjust situation out in the PR. Do any of y'all know a better way to accomplish that? It's only 2 tests and ~60 lines, so I went with the most straightforward way to get it tested.

jnielson94:
@kitten  - Let me know if there's anything that you want me to add to this :) I couldn't think of anything else, so I figured I might as well see what you think of it.

packages/styled-components/src/sheet/test/HoistedTag.test.js
kitten: I think that’s fine. We may want to just change the structure to be more functional / grouped a little nicer so this can be resolved nicely.
It’s one of the many signs that after the feature is done, the code will need some restructuring

kitten:
@jnielson94 this is amazing work already! Thank you so much!
I think all that is missing are some failing tests for SSR. Maybe we can add a separate set of tests that assert that the SSR output is correct (no import after other rules)
And if we’d also have a test that asserts imports-first, font-face after, etc, then we’d have enough to test the eventual, full implementation

jnielson94:
@kitten - Where would you expect the failing tests for SSR to be? I adjusted/added to the tests in [the Rehydration file](packages/styled-components/src/sheet/test/Rehydration.test.js), but I couldn't think of any other places that it would be needed to check that.

Note that I determined not to hoist `font-face`, since from what I could tell in the spec they don't have special rules like `@charset` and `@import` do (though I have no idea if we're handling `@charset`, that seemed outside the scope).

kitten:
@jnielson94 oh sorry! Clearly I’ll have to give this a more thorough read and will add some more comments if I find anything 😅

I’ll double check the font-face rule ordering in the spec. We can definitely ignore charset at-rules, since they’re not really in use anymore by people typically.

jnielson94:
@kitten - No worries! Just wanted to make sure I didn't miss something while I'm still doing work stuff 😆 Looking forward to your review, please feel free to point out tests or things that could improve! It's my first time diving into the styled-components code (and my first time really using flow), so I wouldn't be surprised if you notice things that are off 😄

jnielson94:
@kitten - Anything I can do to make reviewing this easier for you? :)

packages/styled-components/src/sheet/GroupedTag.js
```diff
@@ -51,10 +51,10 @@ export class DefaultGroupedTag implements GroupedTag {
   }
 
   clearGroup(group: number): void {
-    if (group < this.length) {
-      const length = this.groupSizes[group];
+    if (group < this.groups) {
+      const numRules = this.groupSizes[group];
```
kitten: small nit: Let’s maybe name this “groupSize” :)

packages/styled-components/src/sheet/HoistedTag.js
```diff
@@ -4,19 +4,22 @@ import type { HoistedTag, GroupedTag, Tag } from './types';
...
+const hoistedRuleRe = /\s*@import/;
 
 export class DefaultHoistedTag implements HoistedTag {
   hoistedTag: GroupedTag;
 
   normalTag: GroupedTag;
 
-  length: number;
+  /*
+  The number of Groups in the tag
```
kitten: small nit: Let’s keep these as single line comments please 😅 personally is turn off our empty line rule between property type declarations as well, but we haven’t found an opportunity to migrate some rules

packages/styled-components/src/sheet/test/HoistedTag.test.js
```diff
@@ -0,0 +1,185 @@
+// @flow
+
+import { VirtualTag } from '../Tag';
+import { DefaultHoistedTag } from '../HoistedTag';
+
+describe('Works like a groupedTag if there are no hoisted rules', () => {
```
kitten: small nit: lets keep describe descriptions as names of exports or instances, instead of full descriptions like “it”

packages/styled-components/src/sheet/types.js
```diff
@@ -16,12 +16,16 @@ export interface GroupedTag {
...
   getGroup(group: number): string;
-  length: number;
+  groups: number;
 }
 
-export interface HoistedTag extends GroupedTag {
-  hoistedTag: GroupedTag;
-  normalTag: GroupedTag;
+/** Group-aware Tag that hoists @import rules utilizing GroupedWindowedTags */
+export interface HoistedTag {
```
kitten: Not sure why the extends was removed here?

packages/styled-components/src/sheet/Rehydration.js
```diff
@@ -67,10 +65,12 @@ const rehydrateSheetFromTag = (sheet: Sheet, style: HTMLStyleElement) => {
 
   for (let i = 0, l = cssRules.length; i < l; i++) {
     const cssRule = (cssRules[i]: any);
-
     if (typeof cssRule.cssText !== 'string') {
       // Avoid IE11 quirk where cssText is inaccessible on some invalid rules
       continue;
+    } else if (cssRule.type === IMPORT_RULE_TYPE) {
```
kitten: Let’s revert this addition to avoid confusion later on?

Since we don’t have the new hoisting yet, we can’t just leave these out. Since SSR will aggressively avoid insertions, we’d likely remove these forever if we skip them during rehydration.

So let’s remove this change and we’ll figure changes here out once the SSR hoisting is implemented properly :)

packages/styled-components/src/sheet/types.js
jnielson94: I removed the extends mainly because I couldn't think of a way to implement the `getGroup` function that `GroupedTag` has in a way that keeps hoisting possible (Note that this is a PR into your PR), so I removed that function and replaced it with one that can split out the hoisted rules from the normal rules in the return value. I'm not aware enough of how to do things in flow to extend but only implement part of the interface (if that's even reasonable/possible) :)

packages/styled-components/src/sheet/Rehydration.js
jnielson94: What did you have in mind left to do with the SSR hoisting? I adjusted some things in this PR per comments on your PR (which this will merge into). With the adjustments to the hoisting in your PR we do have the imports all hoisted to the top of the style tag, just with no markers since I couldn't find a marker that wouldn't result in something ending up invalid (`@import` rules are imported aggressively so the conditional media rule stuff didn't work out, and no other rules can be intermingled with `@import` and let following `@import` statements be valid). 

So, the flow that works out with ignoring this (from what I understand/tested):
1. `@import` rules are hoisted in the SSR output (see code changes above)
1. `@import` rules are ignored on rehydration since we can't effectively figure out what group they belong to. All other rules are still rehydrated normally
1. `@import` rules would be added (in a hoisted position) during a re-render/whenever the client-side logic takes over the tag

Does that sound right? I'd love to re-hydrate them instead, but I couldn't find an effective way to do that so if you have ideas I'm game! :)

packages/styled-components/src/sheet/HoistedTag.js
jnielson94: I actually just forgot to remove this comment after renaming it, so thanks for pointing it out 😅

packages/styled-components/src/sheet/types.js
kitten: Ah alright :+1: cool, let's leave it then for now. I suppose the API will change anyway once I find time to refactor this

packages/styled-components/src/sheet/Rehydration.js
kitten: We do have them hoisted in the output.
But if we ignore them they will be removed and actually _won't be readded_ since the sheet has the `names` cache, which will ignore entire rule blocks when they've already been inserted.

This is important, because otherwise we'd readd rules on initial rendering/hydration, which is rather slow. Now, we could put in an exception that doesn't cache any rules during SSR with the name-cache that contain `@import` rules, but I'd suspect that this will potentially sometimes cause a flash on the page, as the `@import` is removed and then readded

jnielson94: Ah, I'll have to look at that caching then since from what I could tell on my testing they're being re-added presently.

kitten: That's likely because the tests never contain a full flow (at least not the `Rehydration.test.js`) of adding rules with names then rehydrating, and then readding the rules.

Instead the name cache is assumed to "just work" and not have any special properties apart from preventing components from adding a batch of rules twice.

jnielson94: Sounds good! If you get time today to look at it, I can manage updating this if you update your PR (unless you decide to push here... Which I think you can?).

I'll look at adding a full flow test too, since that sounds handy to have 🙂

jnielson94: Also note that I've been testing this using the sandbox app in the repo, but I haven't committed any changes to that since I wasn't sure if that generally stayed untouched 🙂 Before pushing the adjustments to your logic/PR I also pulled it into a sample app I have at work to make sure those fonts loaded without a flash (and they did), since I'm looking at v4 as well to try and fix the flash 🙂

packages/styled-components/src/sheet/test/GroupedTag.test.js
kitten: I think it’s fine, those are just tests 👍

PR #2780, opened 2019-10-03
Add new param shouldForwardProps to allow the passing in of 'invalid' attributes

kenhkelly:
This PR came up because we were using v3 and were passing in what are being determined as invalid attributes. We upgrade for v4 to fix another styling issue and found out our analytics broke.

We use these attributes for our analytics tracking to have simple event listeners set up globally.

```<button anclick="home_page_button" type="primary" class="buttonstyled__Button-sc-1t8li5e-0 YHsJf">HomePage</button>```

The usage to achieve bring the attributes back is by simply adding `shouldForwardProps` to the styled component.

```
return <StyledButton
          anclick={anclick}
          anevent={anevent}
          atc={atc}
          btnsize={btnsize}
          btnwidth={btnwidth}
          disabled={disabled}
          id={id}
          onClick={this.onClick}
          type={type}
          ref={forwardedRef}
          shouldForwardProps
          {...this.props}
        >
          {children}
        </StyledButton>;
```

### DIFF
--- a/packages/styled-components/package.json
+++ b/packages/styled-components/package.json
@@ -102,7 +102,7 @@
   "bundlesize": [
     {
       "path": "./dist/styled-components.min.js",
-      "maxSize": "16.25kB"
+      "maxSize": "16.26kB"
     }
   ],
   "collective": {
--- a/packages/styled-components/src/models/StyledComponent.js
+++ b/packages/styled-components/src/models/StyledComponent.js
@@ -136,6 +136,7 @@ class StyledComponent extends Component<*> {
 
     const elementToBeCreated = this.props.as || this.attrs.as || target;
     const isTargetTag = isTag(elementToBeCreated);
+    const { shouldForwardProps } = this.props;
 
     const propsForElement = {};
     const computedProps = { ...this.props, ...this.attrs };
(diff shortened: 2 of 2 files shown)


probablyup:
I’d highly recommend using data attributes for this, that’s what they’re
intended for.

kenhkelly:
Hind-sight, but we are a large e-commerce site and that would require a ton of refactor. That's why I made the one prop that is opt-in for us and for those that might also have something similar where white-listing of attributes is a stopper.

probablyup:
@mxstbr @kitten thoughts?

alexandernanberg:
If we were to add something like this, I think [emotions way](https://emotion.sh/docs/styled#customizing-prop-forwarding) is a better choice. 

I'm all for adding this functionality, currently we have to "hack" around this with a custom component passed through the `as` prop. But I don't think a `shouldForwardProps` prop is the way to go

probablyup:
I agree with @alexandernanberg. Thanks for your PR but not the overall solution we're aiming for...

alexandernanberg:
@probablyup If I'd add a PR with the proposed solution, is that something you'd consider merging? Or is more discussion around this needed?

probablyup:
Yeah, alright. We need to make sure the instances fold properly though (see
the attrs implementation.) I’m not 100% on the name at the moment

penx:
As far as I'm aware, styled-components/styled-components#2392 implemented shouldForwardProp in the same way as emotion.

Can this PR be reopened?

karlhorky:
Looks like @stevesims reopened in #3006 👍

PR #2787, opened 2019-10-04
Make ThemeProvider error straightforward

lekterable:
Super small change, just for the sake of keeping errors dead simple and straightforward.
I think that theres a chance that it might save someone a bit of confusion.

### DIFF
--- a/packages/styled-components/src/models/ThemeProvider.js
+++ b/packages/styled-components/src/models/ThemeProvider.js
@@ -17,6 +17,10 @@ export const ThemeContext: Context<Theme | void> = React.createContext();
 export const ThemeConsumer = ThemeContext.Consumer;
 
 function useMergedTheme(theme: ThemeArgument, outerTheme?: Theme): Theme {
+  if (!theme) {
+    return throwStyledError(14);
+  }
+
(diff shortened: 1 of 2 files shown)


probablyup:
This is unfortunately a breaking change. We can't accept this for v4 or v5, but can consider it for v6.

lekterable:
that should do it, let me know what you think @probablyup 

does the `sandbox` work on the canary btw? when I'm trying to access the page after `yarn start` and installing the modules I get:

Zrzut ekranu 2019-10-14 o 22 55 26

I've tried on the clean branch as well so It's not because of my changes.

probablyup:
@lekterable ah yeah just run build first so it generates the file locally

lekterable:
yup, that did it, thanks! I've tried it in sandbox on canary and works as expected so It's ready to merge! :)

Zrzut ekranu 2019-10-23 o 00 03 01

lekterable:
yup, that did it, thanks @probablyup ! I've tried it in sandbox on canary and works as expected so It's ready to merge! :)

Zrzut ekranu 2019-10-23 o 00 03 01

PR #2789, opened 2019-10-06
Attempt to solve unterminated css comment issue

shabbysoft:
Unterminated CSS comments cause pages to hang completely, rather than replacing the affected following styles with comments.

I saw this issue over in the website repo: https://github.com/styled-components/styled-components-website/issues/375

It actually happens even in the react-live styled-component example.  

From my research, counting the split length is allegedly the "fastest" way, but if there's a better solution I would love to fix it!


kitten:
Sorry, I believe none of us have interacted with this yet; This isn't really the responsibility of `styled-components` and I personally didn't even want that comment regular expression to be there. Frankly, the parser should reliably take care of this. Just adding more code, especially with something that may have quadratic impacts on performance isn't the answer though. So I'd defer this to `stylis` if it's still going to be an issue in their v4.

Issue #2774, opened 2019-09-30
Styled components returning too many results in unit tests.


Swazimodo:
## System:
 - OS: macOS Mojave 10.14.6
 - CPU: (8) x64 Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
 - Memory: 169.37 MB / 16.00 GB
 - Shell: 3.2.57 - /bin/bash
## Binaries:
 - Node: 11.14.0 - ~/.nvm/versions/node/v11.14.0/bin/node
 - Yarn: 1.17.3 - /usr/local/bin/yarn
 - npm: 6.7.0 - ~/.nvm/versions/node/v11.14.0/bin/npm
## npmPackages:
 - styled-components: ^4.4.0 => 4.4.0 
 - jest: ^24.9.0
 - react-test-renderer: ^16.10.1
 - ts-jest: ^24.1.0
 - typescript: 3.6.3

## Reproduction
Run `yarn test` on my example repo and look at the output from Demo.test.tsx.
https://github.com/Swazimodo/demo-broken-styled-components

## Steps to reproduce
- set a data tag for UI testing on a styled-component div
- render the component with react-test-renderer
- take the output and find all by props
- check the length of the selection

## Expected Behavior
- get one result for each rendered item

## Actual Behavior
- get 3x the results

## notes
- This works for regular react nodes but seems to blow up when you use styled-components
- I have created a snapshot with the rendered result to look at it and I do not see any duplicates 🤷‍♂

Swazimodo:
Later this week I'll submit a PR with a failing unit test.

Swazimodo:
Can someone add the Hacktoberfest label for reasons

probablyup:
@Swazimodo You might want to try the `renderIntoDocument` API and our test utility instead: https://www.styled-components.com/docs/api#find

Swazimodo:
@probablyup Sorry for the long delay but I'm coming back to this now. The `Find` test utility function does not work for this as you need to specify the styled component your looking for. However, I was not selecting a particular styled component, I was trying to find elements that matched a custom attribute.

PR #2792, opened 2019-10-07
feat: add failing unit test for selection issue

Swazimodo:
There is an open issue where styled-components are breaking unit tests by returning too many items when you select by properties. Here is a failing unit test to show the issue.

https://github.com/styled-components/styled-components/issues/2774


probablyup:
Closing, please see my comment in the OP :)

PR #2819, opened 2019-10-21
append "class" instead of "className" for unidentified DOM elements

### LINKED ISSUE #2802: Styles are not applied when using <custom-tags>
## Environment

`## System:
 - OS: Linux 4.15 Ubuntu 18.04.1 LTS (Bionic Beaver)
 - CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
 - Memory: 18.15 GB / 31.27 GB
 - Container: Yes
 - Shell: 5.4.2 - /usr/bin/zsh
## Binaries:
 - Node: 10.15.2 - /usr/bin/node
 - Yarn: 1.15.2 - /usr/bin/yarn
 - npm: 6.11.3 - /usr/bin/npm
## npmPackages:
 - babel-plugin-styled-components: ^1.10.0 => 1.10.6 
 - styled-components: ^5.0.0-beta.9 => 5.0.0-beta.9 `

## Reproduction

https://codesandbox.io/s/custom-tags-framer-motion-with-styled-components-wgg17

## Steps to reproduce
In this codesandbox, change "custom-tag" to "div" and you can see that the styles are correctly applied. If you change it back to "custom-tag" the styles are lost...

## Expected Behavior
"custom-tag" should be rendered in the DOM with styles applied.

## Actual Behavior
"custom-tag" is correctly rendered in the DOM, but the element loses its styles.

probablyup:
fixes #2802

React doesn't alias className -> class for custom elements, so appending
styles doesn't work properly without this change

### DIFF
--- a/packages/styled-components/src/models/StyledComponent.js
+++ b/packages/styled-components/src/models/StyledComponent.js
@@ -14,6 +14,7 @@ import type {
 import { checkDynamicCreation } from '../utils/checkDynamicCreation';
 import createWarnTooManyClasses from '../utils/createWarnTooManyClasses';
 import determineTheme from '../utils/determineTheme';
+import domElements from '../utils/domElements';
 import { EMPTY_ARRAY, EMPTY_OBJECT } from '../utils/empties';
 import escape from '../utils/escape';
 import generateComponentId from '../utils/generateComponentId';
@@ -154,9 +155,11 @@ function useStyledComponentImpl(
     propsForElement.style = { ...props.style, ...attrs.style };
   }
 
-  propsForElement.className = Array.prototype
+  propsForElement[
+    // handle custom elements which React doesn't properly alias
+    isTargetTag && domElements.indexOf(elementToBeCreated) === -1 ? 'class' : 'className'
+  ] = foldedComponentIds
     .concat(
-      foldedComponentIds,
       styledComponentId,
       generatedClassName !== styledComponentId ? generatedClassName : null,
       props.className,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 ## Unreleased
 
+- Use `class` instead of `className` for unidentified DOM elements (better support for styling web components)
+
+- Upgrade to stylis v4
+
+## [v5.2.0] - 2020-09-04
+
 - Make sure `StyleSheetManager` renders all styles in iframe / child windows (see [#3159](https://github.com/styled-components/styled-components/pull/3159)) thanks @eramdam!
 
 - Rework how components self-reference in extension scenarios (see [#3236](https://github.com/styled-components/styled-components/pull/3236)); should fix a bunch of subtle bugs around patterns like `& + &`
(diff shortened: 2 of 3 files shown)


packages/styled-components/src/models/StyledComponent.js
```diff
@@ -173,7 +174,10 @@ function useStyledComponentImpl<Config: {}, Instance>(
     propsForElement.style = { ...attrs.style, ...props.style };
   }
 
-  propsForElement.className = Array.prototype
+  propsForElement[
+    // handle custom elements which React doesn't properly alias
+    isTargetTag && domElements.indexOf(elementToBeCreated) === -1 ? 'class' : 'className'
```
kitten: This seems... rather odd. I wasn't aware that `class` was allowed in principal at all actually. Couldn't we always use `class` for `isTargetTag` then?

One nit here btw: Accessing the list of element names by `indexOf` is _relatively_ expensive.

packages/styled-components/src/models/StyledComponent.js
```diff
@@ -173,7 +174,10 @@ function useStyledComponentImpl<Config: {}, Instance>(
     propsForElement.style = { ...attrs.style, ...props.style };
   }
 
-  propsForElement.className = Array.prototype
```
kitten: completely unrelated nit, we know that `styledComponentId` is not an array and that `foldedComponentIds` is an array, so why aren't we using `foldedComponentIds.concat(styledComponentId)` or `[styledComponentId].concat(foldedComponentIds)` here?

kitten:
As an overarching problem here, web components aren't well supported through out React and the rest of the ecosystem. Without them there are only a few use-cases for custom elements, and it stands to argue that attaching classes to them is problematic in terms of using semantic HTML and sticking to React through-out.

Of course that leaves us with the point of: What if someone isn't sticking with React but is explicitly looking to integrate a Web Component library into their stack?

That is very valid, but do we want to support this if React chooses not to? It seems fairly close to be an upstream issue for React.

packages/styled-components/src/models/StyledComponent.js
probablyup: I think if you pass `class` to a recognized tag React will warn :/

packages/styled-components/src/models/StyledComponent.js
probablyup: Ah yeah we could do that. I think the concat order used to be different before and a non-array was in the first position.

probablyup:
>  It seems fairly close to be an upstream issue for React.

Totally is, but they've been unwilling to fix it for several majors. Considering this is a low effort fix on our end I think it's worth shipping for DX.

kitten:
It has a potentially large impact though. We could extract the check outside of render as a static “isUnknownTag” flag.

That being said, I’d still say it’s rather odd to introduce a fix for something that is fundamentally buggy in React.

probablyup:
> It has a potentially large impact though.

How so? The only real change is fixing something that is broken since React won't apply `className` to custom elements. I'd like to ship this with v5.

tracy-liucui:
Hi, any updates on this PR? we are looking forward to this change

probablyup:
@tracy-liucui I intend to ship it with v6

PR #2820, opened 2019-10-21
Memoize theme

### LINKED ISSUE #2767: ThemeProvider causes all styled-components to re-render even if not used
ThemeProvider causes all styled-components to re-render even if context is not used in it? 
This seems odd? 

If we just setup our own ContextProvider and handle it manually than ONLY the components that we pass the prop into require a re-render.

Is this true/intended?

probablyup:
Should fix #2767

Even if `props.theme` is a function, the result should be deterministic based on `outerTheme`.


probablyup:
@kitten done

PR #2701, opened 2019-08-07
backport dev SSR classname mismatch fix from canary

probablyup:
fixes #2691

PR #2822, opened 2019-10-21
Added check to flag isStatic to be true if in production or development.

JohnGrisham:
The ssr classname conflict error is still being thrown in development for the latest release of 4.4.0. Upon checking the fix that was implemented for this in that version I noticed that this flag was only true if in a production environment. The commit I am referring to was in #2701


probablyup:
We explicitly want this to be off for development, was on purpose. If you can provide a reproduction sandbox showing the issue can look into it though

probablyup:
Going to close this for now -- please open an issue with a clear reproduction and we can look into it!

Issue #2739, opened 2019-09-07
Global styles lost on Gatsby routing with v5


robinmetral:
On routing with Gatsby, global styles created with `createGlobalStyle` are lost.

I can confirm that this is a v5 error. To reproduce, simply route between pages on the demos:
- [v5 demo with :bug:](https://gatsby-globalstyle-routing.netlify.com/)
- [v4 demo without :bug:](https://5d7346a2093ee6000a1d56d6--gatsby-globalstyle-routing.netlify.com/)

_Note that this might be linked to #2670 (particularly to some comments), but in my case I have a single `createGlobalStyle` declaration and I'm using Gatsby, so I thought I'd open a new issue :slightly_smiling_face:_ 

## Environment
**System:**
 - OS: Linux 4.15 Ubuntu 18.04.3 LTS (Bionic Beaver)
 - CPU: (4) x64 Intel(R) Core(TM) i5-4260U CPU @ 1.40GHz
 - Memory: 1.04 GB / 7.73 GB
 - Container: Yes
 - Shell: 4.4.20 - /bin/bash

**Binaries:**
 - Node: 12.9.0 - ~/.nvm/versions/node/v12.9.0/bin/node
 - Yarn: 1.17.3 - /usr/bin/yarn
 - npm: 6.10.2 - ~/.nvm/versions/node/v12.9.0/bin/npm

**npmPackages:**
 - babel-plugin-styled-components: ^1.10.6 => 1.10.6 
 - styled-components: ^5.0.0-beta.8 => 5.0.0-beta.8-groupsizefix 

## Reproduction

**:bug: Reproduce the bug (with v5)**
- Repo: https://github.com/robinmetral/gatsby-globalstyle-routing
- Demo: https://gatsby-globalstyle-routing.netlify.com/

**:nail_care: Reproduce the expected behavior (with v4)**
- Repo: https://github.com/robinmetral/gatsby-globalstyle-routing/tree/with-v4
- Demo: https://5d7346a2093ee6000a1d56d6--gatsby-globalstyle-routing.netlify.com/

## Steps to reproduce

- visit the v5 demo
- navigate from the index to the second page

**Reproduce locally:**
- clone the repo
- `yarn`
- `yarn develop`
- http://localhost:8000

## Expected Behavior

The background should still be `forestgreen` (declared in global styles).

## Actual Behavior

The background is white because global styles are lost.

SolderHQ:
[...]

PR #2824, opened 2019-10-21
shard cGS by runtime instance

### LINKED ISSUE #2769: Global styles created with createGlobalStyle don't update on unmount
This issue occurs when a component created with `createGlobalStyle` is dynamic (e.g. through props) and more than one instance of the component may be mounted simultaneously.

What seems to happen, as far as I understand it, is this:

1. Instance 1 is mounted, injecting the appropriate CSS
2. Instance 2 is mounted, updating the CSS in question to reflect its own props
3. Instance 2 is unmounted, but the CSS remains unchanged.

### System:
 - OS: macOS Mojave 10.14.6
 - CPU: (8) x64 Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz
 - Memory: 123.61 MB / 16.00 GB
 - Shell: 3.0.2 - /usr/local/bin/fish
### Binaries:
 - Node: 10.16.3 - ~/.config/nvm/10.16.3/bin/node
 - Yarn: 1.17.3 - ~/.config/nvm/10.16.3/bin/yarn
 - npm: 6.9.0 - ~/.config/nvm/10.16.3/bin/npm
### npmPackages:
 - babel-plugin-styled-components: ^1.10.0 => 1.10.1
 - styled-components: ^5.0.0-beta.9 => 5.0.0-beta.8-groupsizefix

## Reproduction

https://codesandbox.io/embed/style-with-styled-components-kvsrk

## Steps to reproduce

In the codesandbox above, check the checkbox and uncheck it again.

## Expected Behavior

When the checkbox is unchecked, the background should go back to white.

## Actual Behavior

The background remains yellow as the styles from the unmounted component are kept around.

probablyup:
fixes #2769 #2739

cGS has a different injection profile than typical components, in that
there is a single component ID for the global style and any number of
instances. This creates problems when rendering multiple of the same
cGS with different props, as it is a defacto singleton but we don't
enforce a single cGS instance at once...

Utilizing a simple sharding strategy will create duplicate styles, but
ensure that when instances are removed from the page they're taken out
of the stylesheet in the proper order such that previous still-live
instances will continue to function and not be removed as well.

Not working: https://codesandbox.io/embed/style-with-styled-components-kvsrk
Working: https://codesandbox.io/s/style-with-styled-components-bs4mw

### DIFF
--- a/packages/styled-components/src/constructors/createGlobalStyle.js
+++ b/packages/styled-components/src/constructors/createGlobalStyle.js
@@ -1,6 +1,6 @@
 // @flow
-import React, { useContext, useEffect } from 'react';
-import { IS_BROWSER, STATIC_EXECUTION_CONTEXT } from '../constants';
+import React, { useContext, useEffect, useMemo } from 'react';
+import { STATIC_EXECUTION_CONTEXT } from '../constants';
 import GlobalStyle from '../models/GlobalStyle';
 import { useStyleSheet } from '../models/StyleSheetManager';
 import determineTheme from '../utils/determineTheme';
@@ -12,11 +12,6 @@ import type { Interpolation } from '../types';
 
 type GlobalStyleComponentPropsType = Object;
 
-// place our cache into shared context so it'll persist between HMRs
-if (IS_BROWSER) {
-  window.scCGSHMRCache = {};
-}
-
 export default function createGlobalStyle(
   strings: Array<string>,
   ...interpolations: Array<Interpolation>
@@ -24,10 +19,12 @@ export default function createGlobalStyle(
   const rules = css(strings, ...interpolations);
   const styledComponentId = `sc-global-${hasher(JSON.stringify(rules))}`;
   const globalStyle = new GlobalStyle(rules, styledComponentId);
+  let count = 0;
 
   function GlobalStyleComponent(props: GlobalStyleComponentPropsType) {
     const styleSheet = useStyleSheet();
     const theme = useContext(ThemeContext);
+    const instance = useMemo(() => String(++count), []);
 
     if (process.env.NODE_ENV !== 'production' && React.Children.count(props.children)) {
       // eslint-disable-next-line no-console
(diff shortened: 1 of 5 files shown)


packages/styled-components/src/test/__snapshots__/ssr.test.js.snap
```diff
@@ -3,20 +3,20 @@
...
 
 exports[`ssr should add a nonce to the stylesheet if webpack nonce is detected in the global scope 2`] = `
-"<style nonce=\\"foo\\" data-styled data-styled-version=\\"JEST_MOCK_VERSION\\">body{background:papayawhip;}
-data-styled.g1[id=\\"sc-global-a\\"]{content:\\"sc-global-a,\\"}
-.c{color:red;}
-data-styled.g2[id=\\"sc-b\\"]{content:\\"c,\\"}
+"<style nonce=\\"foo\\" data-styled data-styled-version=\\"JEST_MOCK_VERSION\\">.c{color:red;}
+data-styled.g1[id=\\"sc-b\\"]{content:\\"c,\\"}
+body{background:papayawhip;}
+data-styled.g2[id=\\"sc-global-a1\\"]{content:\\"sc-global-a1,\\"}
```
probablyup: this switches position in these snapshots because the component ID is registered lazily now for cGS since we can't know the instance # until render time

packages/styled-components/src/constructors/createGlobalStyle.js
```diff
@@ -12,22 +12,19 @@ import type { Interpolation } from '../types';
...
 ) {
   const rules = css(strings, ...interpolations);
   const styledComponentId = `sc-global-${hasher(JSON.stringify(rules))}`;
   const globalStyle = new GlobalStyle(rules, styledComponentId);
+  let count = 0;
 
   function GlobalStyleComponent(props: GlobalStyleComponentPropsType) {
     const styleSheet = useStyleSheet();
     const theme = useContext(ThemeContext);
+    const instance = useMemo(() => String(++count), []);
```
billyjanitsch: `useMemo` is a [hint, not a semantic guarantee](https://reactjs.org/docs/hooks-faq.html#how-to-create-expensive-objects-lazily). It shouldn't contain side effects because React could re-execute it during any render cycle even if the deps haven't changed.

To actually guarantee that this remains stable across renders, you can use a lazily-initialized ref:

```js
const instanceRef = useRef(null)
if (instanceRef.current === null) instanceRef.current = String(++count)
const instance = instanceRef.current
```

probablyup: Fair enough yeah. Probably more concurrent-mode safe too.

probablyup:
@kitten this should be safe enough with the switch to using ref instead of memo. Some people in the OP testing it have been saying it fits the need 👍

Since concurrent mode is constantly in flux and not even released I'd rather ship this fix and then tweak it later if needed.

packages/styled-components/src/constructors/createGlobalStyle.js
```diff
@@ -12,22 +12,23 @@ import type { Interpolation } from '../types';
...
   const styledComponentId = `sc-global-${hasher(JSON.stringify(rules))}`;
   const globalStyle = new GlobalStyle(rules, styledComponentId);
+  let count = 0;
 
   function GlobalStyleComponent(props: GlobalStyleComponentPropsType) {
     const styleSheet = useStyleSheet();
     const theme = useContext(ThemeContext);
+    const instanceRef = useRef(null);
+
+    if (instanceRef.current === null) instanceRef.current = String(++count);
```
kitten: My only nit is that this doesn't have to be a `string`. It's concatenated to the ID (which is already a `string`) later anyway, so it should be fine to just keep the number.

This is also concurrent mode safe now, since if the render repeats without any effects, the number will just already be present.

probablyup:
Oh I just did that because flow was complaining about concatenating
disparate types

PR #2855, opened 2019-11-05
[styled-components] streaming ssr example to devServer

sulmanen:
Problem:
- how to implement streaming ssr example missing
- quickly test if my assumptions on streaming ssr and styled components are correct

Solution:
- add streaming ssr example to devServer in examples from [documentation](https://www.styled-components.com/docs/advanced#streaming-rendering)


probablyup:
@sulmanen you might want to do this against canary instead, or the work will be lost when we do the v5 release

PR #2856, opened 2019-11-05
[styled-components] add ThemeProvider to Example

sulmanen:
Problem:
- quick working example that I can quickly modify the check if theming works correctly

Solution:
- add ThemeProvider to devServer Example


probablyup:
@sulmanen you might want to do this against canary instead, or the work will be lost when we do the v5 release

PR #2859, opened 2019-11-05
(v5) aggregate classNames passed via attrs 

### LINKED ISSUE #2839: className attrs don't append when styling a styled component
There was a breaking change with regard to how className `attrs` behave in styled-components. This change really makes the DX for using something like Tachyons or TailwindCSS that much worse. I've loved mixing Tachyons and Styled Components ever since `attrs` were introduced. Indeed, I think using utility classes was an inspiration for implementing the `attrs` functionality. 

It used to be I could write a simple component like: 
```
export const Title = styled.h1.attrs({
  className: "orange f1 sans-serif"
})``;
```
and then if I need to bespoke style it, I would just append new classes on to it like so:
```
export const AltTitle = styled(Title).attrs({
  className: "serif"
})``;
```

Since the appended classes would come after the first classes, I could also override classes provided through `attrs` this way as well. 

I think the new paradigm is a significant step backward as far as DX for using className `attrs`.

## Reproduction
Previous `styled-components` behavior:
https://codesandbox.io/s/styled-components-base-uujje

Updated behavior:
https://codesandbox.io/s/styled-components-base-57ibz

## Expected Behavior
classNames in `attrs` are appended in order when styling a styled component like:
https://codesandbox.io/s/styled-components-base-uujje

## Actual Behavior
classNames are dropped when styling a styled-component like:
https://codesandbox.io/s/styled-components-base-57ibz

probablyup:
fixes #2839

Normally attrs fully override the previous value, but in the case of
className they're often composed additively and it's desired to retain
the previous value in a HOC or folding scenario.

### DIFF
--- a/packages/styled-components/src/models/StyledComponent.js
+++ b/packages/styled-components/src/models/StyledComponent.js
@@ -57,7 +57,10 @@ function useResolvedAttrs<Config>(theme: any = EMPTY_OBJECT, props: Config, attr
 
     /* eslint-disable guard-for-in */
     for (key in resolvedAttrDef) {
-      context[key] = resolvedAttrs[key] = resolvedAttrDef[key];
+      context[key] = resolvedAttrs[key] =
+        key === 'className'
+          ? [resolvedAttrs[key], resolvedAttrDef[key]].filter(Boolean).join(' ')
+          : resolvedAttrDef[key];
     }
     /* eslint-enable guard-for-in */
   });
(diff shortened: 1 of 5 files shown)


packages/styled-components/src/models/StyledComponent.js
```diff
@@ -57,7 +57,10 @@ function useResolvedAttrs<Config>(theme: any = EMPTY_OBJECT, props: Config, attr
 
     /* eslint-disable guard-for-in */
     for (key in resolvedAttrDef) {
-      context[key] = resolvedAttrs[key] = resolvedAttrDef[key];
+      context[key] = resolvedAttrs[key] =
+        key === 'className'
+          ? [resolvedAttrs[key], resolvedAttrDef[key]].filter(Boolean).join(' ')
```
kitten: This may be a matter of taste, but we could just write a small `joinNames` function, which is then probably inline by Terser and saves us the intermediary array, e.g

```js
const joinNames = (a, b) =>
  a && b ? `${a} ${b}` : (a || b || “”)

PR #2888, opened 2019-11-21
Expose version in API

jamesarmenta:
Inspired by React's API: https://github.com/facebook/react/blob/master/packages/react/src/React.js#L105

My company has multiple teams working with styled-components on a shared eco-system. Although there are warnings for multiple style-component versions in the development build, our collisions often occur in the e2e environment where we use the production bundle of styled-components. This, of course, is a specific use case but I'm sure there are similar situations amongst different organizations and exposing the bundle version seems very low-cost to me.


probablyup:
Let's switch the target branch for this PR to v5 and put version inside secretInternals, ty

PR #2898, opened 2019-11-29
Unnecessary flattening and interleave of css without interpolations

vepor:
Based on previous question on twitter: https://twitter.com/elvepor/status/1200172654340259840

I believe that some performance of `css` construct could be saved when it doesn't have any interpolations. IMHO we can return just the styles that doesn't have to be flattened and interleaved.

Flow suppress comment is not necessary, but then I have to declare new array and return it: `[styles[0]]`.

I didn't add any new tests as this case is already covered.


vepor:
Merge changed to `v5` branch @probablyup @mxstbr 🙂

vepor:
Fixed 🙂

PR #2931, opened 2019-12-18
use funding field instead of postinstall script

koba04:
Thank you for the amazing library!!!

npm supports the`funding` field at v6.13 and Node has updated the bundled npm version to v6.13.4 because of the vulnerability so I think many people are able to use the `fund` subcommand.

- https://github.com/npm/cli/releases/tag/v6.13.0
- https://nodejs.org/en/blog/vulnerability/december-2019-security-releases/

So could you consider to use the `funding` field instead of `postinstall`? 🙏 
I understand that printing a message while installing the package is effective to reach out to many developers but the script is very dangerous so it would be nice if the ecosystem could avoid using this.
Yarn is considering to disable the script by default.
https://github.com/yarnpkg/yarn/issues/7761#issuecomment-565493023

### DIFF
--- a/packages/styled-components/package.json
+++ b/packages/styled-components/package.json
@@ -29,8 +29,7 @@
     "lint:size": "bundlesize",
     "prettier": "prettier */**/*.js --write",
     "prepublishOnly": "npm run build",
-    "dev": "cross-env BABEL_ENV=cjs babel-node example/startServer.js",
-    "postinstall": "node ./scripts/postinstall.js || exit 0"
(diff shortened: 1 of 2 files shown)


probablyup:
Can you rebase this against the v5 branch and switch the merge target? ty

koba04:
@probablyup Thank you! I've rebased this and switched the target branch!

probablyup:
Thanks!

koba04:
Thank you!!! This is huge for the ecosystem!!👏👏👏

PR #2951, opened 2020-01-07
[styled-components]: reworked how DISABLE_SPEEDY is computed

castamir:
I was trying to turn on SPEEDY just to realize it is not easy (or even possible) inside CRA app.
First of all - CRA's env variables are of type `string | undefined`, so turning SPEEDy "on" by setting `REACT_APP_SC_DISABLE_SPEEDY=false` results in truthy `"false"` and keeping it empty defaults to `true` again, since `NODE_ENV` should remain as `"test"`. Setting `NODE_ENV=production` in `beforeAll` or `beforeEach` does not take effect, since babel && typescripts && jest picks all used modules and execute them before any test is run. So there are more issues there that result in split styles or even errors (like [this one](https://www.styled-components.com/docs/faqs#why-am-i-getting-a-warning-about-several-instances-of-module-on-the-page)). 

This PR resolves one of them - ability to actually turn on SPEEDy mode for testing purposes.

### DIFF
--- a/packages/styled-components/src/constants.js
+++ b/packages/styled-components/src/constants.js
@@ -18,11 +18,11 @@ function isSpeedyDisabled() {
     return SC_DISABLE_SPEEDY;
   }
   if (typeof process !== 'undefined') {
-    if (typeof process.env.REACT_APP_SC_DISABLE_SPEEDY !== 'undefined') {
-      return process.env.REACT_APP_SC_DISABLE_SPEEDY && process.env.REACT_APP_SC_DISABLE_SPEEDY.toLowerCase() === 'true';
+    if (process.env.REACT_APP_SC_DISABLE_SPEEDY) {
+      return process.env.REACT_APP_SC_DISABLE_SPEEDY.toLowerCase() === 'true';
     }
-    if (typeof process.env.SC_DISABLE_SPEEDY !== 'undefined') {
-      return process.env.SC_DISABLE_SPEEDY && process.env.SC_DISABLE_SPEEDY.toLowerCase() === 'true';
+    if (process.env.SC_DISABLE_SPEEDY) {
+      return process.env.SC_DISABLE_SPEEDY.toLowerCase() === 'true';
     }
 
     return process.env.NODE_ENV !== 'production';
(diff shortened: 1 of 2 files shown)


packages/styled-components/src/constants.js
```diff
@@ -13,11 +13,25 @@ export const SC_VERSION = __VERSION__;
 
 export const IS_BROWSER = typeof window !== 'undefined' && 'HTMLElement' in window;
 
-export const DISABLE_SPEEDY =
-  (typeof SC_DISABLE_SPEEDY === 'boolean' && SC_DISABLE_SPEEDY) ||
-  (typeof process !== 'undefined' &&
-    (process.env.REACT_APP_SC_DISABLE_SPEEDY || process.env.SC_DISABLE_SPEEDY)) ||
```
kitten: Let's please keep this to a simpler change of `=== "true"` or another hardcoded string.

`typeof process.env.REACT_APP_SC_DISABLE_SPEEDY !== 'undefined'` is redundant,
`process.env.REACT_APP_SC_DISABLE_SPEEDY.toLowerCase()` defeats static optimisations (Webpack / Terser / etc won't be able to statically compile this away anymore)

Cheers :)

castamir: @kitten better?

castamir:
@kitten hey there ;)
could you please take a look if it ok for you now?

castamir:
@probablyup @kitten guys, is there anything I can do to help you merge this PR in?

kitten:
This doesn't address this important point:

> defeats static optimisations (Webpack / Terser / etc won't be able to statically compile this away anymore)

Changing the check to compare to `true` is a breaking change, so I'm afraid we've missed our window of opportunity to release this. The change can't use `toLowerCase` for static optimisations, so it would have to use `=== 'true'` only. It can't use a closure so that if the minifier is not as aggressive, this still can be optimised away by dead-code elimination. etc

cc @probablyup

castamir:
Well, not sure how to test this:

> defeats static optimisations (Webpack / Terser / etc won't be able to statically compile this away anymore)

But yes, this PR introduces a breaking change (in a matter if typings). But it also fixes something, that could never even work in the first place.

castamir:
Guys? What can I do in order to merge this PR in?

kitten:
Probably superseded by: https://github.com/styled-components/styled-components/pull/3289 Let me know if it's not

castamir:
@kitten yes, someone did basically the same PR as I did a year ago

PR #2965, opened 2020-01-14
Support <center> / styled.center

klarstrup:
I searched old issues commits etc. but couldn't find any reason this was omitted, so I propose adding it.

-------

This avoid `TypeError: Cannot read property 'withConfig' of undefined` when doing

```js
const RedCenteredText = styled.center`
  color: red;
`;

export default () => <RedCenteredText>Red Centered Text</RedCenteredText>;
```

or with the babel plugin

```js
export default () => <center css="color: red;">Red Centered Text</center>
```

### DIFF
--- a/packages/styled-components/src/utils/domElements.js
+++ b/packages/styled-components/src/utils/domElements.js
@@ -20,6 +20,7 @@ export default [
   'button',
   'canvas',
   'caption',
+  'center',
   'cite',
   'code',
   'col',


probablyup:
Do you have a use case for it? Those styling tags are strongly discouraged in HTML5

klarstrup:
Is there a use case for `<marquee>` being supported? It seems like a random omission that `<center>` is left out but `<b>`, `<i>`, `<u>`, `<marquee>`, `<big>` etc. are in.

probablyup:
@klarstrup that list was actually copied from React itself prior to v16 where they stripped out the DOM factories. My only real objection to adding more is it's unnecessary bundle weight that almost no one will use 🤷‍♂

klarstrup:
Doing `styled('center')` doesn't solve the problem with `babel-plugin-styled-components`.

What about fixing `babel-plugin-styled-components` such that it transforms `<center css="color:red"/>` in a way that does not create a confusing runtime error?

probablyup:
@klarstrup I'll take a look, it should be transforming using the stringified target syntax if not wrapping a scope variable

PR #2968, opened 2020-01-15
Fix xhtml compat for ssr stylesheet (works in v4)

sulmanen:
by mirroring attribute key in value. This fixes xhtml support as xml doesn't support attribute minimization (excluding the attribute value).

More specifically from [xhtml spec](https://www.w3.org/TR/xhtml1/) 4.5. Attribute Minimization
"XML does not support attribute minimization. Attribute-value pairs must be written in full. Attribute names such as compact and checked cannot occur in elements without their value being specified."

Without this, xhtml is not supported rendering the hbbtv tv platform outside the scope of styled-components. This worked in v4 as we are currently running it and are very pleased so thanks for all the good work!

## Environment
ssr of any page with xhtml doctype

## Steps to reproduce
[Borked v5 with xhtml doctype on codesandbox.io](https://codesandbox.io/s/flamboyant-golick-8fbqe)

## Expected Behavior
styled-components ssr renders xhtml compatible markup
[Working v4 with xhtml doctype](https://h6ffw.sse.codesandbox.io/)
[code](https://codesandbox.io/s/floral-thunder-h6ffw)

## Actual Behavior
styled-components ssr renders xhtml incompatible markup
[Borked v5 with xhtml doctype](https://8fbqe.sse.codesandbox.io/)
[code](https://codesandbox.io/s/flamboyant-golick-8fbqe)

### DIFF
--- a/packages/styled-components/src/models/ServerStyleSheet.js
+++ b/packages/styled-components/src/models/ServerStyleSheet.js
@@ -26,7 +26,7 @@ export default class ServerStyleSheet {
   _emitSheetCSS = (): string => {
     const css = this.instance.toString();
     const nonce = getNonce();
-    const attrs = [nonce && `nonce="${nonce}"`, `${SC_ATTR}="${SC_ATTR}"`, `${SC_ATTR_VERSION}="${SC_VERSION}"`];
+    const attrs = [nonce && `nonce="${nonce}"`, `${SC_ATTR}="true"`, `${SC_ATTR_VERSION}="${SC_VERSION}"`];
     const htmlAttr = attrs.filter(Boolean).join(' ');
 
     return `<style ${htmlAttr}>${css}</style>`;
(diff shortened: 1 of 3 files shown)


packages/styled-components/src/models/ServerStyleSheet.js
```diff
@@ -26,7 +26,7 @@ export default class ServerStyleSheet {
   _emitSheetCSS = (): string => {
     const css = this.instance.toString();
     const nonce = getNonce();
-    const attrs = [nonce && `nonce="${nonce}"`, SC_ATTR, `${SC_ATTR_VERSION}="${SC_VERSION}"`];
+    const attrs = [nonce && `nonce="${nonce}"`, `${SC_ATTR}="${SC_ATTR}"`, `${SC_ATTR_VERSION}="${SC_VERSION}"`];
```
probablyup: maybe just make this `${SC_ATTR}="true"`

sulmanen: sure, though not that's not strictly according to spec if I understand correctly? Works though.

https://www.w3.org/TR/xhtml1/ 4.5.1 

for example
checked="checked"

PR #2976, opened 2020-01-20
Remove workarounds for border radius in (RN)

jacobp100:
So it looks like RN now supports this for the `Image` component! There looks to be issues border width and colour (they don't show up at all), but that's on RN's side.

I think this isn't a breaking change. Technically, you go from `{ borderRadius: 50 }` to `{ borderTopLeftRadius: 50, borderTopRightRadius: … }`, but this shouldn't actually change much

If we remove the blacklist here, we'll remove it in css-to-react-native and there'll be a tiny perf win

### DIFF
--- a/packages/styled-components/src/models/InlineStyle.js
+++ b/packages/styled-components/src/models/InlineStyle.js
@@ -40,16 +40,7 @@ export default (styleSheet: StyleSheet) => {
             console.warn(`Node of type ${node.type} not supported as an inline style`);
           }
         });
-        // RN currently does not support differing values for the corner radii of Image
-        // components (but does for View). It is almost impossible to tell whether we'll have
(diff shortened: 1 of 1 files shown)


probablyup:
@jacobp100 what's the backward compatibility story for this change? We don't currently put a specific minimum RN version on the library so we're currently supporting all the ones that have hooks... which I think is 0.59+

jacobp100:
I think version wise we are fine

There’s a super edge case that you’ll get different results for

```
const A = styled.View`
  border-top-right-radius: 20px
`

const B = styled(A)`
  border-radius: 10px
`
```

probablyup:
I think this will have to wait for next major, sorry. We already got some backlash for the change that was brought into v5

probablyup (approved):

PR #2982, opened 2020-01-22
Add useTheme to the list of native exports

jladuval:
Hi folks, just noticed that useTheme wasn't exported for native so I've added it. First PR to styled-components so apologies if I've missed a contributing rule!


probablyup:
Thanks @jladuval, could you merge from master and then add a changelog entry? :)

PR #2983, opened 2020-01-22
attempt to resolve a SSR deopt in the hashing function

probablyup:
Benchmarking against v5.0.0 I'm seeing a minor improvement

### DIFF
--- a/packages/styled-components/src/utils/hash.js
+++ b/packages/styled-components/src/utils/hash.js
@@ -7,7 +7,6 @@ export const SEED = 5381 | 0;
 // version of djb2 where we pretend that we're still looping over
 // the same string
 export const phash = (h: number, x: string): number => {
-  h |= 0;
   for (let i = 0, l = x.length | 0; i < l; i++) {
     h = (h << 5) + h + x.charCodeAt(i);
   }
--- a/packages/styled-components/src/utils/hash.js
+++ b/packages/styled-components/src/utils/hash.js
@@ -7,8 +7,10 @@ export const SEED = 5381;
 // version of djb2 where we pretend that we're still looping over
 // the same string
 export const phash = (h: number, x: string): number => {
-  for (let i = 0, l = x.length; i < l; i++) {
-    h = (h << 5) + h + x.charCodeAt(i);
+  let i = x.length;
+
+  while (i) {
+    h = (h * 33) ^ x.charCodeAt(--i);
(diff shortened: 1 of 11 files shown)


probablyup:
@kitten this should be good to merge if you approve, seeing a nice little perf bump in general lib usage though it doesn't solve the SSR issue

packages/styled-components/src/utils/hash.js
```diff
@@ -1,21 +1,22 @@
...
 // When we have separate strings it's useful to run a progressive
 // version of djb2 where we pretend that we're still looping over
 // the same string
 export const phash = (h: number, x: string): number => {
-  h |= 0;
-  for (let i = 0, l = x.length | 0; i < l; i++) {
-    h = (h << 5) + h + x.charCodeAt(i);
+  let i = x.length;
+
+  while (i) {
```
kitten: can we work this forwards possibly? That should in theory keep some of the hashes the same as before?

probablyup:
Looping in reverse is generally faster in JS. Why is keeping existing
hashes a goal? Different sc versions don’t conflict at runtime

kitten:
> Looping in reverse is generally faster in JS

@probablyup: That'd be news to me and I highly doubt that. I think you're generally increasing the likelihood of cache misses, if I'm not mistaken, especially if we assume that a string doesn't necessarily have to be stored in a single memory location. Of course, that may not influence the benchmarks much, though I'd prefer forward-looping by default just for readability.

avindra:
Reverse looping historically was faster because you can avoid accessing `.length`. on every iteration. Although, it looks like the previous implementation had the length computed in the first part of the `for` statement.

But nowadays, v8 and other js engines optimize for more typical cases (like forward looping), so generally "readable code" is preferred.  And of course, pinning down the meaning of "readable code" is a separate topic.

packages/styled-components/src/utils/hash.js
```diff
@@ -1,21 +1,22 @@
 // @flow
 /* eslint-disable */
 
-export const SEED = 5381 | 0;
+export const SEED = 5381;
 
 // When we have separate strings it's useful to run a progressive
 // version of djb2 where we pretend that we're still looping over
 // the same string
 export const phash = (h: number, x: string): number => {
```
kmvan: Why not name `hash.ts`?

PR #2996, opened 2020-01-30
Fix upper group limit in GroupedTag

### LINKED ISSUE #2979: getStyleTags method very slow in v5
Hi all,

Upgrading my company's large codebase to v5 and we've seen a performance downgrade. Here are the times I recorded from our server for each `styled-components` specific methods we use:

| v4 | v5 |
|---|---|
|ServerStyleSheet: 5.355ms|ServerStyleSheet: 0.210ms|
|getStyleTags: 0.916ms|getStyleTags: **10.355ms**|
|ServerStyleSheet: 2.883ms|ServerStyleSheet: 0.039ms|
|getStyleTags: 0.789ms|getStyleTags: **9.055ms**|
|ServerStyleSheet: 4.064ms|ServerStyleSheet: 0.030ms|
|getStyleTags: 1.472ms|getStyleTags: **5.485ms**|

Nice to see `ServerStyleSheet` constructor is much faster now! But the worse `getStyleTags` is not worth upgrading to. ~Wondering if it's related to my Node version.~ Just tried latest stable release `Erbium` and have the same results.

## Environment
```
## System:
 - OS: macOS Mojave 10.14.6
 - CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
 - Memory: 1.35 GB / 16.00 GB
 - Shell: 5.3 - /bin/zsh
## Binaries:
 - Node: 10.15.0 - ~/.nvm/versions/node/v10.15.0/bin/node
 - npm: 6.13.4 - ~/.nvm/versions/node/v10.15.0/bin/npm
## npmPackages:
 - babel-plugin-styled-components: ^1.10.6 => 1.10.6
 - styled-components: ^5.0.0 => 5.0.0
```

## Steps to reproduce
Update `styled-components` to v5 and `jest-styled-components` to latest

## Expected Behavior
Expect to see a significant reduction in overall render time

## Actual Behavior
Experience significant reduction in `ServerStyleSheet` time but overwhelming increase in `getStyleTags`

### LINKED ISSUE #2994: Low limit of groups throws error (around 5888)
## System:
 - OS: macOS Mojave 10.14.6
 - CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
 - Memory: 68.19 MB / 16.00 GB
 - Shell: 5.3 - /bin/zsh
## Binaries:
 - Node: 10.16.2 - ~/.nvm/versions/node/v10.16.2/bin/node
 - Yarn: 1.19.1 - /usr/local/bin/yarn
 - npm: 6.13.1 - ~/.nvm/versions/node/v10.16.2/bin/npm
## npmPackages:
 - styled-components: ^5.0.0 => 5.0.0 

## Reproduction
Please, take a look at created repo for this issue:
https://github.com/Ketler13/styled-components-test/blob/master/README.md

Here is _app.js component example:
```
class MyApp extends App {
  static async getInitialProps() {
    if (!isBrowser()) {
      try {
        await getDataFromTree(
          <GlobalStyle theme={{}} />,
          {}
        );
      } catch (error) {
        console.error('Error while running `getDataFromTree`', error);
      }

      if (!process.browser) {
        Head.rewind();
      }
    }
    return {};
  }

  render() {
    const { Component } = this.props;
    return (
      <Container>
        <Head>
          <title>Test Component</title>
          <meta name="viewport" content="initial-scale=1.0, width=device-width" />
        </Head>
        <ThemeProvider theme={{}}>
          <GlobalStyle theme={{}} />
          <Component />
        </ThemeProvider>
      </Container>
    );
  }
}
```
GlobalStyle component is run twice. First - when getting data for apollo in getDataFromTree, second - in render. **I understand** that current problem is caused by one redundant instance of GlobalStyle, but couldn't this issue be reproduced under more real conditions?

## Steps to reproduce
Preconditions: next.js + styled-components + react-apollo.

Latest code of styled-components(5.0 version) is placed in vendor/styled/components. There are a few additional console.logs. This file is taken from the styled-components/dist/styled-components.cjs.js.

This problem is related to a coommit https://github.com/styled-components/styled-components/commit/aca9186920dd4adddb44b76063b75ab1147b480a . There was new counter mechanism introduced to improve SSR handling of multiple instances of global styles components. But under some conditions this mechanism can be broken.

There are 2 logs added: 1 - to detect size of Uint32Array created for groups https://github.com/styled-components/styled-components/blob/147b0e9a1f10786551b13fd27452fcd5c678d5e0/packages/styled-components/src/sheet/GroupedTag.js#L41
2 - to detect istance counter value https://github.com/styled-components/styled-components/blob/147b0e9a1f10786551b13fd27452fcd5c678d5e0/packages/styled-components/src/constructors/createGlobalStyle.js#L31

To reproduse issue, please install all modules and run npm run dev. When server is started, please make big number of requests to index page. Script for loading the server from terminal for ((i=1;i<=5000;i++)); do curl -v --header "Connection: keep-alive" "localhost:3000"; done

You will see incrementing of global style instance number. This is used in registering new id of the group
https://github.com/styled-components/styled-components/blob/147b0e9a1f10786551b13fd27452fcd5c678d5e0/packages/styled-components/src/models/GlobalStyle.js#L45

Then this id will be passed to the GroupIDAllocator and if counter is incremented on each request - created id will never be found in the registry and total groups counter will be increased https://github.com/styled-components/styled-components/blob/147b0e9a1f10786551b13fd27452fcd5c678d5e0/packages/styled-components/src/sheet/GroupIDAllocator.js#L18

Then this counter is used when new rules inserted -
https://github.com/styled-components/styled-components/blob/147b0e9a1f10786551b13fd27452fcd5c678d5e0/packages/styled-components/src/sheet/Sheet.js#L94

When if condition (https://github.com/styled-components/styled-components/blob/147b0e9a1f10786551b13fd27452fcd5c678d5e0/packages/styled-components/src/sheet/GroupedTag.js#L36) is satisfied, new Uint32Array is created https://github.com/styled-components/styled-components/blob/147b0e9a1f10786551b13fd27452fcd5c678d5e0/packages/styled-components/src/sheet/GroupedTag.js#L39

This bitwise operation
```const newSize = BASE_SIZE << ((group / BASE_SIZE) | 0);```
 will return next results under such conditions: 
- When group equals 5362 (5632 requests to the server), newSize will be 1073741824
- When group equals 5888 (5632 requests to the server), newSize will be -2147483648 (Range error comes).

## Expected Behavior
No huge memory consumption

## Actual Behavior
Significant memory consumption (~1-2 GB)

### PR DESCRIPTION
**Pull request by kitten**
Fix #2994 
Fix #2979

Previously the formula for increasing the size of Uint32Arrays
in the GroupedTag was a little aggressive and would cut of very
short of 2^31 groups. Instead it would cut off between 2^12 and
2^13 components, which isn't quite a lot.

This new upper boundary is the absolute limit of how many slots
a Uint32Array is defined to be able to hold. After this limit we'll
see an error.

Instead of letting this error leak out we now throw a more clear
error that hints at this limitation.

I've also refactored `dom.js` to finally use `throwStyledError` as well.

Tests have been updated to reflect all changes.

### DIFF
--- a/packages/styled-components/src/sheet/GroupedTag.js
+++ b/packages/styled-components/src/sheet/GroupedTag.js
@@ -39,11 +39,11 @@ class DefaultGroupedTag implements GroupedTag {
       const oldSize = oldBuffer.length;
 
       let newSize = oldSize;
-      while (group >= newSize && newSize > 0)
-        {newSize <<= 1;}
-
-      if (newSize < 0) {
-        throwStyledError(13, `${group}`);
+      while (group >= newSize) {
+        newSize <<= 1;
+        if (newSize < 0) {
+          throwStyledError(16, `${group}`);
+        }
       }
 
       this.groupSizes = new Uint32Array(newSize);
--- a/packages/styled-components/src/sheet/GroupedTag.js
+++ b/packages/styled-components/src/sheet/GroupedTag.js
@@ -2,13 +2,14 @@
 /* eslint-disable no-use-before-define */
 
 import type { GroupedTag, Tag } from './types';
+import throwStyledError from '../utils/error';
 
 /** Create a GroupedTag with an underlying Tag implementation */
 export const makeGroupedTag = (tag: Tag): GroupedTag => {
   return new DefaultGroupedTag(tag);
 };
 
-const BASE_SIZE = 1 << 8;
+const BASE_SIZE = 1 << 9;
 
 class DefaultGroupedTag implements GroupedTag {
   groupSizes: Uint32Array;
@@ -36,7 +37,14 @@ class DefaultGroupedTag implements GroupedTag {
     if (group >= this.groupSizes.length) {
       const oldBuffer = this.groupSizes;
       const oldSize = oldBuffer.length;
-      const newSize = BASE_SIZE << ((group / BASE_SIZE) | 0);
+
+      let newSize = oldSize;
+      while (group >= newSize && newSize > 0)
+        {newSize <<= 1;}
+
+      if (newSize < 0) {
+        throwStyledError(13, `${group}`);
+      }
 
       this.groupSizes = new Uint32Array(newSize);
       this.groupSizes.set(oldBuffer);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,11 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 ## Unreleased
 
-- Added useTheme hook to named exports for react native  
+- Added useTheme hook to named exports for react native
+
+- Performance enhancements
+  - Refactored hashing function that is a bit faster in benchmarks
+  - Fixed a bitwise math issue that was causing SSR performance degradations due to how we allocate typed arrays under the hood
 
 ## [v5.0.0] - 2020-01-13
 
--- a/packages/styled-components/src/sheet/GroupIDAllocator.js
+++ b/packages/styled-components/src/sheet/GroupIDAllocator.js
@@ -1,5 +1,9 @@
 // @flow
 
+import throwStyledError from '../utils/error';
+
+const MAX_SMI = 1 << 31 - 1;
+
 let groupIDRegister: Map<string, number> = new Map();
 let reverseRegister: Map<number, string> = new Map();
 let nextFreeGroup = 1;
@@ -16,6 +20,13 @@ export const getGroupForId = (id: string): number => {
   }
 
   const group = nextFreeGroup++;
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    ((group | 0) < 0 || group > MAX_SMI)
+  ) {
+    throwStyledError(16, `${group}`);
+  }
+
   groupIDRegister.set(id, group);
   reverseRegister.set(group, id);
   return group;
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: packages/styled-components/src/sheet/GroupedTag.js [hunk 1]
```diff
@@ -2,13 +2,14 @@
...
 import type { GroupedTag, Tag } from './types';
+import throwStyledError from '../utils/error';
 
 /** Create a GroupedTag with an underlying Tag implementation */
 export const makeGroupedTag = (tag: Tag): GroupedTag => {
   return new DefaultGroupedTag(tag);
 };
 
-const BASE_SIZE = 1 << 8;
+const BASE_SIZE = 1 << 9;
```
kitten: I've increased the base-size here since I find it highlight likely that a lot of apps will actually have around 512 components usually. Larger apps will of course see this grow as usual

packages/styled-components/src/sheet/GroupedTag.js
```diff
@@ -36,7 +37,14 @@ class DefaultGroupedTag implements GroupedTag {
     if (group >= this.groupSizes.length) {
       const oldBuffer = this.groupSizes;
       const oldSize = oldBuffer.length;
-      const newSize = BASE_SIZE << ((group / BASE_SIZE) | 0);
+
+      let newSize = oldSize;
+      while (group >= newSize) {
+        newSize <<= 1;
+        if (newSize < 0) {
```
kitten: This works because eventually we'll hit `1 << 31` the upper limit of SMIs, where we overflow into the signed bit, which turns `newSize` negative.

kitten:
This is published with as the prerelease `5.0.0-testgrouplimit.0` under the `test` tag.

probablyup:
@kitten throw a changelog entry in there too when you get a moment :)

PR #2997, opened 2020-01-31
add dev-time warning discouraging usage of @import

probablyup:


packages/styled-components/src/constructors/createGlobalStyle.js
```diff
@@ -40,6 +40,15 @@ export default function createGlobalStyle(
       );
     }
 
+    if (
+      process.env.NODE_ENV !== 'production' &&
+      rules.some(rule => typeof rule === 'string' && rule.includes('@import'))
```
kitten: Nit: includes isn’t available on IE11

PR #2998, opened 2020-01-31
Add warning for components that are created dynamically

kitten:


packages/styled-components/src/models/StyledComponent.js
```diff
@@ -277,6 +278,8 @@ export default function createStyledComponent(
   });
 
   if (process.env.NODE_ENV !== 'production') {
+    checkDynamicCreation(displayName, styledComponentId);
```
probablyup: Need to add this to keyframes as well yeah?

probablyup:
Code LGTM

packages/styled-components/src/models/StyledComponent.js
kitten: Key frames doesn’t inject its rules on its own anymore and generates a stable name and ID so that ones safe to inline

packages/styled-components/src/utils/checkDynamicCreation.js
```diff
@@ -3,31 +3,34 @@
...
+
+      if (!seen.has(message)) {
+        // eslint-disable-next-line no-console
+        console.warn(message);
+        seen.add(message);
+      }
+    } catch (error) {
+      // The error here is expected, since we're expecting anything that uses `checkDynamicCreation` to
+      // be called outside of a React component.
+      if (invalidHookCallRe.test(error.message)) {
```
probablyup: I think this might be a red herring for regex stickiness? If you use regex.exec instead of test there shouldn't be a 50/50 match miss

kitten: Do you mean `RegExp#lastIndex`? That doesn't apply to `test`. Also the `seen` Set is intended to filter by `displayName` & `componentId` duplicates.

oliviertassinari:
I have discovered a downside with this approach, in use my case (simplified version), I do

```jsx
const req = require.context('components', false, /\.(tsx)$/);

export default function Page() {
  const Component = req('components/Button').default;
  return <Component />;
}
```

Which leads to the warning from styled-components:

> You may see this warning because you've called styled inside another component.
To resolve this only create new StyledComponents outside of any render method and function component.

then a warning from React:

> Warning: React has detected a change in the order of Hooks called by MarkdownDocs. This will lead to bugs and errors if not fixed. For more information, read the Rules of Hooks:

and finally a crash:

> Error: Should have a queue. This is likely a bug in React. Please file an issue.

oliviertassinari:
The best fix I could find:

```diff
const req = require.context('components', false, /\.(tsx)$/);

+// Run styled-components ref logic
+// https://github.com/styled-components/styled-components/pull/2998
+req.keys().map(req);

export default function Page() {
  const Component = req('components/Button').default;
  return <Component />;
}
```

bihongbin3027:
@oliviertassinari  You, I’m facing the problem of dynamically loading components, use require().default

bihongbin3027:
https://github.com/styled-components/styled-components/issues/3328

PR #3006, opened 2020-02-07
shouldForwardProp support

### LINKED ISSUE #439: Separate HTML attributes from styling props

### LINKED ISSUE #2878: RFC filtering forwarded props
I am fan of styled-components, but there is a big pain point for me: filtering forwarded props (that pushed me to consider other librairies like emotion).

For example, if I want a `div` to have a controlable `height` and `width`, I will have to do this in order to ensure my `div` doesn't end up with `height` and `width` attributes:

```js
const StyledDiv = styled(({height, width, ...props}) => <div {...props} />)`
    width: ${props => props.width}px;
    height: ${props => props.height}px;
`
```

This is painful to write and makes the code unreadable. It seems to be a big issue with the library if we look at the number of issues opened for this (https://github.com/styled-components/styled-components/issues/135, https://github.com/styled-components/styled-components/issues/2467, https://github.com/styled-components/styled-components/issues/411, https://github.com/styled-components/styled-components/issues/439), the number of articles/spectrum questions about it, the aborted attempt of transient prop with v5 (https://github.com/styled-components/styled-components/pull/2093), etc...

Hence, I would like to propose a new solution, that I have already implemented and works well for me.

## Preliminary remark

A lot of people use styled components this way (because it is the only way described by the doc):

```js
const StyledDiv = styled.div`
    width: ${props => props.width}px;
    height: ${props => props.height}px;
`
```

But this quickly leads to long properties that are hard to understand. I usually use it this way, which can also fully benefit from props destructuring:

```js
const StyledDiv = styled.div(({ width, height }) => `
    width: ${width}px;
    height: ${height}px;
`)
```

**Remark:** These codes doesn't handle the issue with props forwarding.

## The base solution

Currently, `styled.div` takes a function that transforms props into **css**. The idea is to rather take a function that transform props into **props**, where the `css` prop will represent the styling:

```js
const StyledDiv = styled.div(({ width, height, ...props }) => ({
    css: `
        width: ${width}px;
        height: ${height}px;
    `,
    ...props
})
```
You can notice that this code handles the issue with props forwarding! `width` and `height` will not be forwarded to the div. If you want to forward `width` but not `height`, you can do:

```js
const StyledDiv = styled.div(({ width, height, ...props }) => ({
    css: `
        width: ${width}px;
        height: ${height}px;
    `,
    width,
    ...props
})
```

If you want to forward all the props, you can just use the current solution:

```js
const StyledDiv = styled.div(({ width, height }) => `
    width: ${width}px;
    height: ${height}px;
`)
```
etc...

## Variants

The solution I have just proposed illustrate perfectly the design shift. However, it may not be a good solution in practice because painful to use. Hence, it may be a good idea to also implement several alternatives for convenience:

### Variant 1 (css string and other props)

Instead of returning the props, a better solution could be to return an array containing the css and then the other props we want to forward. For example, if we only want to forward `width`, it would be:

```js
const StyledDiv = styled.div(({ width, height, ...props }) => [`
    width: ${width}px;
    height: ${height}px;
`, {width, ...props}])
```

### Variant 2 (tagged template litteral)

The previous solutions don't use tagged template litteral, what styled-components heavily use. Alternative 1 can be adapted to use them:

```js
const StyledDiv = styled.div([`
    width: ${props => props.width}px;
    height: ${props => props.height}px;
`, ({ width, height, ...props }) => props ])
```

But, now, the `[ ]` doesn't seem useful anymore. This leads to third and last variant:

### Variant 3 (simplified tagged template litteral)

```js
const StyledDiv = styled.div(`
    width: ${props => props.width}px;
    height: ${props => props.height}px;
`, ({ width, height, ...props }) => props)
```

## To sum up

The current solution is:

```js
const StyledDiv = styled(({height, width, ...props}) => <div {...props} />)`
    width: ${props => props.width}px;
    height: ${props => props.height}px;
`
```

which is painful to write and unreadable (what kind of component is it?).

The solution I propose is to add a second argument for filtering props:

```js
const StyledDiv = styled.div(`
    width: ${props => props.width}px;
    height: ${props => props.height}px;
`, ({width, height, ...props}) => props);
```

It is easy to write and to understand, and props filtering (which is not the important part of the definition) is sent at the end of the definition.

For people currently writing components like this:

```js
const StyledDiv = styled.div(({ width, height }) => `
    width: ${width}px;
    height: ${height}px;
`)
```

and wanting to filter props, the solution would be:

```js
const StyledDiv = styled.div(({ width, height, ...props }) => [`
    width: ${width}px;
    height: ${height}px;
`, {width, ...props}])
```

stevesims:
Following discussion elsewhere, such as on #2780, re-opening this PR (previously #2392)

adds support for use of `shouldForwardProp` config method, which can be used to explicitly determine which props should be forwarded on
if this config is not provided then existing functionality is preserved (i.e. all props passed to components, or for HTML elements check using `validAttr`)

Inspired by the same feature in emotion

Resolve #439 
Resolve #2878

### DIFF
--- a/packages/styled-components/src/models/StyledComponent.js
+++ b/packages/styled-components/src/models/StyledComponent.js
@@ -75,7 +75,7 @@ interface StyledComponentWrapperProperties {
   displayName: string;
   foldedComponentIds: Array<string>;
   target: Target;
-  shouldForwardProp: Function;
+  shouldForwardProp: ?(prop: string) => boolean;
   styledComponentId: string;
   warnTooManyClasses: $Call<typeof createWarnTooManyClasses, string, string>;
 }
@@ -147,25 +147,16 @@ function useStyledComponentImpl<Config: {}, Instance>(
 
   const isTargetTag = isTag(elementToBeCreated);
   const computedProps = attrs !== props ? { ...props, ...attrs } : props;
-  const shouldFilterProps = shouldForwardProp || isTargetTag || 'as' in computedProps || 'forwardedAs' in computedProps;
-  const propsForElement = shouldFilterProps ? {} : { ...computedProps };
+  const shouldFilterProps = 'as' in computedProps || 'forwardedAs' in computedProps;
+  const propFilterFn = shouldForwardProp || (isTargetTag && validAttr);
+  const propsForElement = propFilterFn || shouldFilterProps ? {} : { ...computedProps };
 
-  // use explicit `shouldForwardProp` function if one is set, otherwise
-  // if no explicit `shouldForwardProp` set then
-  //   if we're on a tag, use `validAttr`
-  //     i.e. Don't pass through non HTML tags through to HTML elements
-  // else (we're on a component - pass through all by default)
-  //   () => true
-  const shouldReallyForwardProp = shouldForwardProp || (
-    isTargetTag ? validAttr : () => true
-  );
-
-  if (shouldFilterProps) {
+  if (shouldFilterProps || propFilterFn) {
     // eslint-disable-next-line guard-for-in
     for (const key in computedProps) {
       if (key === 'forwardedAs') {
         propsForElement.as = computedProps[key];
-      } else if (key !== 'as' && key !== 'forwardedAs' && (shouldReallyForwardProp(key))) {
+      } else if (key !== 'as' && key !== 'forwardedAs' && (!propFilterFn || propFilterFn(key))) {
         // Don't pass through non HTML tags through to HTML elements
         propsForElement[key] = computedProps[key];
       }
@@ -190,20 +181,6 @@ function useStyledComponentImpl<Config: {}, Instance>(
   propsForElement.ref = refToForward;
 
   return createElement(elementToBeCreated, propsForElement);
-}
-
-function getShouldForwardProp(options, target, isTargetStyledComp) {
-  // $FlowFixMe
-  if (isTargetStyledComp && target.shouldForwardProp) {
-    if (options.shouldForwardProp) {
-      // compose nested shouldForwardProp calls
-      // $FlowFixMe
-      return prop => target.shouldForwardProp(prop) && options.shouldForwardProp(prop);
-    }
-    return target.shouldForwardProp;
-  }
-
-  return options.shouldForwardProp;
 }
 
 export default function createStyledComponent(
@@ -232,6 +209,17 @@ export default function createStyledComponent(
       ? Array.prototype.concat(target.attrs, attrs).filter(Boolean)
       : attrs;
 
+  let { shouldForwardProp } = options;
+  // $FlowFixMe
+  if (isTargetStyledComp && target.shouldForwardProp) {
+    if (shouldForwardProp) {
+      // compose nested shouldForwardProp calls
+      // $FlowFixMe
+      shouldForwardProp = prop => target.shouldForwardProp(prop) && options.shouldForwardProp(prop);
+    }
+    ({ shouldForwardProp } = target);
+  }
+
   const componentStyle = new ComponentStyle(
     isTargetStyledComp
       ? // fold the underlying StyledComponent rules up (implicit extend)
@@ -258,7 +246,7 @@ export default function createStyledComponent(
   WrappedStyledComponent.attrs = finalAttrs;
   WrappedStyledComponent.componentStyle = componentStyle;
   WrappedStyledComponent.displayName = displayName;
-  WrappedStyledComponent.shouldForwardProp = getShouldForwardProp(options, target, isTargetStyledComp);
+  WrappedStyledComponent.shouldForwardProp = shouldForwardProp;
 
   // this static is used to preserve the cascade of static classes for component selector
   // purposes; this is especially important with usage of the css prop
--- a/packages/styled-components/src/models/StyledComponent.js
+++ b/packages/styled-components/src/models/StyledComponent.js
@@ -75,6 +75,7 @@ interface StyledComponentWrapperProperties {
   displayName: string;
   foldedComponentIds: Array<string>;
   target: Target;
+  shouldForwardProp: Function;
   styledComponentId: string;
   warnTooManyClasses: $Call<typeof createWarnTooManyClasses, string, string>;
 }
@@ -119,6 +120,7 @@ function useStyledComponentImpl<Config: {}, Instance>(
     // $FlowFixMe
     defaultProps,
     foldedComponentIds,
+    shouldForwardProp,
     styledComponentId,
     target,
   } = forwardedComponent;
@@ -145,15 +147,25 @@ function useStyledComponentImpl<Config: {}, Instance>(
 
   const isTargetTag = isTag(elementToBeCreated);
   const computedProps = attrs !== props ? { ...props, ...attrs } : props;
-  const shouldFilterProps = isTargetTag || 'as' in computedProps || 'forwardedAs' in computedProps;
+  const shouldFilterProps = shouldForwardProp || isTargetTag || 'as' in computedProps || 'forwardedAs' in computedProps;
   const propsForElement = shouldFilterProps ? {} : { ...computedProps };
+
+  // use explicit `shouldForwardProp` function if one is set, otherwise
+  // if no explicit `shouldForwardProp` set then
+  //   if we're on a tag, use `validAttr`
+  //     i.e. Don't pass through non HTML tags through to HTML elements
+  // else (we're on a component - pass through all by default)
+  //   () => true
+  const shouldReallyForwardProp = shouldForwardProp || (
+    isTargetTag ? validAttr : () => true
+  );
 
   if (shouldFilterProps) {
     // eslint-disable-next-line guard-for-in
     for (const key in computedProps) {
       if (key === 'forwardedAs') {
         propsForElement.as = computedProps[key];
-      } else if (key !== 'as' && key !== 'forwardedAs' && (!isTargetTag || validAttr(key))) {
+      } else if (key !== 'as' && key !== 'forwardedAs' && (shouldReallyForwardProp(key))) {
         // Don't pass through non HTML tags through to HTML elements
         propsForElement[key] = computedProps[key];
       }
@@ -178,6 +190,20 @@ function useStyledComponentImpl<Config: {}, Instance>(
   propsForElement.ref = refToForward;
 
   return createElement(elementToBeCreated, propsForElement);
+}
+
+function getShouldForwardProp(options, target, isTargetStyledComp) {
+  // $FlowFixMe
+  if (isTargetStyledComp && target.shouldForwardProp) {
+    if (options.shouldForwardProp) {
+      // compose nested shouldForwardProp calls
+      // $FlowFixMe
+      return prop => target.shouldForwardProp(prop) && options.shouldForwardProp(prop);
+    }
+    return target.shouldForwardProp;
+  }
+
+  return options.shouldForwardProp;
 }
 
 export default function createStyledComponent(
@@ -232,6 +258,7 @@ export default function createStyledComponent(
   WrappedStyledComponent.attrs = finalAttrs;
   WrappedStyledComponent.componentStyle = componentStyle;
   WrappedStyledComponent.displayName = displayName;
+  WrappedStyledComponent.shouldForwardProp = getShouldForwardProp(options, target, isTargetStyledComp);
 
   // this static is used to preserve the cascade of static classes for component selector
   // purposes; this is especially important with usage of the css prop
@@ -296,6 +323,7 @@ export default function createStyledComponent(
       componentStyle: true,
       displayName: true,
       foldedComponentIds: true,
+      shouldForwardProp: true,
       self: true,
       styledComponentId: true,
       target: true,
--- a/packages/styled-components/src/models/StyledComponent.js
+++ b/packages/styled-components/src/models/StyledComponent.js
@@ -156,7 +156,7 @@ function useStyledComponentImpl<Config: {}, Instance>(
     for (const key in computedProps) {
       if (key === 'forwardedAs') {
         propsForElement.as = computedProps[key];
-      } else if (key !== 'as' && key !== 'forwardedAs' && (!propFilterFn || propFilterFn(key))) {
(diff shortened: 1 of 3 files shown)


stevesims:
Apologies - I've not had a chance to rebase this yet - I'll see if I can do that this afternoon

penx:
Fixes #439

diondiondion:
Does this support statically forwarding the `as` prop, too?

In other words, will a component as defined below pass the `as` prop on to the wrapped styled component without having to use the `forwardedAs` prop instead?
 
```js
const Comp = styled(AnotherStyledComponent).withConfig({
  shouldForwardProp: prop => !['as'].includes(prop)
})`
  /* ... */
`;
```

penx:
@diondiondion what would a unit test look like for that? Something like this?

```jsx
    it('supports statically forwarding the as prop', () => {
       const StyledDiv = styled('div')`
         color: red;
       `;
       const StyledDiv2 = styled('div')`
         color: blue;
       `;

       const ComposedDiv = styled(StyledDiv).withConfig({
         shouldForwardProp: prop => !['as'].includes(prop)
       })``;
       const wrapper = TestRenderer.create(<ComposedDiv as={StyledDiv2} />);
       const { props } = wrapper.root.findByType('div');
       // expect ?
     });
```

stevesims:
@diondiondion good question - and to be honest, I'm not sure.  I did wrote some tests concerning how `as` works, but I'm not entirely sure they cover what you're asking about.  Unfortunately I don't have the time to check on this right now, and won't until Monday at the earliest.

probablyup:
Fwiw we would not deprecate forwardedAs so might not be an issue

diondiondion:
I realise now that the example in my previous question had a typo in it, it was set up to _prevent_ the `as` prop from being passed on. I've edited it now.

> what would a unit test look like for that? Something like this?

@penx Let's use an intermediate JSX component as an example as that's the most likely real world use case I can think of.

```jsx
it('supports statically forwarding the as prop', () => {
	const Clickable = styled('a')`
		background-color: pink;
	`;
	const Button = ({as, className, children}) => (
		<Clickable as={as} className={className}>Hello {children}</Clickable>
	);

	const MenuButton = styled(Button).withConfig({
		shouldForwardProp: prop => prop === 'as'
	})`
		margin-right: 4px;
	`;

	const wrapper = TestRenderer.create(<MenuButton as="button">world</MenuButton>);
});
```

Given this example, I'd expect The rendered component's
- element type to be "button"
- text content to equal "Hello world"
- styles to be `background-color: pink; margin-right: 4px;`


> Fwiw we would not deprecate forwardedAs so might not be an issue

@probablyup In my mind `forwardedAs` has one valid use case – when it's used at the same time as the `as` prop.

However, the scenario where you _have_ to use it _instead of_ the `as` prop because not doing so would break the component or its styles is why I'm advocating for it. In those cases, forwarding the `as` prop statically would be a MUCH safer option. The lack of this ability creates inconsistent component APIs that unnecessarily require knowing the internals of a component.

probablyup:
We’re not making subtractive API changes at this point so we can
re-evaluate at a later date

diondiondion:
Not sure how this is a "subtractive API change", but alright. But wouldn't _not_ introducing this capability from the get go make adding it later a potential breaking change? It seems like that's how it _should_ work. `shouldForwardProp: prop => prop === 'as'` reads pretty obvious to me.

karlhorky:
> @penx: Fixes #439

Also:

Fixes #2878

@stevesims you need to add these to the original issue description, and then those issues will be automatically closed in case this gets accepted.

packages/styled-components/src/models/StyledComponent.js
```diff
@@ -145,15 +147,25 @@ function useStyledComponentImpl<Config: {}, Instance>(
...
+  const shouldFilterProps = shouldForwardProp || isTargetTag || 'as' in computedProps || 'forwardedAs' in computedProps;
   const propsForElement = shouldFilterProps ? {} : { ...computedProps };
 
+  // use explicit `shouldForwardProp` function if one is set, otherwise
+  // if no explicit `shouldForwardProp` set then
+  //   if we're on a tag, use `validAttr`
+  //     i.e. Don't pass through non HTML tags through to HTML elements
+  // else (we're on a component - pass through all by default)
+  //   () => true
+  const shouldReallyForwardProp = shouldForwardProp || (
```
kitten: I think the comment here is redundant if this was a little clearer in the first place 😅 

If we separate this out a little then we can make this very clear.

- `shouldFilterProps` could be reverted and we can remove `isTargetTag` from it.
- We can introduce a separate variables: `const propFilterFn = shouldForwardProp || (isTargetTag && validAttr);`.
- `if (shouldFilterProps)` becomes `if (shouldFilterProps || propFilterFn)`
- Then `shouldReallyForwardProp(key)` (which btw has unnecessary parens) can be replaced by: `(!propFilterFn || propFilterFn(key))`
- lastly we can now rename `shouldFilterProps` to `shouldFilterComputedProps`

packages/styled-components/src/models/StyledComponent.js
```diff
@@ -180,6 +192,20 @@ function useStyledComponentImpl<Config: {}, Instance>(
...
+  if (isTargetStyledComp && target.shouldForwardProp) {
+    if (options.shouldForwardProp) {
+      // compose nested shouldForwardProp calls
+      // $FlowFixMe
+      return prop => target.shouldForwardProp(prop) && options.shouldForwardProp(prop);
+    }
+    return target.shouldForwardProp;
+  }
+
+  return options.shouldForwardProp;
```
kitten: @probablyup This can probably be done separately but we can totally add `shouldForwardProp` as a shortcut to replace the manual `withConfig`, right? Just to have alignment with Emotion.

I'd also argue now that we're touching filtering props, "transient props" (props that are always filtered) could also be revived after merging this. (Ref: #2093)

packages/styled-components/src/models/StyledComponent.js
```diff
@@ -75,6 +75,7 @@ interface StyledComponentWrapperProperties {
   displayName: string;
   foldedComponentIds: Array<string>;
   target: Target;
+  shouldForwardProp: Function;
```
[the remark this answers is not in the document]
stevesims: :+1:

packages/styled-components/src/models/StyledComponent.js
```diff
@@ -180,6 +192,20 @@ function useStyledComponentImpl<Config: {}, Instance>(
   return createElement(elementToBeCreated, propsForElement);
 }
 
+function getShouldForwardProp(options, target, isTargetStyledComp) {
```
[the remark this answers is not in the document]
stevesims: :+1:

packages/styled-components/src/models/StyledComponent.js
stevesims: done - was a little bit more involved, but essentially followed your suggestions - also didn't rename `shouldFilterProps` cos that didn't feel entirely worth it :grin:

packages/styled-components/src/models/StyledComponent.js
```diff
@@ -206,6 +209,17 @@ export default function createStyledComponent(
...
 
+  let { shouldForwardProp } = options;
+  // $FlowFixMe
+  if (isTargetStyledComp && target.shouldForwardProp) {
+    if (shouldForwardProp) {
+      // compose nested shouldForwardProp calls
+      // $FlowFixMe
+      shouldForwardProp = prop => target.shouldForwardProp(prop) && options.shouldForwardProp(prop);
+    }
+    ({ shouldForwardProp } = target);
```
kitten: This doesn’t seem right maybe, 🤔 was this supposed to be in an `else` statement?

nit: Also can we drop the destructuring here please? We’re also not destructuring for the arrow function composition above, so let’s maybe settle on one pattern for both?

stevesims: heh - i didn't want to use destructuring here - eslint complained at me :wink: happy to drop

stevesims: and yes - should have had an `else` in there - good spot :grin:

ajkl2533:
Hi there do you have any timeline when this will be released?

stevesims:
@diondiondion FYI i've not been ignoring your question, and have looked into it quite deeply.

There's an issue with the test that you've proposed:

> ```js-jsx
> it('supports statically forwarding the as prop', () => {
> 	const Clickable = styled('a')`
> 		background-color: pink;
> 	`;
> 	const Button = ({as, className, children}) => (
> 		<Clickable as={as} className={className}>Hello {children}</Clickable>
> 	);
> 
> 	const MenuButton = styled(Button).withConfig({
> 		shouldForwardProp: prop => prop === 'as'
> 	})`
> 		margin-right: 4px;
> 	`;
> 
> 	const wrapper = TestRenderer.create(<MenuButton as="button">world</MenuButton>);
> });
> ```
> 
> Given this example, I'd expect The rendered component's
> 
> * element type to be "button"
> * text content to equal "Hello world"
> * styles to be `background-color: pink; margin-right: 4px;`

The problem is that by providing an `as` prop to `MenuButton` you're telling *that* component to use a different component to render than the specified `Button` component.  As a result your text content will be just `world`, with only the `margin-right` style applied.

Without using `forwardedAs`, the only way for the system to know that it should ignore the `as` prop being passed to `MenuButton` would be if the passed in `shouldForwardProp` function is analysed to see what it does when presented with an `as` prop.  This doesn't feel entirely wise to me.

It seems to me that the current `forwardedAs` approach is the best compromise to this issue.

diondiondion:
> FYI i've not been ignoring your question, and have looked into it quite deeply.

@stevesims No worries, thanks for looking into it.

> The problem is that by providing an as prop to MenuButton you're telling that component to use a different component to render than the specified Button component.

Yes, but that's exactly the point of my request. I want to be able to tell styled-components that `MenuButton` should never "act" on the `as` prop and instead directly forward it to the `Button` component that it wraps (see #2129 for more detail).

The `shouldForwardProp` option _sounds_ to me like it should be able to handle this, but maybe it's not the right tool for the job. (In which case we'd need an additional option, something like `ignoreAs` or [`allowAs` like in astroturf](https://github.com/4Catalyzer/astroturf#as-prop); but that'd be a discussion for another place).

> Without using `forwardedAs`, the only way for the system to know that it should ignore the `as` prop being passed to `MenuButton` would be if the passed in `shouldForwardProp` function is analysed to see what it does when presented with an `as` prop. This doesn't feel entirely wise to me.

Isn't every prop passed to the `shouldForwardProp` function? I thought the complexity came from making sure that when the `as` prop is _not_ handled in `shouldForwardProp`, the current behaviour stays the same, but it didn't seem like an unsurmountable problem.

Steveeeie:
Any updates on when this will be released?

Martinnord:
I feel with @Steveeeie. Any plans on this being released soon?

lucastobrazil:
Ppppplllleeease can we merge this? i was about to start migrating to Emotion for this very reason but already started missing the feel of s-c.

aulneau:
I've been hoping for this feature to be implemented for so long, would absolutely love to see it merged! Thanks for all the work on this.

lucastobrazil:
@kitten sorry for ping but we have a lot of excited people here. Anything we could do to help get it merged? Docs need to be written or anything?

probablyup:
@lucastobrazil Yes a PR for the website to add this API documentation would be great. I think this is ready to be merged though.

probablyup:
We also need a PR for `StyledNativeComponent` with the same changes.

peduarte:
Ohhhh yes! So so so happy about this. Thank you everyone for your work <3

lucastobrazil:
Just trying this out locally - based on the tests written, i've tried doing this:

```jsx
const Box = styled.div.withConfig({
  shouldForwardProp: prop => !['font-family'].includes(prop)
})(
  { boxSizing: 'border-box' },
  compose(
    space,
    // ...
  )
);
```

But it doesn't seem to call the `withConfig` method?

Martinnord:
I'm facing the same issue as @lucastobrazil. Is it something we are missing?

lucastobrazil:
@Martinnord I managed to get it working with the local styled components sandbox app that is bundled with the mono repo. Following the example in the test it worked. 

Was trying to use `yarn link` to link my project locally to this package, but it was failing. 

Do we have any visibility into the release plan for the next official npm release with this feature in it?

Steveeeie:
@probablyup Are you able to give an update as to when this will be part of a release?

probablyup:
Near future. There’s an SSR bug I’m looking to resolve before cutting 5.1

tuck2002:
Near future meaning? Rebuilding a component library with Styled System and this is my biggest hang up at the moment.

Steveeeie:
@probablyup can we get an update on when this will be released?

Martinnord:
image

aulneau:
@probablyup anything we can do to help? seems like a lot of folks are waiting for this and would probs be happy to assist :)

probablyup:
I was waiting on https://github.com/styled-components/styled-components/pull/3061 but I think I'll just cut a release later tonight since it's taking a while.

StevenLangbroek:
Yeah sorry I wouldn't wait on that, need to figure out what the hell is going on before I can submit a fix, and since it _is_ a fix we can cut a patch when it's ready.

probablyup:
Out in 5.1, thanks everyone for pitching in! https://github.com/styled-components/styled-components/releases/tag/v5.1.0

kachkaev:
Discussion of typings for this new feature: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/43693

Issue #1617, opened 2018-03-18
RFC: Temporarily fork and develop a separate native version


kitten:
## Summary

Let's start a separate "styled-native-next" fork that'd be used to:

- Rewrite large parts of our codebase (especially the ones providing legacy hacks and support)
- Provide the opportunity to split out a "core" package which can be used for the web library eventually from this rewrite
- Improve React Native support and performance
- Provide a breeding ground for our ISTF/Sweetsour effort

## Motivation

Currently there's two problems that we're aiming to solve in the future. Our codebase's organisation and some bundlesize and performance concerns.

Most of these can be addressed by larger refactors, which are currently hard to accomplish and navigate. Instead of revamping the entire codebase in one fell swoop, we can opt to develop a "styled-native-next" package.

This package actually has another motivation entirely: Provide a more solid and performant styled-components version for React Native. It would address the problems around forwarding interpolations to the final styles that reach React Native, which would improve our support for the `Animated` API.

This is all possible and combined with out Sweetsour/ISTF effort (See: https://github.com/philpl/sweetsour). Which is easier to start with on React Native, since it won't need any CSS rule flattening or autoprefixing — only shorthands will need to be implemented.

Also, at the end of this refactor we'd hopefully be in a position to create a "core" package that can be used by the web library, which can eventually lead to a "web", "native", and "core" repo setup.

## Plan

- Fork the current library
- Remove web-specific bits
- Integrate the sweetsour-parser, optimally with a babel plugin
- Write a shorthand transformer, in parallel to other tasks
- Rewrite core parts of the library, like the interpolation handling and StyledComponent file
- Split generated ISTF styles into static, dynamic, and animated
- Segregate into a core/native package structure
- Reintegrate core package and refactors into a web shard of the library
- Combine new code into current repo

## Champions

- Myself (@kitten)
- Jani Eväkallio (@jevakallio) has helped to form this plan and will potentially be able to help out a little
- Evan Scott (@probablyup) is motivated in finding parts that can be refactored into the babel plugin and other parts of the source that can be rewritten for bundle size benefits

Noitidart:
Thank you team!!!

If possible, it would be awesome if as much as possible used `StyleSheet.create`. That is a huge perf boost and one of the major reasons why we had to pull out styled-components in our RN project (i fought hard to keep it! :( i lost out). I know `StyleSheet.create` can't be used with Animated API, so this is aside from that please.

kitten:
[...]

PR #3009, opened 2020-02-08
Failing test for comments with styled-components/native

Zn4rK:
Posted this in https://github.com/styled-components/css-to-react-native/issues/133, but later realised that it probably fits better here.

```ts
styled.View`
  // background: green;
  background: red;
`;
```

Yields:
```
Invariant Violation: "//" is not a valid style property.
StyleSheet generated: {
  "//": "green",
  "backgroundColor": "red"
}
Valid style props: [
  "alignContent",
  "alignItems",
  "alignSelf",
  "aspectRatio",
   ...etc
]
```

### DIFF
--- a/packages/styled-components/src/native/test/native.test.js
+++ b/packages/styled-components/src/native/test/native.test.js
@@ -140,7 +140,6 @@ Object {
       console.warn = oldConsoleWarn;
     }
   });
-
   // https://github.com/styled-components/styled-components/issues/1266
   it('should update when props change', () => {
     const Comp = styled.View`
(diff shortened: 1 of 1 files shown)


kitten:
It may make sense to move this over to this repository, which does the CSS transformations on React Native in styled-components: https://github.com/styled-components/css-to-react-native

Zn4rK:
> It may make sense to move this over to this repository, which does the CSS transformations on React Native in styled-components: https://github.com/styled-components/css-to-react-native

I started there, but with regards to @jacobp100's [latest comment](https://github.com/styled-components/css-to-react-native/issues/133#issuecomment-586965845) on the original issue, I think it should be handled by styled-components.

kitten:
Closing since the implementation of the React Native feature is frozen unless we start addressing: #1617

PR #3018, opened 2020-02-12
revise SSR+hydration functionality to be more fault-tolerant

### LINKED ISSUE #3017: Fix Rehydration regex causing infinite loops for malformed inputs
The regex based rehydration can fall into an infinite loop in some edge cases. As part of this issue we should:

- Update the regex to prevent the edge case in the first place
- Investigate why it caused an infinite loop and protect against similar loops to prevent this issue in the future

Edit: I'm now actually not convinced it's an infinite loop. I've got a minimal reproduction that eventually completes after a very long time, so it may be a Regex catastrophic backtracking issue.

This issue was discovered in #2972. There some CSS was being output (malformed by stylis due to `/*!` comments) and caused the following rule to be input into rehydration:

```css
{/*! normalize.css v8.0.1 | MIT License | github.com/necolas/normalize.css *//*!                                                                                                                               
 *  Font Awesome 4.7.0 by @davegandy - http://fontawesome.io - @fontawesome                                                                                                                                    
 *  License - http://fontawesome.io/license (Font: SIL OFL 1.1, CSS: MIT License)                                                                                                                              
 */} 
```

This causes an infinite loop since (presumably) it's a selector-less rule body.

Edit: Because I forgot as well, to recap why we switched away from CSSOM-based rehydration in #2872: There are quirks and bugs in browsers that cause `cssText` on rules to be unreliable 🤦‍♂ See: https://bugzilla.mozilla.org/show_bug.cgi?id=1594241

Edit 2: This test here gets slower as more characters (one by one!) are added to the invalid `normalize.css` comment line:

```js
  it.only('tolerates malformed CSS', () => {
    document.head.innerHTML = `
      <style ${SC_ATTR} ${SC_ATTR_VERSION}="${SC_VERSION}">
        {/*! normalize.css v8.
          }
        .rule {}
        ${SC_ATTR}.g1[id="idA"]{content:""}
      </style>
    `;

    const sheet = new StyleSheet({ isServer: true });
    rehydrateSheet(sheet);
  });
```

Edit 3: Content doesn't matter so it seems to somehow be the newline in the rule, which usually is stripped out by stylis, so this may be why it's such a rare edge case (?)

Edit 4: Using `/([^{]*){((?:{[^}]*}|(?!{).*?)*)}/g;` instead (removing `.*`) and adding `trim` later helps to keep the time down but after a couple more characters the time grows again. So there's another backtracking issue in here.

kitten:
Fix #3017

### DIFF
--- a/packages/styled-components/src/sheet/Rehydration.js
+++ b/packages/styled-components/src/sheet/Rehydration.js
@@ -5,7 +5,7 @@ import { getIdForGroup, setGroupForId } from './GroupIDAllocator';
 import type { Sheet } from './types';
 
 const SELECTOR = `style[${SC_ATTR}][${SC_ATTR_VERSION}="${SC_VERSION}"]`;
-const RULE_RE = /(?:\s*)?(.*?){((?:{[^}]*}|(?!{).*?)*)}/g;
+const RULE_RE = /([^{]*){((?:[^{}]+(?:{[^}]*})?)*)}/g;
 const MARKER_RE = new RegExp(`^${SC_ATTR}\\.g(\\d+)\\[id="([\\w\\d-]+)"\\]`);
 
 export const outputSheet = (sheet: Sheet) => {
@@ -60,7 +60,7 @@ const rehydrateSheetFromTag = (sheet: Sheet, style: HTMLStyleElement) => {
   // parts = [match, selector, content]
   // eslint-disable-next-line no-cond-assign
   while ((parts = RULE_RE.exec(rawHTML))) {
-    const marker = parts[1].match(MARKER_RE);
+    const marker = parts[1].trim().match(MARKER_RE);
 
     if (marker) {
       const group = parseInt(marker[1], 10) | 0;
--- a/packages/styled-components/src/sheet/Rehydration.js
+++ b/packages/styled-components/src/sheet/Rehydration.js
@@ -1,10 +1,9 @@
 // @flow
 
-import { SC_ATTR, SC_ATTR_ACTIVE, SC_ATTR_VERSION, SC_VERSION } from '../constants';
+import { SPLITTER, SC_ATTR, SC_ATTR_ACTIVE, SC_ATTR_VERSION, SC_VERSION } from '../constants';
 import { getIdForGroup, setGroupForId } from './GroupIDAllocator';
 import type { Sheet } from './types';
 
-const SPLITTER = '/*!sc*/';
 const SELECTOR = `style[${SC_ATTR}][${SC_ATTR_VERSION}="${SC_VERSION}"]`;
 const MARKER_RE = new RegExp(`^${SC_ATTR}\\.g(\\d+)\\[id="([\\w\\d-]+)"\\].*?"([^"]*)`);
 
@@ -34,7 +33,7 @@ export const outputSheet = (sheet: Sheet) => {
 
     // NOTE: It's easier to collect rules and have the marker
     // after the actual rules to simplify the rehydration
-    css += `${rules}${SPLITTER}${selector}{content:"${content}"}${SPLITTER}`;
+    css += `${rules}${selector}{content:"${content}"}${SPLITTER}\n`;
   }
 
   return css;
@@ -53,13 +52,12 @@ const rehydrateNamesFromContent = (sheet: Sheet, id: string, content: string) =>
 };
 
 const rehydrateSheetFromTag = (sheet: Sheet, style: HTMLStyleElement) => {
-  const rawHTML = style.innerHTML;
+  const parts = style.innerHTML.split(SPLITTER);
   const rules: string[] = [];
-  const parts = rawHTML.split(SPLITTER);
 
-  parts.forEach(part => {
-    const marker = part.trim().match(MARKER_RE);
-
+  for (let i = 0, l = parts.length; i < l; i++) {
+    const part = parts[i].trim();
+    const marker = part.match(MARKER_RE);
     if (marker) {
       const group = parseInt(marker[1], 10) | 0;
       const id = marker[2];
@@ -75,9 +73,9 @@ const rehydrateSheetFromTag = (sheet: Sheet, style: HTMLStyleElement) => {
 
       rules.length = 0;
     } else {
-      rules.push(part.trim());
+      rules.push(part);
     }
-  });
+  }
 };
 
 export const rehydrateSheet = (sheet: Sheet) => {
--- a/packages/styled-components/src/sheet/Rehydration.js
+++ b/packages/styled-components/src/sheet/Rehydration.js
@@ -57,7 +57,10 @@ const rehydrateSheetFromTag = (sheet: Sheet, style: HTMLStyleElement) => {
 
   for (let i = 0, l = parts.length; i < l; i++) {
     const part = parts[i].trim();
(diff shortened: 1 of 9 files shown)


packages/styled-components/src/sheet/Rehydration.js
```diff
@@ -5,7 +5,7 @@ import { getIdForGroup, setGroupForId } from './GroupIDAllocator';
 import type { Sheet } from './types';
 
 const SELECTOR = `style[${SC_ATTR}][${SC_ATTR_VERSION}="${SC_VERSION}"]`;
-const RULE_RE = /(?:\s*)?(.*?){((?:{[^}]*}|(?!{).*?)*)}/g;
+const RULE_RE = /([^{]*){((?:[^{}]+(?:{[^}]*})?)*)}/g;
```
kitten: - Leading `(?:\s*)?` is gone. This isn't needed because `exec` will keep walking forwards character by character. Also we want to start matching as soon as possible and if we later don't match, then this will slow the regex down because it'll start backtracking.
- Replaced `.*?` (btw `*?` is redundant) with `[^{]*` because we know that a `{` must follow. So we safe some time here by telling the Regex that we're matching as much as possible until that char instead of eagerly matching when the following characters don't quite match.
- Internally in the regex we had the same problems but more pronounced.

probablyup: This doesn't work for certain types of deeply nested media queries unfortunately, use this as a test string:

```
@media (max-width:300px){.fvsNoc{display:block;}@supports (display:-webkit-box) or (display:-webkit-flex) or (display:-ms-flexbox) or (display:flex){.fvsNoc{display:-webkit-box;display:-webkit-flex;display:-ms-flexbox;display:flex;}}@media (max-width:320px){.fvsNoc{display:block;}}}
data-styled.g1[id="pages__Container-sc-4wa63o-0"]{content:"fvsNoc,"}
.dBiKfw{font-size:8rem;font-weight:900;color:darkgreen;text-align:center;}
data-styled.g2[id="pages__Heading-sc-4wa63o-1"]{content:"dBiKfw,"}
```

(it's from the original issue I created the regex from)

The revised regex breaks at the second `@media`

probablyup: My suggestion here is to change the SSR emitter to add a special character CSS doesn't process to serve as the "split" point (for use in `string.split()` and just ditch the splitting regex.

kitten: We could likely just add `/*!sc*/` as splitting markers

packages/styled-components/src/sheet/Rehydration.js
```diff
@@ -53,15 +52,12 @@ const rehydrateNamesFromContent = (sheet: Sheet, id: string, content: string) =>
...
-
-  // parts = [match, selector, content]
-  // eslint-disable-next-line no-cond-assign
-  while ((parts = RULE_RE.exec(rawHTML))) {
-    const marker = parts[1].match(MARKER_RE);
 
+  for (let i = 0, l = parts.length; i < l; i++) {
+    const part = parts[i].trim();
+    const marker = part.match(MARKER_RE);
     if (marker) {
```
probablyup: Let's add a `continue` shortcircuit at the beginning if `part` is empty. Since we're using the `split` method it's possible there might be crud in there

probablyup:
Published for testing under `styled-components@5.0.1-revisedssr`

packages/styled-components/src/sheet/test/GroupedTag.test.js
```diff
@@ -89,7 +89,7 @@ it('does supports large group numbers', () => {
   expect(groupedTag.length).toBeGreaterThan(group);
   expect(tag.length).toBe(1);
   expect(groupedTag.indexOfGroup(group)).toBe(0);
-  expect(groupedTag.getGroup(group)).toBe('.test {}\n');
+  expect(groupedTag.getGroup(group)).toBe('.test {}/*!sc*/\n');
```
abbyhowell: I don't understand why this change adds this string to the output of this method. How do we know this change isn't introducing a new, unwanted behavior, if all these tests needed to change?

packages/styled-components/src/sheet/test/Rehydration.test.js
```diff
@@ -94,4 +100,19 @@ describe('rehydrateSheet', () => {
...
   });
+
+  it('tolerates long, malformed CSS', () => {
+    document.head.innerHTML = `
+      <style ${SC_ATTR} ${SC_ATTR_VERSION}="${SC_VERSION}">
+        {xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+          xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+          }
+        .rule {}/*!sc*/
+        ${SC_ATTR}.g1[id="idA"]{content:""}/*!sc*/
```
abbyhowell: Maybe this test CSS should include a real inline comment with `/*!` wrapped around it, rather than only `sc`? What am I missing here?

packages/styled-components/src/sheet/test/Rehydration.test.js
```diff
@@ -23,31 +23,37 @@ describe('outputSheet', () => {
...
-      '.b {}',
-      `${SC_ATTR}.g22[id="idB"]{content:"nameB,"}`,
-    ]);
+      .split('/*!sc*/');
+
+    expect(output).toMatchInlineSnapshot(`
+      Array [
+        ".a {}",
+        "
+      data-styled.g11[id=\\"idA\\"]{content:\\"nameA,\\"}",
```
abbyhowell: `data-styled` should probably still be `${SC_ATTR}` here, to accommodate the different `process.env` values it could be (https://github.com/styled-components/styled-components/blob/d0e8eab8cf391630294c6ac9e81f206eb64e5e53/packages/styled-components/src/constants.js#L6:14)

packages/styled-components/src/sheet/test/Rehydration.test.js
```diff
@@ -23,31 +23,37 @@ describe('outputSheet', () => {
...
+        "
+      .b {}",
+        "
+      data-styled.g22[id=\\"idB\\"]{content:\\"nameB,\\"}",
```
abbyhowell: Why is the output double-escaped now when it wasn't before? That seems like an unwanted change.

probablyup: I think this is just a jest snapshot serialization artifact

packages/styled-components/src/sheet/test/Rehydration.test.js
probablyup: This is just the new "splitter" marker we are adding. You can consider it an implementation detail for SSR.

packages/styled-components/src/sheet/test/GroupedTag.test.js
probablyup: This is used by `outputSheet` in the SSR functionality. We need to add the marker at this point so the individual rules are split correctly during rehydration. This PR generally changes the whole rehydration mechanic, which previously used regexes to do the splitting.

PR #3020, opened 2020-02-13
Keep the style prop stable for React Native

Federkun:
When trying to styling a `memo` component in React Native, the identity of `style` prop makes it to re-render nonetheless. This behaviour is different from ReactJS.

What happens on reactjs: https://codesandbox.io/s/react-hooks-counter-demo-dx4c2
The className doesn't change between renders, so the component skip rendering correctly.

What happens on react native: https://codesandbox.io/s/react-hooks-counter-demo-edh0d
Since the `style` props change every time, the component will render as a normal non-pure component.

This PR is a PoC, and try to address that issue.

Cons:

- Adding two extra deps
- Styling a `React.memo` component is rare, and having a deep compare operation may hurt performance more than less.

I had this unexpected issue at work, when working with some animation. Even providing some sort of warning as an alternative would do the job.

### DIFF
--- a/packages/styled-components/package.json
+++ b/packages/styled-components/package.json
@@ -69,6 +69,8 @@
     "babel-plugin-styled-components": ">= 1",
     "css-to-react-native": "^3.0.0",
     "hoist-non-react-statics": "^3.0.0",
+    "lodash.isequal": "^4.5.0",
+    "memoize-one": "^5.1.1",
     "shallowequal": "^1.1.0",
     "supports-color": "^5.5.0"
   },
--- a/packages/styled-components/src/models/StyledNativeComponent.js
+++ b/packages/styled-components/src/models/StyledNativeComponent.js
@@ -24,7 +24,10 @@ class StyledNativeComponent extends Component<*, *> {
 
   attrs = {};
 
-  generateStyles = memoize((...styles) => styles, isDeepEqual);
+  generateStyles = memoize(
+    (generatedStyles, style) => [generatedStyles].concat(style),
+    isDeepEqual
+  );
 
   render() {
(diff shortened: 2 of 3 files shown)


kitten:
Right now the styles-components native variant isn’t well supported admittedly btw,  but that’s just a side note 😅

I’d argue that this can indeed be confusing but you’re making the assumption here that the check is worth it because `styled(memo(AComponent))` will redender less.

I’d argue that `styled(AComponent)` should be used sparingly and if StyledComponents after kept towards the lead nodes that they’re a very small minority of cases. All other cases of `styled.View` etc will likely not benefit from this and will just incur the comparison performance penalty.

There are however faster deep comparison utilities like `react-fast-compare`. That being said though we still incur a cost of comparison.

The other point you’re making though is that a Terences may cause unexpected behaviour for some of your components. I’d say that means that some of your components are not resilient enough to rerenders where styles change, which is a separate problem. This can be solved instead with a deep-equal comparison hook if you’re reading `style` directly.

Alternatively `React.memo` actually allows you to pass a comparison function though, which could be a deep-equal comparison function.

So I’d say this isn’t a necessary change, but happy to keep this open for further discussion ♥️

Federkun:
I do agree that the overhead here is enough to hold back on this.

> Alternatively React.memo actually allows you to pass a comparison function though, which could be a deep-equal comparison function.

I think this is the solution.
I was more interested in what I'd perceive as an "unexpected behaviour". I'm always of the idea that your application should work the same even if the entire tree render every time, but unfortunately there are application out there that elevate react's optimisations from implementation details to abstraction that you depends on.

Thanks anyway!

Issue #3032, opened 2020-02-21
Refactor isElement component selector warning to be an error


juansalvatore:
### Code example
Here's an example of how this warning works now:
https://codesandbox.io/s/iselement-warning-example-32xrd

### Warning
![image](https://user-images.githubusercontent.com/11010928/75061288-32716380-54bf-11ea-94c5-d311abe7ba48.png)

### Suggestion
I would like to suggest to make this an error to prevent adding styles that won't work.

GMaiolo:
This has caused major issues in the past as nested styles inside one of these invalid selectors break or have unexpected behavior

Making this an error would safe so many dev hours and hard to track bugs

Agree 100% 👍

probablyup:
This would be a breaking change, so it'll need to wait until v6.

quantizor:
Closing — this issue is stale (2+ years without resolution) and relates to an older version or has been superseded by v6.x changes. If this is still reproducible on the latest v6.x release, please open a new issue with a minimal reproduction.

PR #3033, opened 2020-02-21
Refactor isElement component selector warning to be an error

juansalvatore:
### Code example
Here's an example of how this warning works now:
https://codesandbox.io/s/iselement-warning-example-32xrd

### Warning
image

### Suggestion
I would like to suggest to make this an error to prevent adding styles that won't work.

#3032


probablyup:
This is a breaking change so it’ll have to wait for v6. Thanks!

PR #3041, opened 2020-03-01
Migrate to import * as React from 'react'

sayjeyhi:
Ref

- facebook/react#18102
- https://twitter.com/sebmarkbage/status/1231673639080038400


kitten:
It’s a little too early to do this since React doesn’t actually publish an ES module. Even if they do, there’s some things we can do to apply this on a bundling level instead I’d say 😅

For reference, I contributed `es-react` which is a rebundled version of React as an ES module which @lukejacksonn maintains

PR #3052, opened 2020-03-05
transient props

probablyup:
As an alternate (but complementary) pattern to `shouldFilterProp`, this implements the proposal detailed here: https://medium.com/@probablyup/introducing-transient-props-f35fd5203e0c


Greenheart:
Just wanted to give a big thanks for this feature! It allowed my team to work much better compared to when we used workarounds presented in https://github.com/styled-components/styled-components/issues/439

Keep up the good work! 🌍✨

PR #3055, opened 2020-03-05
[BREAKING] remove withConfig API, add options argument to styled()

### LINKED ISSUE #3053: Remove withConfig API in favor of a second argument to styled()
This is a v6 proposal and aligns our API a little more cleanly with emotion's styled API so people can migrate back to s-c more easily.

Need to modify the babel plugin accordingly.

probablyup:
closes #3053

This change aligns our API more closely to emotion's styled API to allow more easy portability between libraries. As an aside, this also removes the ability to progressively chain different options together, though I think this is a rare edge case. A corresponding babel plugin change is also needed.

### DIFF
--- a/packages/styled-components/src/constructors/constructWithOptions.js
+++ b/packages/styled-components/src/constructors/constructWithOptions.js
@@ -19,10 +19,6 @@ export default function constructWithOptions(
   // $FlowFixMe: Not typed to avoid destructuring arguments
   const templateFunction = (...args) => componentConstructor(tag, options, css(...args));
 
-  /* If config methods are called, wrap up a new template function and merge options */
-  templateFunction.withConfig = config =>
-    constructWithOptions(componentConstructor, tag, { ...options, ...config });
-
   /* Modify/inject new props at runtime */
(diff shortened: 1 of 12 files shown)


mxstbr:
This is :fire: :fire: :fire: thanks for tackling it! What if we made it a depreciation first and removed it in another major down the line, similar to the other breaking changes we've done?

1. Introduce support for `styled(tag, config)` but keep `.withConfig` for backwards-compatibility (tbd: what happens if both are used at the same time? Err out?)
1. Change the Babel plugin to transpile to the new `styled(tag, config)` API
1. Publish and announce the new major version of the Babel plugin
1. Make `.withConfig` usage log a warning to upgrade the Babel plugin
1. If possible, make the upgraded Babel plugin log a warning if you're using a version of styled-components that doesn't support the new `styled(tag, config)` API and fallback to `.withConfig`? Then we wouldn't even need a major version initially but could gradually transition people over a couple of months

That seems doable and would work well with our ethos of never breaking any users, right? WDYT?

Issue #2988, opened 2020-01-25
<video> tag does not working properly with SC


luiz504:
Sc version: 4.4.1 
Bug report - >  using the tag video throw the SC the attribute  *disablePictureInPicture* does not work.

export const Player = styled.video``;

<Player
      id="video-viewer"
      controls
      height={files.original.height}
      width={files.original.width}
      disablePictureInPicture
    >.... 

using the pure HTML tag video its work fine.

probablyup:
It needs to be added here: https://github.com/emotion-js/emotion/blob/master/packages/is-prop-valid/src/props.js

PR #3058, opened 2020-03-10
Bump is-prop-valid

egdbear:
fixes: https://github.com/styled-components/styled-components/issues/2988

https://github.com/emotion-js/emotion/blob/master/packages/is-prop-valid/src/props.js#L66


probablyup:
Thanks! This will go out with 5.1 in the next week or so

PR #3059, opened 2020-03-10
Improve support for css`` helper with css prop

zacharyliu:
When using the `css` prop in combination with the `css` tagged template helper, I noticed performance suffered if I used the macro to compile the prop. It seems that this is because a component like this:

```jsx
import React from 'react'
import { css } from 'styled-components/macro'
function Foo() {
  return <div css={css`color: red;`} />
}
```

gets transpiled into something like this:

```js
import _styled2 from "styled-components";
import _styled from "styled-components";
import { css as _css } from "styled-components";
import React from 'react';

var _StyledDiv = _styled("div").withConfig({
  displayName: "macrotest___StyledDiv",
  componentId: "sc-11gvsec-0"
})(["", ""], p => p._css2);

function Foo() {
  return React.createElement(_StyledDiv, {
    _css2: _css(["color:red;"])
  });
}
```

Note how the css helper gets re-run on every render of Foo! This happens because `babel-plugin-styled-components` expects to see a tagged template literal named "css", but the modified import name `_css` breaks this check:

https://github.com/styled-components/babel-plugin-styled-components/blob/55b43ee70bdd8cd20ba942577e3747879cf1b44b/src/visitors/transpileCssProp.js#L82

This PR changes the import processing behavior of the macro to avoid changing the names of imports. Instead of using `@babel/helper-module-imports` to generate unique import names, we should be able to keep the original names like `styled` and `css`. This allows the macro to more transparently change the `styled-components/macro` import into `styled-components`.

With the change, we get this new output:

```js
import _styled from "styled-components";
import { css } from "styled-components";
import React from 'react';

var _StyledDiv = _styled("div").withConfig({
  displayName: "macrotest___StyledDiv",
  componentId: "sc-11gvsec-0"
})(["color:red;"]);

function Foo() {
  return React.createElement(_StyledDiv, null);
}
```


probablyup:
Fwiw you don’t need to use the helper with the css prop unless you’re doing
an interpolation

zacharyliu:
I think with the `css` prop you don't even need the helper at all, it seems like the plugin extracts both plain template literals and tagged template literals in the same way. But using the `css` helper integrates better with tools like Prettier and syntax highlighters which can recognize it as CSS source.

zacharyliu:
We found one significant use case across our codebase. Without using the `css` helper with the `css` prop, there is no TypeScript typing on interpolations, so you can't access the `theme`. I'm guessing this is because if you omit the helper, the [prop value is interpreted as a plain string](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0e208f26febf93c4f412b181491b11c863118f27/types/styled-components/index.d.ts#L496), so there's no way for TypeScript to infer that the theme prop is accessible.

eg this is not possible:

```tsx
<div css={`color: ${props => props.theme.color}`} /> // does not typecheck
```

and this is:
```tsx
<div css={css`color: ${props => props.theme.color}`} /> // works
```

probablyup:
@zacharyliu that's a totally different issue and should be filed at DefinitelyTyped

zacharyliu:
I'll file an issue, but I'm not sure if this is technically fixable from the DefinitelyTyped side. As far as I'm aware there's no way to specify types for the interpolations in a plain template string, so this would need to be a new TypeScript feature. But even in the styled-components case, the function interpolations with the css prop only work because of the compile-time transform, and I wonder if TS would consider this out of scope to implement.

zacharyliu:
The TypeScript use case aside, it does seem like something should be done about this issue, since it affects any code that tries to use this API with the macro. If we don't fix the macro, some alternatives:

- Add documentation that the css helper should never be used with the css prop with the macro.
- Emit a warning when the macro encounters a css helper.

LoiKos:
Hi ! With v6 of styled-components it seems that you aimed to refactor to typescript. Will that change something for this because using `<div css={`color: ${props => props.theme.color}`} /> ` in Typescript is really a pain ?

kitten:
This looks a bit like something the Babel plugin itself could transpile and change — correct me if I'm wrong. But I'll close this for now since it's gotten stale and inactive.

PR #3068, opened 2020-03-19
Change isPlainObject to support objects created in a different context

### LINKED ISSUE #3066: Styles not generated for Style Objects in a Node VM context
Styles are not generated for components styled with [Style Objects](https://styled-components.com/docs/advanced#style-objects) when run in a [Node VM context](https://nodejs.org/api/vm.html#vm_vm_executing_javascript). Components styled with tagged template literals work perfectly well in the same Node VM context.

This is because the [`isPlainObject`](https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/utils/isPlainObject.js) function tests whether the constructor of an object is `Object`. If an object is created in a VM context, it's constructor `Object` is not equal to the `Object` prototype object in the main context.

It would be great if this function could be updated so that it doesn't rely on being executed in the same JavaScript context that the component was created in.

## Test case

```
npm i react react-dom styled-components
```

Run the following in Node:
```js
const React = require("react");
const { renderToString } = require("react-dom/server");
const { default: styled, ServerStyleSheet } = require("styled-components");
const vm = require("vm");

const context = vm.createContext({ styled, React, App: null });

vm.runInContext(`
  const Green = styled.div\`
    color: green;
  \`;

  const Red = styled.div(() => ({
    color: "red"
  }));

  App = React.createElement("div", null, [
    React.createElement(Green, { key: 0 }, null),
    React.createElement(Red, { key: 1 }, null)
  ]);
`, context);

renderToString(sheet.collectStyles(context.App));
const styleTags = sheet.getStyleTags();
console.log(styleTags);
```

## Expected output

```html
<style data-styled="true" data-styled-version="5.0.1">.lbxPDC{color:green;}
data-styled.g1[id="sc-AxjAm"]{content:"lbxPDC,"}
.dahFIE{color:red;}
data-styled.g2[id="sc-AxirZ"]{content:"dahFIE,"}
</style>
```

## Actual output

`[object Object];` is being rendered instead of `color:red;`.

```html
<style data-styled="true" data-styled-version="5.0.1">.lbxPDC{color:green;}
data-styled.g1[id="sc-AxjAm"]{content:"lbxPDC,"}
.dahFIE{[object Object];}
data-styled.g2[id="sc-AxirZ"]{content:"dahFIE,"}
</style>
```

keeganstreet:
To resolve #3066

### DIFF
--- a/packages/styled-components/src/utils/isPlainObject.js
+++ b/packages/styled-components/src/utils/isPlainObject.js
@@ -1,2 +1,8 @@
 // @flow
-export default (x: any): boolean => typeof x === 'object' && x.constructor === Object;
+import { typeOf } from 'react-is';
+
+export default (x: any): boolean =>
+  x !== null &&
+  typeof x === 'object' &&
+  x.toString() === '[object Object]' &&
+  !typeOf(x);
--- a/packages/styled-components/src/utils/isPlainObject.js
+++ b/packages/styled-components/src/utils/isPlainObject.js
@@ -4,5 +4,5 @@ import { typeOf } from 'react-is';
 export default (x: any): boolean =>
   x !== null &&
   typeof x === 'object' &&
-  x.toString() === '[object Object]' &&
+  (x.toString ? x.toString() : Object.prototype.toString.call(x)) === '[object Object]' &&
   !typeOf(x);
--- a/packages/styled-components/src/utils/test/isPlainObject.test.js
+++ b/packages/styled-components/src/utils/test/isPlainObject.test.js
@@ -1,0 +1,36 @@
+// @flow
+import vm from "vm";
+import React from "react";
+import isPlainObject from '../isPlainObject';
+
+it('returns true for an object literal', () => {
+  expect(isPlainObject({})).toEqual(true);
+});
+
+it('returns false for an instance of a class with its own toString method', () => {
+  class SomeClass {
+    toString() {
+      return 'some: thing;';
+    }
+  }
+  expect(isPlainObject(new SomeClass())).toEqual(false);
+});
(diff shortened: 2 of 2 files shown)


packages/styled-components/src/utils/isPlainObject.js
```diff
@@ -1,2 +1,8 @@
 // @flow
-export default (x: any): boolean => typeof x === 'object' && x.constructor === Object;
+import { typeOf } from 'react-is';
+
+export default (x: any): boolean =>
+  x !== null &&
+  typeof x === 'object' &&
+  x.toString() === '[object Object]' &&
```
kitten: Like indicated in the PR, this will need to be `Object.prototype.toString` to support objects with `toStrung`, including `Object.create(null)` 🙌

keeganstreet: Ah yes of course, we can't call `toString()`if it doesn't exist on that object.

The problem I was trying to address by using `x.toString()` instead of `Object.prototype.toString(x)` was to call a custom `toString` when it exists:

```js
class MyClass {
  toString() {
    return "[object MyClass]"
  }
}
const m = new MyClass();
Object.prototype.toString(m); // "[object Object]" <-- does not call the custom toString method
m.toString(); // "[object MyClass]"
```

I can solve this by using: `(x.toString ? x.toString() : Object.prototype.toString(x)) === '[object Object]'`

keeganstreet:
Hi @kitten, are you happy with this now that it has been updated?

shadforth:
Looking forward to seeing this in action! 👏

PR #3087, opened 2020-04-03
Implements rem, em, vh vw, vmin and vmax units in React Native (#1309)

### LINKED ISSUE #1309: [React Native] Expand supported CSS for React Native
I would like to start a discussion on expanding the CSS supported by `styled-components/native`.
Primary items of interest:
* rem/em/viewport units
* global syles
* media queries
* subset of pseudo classes

It would seem that some of this functionality could be included in [css-to-react-native](https://github.com/styled-components/css-to-react-native)... further improving the convenience of writing styles for React Native components with styled-components. @jacobp100, this looks to be your domain.

It appears this has been [discussed in at least a limited manner](https://github.com/styled-components/styled-components/pull/489), but that was in context of a specific method of accomplishing these features and was some time ago. 

## Looking to precedent in styled-components
Based on the inclusion of shortand and convenient syntax for properties like `transform`, it seems that styled-components is not limiting itself to simple mapping of styles to RN. (https://www.styled-components.com/docs/basics#react-native)

## Looking elsewhere in the React Native ecosystem
* https://github.com/vitalets/react-native-extended-stylesheet adds extended Stylesheets for React Native and adds a lot of convenient CSS functionality (to name a few: global styles, rem units, media queries, some pseudo classes).
* Adds viewport units: https://github.com/jmstout/react-native-viewport-units (used w/ multiplication).
* Method to add em units: https://medium.com/@elieslama/responsive-design-in-react-native-876ea9cd72a8
* It would seem that keyframes could be supported by mapping to usage of the Animated API. This may not be on the "low hanging fruit" side of the difficulty spectrum though.
* Linear gradients: https://github.com/styled-components/styled-components/issues/1170

## Addressing template questions
* How is this product-centric? By allowing usage of em/rem units, viewport units, and media queries... creating native apps in a responsive manner (so they look ideal on all devices) would become much easier. Other included items improve convenience as well.

* How does it make workarounds straight forward? Any functionality that is included eliminates the need for using helper libraries or each app having their own method to deal with responsiveness of their styles.

## Limitations of React Native in regards to styling.
As discussed here: https://github.com/styled-components/styled-components/issues/1243, some css properties (such as `text-transform`) aren't going to be easy to support, as limitations in React Native prevent them. This discussion is meant for CSS that can be supported (as evidenced elsewhere in the React Native ecosystem) with the current state of React Native.

## Related issues
* https://github.com/styled-components/styled-components/issues/1243
* https://github.com/styled-components/styled-components/issues/1170
* https://github.com/styled-components/styled-components/pull/489
*https://github.com/styled-components/styled-components/issues/835#issuecomment-313136838
* I'm sure others that I don't have the time to hunt down.

Sharcoux:
Trying to fix #1309

Ok, vh and vw are basically working. I just need to find the best way to listen to orientation/layout changes. What do you think is the best approach?

About em, I think that it should work, but I could not use the TestRenderer to test that as it doesn't handle well the Context.Consumer. Do you see a way to automate this last test ?

I'll give it a try in a real application this afternoon.

One last thing: flow generates an error because it thinks that "Dimensions" doesn't exist in react-native (which is wrong). I'm not very friendly with flow. Any idea how I should fix that?

Any thoughts are welcomed!


packages/styled-components/src/models/InlineStyle.js
```diff
@@ -34,6 +34,26 @@ export default (styleSheet: StyleSheet) => {
...
+                if(fontSizeUnit[0] === '%') value /= 100
+                // eslint-disable-next-line no-param-reassign
+                node.value = `${value}`
+                // We need to warn that the font size is specified in em
+                // eslint-disable-next-line no-param-reassign
+                unitsInformation.fontSizeInEm = true
+              }
+            }
+            // We strip the special units and mark them for later treatment
+            const extendedUnits = node.value.match(/(em|vh|vw)$/)
```
jacobp100, line 51: This isn't quite right, as it's possible to have multiple units in one definition - like `padding: 10vw 10vh`

Sharcoux: Damn. Hard to keep in mind all the possible cases...
I will need to keep track of the properties that need to be updated after the call to css-to-react-native. Do you guys have a list of this kind of properties somewhere ?

Sharcoux: Ok, here is the list I can think of:
  * can receive multiple unit values : ['padding', 'margin']
  * can receive multiple values among which there may be one unit : ['font', 'border', 'border-top', 'border-bottom', 'border-left', border-right', 'flex', 'background']
The current code should be taking care correctly of other properties.

jacobp100: If you pass in the viewport dimensions, you can do a regexp on each value and convert it to px values, then hand it off to css-to-react-native

E.g. `padding: 10vw` -> `padding: 48px`

If you use the callback String.replace, you can even track whether you did this replacement so you know whether the stylesheet depended on viewport dimensions

If you need some regexps, you can also find them [here](https://github.com/jacobp100/cssta/blob/8c9697e8d4f5ddd6c2ff9333af42ec62e31a6773/runtime/cssRegExp.ts)

Sharcoux: Ok, I think I got them all, but if I forgot some, please tell me!
It would indeed be a good way to provide the width and height of the screen and make the calculation within the script. But, I need to take care of the em anyway and this can only be done in the component...

Sharcoux:
Can someone help me fix the remaining flow errors?

packages/styled-components/src/models/InlineStyle.js
probablyup: I think this is an exhaustive list of CSS properties that can receive one of those units:

background
background-position
background-size
border
border-bottom
border-bottom-left-radius
border-bottom-right-radius
border-bottom-width
border-image
border-image-outset
border-image-slice
border-image-width
border-left
border-left-width
border-radius
border-right
border-right-width
border-spacing
border-top
border-top-left-radius
border-top-right-radius
border-top-width
border-width
bottom
box-shadow
clip
font
font-size
font-size-adjust
height
left
letter-spacing
line-height
margin
margin-bottom
margin-left
margin-right
margin-top
max-height
max-width
min-height
min-width
outline
outline-width
padding
padding-bottom
padding-left
padding-right
padding-top
right
text-indent
top
transform
transform-origin
width
word-spacing

From: https://htmldog.com/references/css/properties/

Sharcoux: Great, thx! I'll add this, but I'd like to be sure that this has a chance to be merged before working more on the details.
I need some help from anyone knowing flow, and ideally a principe approval from anyone maintaining this repo...

probablyup:
@kitten is probably most knowledgeable about flow out of us

Sharcoux:
@probablyup , except for flow, do you think that it can be merged like this? I had to make some choices because things are not that easy, so before going further I'd like to hear out your thoughts on what's already there.

probablyup:
Honestly I don't use the React Native integration at work or in side projects, so I'm not the one to review this.

Sharcoux:
Should it be @kitten ?

packages/styled-components/src/native/test/native.test.js
```diff
@@ -446,4 +446,45 @@ Object {
...
   });
+
+  describe('extended CSS support', () => {
+    it('should handle vh, vw, vmin, vmax and rem units', () => {
+      const Comp = styled(View)`
+        width: 10vmin;
+        height: 10vmax;
+        padding: 10vw 10vh;
+        border: 1rem solid black;
+      `;
```
kristerkari, line 457: I'm not sure what's the policy with unit tests in styled-components, but these two tests that you have don't properly cover the changes that you have made in the code.

What about adding tests for `vmin`/`vmax`, for the `Dimensions` event listener, for `rem` units, for different CSS property combinations, for invalid code, etc.?

kristerkari: oh sorry, vmin/vmax is already tested.

Sharcoux: Good remark. I'm not sure about how to emulate screen resize, but I'll try. rem is tested already in the border width. What should I do in case of invalid code?

Sharcoux: Sorry, I don't find how to emulate a resize of the view...

Sharcoux: About bad inputs, it seems that the current codebase doesn't handle them, so I don't think that it should be part of this PR...

kristerkari: > What should I do in case of invalid code?

I was thinking of some tests that verify that your code path only runs for certain properties and gets skipped with a property that is not supported. React Native will throw an error if you try to give it invalid styling, so I'm not sure if it actually would make any sense to test that.

kristerkari: > Sorry, I don't find how to emulate a resize of the screen or an orientation change...

I don't remember how to do it exactly, but I guess that you would need to mock the functions in the `Dimensions` module, trigger the event listener and then re-render the component. I'm not sure if there is some other way to do that easier.

Sharcoux: > > What should I do in case of invalid code?
> 
> I was thinking of some tests that verify that your code path only runs for certain properties and gets skipped with a property that is not supported.
> 
> React Native will throw an error if you try to give it invalid styling, so I'm not sure if it actually would make any sense to test that.

I tried

```
       const Comp = styled(View)`
        width: 10vwf;
        padding: 10vwqsd 10fdsvw;
        border: 10vw soli black;
        heigsh: 10vw;
      `;
```

But I get :

```
Failed to parse declaration "padding: 10vwqsd 10fdsvw"

      137 |         // support, so we'll just disable multiple values here.
      138 |         // https://github.com/styled-components/css-to-react-native/issues/11
    > 139 |         const styleObject = transformDeclPairs(declPairs, [
```

Which is not my doing...

Sharcoux: > > Sorry, I don't find how to emulate a resize of the screen or an orientation change...
> 
> I don't remember how to do it exactly, but I guess that you would need to mock the functions in the `Dimensions` module, trigger the event listener with updated screen size and then re-render the component. I'm not sure if there is some other way to do that easier.

I don't think that I can manually fire an event on Dimensions listeners. Check [the code](https://github.com/facebook/react-native/blob/bcc62a6d179e673754a46f9d93ab61dbe0576894/Libraries/Utilities/Dimensions.js) . I have no way to access the eventEmitter...

Sharcoux: I'll try something. I have an idea

Sharcoux: Nope. Didn't work. This code:
```
  React.useEffect(() => {
    Dimensions.addEventListener('change', sizeListener)
    return () => Dimensions.removeEventListener('change', sizeListener)
  }, [])
```
is executed only after the tests are over.
I don't see how I can make the tests wait for this to happen...

kristerkari: > I tried

As I said previously, it does not make sense to test invalid CSS, because the CSS parser or React Native will throw an error anyway.

What could make sense would be to test that the properties that can have those vm/vh/vmin/vmax/em units are actually going through the code path that you added, but as I have not been working with the styled-components, I'm not sure if those kind of things are usually covered with tests.

kristerkari: > I don't think that I can manually fire an event on Dimensions listeners.

What if you would just mock the whole `Dimensions` API? Then you shouldn't have to worry about actually dispatching events.

...and looking at the test renderer docs, it seems that you can mount a component and the just update it after: https://reactjs.org/docs/test-renderer.html

Sharcoux: That was my idea. This is what I tried:

```javascript
    it('should handle window resizing', () => {
      // Create the component to be tested
      const Comp = styled(View)`
        width: 10vw;
        padding: 10vw 10vw;
        border: 10vw solid black;
      `;

      // Getting current window dimension
      const { width, height } = Dimensions.get('window')
      const vw = width / 10

      // We want to restore the Dimensions object after our test
      const oldDimensions = {...Dimensions}

      // Mock the Dimensions object
      const listeners = []
      Dimensions.addEventListener = (type, listener) => type==='change' && listeners.push(listener)
      Dimensions.removeEventListener = (type, listener) => type==='change' && listeners.splice(listeners.indexOf(listener), 1)
      Dimensions.get = () => ({ width, height })

      // Initial render of the View
      const wrapper = TestRenderer.create(<Comp />);
      const view = wrapper.root.findByType('View');
      expect(view.props.style).toEqual([{
        width: vw,
        paddingTop: vw, paddingBottom: vw, paddingLeft: vw, paddingRight: vw,
        borderColor: 'black', borderStyle: 'solid', borderWidth: vw
      }]);

      // Emulate a resize
      Dimensions.get = () => ({ width: 200, height })
      listeners.forEach(listener => listener())

      // Trying to update the view
      wrapper.update()
      const updatedView = wrapper.root.findByType('View');
      // /!\ The line above generate the error: Can't access .root on unmounted test renderer
      expect(updatedView.props.style).toEqual([{
        width: 20,
        paddingTop: 20, paddingBottom: 20, paddingLeft: 20, paddingRight: 20,
        borderColor: 'black', borderStyle: 'solid', borderWidth: 20
      }]);

      // Restore the old Dimensions object.
      Object.assign(Dimensions, oldDimensions)
    })
```

I get `Can't access .root on unmounted test renderer`.

Am I doing something wrong?

packages/styled-components/src/models/InlineStyle.js
Sharcoux: These are the properties that don't support other units for now:

- background
- background-position
- background-size
- border-image
- border-image-outset
- border-image-slice
- box-shadow
- clip
- outline
- transform
- transform-origin

I have no solution for transform and clip. I think that the rest can be fixed as soon as a maintainer answers me on this PR...

packages/styled-components/src/native/test/native.test.js
Sharcoux: Ok @kristerkari I got it fixed after hours of bumping my head on the wall. react-test-renderer v16.9 let you use await act(async () => {}) to wait for useEffects to happen. In the end I could successfully emulate a resize and confirm that the code is working.

Sharcoux:
@probablyup Please, I'm begging you, can you help me merge this? Each new commit on master, I have to resolve conflicts to keep this PR on track. It's been 3 weeks now. I'm getting desperate... I mean, if anyone had any argument against it I wouldn't mind, but here it works, it is tested, and I'm using it. I can't spend more weeks resolving conflicts...

bsehovac:
Well I think it would be better if we could provide our own function for custom unit. In my last project, I had design for iPhone X, and it required a lot of adjustment for screens smaller than 375pt. So i've made this scaled unit which would scale down everything only on smaller screens:

```
import { Dimensions } from 'react-native'
const { width } = Dimensions.get('window')
const scale = width / 375 > 1 ? 1 : width / 375
const scaledUnit = size => Number(size) * scale
export default scaledUnit
```

So maybe it would be better if we could provide custom units to ThemeProvider:

```
<ThemeProvider units={{ su: scaledUnit }}>
```

And use that in [src/models/InlineStyle.js](https://github.com/styled-components/styled-components/blob/4add697ac770634300f7775fc880882b5497bdf4/packages/styled-components/src/models/InlineStyle.js#L29), something like this:

```
generateStyleObject(executionContext: Object, customUnits: Object) {
  let flatCSS = flatten(this.rules, executionContext).join('');

  if (isPlainObject(customUnits)) {
    Object.keys(customUnits).forEach(unit => {
      flatCSS = flatCss.replace(new RegExp(`-?\d*[.\d]*${unit}`, 'g'), customUnits[unit])
    })
  }
```

Of course we would generate regexp somewhere in ThemeProvider, this is just an idea...

Sharcoux:
But vw is supposed to be relative to the screen size. If it doesn't look good on small screens, isn't it a problem of design? Maybe we should use media queries for this case?
You could also consider using em to easily apply a ratio on the problematic parts of your app?

Sharcoux:
I think that the goal of react native support of styled component is for the developer to be able to use the same code for all platform. Moreover, not everyone is using the theme provider.

bsehovac:
Than maybe we could add new wrapper like <UnitsProvider>, or something other that could manipulate `flatCSS`for RN. Why would we limit to vw, wh, em, rem, when we could have more freedom this way...

Sharcoux:
I'm not against, I'm just saying that it is a completely new and unrelated feature. Right now, the objective is that a developer that only knows CSS can make a react native app using CSS. I think that this part at least should work out of the box, without requiring to create and parameter a UnitsProvider.

On the other hand, if you want to implement such a system, I think that you should get in touch with the team at css-to-react-native. They were thinking about adding something like that. It could lead to a PR in css-to-react-native, and the creation of a UnitsProvider in styledComponents if the community wants that.

likern:
> I'm not against, I'm just saying that it is a completely new and unrelated feature. Right now, the objective is that a developer that only knows CSS can make a react native app using CSS. I think that this part at least should work out of the box, without requiring to create and parameter a UnitsProvider.
> 
> On the other hand, if you want to implement such a system, I think that you should get in touch with the team at css-to-react-native. They were thinking about adding something like that. It could lead to a PR in css-to-react-native, and the creation of a UnitsProvider in styledComponents if the community wants that.

Agree it's irrelevant. Merging this PR for right now will be more than enough at this stage.
Ideally css should just have features if the web.

bsehovac:
I just think that this PR will give performance downgrade to styled-components in RN, and allowing some kind of extensions would be more beneficial...

Sharcoux:
@probablyup , we have no news from @kitten and the PR got reviewed by @kristerkari . Please, can we move on this?

Sharcoux:
It's been a month 🎉🎉🎉 -_-

packages/styled-components/src/models/StyledNativeComponent.js
```diff
@@ -1,6 +1,6 @@
-// @flow
-import React, { createElement, Component } from 'react';
+import React, { createContext, createElement, Component } from 'react';
 import hoist from 'hoist-non-react-statics';
+import { Dimensions } from 'react-native';
```
macintoshhelper, line 3: I'm pretty sure this will break `styled-components/primitives`, as it depends on this component, and React Primitives doesn't have a `Dimensions` API. `react-sketchapp` now has a `useWindowDimensions` API, but am not sure if this is something that will eventually land in React Primitives.

macintoshhelper:
> I just need to find the best way to listen to orientation/layout changes. What do you think is the best approach?

`useWindowDimensions` is probably the best way to do this. This will require adding hooks to the native/primitives code though which will break backwards compatibility with pre-hooks React Native/Primitives versions.

packages/styled-components/src/models/StyledNativeComponent.js
Sharcoux: I don't understand at all why importing Dimensions from react native would break anything. Isn't there a test for /primitives anyway? It should have fail. Otherwise could you think of a test that would fail to help me figure out what you mean and what is your concern?
I am using react-native and react-native-Web in one of my projects. I will check if there is an issue.

macintoshhelper: React Primitives codebases are likely to not have the React Native package available, some more info here: https://medium.com/styled-components/announcing-primitives-support-for-truly-universal-component-systems-5772c7d14bc7 and https://github.com/lelandrichardson/react-primitives/ .

The React Primitives entrypoint can be used in React Native, React web, React Sketch.app, React 360, etc codebases.

It'll fail in the import because the package 'react-native' will not be found. `react-native-web` codebases generally alias `react-native`, but there is no equivalent usage for other platforms. The test will pass because `react-native` is a dev dependency in the overall project, but won't exist in a React Sketch.app or 360 project.

macintoshhelper: Nothing is being imported directly from the `react-native` package in `StyledNativeComponent.js`. Native components/APIs are passed in as arguments in the entrypoint (i.e. `src/native/index.js` or `src/primitives/index.js`)

If you look at https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/primitives/index.js#L19 , you can see that `reactPrimitives.StyleSheet` is being passed as an argument.  And `reactNative.StyleSheet` is used in https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/native/index.js#L19

What could be done to address this, is add `Dimensions` or `useWindowDimensions` as an argument to the `_StyledNativeComponent` function, and add some feature detection that handles the case of it being undefined for the primitives entrypoint.

I hope this makes sense. Would be ideal I think to use `useWindowDimensions` instead of `Dimensions`, since that handles rotation/layout changes, and may be easier to standardise on different platforms where there are no clear dimensions – could always be changed later on though.

Sharcoux: Ok, I can use useWindowDimensions. It seems that it is supported in react-native-web.
I'm a little concerned about the current architecture of the project. What you describe here makes the project very vulnerable, don't you think?

Sharcoux: How am I supposed to import useWindowDimensions? I guess that I can't use `import { useWindowDimensions } from 'reat-native'` can I ?

macintoshhelper: Yes, you're right. There should be better documentation on this, at least in the form of code comments. Open an issue: #3140

macintoshhelper: > How am I supposed to import useWindowDimensions? I guess that I can't use `import { useWindowDimensions } from 'reat-native'` can I ?

If using `useWindowDimensions` would help simplify your implementation i.e. https://github.com/styled-components/styled-components/blob/422cf8c23b4f3cc7b7b162a50695782e684acf7b/packages/styled-components/src/models/StyledNativeComponent.js#L26-L34 then that would be great, I don't want to create extra work for you though 🙂, this feature could be left in for React Native only and not do anything if Dimensions is undefined. Main thing is just for it to not crash React Primitives projects.

You could pass `Dimensions` or `useWindowDimensions` as an argument here: https://github.com/styled-components/styled-components/blob/422cf8c23b4f3cc7b7b162a50695782e684acf7b/packages/styled-components/src/models/StyledNativeComponent.js#L232 from here https://github.com/styled-components/styled-components/blob/422cf8c23b4f3cc7b7b162a50695782e684acf7b/packages/styled-components/src/native/index.js#L19

Some context/code on having platform specific imports here that might be helpful: https://github.com/styled-components/styled-components/pull/904

mathieudutour: Regardless of react-primitives support, you are defining a dependency to `react-native`, so you need to pass it to the `dependency` (from `dev-dependency`).
This is a major change in that every `styled-components` install would then install `react-native`, even for a pure web app. Something non-negligible.

Sharcoux: No problem for the extra work. I'm just not quite sure about how useWindowDimensions should be passed as an argument, but I'll make a try (during the week) and you tell me if I got it right.

packages/styled-components/package.json
```diff
@@ -89,13 +89,13 @@
     "eslint-plugin-react": "^7.12.4",
     "flow-bin": "^0.100.0",
     "prop-types": "^15.7.2",
-    "react": "^16.8.6",
+    "react": "^16.13.1",
```
macintoshhelper, line 92: Can this not stay at `16.8.6`? I can't see any code that's depending on features past 16.8. 16.8.0 supports hooks

macintoshhelper:
> One last thing: flow generates an error because it thinks that "Dimensions" doesn't exist in react-native (which is wrong). I'm not very friendly with flow. Any idea how I should fix that?

Example of global typing for `react-native` here: https://github.com/macintoshhelper/styled-components/commit/8d89f9735add7954156ee550e37ac30377b0d61c

packages/styled-components/src/models/StyledNativeComponent.js
macintoshhelper: Untested proof-of-concept demo here: https://github.com/macintoshhelper/styled-components/commit/be26f6be789c15113351439e0db5bd325f50c077

Maybe best to wait for a review/feedback from someone else before implementing though, not sure if I'm doing something horribly bad by passing hooks around across components like that 😄.

packages/styled-components/src/models/StyledNativeComponent.js
```diff
@@ -32,43 +125,42 @@ class StyledNativeComponent extends Component<*, *> {
...
             ...props
           } = this.props;
 
-          const { defaultProps, target, shouldForwardProp } = forwardedComponent;
+          const { defaultProps, target } = forwardedComponent;
           const elementToBeRendered =
             this.attrs.$as || this.attrs.as || transientAsProp || renderAs || target;
 
-          const generatedStyles = this.generateAndInjectStyles(
+          const generatedStyles = {...this.generateAndInjectStyles(
```
macintoshhelper, line 135: Is the `{...` required for it to work? In my attempt to get this rendering with React Primitives, this appears to be breaking the render.

Sharcoux: This is required because we will edit the object and overwrite the values. Check below for instance:
`this.remStyles.forEach(key => generatedStyles[key]*=16)`

I think that removing this could generate problematic border effects...

Sharcoux: Can you tell me what the issue is? I don't see why it would break the render... It is the same object in the end... Unless there are some shady operations involved?

packages/styled-components/package.json
Sharcoux: I'll check again. Maybe I was wrong.

Sharcoux: You were right. Well done.

packages/styled-components/src/models/StyledNativeComponent.js
Sharcoux: TypeError: (0 , _reactNative.useWindowDimensions) is not a function
:/
Should I update React-Native deps?

macintoshhelper: Ah, it's only available in `0.61.0`, maybe leave it as is for now.

packages/styled-components/src/models/StyledNativeComponent.js
macintoshhelper: Ah, ok. If it's working fine in React Native, should be fine to leave it as is. I was trying to get it working with `react-sketchapp` (primitives platform) and it was breaking here, but that use-case is probably outside the scope of this PR anyway.

packages/styled-components/src/models/StyledNativeComponent.js
Sharcoux: Well, I'm not very convinced either by this approach... :/

macintoshhelper: Could update the React Native `peerDependency` to `0.61.0` and use feature detection (i.e. `typeof useWindowDimensions !== 'undefined' && withDimensions(Component)`, so for older versions, the `rem`, `em`, `vh`, etc will not work, but everything else will. Probably best to wait for a review by a maintainer though, I'm not really sure what's best in terms of dependencies and feature detection.

Sharcoux:
Ok guys, I gave up this repo and developped my own : https://github.com/Sharcoux/rn-css

I think that the current approach of styled-components for native is doomed to fail at some points or to become more and more complicated everyday. I successfully reproduced css-to-react-native and styled-component API with some improvements, but as I had only 2 days to spend on that there is definitely room for improvement.

I still need to finish some bugs, and I didn't try it in real situations, but that should come soon.

Sharcoux:
This PR should probably not be merged as it is now. I noticed performance impacts with the current implementation.

Sharcoux:
Fixing the performances issues would need refactoring the project. This was done in [rn-css](https://github.com/Sharcoux/rn-css). The current PR can be discarded as I won't be able to replicate the fix here.

Issue #3097, opened 2020-04-08
`createGlobalStyle` generate the same id and global style is missing.


ykan:
# Environment
## System:
 - OS: macOS 10.15.3
 - CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
 - Memory: 56.37 MB / 32.00 GB
 - Shell: 5.7.1 - /bin/zsh
## Binaries:
 - Node: 10.19.0 - /usr/local/bin/node
 - Yarn: 1.22.0 - ~/.yarn/bin/yarn
 - npm: 6.13.4 - /usr/local/bin/npm
## npmPackages:
 - babel-plugin-styled-components: ^1.10.6 => 1.10.6
 - styled-components: ^4.3.2 => 4.4.1

## Reproduction
https://codesandbox.io/s/styled-components-base-9w9x5

```tsx

const Style1 = createGlobalStyle`
${props => css`
  .test {
    color: #999;
  }
`}
`;

const Style2 = createGlobalStyle`
${props => css`
  .test {
    background: #000;
  }
`}
`;

const App = () => (
  <div>
    <Style2 />
    <Style1 />
    <div className="test">test</div>
  </div>
);
```
You will see only Style1 is rendered, Style2 is missing.

I found in `styled-components`, the `createGlobalStyle`, the `id` of GlobalStyle is generated by `JSON.stringify(rules)`, if you pass functions, you will see `rules` like this `['\n', f..., '\n']`, and `JSON.stringify(rules)` will return the same `["\n  ",null,"\n"]`.

## Expected Behavior
different rules generate different id.

## Actual Behavior
different rules generate the same id.

## Solution

```tsx

const Style1 = createGlobalStyle`
// add id of this global style using comment, id * global style name *
${props => css`
  .test {
    color: #999;
  }
`}
`;
```

xxn520:
hulp!

littlecold233:
same problem

natterstefan:
Short excursion: Do you also experience this issue https://github.com/styled-components/styled-components/issues/3076? I just wanna know if other developers see this too. Thanks :)

ykan:
> Short excursion: Do you also experience this issue #3076? I just wanna know if other developers see this too. Thanks :)

We have not upgraded to v5 yet.

PR #3101, opened 2020-04-09
fix: fix createGlobalStyle generate same id when rules includes function

ykan:
fix https://github.com/styled-components/styled-components/issues/3097

### DIFF
--- a/packages/styled-components/src/constructors/createGlobalStyle.js
+++ b/packages/styled-components/src/constructors/createGlobalStyle.js
@@ -5,6 +5,7 @@ import GlobalStyle from '../models/GlobalStyle';
 import { useStyleSheet, useStylis } from '../models/StyleSheetManager';
 import { checkDynamicCreation } from '../utils/checkDynamicCreation';
 import determineTheme from '../utils/determineTheme';
+import stringifyRules from '../utils/stringifyRules';
 import { ThemeContext } from '../models/ThemeProvider';
 import { EMPTY_ARRAY } from '../utils/empties';
 import generateComponentId from '../utils/generateComponentId';
@@ -19,7 +20,7 @@ export default function createGlobalStyle(
   ...interpolations: Array<Interpolation>
 ) {
   const rules = css(strings, ...interpolations);
-  const styledComponentId = `sc-global-${generateComponentId(JSON.stringify(rules))}`;
+  const styledComponentId = `sc-global-${generateComponentId(stringifyRules(rules))}`;
   const globalStyle = new GlobalStyle(rules, styledComponentId);
 
   if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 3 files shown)


packages/styled-components/src/constructors/createGlobalStyle.js
```diff
@@ -19,7 +20,7 @@ export default function createGlobalStyle(
   ...interpolations: Array<Interpolation>
 ) {
   const rules = css(strings, ...interpolations);
-  const styledComponentId = `sc-global-${generateComponentId(JSON.stringify(rules))}`;
+  const styledComponentId = `sc-global-${generateComponentId(stringifyRules(rules))}`;
```
probablyup, line 23: There shouldn't be any functions in `rules` because of the `css` utility that's used to flatten the styles: https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/constructors/css.js

packages/styled-components/src/utils/stringifyRules.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+import isFunction from './isFunction';
+
+export default function stringifyRules(rules: Array<any>): string {
+  const newRules = rules.map(rule => {
+    if (isFunction(rule)) {
```
probablyup, line 6: This unfortunately is not SSR-safe. Code can differ between server and client depending on what babel flags are enabled etc.

ykan: Do you mean `isFunction` or `rule.toString`?  It just use to generate id, or just try to add `displayName` for `GlobalStyleComponent` to generate different id ?

probablyup: > Do you mean isFunction or rule.toString?

`rule.toString()`, there can be subtle differences

> maybe we can try to add displayName for GlobalStyleComponent to generate different id like withConfig in StyledComponent?

That's possible but requires a bunch of architectural changes because cGS doesn't have the same construction pattern as `styled()`.

ykan: Or，just try to generate different id for different function: 

```
const fnIDMap = new WeakMap<Function, number>();
let fnCount = 1;

function getId(fn: Function) {
  let fnId = fnIDMap.get(fn);
  if (!fnId) {
    fnId = fnCount ++;
    fnIDMap.set(fn, fnId);
  }

  return `fnId-${fnId}`;
}

export default function stringifyRules(rules: Array<any>): string {
  const newRules = rules.map(rule => {
    if (isFunction(rule)) {
      return getId(rule);
    }
    return rule;
  });
  return JSON.stringify(newRules);
}
```

ykan: OK，I've got it. In our project, we try another way, just add id in comment:

```
const Style1 = createGlobalStyle`
// Id of this global style, don't remove
// Id `my global style name`
${props => css`
  .test {
    color: #999;
  }
`}
`;
```

Issue #3076, opened 2020-03-27
createGlobalStyle adds stylings to the <head> twice in 5.0.1


natterstefan:
## Environment

- `react@16.12.0`
- `react-dom@16.12.0`
- `react-scripts@3.0.1`
- `styled-components@5.0.1`

⬇️ see codesandbox example.

## Reproduction

- [codesandbox - buggy w/ styled-components@5.0.1](https://codesandbox.io/s/epic-firefly-gzdfx)
- [codesandbox - working w/ styled-components@4.4.1](https://codesandbox.io/s/icy-sun-s88db)

## Steps to reproduce

Open the provided [codesandbox](https://codesandbox.io/s/epic-firefly-gzdfx) and inspect the styling of either of the headlines. The `h1` (or `h2`) styling is applied twice because it was added twice to the document.

If you downgrade styled-components ([see codesandbox](https://codesandbox.io/s/icy-sun-s88db)) to the latest 4.x.x version and inspect the elements again, everything is fine.

## Expected Behavior

Global style is added only once.

## Actual Behavior

Global style is added twice to the document (also in a SSR scenario, which happens in one of our apps right now).

### Bug with version 5.0.1

![image](https://user-images.githubusercontent.com/1043668/77751757-95a66680-7026-11ea-8fe6-9184a7222fe3.png)

### Working w/ version 4.4.1

![image](https://user-images.githubusercontent.com/1043668/77751779-9fc86500-7026-11ea-90ff-812594a44392.png)

bfields-cpi:
Seeing the same doubling of styles created with `createGlobalStyle` on v5.1.0.
Gentle bump ;)

EduardAkhmetshin:
Global styles are rendered twice because of React.StrictMode, so you won't have this issue in production mode. I provided a pull request to solve this.

dilipp02:
Issue still exists and its annoying

lakaniemi:
This is indeed caused by `React.StrictMode`, with more details described by @EduardAkhmetshin in his PR #3121. The issue is quite annoying, but I managed to work around it by moving `ThemeProvider` and the global CSS component outside of `React.StrictMode`. Not the most ideal solution, but I guess the biggest benefit of `React.StrictMode` is finding faults in your own code, and thus it shouldn't matter that it's deeper in the component tree after all the providers :)

Silon:
@natterstefan I solve this problem by moving GlobalStyle outside React.StrictMode:

```
[...]

PR #3121, opened 2020-04-25
Render global styles with useEffect hook

EduardAkhmetshin:
Fixes [#3076](https://github.com/styled-components/styled-components/issues/3076)

Rendering of styles is a side-effect and should not be performed before mounting.
React.StrictMode helps to enforce this rule by running various functions (lifecycle methods or the function itself in case of function components) twice to make sure that they are idempotent. GlobalStyleComponent was not idempotent, calling Sheet#allocateGSInstance on render, which led to generating a different id on the second render and absence of corresponding unmount callback for the first one.

### DIFF
--- a/packages/styled-components/src/constructors/createGlobalStyle.js
+++ b/packages/styled-components/src/constructors/createGlobalStyle.js
@@ -1,5 +1,5 @@
 // @flow
-import React, { useContext, useEffect, useRef } from 'react';
+import React, { useContext, useEffect, useLayoutEffect, useRef } from 'react';
 import { STATIC_EXECUTION_CONTEXT } from '../constants';
 import GlobalStyle from '../models/GlobalStyle';
 import { useStyleSheet, useStylis } from '../models/StyleSheetManager';
@@ -13,6 +13,8 @@ import css from './css';
 import type { Interpolation } from '../types';
 
 type GlobalStyleComponentPropsType = Object;
+
+declare var __SERVER__: boolean;
 
 export default function createGlobalStyle(
   strings: Array<string>,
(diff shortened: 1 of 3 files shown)


EduardAkhmetshin:
I need help with fixing the failed tests. They pass on my local machine though.

probablyup:
I tried doing this originally and we ultimately moved away from it because of issues. I don't remember what they were though.

kitten:
I believe we had an issue with SSR, and we also need to avoid issues with concurrent mode.

One thing that needs to happen is that during SSR the insertion must be synchronous. Luckily we won't have to handle unmounts in that case.

In the browser timing is important. We don't want to let a React render without stylings in place, so we must use `useLayoutEffect` if we want an effect to insert styles with an element.

Now comes the tricky part: Concurrent Mode and Strict Mode. I'll skip over the details but the important thing is that an effect may never run, which is fine, but is becoming a problem if we're doing non-symmetrical effects, e.g. synchronous insertion like we're doing now but removal in an effect's teardown.

So to solve this we need to insert synchronously for SSR and otherwise use `useLayoutEffect` to solve this comprehensively

EduardAkhmetshin:
@kitten so that's essentially what I tried to do. I agree that `useLayoutEffect` is more suitable than `useEffect` that I used. I might try again, using a library to detect the environment this time. Something like `exenv`. It will also facilitate testing.
Also, what do you mean by issues with concurrent mode? What should I be aware of?

EduardAkhmetshin:
I rewrote the code - it uses `useLayoutEffect` now. Moreover, it turned out that the project already had means of detecting the environment, so I didn't need any libraries.

kitten:
@probablyup Do we want to pick this back up and get this ready or just defer it for a rewrite? I have some more nits, since this has some other problems, but depends on whether it's worth getting this ready

kitten:
Obsolete / Resolved

Issue #3149, opened 2020-05-20
Feature Request: Ability to maintain stable style prop references in React Native


philipbulley:
## Problem
Currently, in `styled-components/native` the `style` prop passed into a styled custom component is not stable — a new object reference is passed in on every render, regardless of whether any of the styling has changed.

#### Example
https://snack.expo.io/@pbcodes/7e9d69
Please run this snack using the Android or iOS devices. The issue doesn't apply to React Native Web.

## Motivation
Stable object references are a fundamental part of React and its performant application. We should make provisions for this wherever possible.

Without this ability, we cannot make optimisations using `React.memo` in an idiomatic way.

## Prior Art
This problem was raised in #3020 by @Federkun but was ultimately closed with only a workaround suggested.

@kitten correctly [pointed out](https://github.com/styled-components/styled-components/pull/3020#issuecomment-585487853) that `styled(CustomComponent)` should be used sparingly, however, this doesn't mean that this usage is wrong. The forwarding of the `style` prop can often be a useful pattern in adding layout properties such as `margin` from the outside of a custom component without having to create wrapper containers.

In lieu of stable object references, it was also suggested that `React.memo` could be used in conjunction with a comparator function, such as `isEqual` from `react-fast-compare`. However, there is a separation of concerns downside to this as any custom component forwarding the `style` prop and also exposed via `React.memo` must assume that there is a possibility that `styled-components` may be used and thus preemptively employ a deep equality check. If the custom component is part of a shared component library, it is unreasonable for the component to expect to be styled via `styled-components`. In cases where `StyleSheet` is used, there will be an unnecessary baked-in performance hit to the custom component via its usage of `isEqual`.

## Conslusion
I think it's reasonable to ask that `styled-components` itself should provide support for the fundamental ability to work with stable object references on an opt-in basis.

philipbulley:
## Proposal
Instead of forcing all components to take on the overhead of memoization, can we export an additional form of `styled` as:

```js
import { styledMemo } from 'styled-components/native';
import CustomComponent from './CustomComponent';

const CustomComponentStyled = styled(CustomComponent)`
...
`;
```

I've created a WIP PR #3150 exploring a way that this might be achieved.

probablyup:
I think we'll look into this after https://github.com/styled-components/styled-components/pull/3152 is merged

philipbulley:
[...]

PR #3150, opened 2020-05-20
Native supports a stable style prop reference with React.memo

philipbulley:
This is a follow up to the closed #3020 (cc: @Federkun, @kitten) and I have raised the following new issue which provides a rationale for the proposed changes: #3149

In #3020, the consensus was that memoization of the resulting style prop would be too expensive to run on every instance when only a subset of use cases would benefit.

**Update:**
This PR initially explored this approach:
```js
import { styledMemo } from 'styled-components/native';
```

But then after feedback, that was reverted and it now only employs a memoize optimisation for `React.memo` components.

#### Caveats
* I haven't used flowtype before, so might need a hand with anything flow specific

### DIFF
--- a/packages/styled-components/package.json
+++ b/packages/styled-components/package.json
@@ -69,6 +69,8 @@
     "babel-plugin-styled-components": ">= 1",
     "css-to-react-native": "^3.0.0",
     "hoist-non-react-statics": "^3.0.0",
+    "memoize-one": "^5.1.1",
+    "react-fast-compare": "^3.1.1",
     "shallowequal": "^1.1.0",
     "supports-color": "^5.5.0"
   },
--- a/packages/styled-components/src/models/StyledNativeComponent.js
+++ b/packages/styled-components/src/models/StyledNativeComponent.js
@@ -1,6 +1,8 @@
 // @flow
 import React, { createElement, Component } from 'react';
 import hoist from 'hoist-non-react-statics';
+import memoizeOne from 'memoize-one';
+import isEqual from 'react-fast-compare';
 import merge from '../utils/mixinDeep';
 import determineTheme from '../utils/determineTheme';
 import { EMPTY_ARRAY, EMPTY_OBJECT } from '../utils/empties';
@@ -35,6 +37,7 @@ class StyledNativeComponent extends Component<*, *> {
             forwardedAs,
             forwardedRef,
             style = [],
+            __scIsMemo,
             ...props
           } = this.props;
 
@@ -63,7 +66,7 @@ class StyledNativeComponent extends Component<*, *> {
             }
           }
 
-          propsForElement.style = [generatedStyles].concat(style);
+          propsForElement.style = __scIsMemo ? this.getStyles(generatedStyles, style) : this.getMemoStyles(generatedStyles, style);
 
           if (forwardedRef) propsForElement.ref = forwardedRef;
           if (forwardedAs) propsForElement.as = forwardedAs;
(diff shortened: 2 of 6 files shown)


Federkun:
Hi @philipbulley, thanks for this PR!
My opinion is that a `styledMemo`, while allowing to support this use case, may still not be the best abstraction? It may leak to the client code the concept that a memoized component should be handled in a different way than normal. Usually using any kind of "decorator" should be transparent to the client code. If we have a `styledMemo`, removing the memoization and/or add it back in may require to change `styledMemo` with `styled` and vice versa as well.

To be honest, I'd like much more to relaying on the implementation details of React.memo, to detect if the component is memoized or not ( something along the line of `ReactComponent.compare !== undefined` ) and act accordingly. While still not ideal, at least it shield all this from the client

philipbulley:
@Federkun that's a good point and would be ideal, thanks!

`ReactComponent.compare !== undefined` is definitely a little obscure, especially as it relies on the fact that `compare === null` when no compare fn is passed.

An alternative (also not ideal) is to use introspection on the Element Type `elementToBeRendered`. 

<details>
<summary>Expand for commentary on introspection woes...</summary>

1️⃣ This would be great, except that it doesn't work, as `ReactIs.isMemo` expects to be passed the return value of `React.createElement`:

```js
ReactIs.typeOf(CustomComponent) === ReactIs.Memo
ReactIs.isMemo(CustomComponent) // sugar coated version of above
```

Several people have proposed a solution that can introspect an element type
* facebook/react/pull/12932
* facebook/react/pull/15349

but neither of these have landed.

2️⃣ This works 🤮 :
```js
ReactIs.isMemo(React.createElement(CustomComponent)) // true
```
But I don't imagine it would be at all healthy for perf, especially as this would run for every custom component. Let's not go there!

3️⃣ Another option — if we don't mind looking at implementation details — is this approach: facebook/react/issues/12882#issuecomment-440227651
I'd argue that checking `$$typeof` is more or less as risky as `ReactComponent.compare !== undefined`, but is a little more explicit as to its intention.

See the usage of this technique in MobX:
https://github.com/mobxjs/mobx-react/blob/376bff72d84ae8d342c55d1fb25ff81a223624d2/src/observer.tsx#L14-L16
</details>

I've implemented the technique that MobX uses and it works a treat! I've added tests too. Please see the latest commit.

⚠️ Finally, the build will fail as I haven't sorted out the types for `memoize-one` — I don't know much about the FlowType ecosystem, what's the easiest acceptable course of action here? Write a custom definition?

philipbulley:
Can anyone who better understands Flow help with this issue please?

Why does this happen?

```
$ flow check
Error ┈┈┈┈┈ packages/styled-components/src/models/StyledNativeComponent.js:4:24

Cannot resolve module memoize-one.

     1│ // @flow
     2│ import React, { createElement, Component } from 'react';
     3│ import hoist from 'hoist-non-react-statics';
     4│ import memoizeOne from 'memoize-one';
     5│ import isEqual from 'react-fast-compare';
     6│ import merge from '../utils/mixinDeep';
     7│ import determineTheme from '../utils/determineTheme';
```

Despite memoize-one [having flow types](https://github.com/alexreardon/memoize-one/blob/master/src/memoize-one.js.flow).

kitten:
Closing due to inacitivity. I think we could consider _changing_ the approach of dynamic styling in the implementation if we do split off the package to avoid complexity / differences in the CSS pipeline to the main package, see #1617

PR #3151, opened 2020-05-21
Template for Pull Request

N-Verma:
Added **`pull_request_template.md`** in `.github` . Template for pull request which will be used when a pull request in made

### Type of change
- [x] Added Template

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,15 @@
+# Pull Request Template
+
+### Description
+ Please include required summary for the changes made and which issue has been fixed. Also mention any dependencies required.
+ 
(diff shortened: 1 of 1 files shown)


kitten:
Sorry, I understand the motivation of contribution here, but pull request templates are meant to be to make reviews by maintainers / core contributors easier, hence they're created by core contributors. This just isn't what we'd want from a template, if we need one. As it stands we also don't get a lot of discussions that lead to PRs. We either "just get a PR" without notice and that often doesn't lead to a merge, or leads to some back and forth, or we get a PR that has been discussed prior, where the author and reviewers are already aware of what's happening.

PR #3152, opened 2020-05-22
Update StyledNativeComponent to hooks

janicduplessis:
This tries to follow the same structure as `StyledComponent.js` while making the required changes for react-native.

Differences from `StyledComponent.js`

- Use `InlineStyle` instead of `ComponentStyle`
- Remove special `className` handling
- Remove too many classes warning

Differences from class implementation:

- Removed `setNativeProps` warning. This is actually tricky to do with hooks and I didn't see much value added from this warning.

### Test plan

- Tests / flow pass
- Tested the a large RN app using styled components still renders as before
- Tested that styled components render nicely in React DevTools (No more context provider 🎉 )

### DIFF
--- a/packages/styled-components/src/models/StyledNativeComponent.js
+++ b/packages/styled-components/src/models/StyledNativeComponent.js
@@ -15,7 +15,6 @@ import generateDisplayName from '../utils/generateDisplayName';
 import getComponentName from '../utils/getComponentName';
 import isFunction from '../utils/isFunction';
 import isStyledComponent from '../utils/isStyledComponent';
-import isTag from '../utils/isTag';
 import merge from '../utils/mixinDeep';
 import { ThemeContext } from './ThemeProvider';
 
@@ -103,9 +102,7 @@ function useStyledComponentImpl<Config: {}, Instance>(
 
   const elementToBeCreated: Target = attrs.$as || props.$as || attrs.as || props.as || target;
 
-  const isTargetTag = isTag(elementToBeCreated);
   const computedProps = attrs !== props ? { ...props, ...attrs } : props;
-  const propFilterFn = shouldForwardProp || (isTargetTag && validAttr);
   const propsForElement = {};
 
   // eslint-disable-next-line guard-for-in
@@ -113,8 +110,7 @@ function useStyledComponentImpl<Config: {}, Instance>(
     if (key[0] === '$' || key === 'as') continue;
     else if (key === 'forwardedAs') {
       propsForElement.as = computedProps[key];
-    } else if (!propFilterFn || propFilterFn(key, validAttr)) {
-      // Don't pass through non HTML tags through to HTML elements
+    } else if (!shouldForwardProp || shouldForwardProp(key, validAttr)) {
       propsForElement[key] = computedProps[key];
     }
   }
@@ -129,7 +125,6 @@ function useStyledComponentImpl<Config: {}, Instance>(
 export default (InlineStyle: Function) => {
   const createStyledNativeComponent = (target: Target, options: Object, rules: RuleSet) => {
     const isTargetStyledComp = isStyledComponent(target);
-    const isCompositeComponent = !isTag(target);
 
     const {
       displayName = generateDisplayName(target),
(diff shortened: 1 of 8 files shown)


packages/styled-components/src/models/StyledNativeComponent.js
```diff
@@ -1,152 +1,148 @@
...
-
-    // $FlowFixMe
-    const WrappedStyledNativeComponent = React.forwardRef((props, ref) => (
-      <ParentComponent
-        {...props}
-        forwardedComponent={WrappedStyledNativeComponent}
-        forwardedRef={ref}
-      />
-    ));
+    const styledComponentId =
```
janicduplessis, line 140: I’m not sure if this is needed for native, it seems to be only used by useDebugValue.

probablyup:
I'll review this soon, thanks!

packages/styled-components/src/models/StyledNativeComponent.js
```diff
@@ -1,152 +1,148 @@
...
-    } else if (process.env.NODE_ENV !== 'production') {
-      // eslint-disable-next-line no-console
-      console.warn(
-        'setNativeProps was called on a Styled Component wrapping a stateless functional component.'
-      );
+  const refToForward = forwardedRef;
+
+  const elementToBeCreated: Target = attrs.$as || props.$as || attrs.as || props.as || target;
+
+  const isTargetTag = isTag(elementToBeCreated);
```
kitten, line 106: I know we had this in the codebase before, but I believe that was and is a mistake, since string tags aren't a thing in React Native, so we can assume `isTargetTag === false`

packages/styled-components/src/models/StyledNativeComponent.js
```diff
@@ -1,152 +1,148 @@
...
+
+  propsForElement.ref = refToForward;
+
+  return createElement(elementToBeCreated, propsForElement);
 }
 
 export default (InlineStyle: Function) => {
   const createStyledNativeComponent = (target: Target, options: Object, rules: RuleSet) => {
+    const isTargetStyledComp = isStyledComponent(target);
+    const isCompositeComponent = !isTag(target);
```
kitten, line 132: Same here, we can assume `isCompositeComponent === true`

packages/styled-components/src/models/StyledNativeComponent.js
```diff
@@ -173,42 +169,58 @@ export default (InlineStyle: Function) => {
...
 
-    // $FlowFixMe
-    WrappedStyledNativeComponent.styledComponentId = 'StyledNativeComponent';
-    // $FlowFixMe
-    WrappedStyledNativeComponent.target = isTargetStyledComp
+    // this static is used to preserve the cascade of static classes for component selector
+    // purposes; this is especially important with usage of the css prop
+    WrappedStyledComponent.foldedComponentIds = isTargetStyledComp
+      ? // $FlowFixMe
+        Array.prototype.concat(target.foldedComponentIds, target.styledComponentId)
```
kitten, line 194: I'm not sure why this is using `Array.prototype.concat` since `target.foldedComponentIds` should be known to be an array, so this can just become a `concat` method call.

packages/styled-components/src/models/StyledNativeComponent.js
probablyup: Yeah this isn't needed, we don't support component selectors in native because of the lack of cascade

probablyup:
Marking this for v6 since I think it's a breaking change (requires newer react version which didn't affect RN for the v5 release.) We will likely release v6 this Fall.

janicduplessis:
Removed foldedComponentIds and styledComponentId as those are only use for the web implementation to generate class names [here](https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/models/StyledComponent.js#L169).

janicduplessis:
Actually `styledComponentId` is needed for folding to work, I'll leave it as is.

packages/styled-components/src/models/StyledNativeComponent.js
```diff
@@ -219,20 +206,22 @@ export default (InlineStyle: Function) => {
...
-        displayName: true,
-        shouldForwardProp: true,
-        inlineStyle: true,
-        styledComponentId: true,
-        target: true,
-        withComponent: true,
-      });
-    }
+    // $FlowFixMe
+    WrappedStyledComponent.toString = () => `.${WrappedStyledComponent.styledComponentId}`;
```
kitten, line 210: I believe this is obsolete however

probablyup: Yes, native doesn't have component selectors

janicduplessis:
Removed toString and ported over improvements to flow types that landed in master (4ad5f74ec44d8041291599dcfb2bf7b1bc325998)

Issue #2973, opened 2020-01-16
StylesheetManager does not render all styles in iframes


dcastil:
## Environment

`npx envinfo --system --binaries --npmPackages styled-components,babel-plugin-styled-components --markdown --clipboard`

### System:

 - OS: macOS 10.15.2
 - CPU: (8) x64 Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
 - Memory: 385.48 MB / 16.00 GB
 - Shell: 5.7.1 - /bin/zsh

### Binaries:

 - Node: 12.14.1 - /usr/local/bin/node
 - Yarn: 1.21.1 - /usr/local/bin/yarn
 - npm: 6.13.6 - /usr/local/bin/npm

## Reproduction

`styled-components v4.3.2`: https://primer.style/doctocat/components/do-dont
`styled-components v5.0.0`: https://doctocat-ajl99nptm.now.sh/doctocat/components/do-dont

Changes made between both deployments: https://github.com/primer/doctocat/pull/108
Source code where `StylesheetManager` is used: https://github.com/dcastil/doctocat/blob/styled-components-v5/theme/src/components/frame.js#L17

## Steps to reproduce

1. Clone https://github.com/dcastil/doctocat
2. Switch to branch `styled-components-v5`
3. Run `yarn && yarn start` in repo root
4. Navigate to `http://localhost:8000/components/do-dont` in browser

## Expected Behavior

CSS is rendered into head of each iframe.

## Actual Behavior

Some CSS is rendered and some not. I couldn't find a pattern in which CSS is rendered and which isn't unfortunately. Also if I delete the first iframe in https://github.com/dcastil/doctocat/blob/styled-components-v5/docs/content/components/do-dont.mdx, other CSS is being rendered in the second one. So the `StylesheetManager` instances seem to interfere with each other.

kitten:
@probablyup looks like we’re not iFrame safe anymore, since even if a new target is passed the names record is shared and hence a rule group can only be inserted once

- https://github.com/styled-components/styled-components/blob/63b083d43ab577a2203c4975702b0fd064d55fd6/packages/styled-components/src/sheet/Sheet.js#L60

probablyup:
On it

probablyup:
[...]

PR #3159, opened 2020-05-30
Pass empty `names` option to Sheet when using StyleSheetManager with custom target

eramdam:
Fixes https://github.com/styled-components/styled-components/issues/2973

When looking at the thread for the above issue, I noticed that after undoing the changes from https://github.com/styled-components/styled-components/pull/2937, the issue was fixed.

It seems obvious that the `names` map change was made for a reason and I didn't want to break SSR so I settled with that solution which looks "fine" 😅

But maybe there's a better approach, if so I'd be happy to oblige 🙇‍♂️ 

I also updated the test for StyleSheetManager in order to mount `Title` both in the parent document _and_ in the iframes, as to trigger the bug in a test environment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 ## Unreleased
 
+### Bugfixes
+
+- Make sure `StyleSheetManager` renders all styles in iframe / child windows (see [#3159](https://github.com/styled-components/styled-components/pull/3159)) thanks @eramdam!
+
 ## [v5.1.1] - 2020-04-07
 
 ### New Functionality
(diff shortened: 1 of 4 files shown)


probablyup:
@eramdam changes look ok, can you add a regression test?

eramdam:
> @eramdam changes look ok, can you add a regression test?

That's what I did by adding `<Title>` in the test in `packages/styled-components/src/models/test/StyleSheetManager.test.js` test file. Should I create a new test entirely?

eramdam:
> Yes, adding that one line isn't obvious for what it's testing.

Fair enough! 😄 Will do!

eramdam:
And done! https://github.com/styled-components/styled-components/pull/3159/commits/8ceffa487e73e61d285214ba4a064f1bf089bee7

eramdam:
> Looks good, thanks!
> 
> Could you also add a changelog entry?

[Done](https://github.com/styled-components/styled-components/pull/3159/commits/f32afd243276373fe898a1a675d0b78651b82a35)! Wasn't completely sure if there was a specific format to follow so I used something similar to the already existing entries.

macrozone:
any chance that this can be released?

PR #3194, opened 2020-07-08
feat: add warning for repeat global style id

ykan:
According to https://github.com/styled-components/styled-components/issues/3097 ，add warning for repeat global style id

### DIFF
--- a/packages/styled-components/src/constructors/createGlobalStyle.js
+++ b/packages/styled-components/src/constructors/createGlobalStyle.js
@@ -14,6 +14,8 @@ import type { Interpolation } from '../types';
 
 type GlobalStyleComponentPropsType = Object;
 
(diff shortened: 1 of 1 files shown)


packages/styled-components/src/constructors/createGlobalStyle.js
```diff
@@ -14,6 +14,8 @@ import type { Interpolation } from '../types';
 
 type GlobalStyleComponentPropsType = Object;
 
+const seen = new Set();
```
probablyup, line 17: I think this will warn excessively for SSR, which is undesirable. It's also possible and OK to mount 2 instances of the same global style which would trigger this warning.

kitten:
Closing due to inactivity

PR #3215, opened 2020-07-30
Add coverage folders to .gitignore

janicduplessis:
When running tests locally these folders are created and should be ignored.


probablyup:
Thanks, I ended up doing this in a PR I merged earlier

Issue #3186, opened 2020-06-30
[Feature] Prefix hash class


gcoombe:
I'm hoping to add a prefix to the generated hash classes. Would this be something that you would consider accepting as a PR? 

Our use case is that we integrate with a tool called pendo (https://www.pendo.io/) for user analytics. They have tools to allow non-developers to tag parts of our app. These tools will auto-generate a css selectors to target the feature. For obvious reasons we wouldn't want them to use styled component generated classes in the css selector so need to be able to provide known patterns of classes that are auto generated ex: sc-*.

A potential solution: add an environment variable `process.env.SC_HASH_PREFIX` which we could prepend to the hash classname:

Ex:
```
// Add process.env.SC_HASH_PREFIX = 'prefix-'

// Results in
<div class="sc-xyz prefix-mnb">
```

TianlunXiong:
+1, i would like to have a similar method to replace the default prefix 'sc' too, for example: 
```javascript
import styled from 'styled-components'
styled.config.SC_HASH_PREFIX = 'prefix' // => <div class="prefix-xyz">
```

tu4mo:
This would be a very welcome feature. My use-case is a component library built with styled-components. Sometimes the library is used in multiple React apps in one page and we are getting classname collisions/conflicts. Having the ability to prefix the class name would solve the issue. Any thoughts or workarounds, @kitten?

probablyup:
There's a feature in the Babel plugin for this purpose: https://styled-components.com/docs/tooling#namespace

probablyup:
> Would be better if sc just produced unique classes automatically.

We do, but there are tradeoffs between different methods of hashing and we chose ours primarily balancing for performance.

natearmagost:
@probablyup The namespace seems to change the first class... usually the `sc-` prefix.

```
styledComponents: {
  "namespace": "MYPREFIX"
}
```
Results in:
```
<div class="App__Thing-MYPREFIX__cobd8z-0 fUSGm">
```
We are having issues with the 2nd class `fUSGm`. Is there any way to prefix that class like @gcoombe's example:
`<div class="sc-xyz prefix-mnb">`?

mattleonowicz:
@probablyup Please reopen. I think it's a bug that `sc_xxx` get's prefixed instead of the other class (more random one) `sc-` doesn't really contain any styles so it's name is not that crucial as the other one.

seahorsepip:
@mxstbr 
[...]

PR #3225, opened 2020-08-14
Add SC_CLASS_PREFIX which can be configured during bundle time to avo…

justnewbee:
Add `SC_CLASS_PREFIX` which can be configured during bundle time to avoid class name collision issue occurred when multiple styled-components co-exist in one page.

### DIFF
--- a/packages/styled-components/src/constants.js
+++ b/packages/styled-components/src/constants.js
@@ -7,6 +7,12 @@ export const SC_ATTR =
   (typeof process !== 'undefined' && (process.env.REACT_APP_SC_ATTR || process.env.SC_ATTR)) ||
   'data-styled';
 
+// Allowing user to add class prefix to all generated class names to avoid class name collision.
+// This is very handy when it comes to situation where different styled-components might co-exist
+// in different JS bundles.
+export const SC_CLASS_PREFIX =
+  (typeof process !== 'undefined' && process.env.SC_CLASS_PREFIX) || '';
+
 export const SC_ATTR_ACTIVE = 'active';
 export const SC_ATTR_VERSION = 'data-styled-version';
 export const SC_VERSION = __VERSION__;
--- a/packages/styled-components/src/utils/generateAlphabeticName.js
+++ b/packages/styled-components/src/utils/generateAlphabeticName.js
@@ -1,6 +1,8 @@
 // @flow
 /* eslint-disable no-bitwise */
 
+import { SC_CLASS_PREFIX } from '../constants';
+
 const AD_REPLACER_R = /(a)(d)/gi;
 
 /* This is the "capacity" of our alphabet i.e. 2x26 for all letters plus their capitalised
@@ -21,5 +23,5 @@ export default function generateAlphabeticName(code: number): string {
     name = getAlphabeticChar(x % charsLength) + name;
   }
 
-  return (getAlphabeticChar(x % charsLength) + name).replace(AD_REPLACER_R, '$1-$2');
+  return SC_CLASS_PREFIX + (getAlphabeticChar(x % charsLength) + name).replace(AD_REPLACER_R, '$1-$2');
 }


justnewbee:
in our case, we have a topbar bundle which is loaded by many different web apps.
both topbar bundle and app bundle use styled-components, versions might be same or not.
there are bugs reported on styles for styled-components in the two bundles generate same class names.
that is why it is important for at least one of them to specify a class name prefix.

for example, the class name used to be like this `sc-AxiKw dsoEjo`, after added webpack plugin like below

```
new webpack.DefinePlugin({
  'process.env.SC_CLASS_PREFIX': JSON.stringify('XX-')
})
```

the class generated will be like `sc-XX-AxiKw XX-dsoEjo`.

class name collision? solved! 🎉🎉🎉

probablyup:
This is handled by https://styled-components.com/docs/tooling#namespace, thanks!

justnewbee:
> This is handled by https://styled-components.com/docs/tooling#namespace, thanks!

wow, i didn't know... that is cool and i will give it a try

justnewbee:
sorry to say... the namespace does NOT work... i think i may have tried it earlier, and that is why i have to make a PR here

my .babelrc file will look like below

```
{
  "presets": [
    ["@babel/preset-env", {
      "corejs": 3,
      "useBuiltIns": "entry"
    }],
    "@babel/preset-typescript",
    "@babel/preset-react"
  ],
  "plugins": [
    ["@babel/plugin-transform-runtime", {
      "corejs": 3
    }],
    "@babel/plugin-proposal-object-rest-spread",
    ["@babel/plugin-proposal-decorators", {
      "legacy": true
    }],
    ["@babel/plugin-proposal-class-properties", {
      "loose": true
    }],
    ["babel-plugin-styled-components", {
      "minify": true,
      "transpileTemplateLiterals": true,
      "namespace": "whatever" // i tried it, no luck
    }]
  ],
  "comments": false
}
```

mattleonowicz:
@probablyup I think this should be reopened. Same as https://github.com/styled-components/styled-components/pull/3225
The class that actually contains css rules does not get the namespace in its name, I think it's a bug.

justnewbee:
> @probablyup I think this should be reopened. Same as #3186
> The class that actually contains css rules does not get the namespace in its name, I think it's a bug.

https://github.com/styled-components/babel-plugin-styled-components/issues/299 there seem to be an open issue, could u please fix that?

justnewbee:
> This is handled by https://styled-components.com/docs/tooling#namespace, thanks!

tried again, still not working...

a class name with some prefix is attached to the node, however there is no solid styles attached to it

part of the bablerc:

```
["babel-plugin-styled-components", {
  "minify": true,
  "transpileTemplateLiterals": true,
  "namespace": "XXX"
}]
```

the result:

image

i still have to use the way i did in this pull request, in which case, it really prevented class name collision:

image

BartJanvanAssen:
Hey! Any update on this? this still seems very valid and i'm facing the same issue :( @justnewbee do you found some other workaround?

BartJanvanAssen:
@probablyup is this worth another look?

khyberspache:
We are hitting the same issue. Namespace does not work

taranek:
@probablyup It's been a while. Is there any chance to check it out once again?

luzannew:
Yeah, it would really help if we can change this. We ended up in using something called 'patch-package' to add this PR changes into it, to fix our issues with naming collisions. Would be nice if it is out of the box supported

BartJanvanAssen:
+ @kitten

khyberspache:
@justnewbee we are using this pull request in production and it works perfectly

motevallian:
I think a wiser solution would be to NOT provide a bundle-time solution (a bundle time prefix) but rather a dynamic one. This gives us maximum flexibility and we only decide on the right prefix when we render components.
My suggestion is something like what Material-UI does via JSS, i.e. using a `StyleProvider` where any descendant components of this provider will automatically be prefixed by the value it provides:
```
  <StyleProvider classNamePrefix="mfA">
  {/* All classnames generated by children here are prefixed by mfA, e.g. mfa-2cb5fe */}
    ...
  </StyleProvider>
```

Let us decide on prefixing as late as possible. This also gives us the flexibility to run a logic to decide on the appropriate prefix, e.g. a piece of code might be able to auto-generate it for each micro-frontend loading on the page.

Issue #2952, opened 2020-01-07
CSS class order is mixed up when I extend a conditional rendered component's style in SSR.


cheolee:
## Environment

```
## System:
 - OS: macOS Mojave 10.14.6
 - CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
 - Memory: 12.55 GB / 32.00 GB
 - Shell: 2.7.1 - /usr/local/bin/fish
## Binaries:
 - Node: 12.14.0 - /usr/local/bin/node
 - Yarn: 1.12.3 - /usr/local/bin/yarn
 - npm: 6.13.4 - /usr/local/bin/npm
```

## Reproduction

I made a minimal reproducible example using razzle and nextjs.
https://github.com/cheolee/styled-components-ssr-order-issue

- AnchorOrBasicButton renders one of two components(AnchorButton, BasicButton) depending on `useAnchorTag` prop.
- StyledButton extends AnchorOrBasicButton.
- If the StyledButton that use BasicButton is rendered on the server side and later the StyledButton that use AnchorButton is rendered on the client side, the AnchorButton's style overrides the StyledButton's style.

### razzle example

1. Run project with `yarn start` and open localhost:3000.

2. Click `AnchorButtonPage` link.

3. AnchorButton will be red even though the color is specified.

### nextjs example

1. Run project with `yarn dev` and open localhost:3000.

2. Click `AnchorButtonPage` link.

3. AnchorButton will be red even though the color is specified.

## Expected Behavior

- `Yellow AnchorButton` should have yellow background-color as computed value
<img width="226" alt="image" src="https://user-images.githubusercontent.com/5275823/71929629-662d4e00-31dd-11ea-99a4-9bbc01e09eb8.png">

<img width="493" alt="image" src="https://user-images.githubusercontent.com/5275823/71929638-6af20200-31dd-11ea-899e-ea3cd690430d.png">

```
<a class="sc-iwsKbI ctruER sc-gZMcBi ciAlfw">Yellow AnchorButton</a>
```

```
<style data-styled="" data-styled-version="4.4.1">
/* sc-component-id: sc-iwsKbI */
.ctruER{background-color:red;}
/* sc-component-id: sc-gZMcBi */
.ciAlfw{background-color:yellow;}
/* sc-component-id: sc-gqjmRU */
.iRDvLn{display:-webkit-box;display:-webkit-flex;display:-ms-flexbox;display:flex;}
[...]

Issue #2950, opened 2020-01-07
SSR. wrong css priority using styled.css


ddalpange:
### System:
 - OS: macOS Mojave 10.14.6
 - CPU: (4) x64 Intel(R) Core(TM) i5-7600 CPU @ 3.50GHz
 - Memory: 39.66 GB / 56.00 GB
 - Shell: 5.3 - /bin/zsh

### Binaries:
 - Node: 12.13.1 - ~/.nvm/versions/node/v12.13.1/bin/node
 - Yarn: 1.17.3 - ~/Projects/class101/node_modules/.bin/yarn
 - npm: 6.12.1 - ~/.nvm/versions/node/v12.13.1/bin/npm

## Reproduction

There are two buttons using styled.css.

ContainButton = ContainButton Css > CommonButtonCss 
AnchorButton = AnchorButtonCss > CommonButtonCss 

If render only ContainButton in server side. AnchorButton is broken in client side.

because CommonButtonCss is created in index.html(header css). (server side).
so CommonButtonCss priority is bigger than AnchorButtonCss priority.

how can I fix it?

you can check button '결제하기' here.

https://class101.net/products/t0xqIJItJNxhobu1MAjH/confirmOrder (o)
https://ssr.class101.net/products/t0xqIJItJNxhobu1MAjH/confirmOrder (x)

I'm using @class101/ui. 

## Steps to reproduce

1. use @class101/ui
2. render only ContainButton in server side. and render only AnchorButton in client side.
3. you can see ugly button :( 

## Expected Behavior

CommonButtonCss should have difference classNames for each components.

## Actual Behavior

CommonButtonCss should have same classNames for each components.

probablyup:
You need to render the same component tree between SSR and client for things to work properly. Do any client-specific changes in a mount effect to ensure it happens after rehydration. React does not handle server vs client differences during hydration smoothly.

PR #3233, opened 2020-08-24
Enable new style rules can be inserted in the middle of existing sheet when rendering on client after rehydrate

mu29:
Solves #2950, #2952 
For more details of the problem, see #2952 

## Actual Behavior
스크린샷 2020-09-21 오후 6 42 49

The left side is the `reverseRegister` of the client, and the right side is the `reverseRegister` of the server.

스크린샷 2020-09-21 오후 6 43 30

The class `.fCnWMK` should be located 2nd, right after `.ldJpNF`. However it is in 5th on the client.

## Expected Behavior
스크린샷 2020-09-21 오후 6 51 35

So I modified the `nextFreeGroup` setting method - to find in `reverseRegister`, not with a value greater than the added group. Also deleted the `GroupIDAllocator`'s "throws early if the group ID is too large" test which is now less significance.

### DIFF
--- a/packages/styled-components/src/sheet/GroupIDAllocator.js
+++ b/packages/styled-components/src/sheet/GroupIDAllocator.js
@@ -19,7 +19,12 @@ export const getGroupForId = (id: string): number => {
     return (groupIDRegister.get(id): any);
   }
 
+  while (reverseRegister.has(nextFreeGroup)) {
+    nextFreeGroup++;
+  }
+
   const group = nextFreeGroup++;
+
   if (
     process.env.NODE_ENV !== 'production' &&
     ((group | 0) < 0 || group > MAX_SMI)
@@ -37,10 +42,6 @@ export const getIdForGroup = (group: number): void | string => {
 };
 
 export const setGroupForId = (id: string, group: number) => {
-  if (group >= nextFreeGroup) {
-    nextFreeGroup = group + 1;
-  }
-
   groupIDRegister.set(id, group);
   reverseRegister.set(group, id);
 };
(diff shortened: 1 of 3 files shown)


packages/styled-components/src/models/StyleSheetManager.js
```diff
@@ -38,7 +38,7 @@ export default function StyleSheetManager(props: Props) {
 
     if (props.sheet) {
       // eslint-disable-next-line prefer-destructuring
-      sheet = props.sheet;
+      sheet = props.sheet.rehydrate();
```
probablyup, line 41: @mu29 do you have an example of this in use? We call `rehydrateSheet` from the Sheet constructor, so realistically it should already be called by the time you pass `sheet` into `StyleSheetManager`

mu29:
@probablyup Hi! I fixed real issue and updated the comment. Please check it out :)

kitten (approved):
This is really nice!
When I wrote this code I wasn't quite sure whether this empty group inserting would be safe, but I believe it must be since groups will still be registered in their right order.

I still think we have to further test and think whether it holds true though, since SSR results may be very much out of order, if preregistration fails.

packages/styled-components/src/sheet/GroupIDAllocator.js
```diff
@@ -19,7 +19,12 @@ export const getGroupForId = (id: string): number => {
     return (groupIDRegister.get(id): any);
   }
 
-  const group = nextFreeGroup++;
+  while (reverseRegister.has(nextFreeGroup)) {
+    nextFreeGroup++;
+  }
+
+  const group = nextFreeGroup;
```
kitten, line 26: This is technically more correct since we know that this group will be taken, so there's no need to check the reverse register on the next run against it.

```suggestion
  const group = nextFreeGroup++;
```

mu29: You're right! I fixed it :)

probablyup (approved):
Can you add a changelog entry?

probablyup:
@mu29 Thank you!

Issue #3163, opened 2020-06-04
stylis plugin not working for keyframes since 5.0.0-rc3 #2936


mzvast:
## System:
 - OS: macOS 10.15.5
 - CPU: (4) x64 Intel(R) Core(TM) i5-6500 CPU @ 3.20GHz
 - Memory: 69.18 MB / 8.00 GB
 - Shell: 5.7.1 - /bin/zsh
## Binaries:
 - Node: 10.21.0 - ~/.nvm/versions/node/v10.21.0/bin/node
 - Yarn: 1.22.4 - ~/.nvm/versions/node/v10.21.0/bin/yarn
 - npm: 6.14.4 - ~/.nvm/versions/node/v10.21.0/bin/npm

## Reproduction
I wrote a stylis plugin to transform px to vm.
It works fine before following commit:

![image](https://user-images.githubusercontent.com/8097465/83751693-d6130800-a699-11ea-97fb-8f8bbf54131a.png)

My plugin stop working when it encounters @keyframes mark, which can't receive any input then.

### Working
 [workspace](https://codesandbox.io/s/styled-components-working-example-6hjr3)
![image](https://user-images.githubusercontent.com/8097465/83749118-c72a5680-a695-11ea-91ab-bca97d4f07ee.png)

### Broken
[workspace](https://codesandbox.io/s/styled-components-borken-example-r6xy7?file=/index.js)
![image](https://user-images.githubusercontent.com/8097465/83748935-87636f00-a695-11ea-95f6-cf04d470759f.png)

## Steps to reproduce
change styled-components version to greater than rc3 .
## Expected Behavior
My plugin should receive @keyframes input.
## Actual Behavior
 My plugin didn't receive @keyframes input.

## My attempts
I used binary searching to accurately allocate the first breaking commit.
And I spend some time to compare the behavior and output between fine version and broken version.
However, I'm not quite sure how it works, especially on following lines.
https://github.com/styled-components/styled-components/blob/37f472da70af60ff3e02e7f9a0073c895db4eb04/packages/styled-components/src/models/Keyframes.js#L20-L24

I add a debugger right here, and hit refresh.
![image](https://user-images.githubusercontent.com/8097465/83750940-a0215400-a698-11ea-93e9-909d5be89533.png)

To my surprise, the break point didn't hit at all. But still injected @keyframes to style tag as below.
[...]

Issue #2993, opened 2020-01-28
GlobalStyles using an attribute selector are loaded after regular styles


joshuajaco:
## System:
 - OS: macOS 10.15.2
 - CPU: (8) x64 Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
 - Memory: 1.12 GB / 16.00 GB
 - Shell: 5.7.1 - /bin/zsh
## Binaries:
 - Node: 13.7.0 - ~/.asdf/installs/nodejs/13.7.0/bin/node
 - Yarn: 1.21.1 - ~/.asdf/installs/nodejs/13.7.0/.npm/bin/yarn
 - npm: 6.13.6 - ~/.asdf/installs/nodejs/13.7.0/bin/npm
## npmPackages:
 - babel-plugin-styled-components: ^1.10.7 => 1.10.7
 - styled-components: ^5.0.0 => 5.0.0
## Reproduction
v5.0.0: https://codesandbox.io/s/thirsty-kalam-hi5s8
v4.4.1:  https://codesandbox.io/s/strange-mendeleev-pindz

## Steps to reproduce
1. Create a global stylesheet using an attribute selector to set some styles.
2. Try to overwrite those styles on a component

## Expected Behavior
The global styles are overwritten by the component styles, as in v4.4.1

## Actual Behavior
The component styles are overwritten

4nduril:
I can confirm that issue. Here is a sandbox where I tried to load the `createGlobalStyle` component from another file before the "App". But that doesn't help either. https://codesandbox.io/s/crimson-violet-mv5xy

 In the `<style>` tag in devtools here https://mv5xy.csb.app/ you can see that the global styles are coming after the StyledH1 declarations.

probablyup:
We currently don't guarantee an insertion order for global styles. I agree this should be fixed though.

Leonya:
Really hoping that this gets fixed, it's a major bummer for us 🙏

davidx1:
I have just ran into the same issue. Replicated here:
https://codesandbox.io/embed/great-grass-l9y63

Since attribute selector and class selector have the same specificity, and global style is meant to set a base-line for individual components to override, I think the library should gurantee that global styles are inserted first before normal styles.

Leonya:
[...]

PR #3239, opened 2020-08-28
misc updates

### LINKED ISSUE #2074: Problem with hot module reloading & createGlobalStyle
## System:
 - OS: macOS Sierra 10.12.6
 - CPU: x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
 - Memory: 105.95 MB / 16.00 GB
 - Shell: 3.2.57 - /bin/bash
## Binaries:
 - Node: 10.1.0 - ~/.nvm/versions/node/v10.1.0/bin/node
 - Yarn: 1.6.0 - /usr/local/bin/yarn
 - npm: 6.1.0 - ~/.nvm/versions/node/v10.1.0/bin/npm
 - Watchman: 4.7.0 - /usr/local/bin/watchman
## npmPackages:
 - styled-components: ^4.0.0-beta.10 => 4.0.0-beta.10 

Hi there, not sure if I would call this a bug but thought it was worth pointing out.

I have started using the `createGlobalStyle` function in v4, but have run into some issues relating to Hot Module Reloading. Whenever a reload is triggered by something outside the `createGlobalStyle` function, all of its comprising css rules are lost. Here is a simple project which demonstrates this:
https://github.com/frednomoon/styled-hot-reload

It seems from this behaviour that its necessary to reload the module containing the global styles every time the page changes. But, of course, this doesn't happen unless you actually change something inside that module.

I'm guessing its probably possible to solve this with some additional config in my bundler, but obviously would be great to avoid things like that.

### LINKED ISSUE #3008: Global styles does not work in React.StrictMode
## Environment

System:
OS: macOS 10.15.1
CPU: (8) x64 Intel(R) Core(TM) i7-8569U CPU @ 2.80GHz
Memory: 857.00 MB / 16.00 GB
Shell: 3.2.57 - /bin/bash
Binaries:
Node: 10.15.3 - ~/.nvm/versions/node/v10.15.3/bin/node
Yarn: 1.16.0 - /usr/local/bin/yarn
npm: 6.4.1 - ~/.nvm/versions/node/v10.15.3/bin/npm
npmPackages:
styled-components: ^5.0.1 => 5.0.1
Reproduction
https://codesandbox.io/s/wonderful-hill-6glof

## Steps to reproduce

Load the code sandbox
Click the button "to black" and then click again "to white" and notice how the background did not change back to black.

## Expected Behavior

The should change between black and white

## Actual Behavior

The background stays black after the first press on the toggle button. The global CSS (rendered by GlobalStyle in ComponentBlack) is added twice and only one is removed when unmounting when in Strict Mode. This is because to Strict Mode "intentionally double-invoking" some lifecycle methods to determine side effects that could be a problem in the upcoming async mode in React. When concurrent mode becomes available and React might call render function multiple times the side effect of setting global styles on render will cause issues like this.

As far as I can tell `styled-components` should be `React.Strict` compatible and using the `createGlobalStyle` util is the recommended way to do global styling with `styled-components`, so this seems to be a bug to me.

probablyup:
1. pay attention to SC_VERSION when hashing to provide more effective component isolation when multiple SC instances are running on the page

2. Bundle size improvements

3. Fix `keyframes` not receiving nested stylis context, and therefore missing plugins if applicable

Fixes #3008 #3076 #2993 #3163
Probably fixes #2074


packages/styled-components/rollup.config.js
```diff
@@ -123,7 +131,8 @@ const browserConfig = {
   plugins: configBase.plugins.concat(
     replace({
       __SERVER__: JSON.stringify(false),
-    })
+    }),
+    minifierPlugin
```
kitten, line 135: If we're minifying all output bundles (which I'm a fan of) I'd love to make sure to also try out Closure Compiler, which is pretty amazing https://github.com/styled-components/styled-components/blob/d220a202ffc590447d32ce3b5b79e61fe939308b/packages/styled-components/rollup.config.js#L111-L137

This is what we're currently doing for `urql` at least, where we try to squeeze out a lot of bytes here and there https://github.com/FormidableLabs/urql/blob/e031f9b887960a8113890777e8a100973068a36c/scripts/rollup/plugins.js#L98-L137

packages/styled-components/src/models/ComponentStyle.js
```diff
@@ -29,7 +29,11 @@ export default class ComponentStyle {
     this.staticRulesId = '';
     this.isStatic = process.env.NODE_ENV === 'production' && isStaticRules(rules);
     this.componentId = componentId;
-    this.baseHash = hash(componentId);
+
+    // SC_VERSION gives us isolation between multiple runtimes on the page at once
+    // this is improved further with use of the babel plugin "namespace" feature
+    this.baseHash = hash(SC_VERSION + componentId);
```
kitten, line 35: Do you think it's worth running `hash` on a seed value separately and then continue using `phash` here?

packages/styled-components/src/utils/error.js
```diff
@@ -31,8 +31,8 @@ export default function throwStyledComponentsError(
 ) {
   if (process.env.NODE_ENV === 'production') {
     throw new Error(
-      `An error occurred. See https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/utils/errors.md#${code} for more information.${
-        interpolations.length > 0 ? ` Additional arguments: ${interpolations.join(', ')}` : ''
+      `An error occurred. See https://git.io/JUIaE#${code} for more information.${
```
kitten, line 34: Nice!

packages/styled-components/src/utils/flatten.js
```diff
@@ -8,45 +8,47 @@ import isStyledComponent from './isStyledComponent';
...
-          rules.push(`${hyphenate(key)}:`, obj[key], ';')
-
-          return rules
-        }
-        rules.push(`${hyphenate(key)}: ${addUnitIfNeeded(key, obj[key])};`);
+  const rules = [];
+  const keys = Object.keys(obj);
+
+  keys.forEach(key => {
+    if (!isFalsish(obj[key])) {
```
kitten, line 23: - can we get rid of the nested statements?
- why are we returning `rules`?
- should we just use for-in and `hasOwnProperty` here?

kitten (approved):

packages/styled-components/rollup.config.js
probablyup: Shaved off 0.4kB!

packages/styled-components/src/models/ComponentStyle.js
probablyup: Will bench it

packages/styled-components/src/utils/flatten.js
probablyup: This actually made a massive difference in the performance of our object style implementation! Awesome review

packages/styled-components/src/models/ComponentStyle.js
probablyup: Minor but noticeable improvement 👍

packages/styled-components/rollup.config.js
probablyup: So although it reduced the bundle size, something that terser doing actually regressed perf here. Going to hold off on it for now.

PR #3241, opened 2020-08-28
[breaking] stylis v4

probablyup:
moves off @emotion/stylis and attempts to retain backward-compatibility
for everything except stylis plugins, which cannot be made compatible
due to a vastly different middleware setup in v4

note that some of the tests will be broken until I revise `stylis-plugin-rtl` for the v4 middleware signature


packages/styled-components/src/constructors/test/createGlobalStyle.test.js
```diff
@@ -301,7 +301,7 @@ describe(`createGlobalStyle`, () => {
     expect(getCSS(document).trim()).toMatchInlineSnapshot(`"body{background:red;}"`);
 
     render(<A />);
-    expect(getCSS(document).trim()).toMatchInlineSnapshot(`""`);
+    expect(getCSS(document).trim()).toMatchInlineSnapshot(`"body{background:;}"`);
```
probablyup, line 304: This was interesting, it should have been rendering this previously and wasn't...

packages/styled-components/src/sheet/GroupedTag.js
```diff
@@ -34,7 +34,7 @@ class DefaultGroupedTag implements GroupedTag {
     return index;
   }
 
-  insertRules(group: number, rules: string[]): void {
+  insertRules(group: number, rules: string | string[]): void {
```
kitten, line 37: Are we sure that we want to make this function more polymorphic? And is this a special case? Afaik if we know that we only have one rule then `[x]` does the trick, so is this because of a change in Stylis?

probablyup: > Are we sure that we want to make this function more polymorphic? And is this a special case? Afaik if we know that we only have one rule then `[x]` does the trick, so is this because of a change in Stylis?

Yes stylis only emits a single rule now instead of an array. In my benchmarking this was more performant...

kitten: This could depend though. Are we sure that a variable array sometimes and a string most of the time is properly benchmarked with attempts to evenly switch between the two to check for V8 performance cliffs? 😅 just making sure

probablyup: Good question, It's mostly rehydration where multirule inserts happen and we don't have any benchmarks for that.

kitten: I guess making this work goes first, making it fast goes second. There's a lot of refactoring we could do anyway to slim thing down generally.

probablyup:
@kitten this is ready for final review

PR #3248, opened 2020-09-01
update flow, rework types for StyledComponent model

probablyup:


packages/styled-components/src/models/StyledComponent.js
```diff
@@ -211,24 +203,24 @@ export default function createStyledComponent(
 
   // fold the underlying StyledComponent attrs up (implicit extend)
   const finalAttrs =
-    // $FlowFixMe
-    isTargetStyledComp && target.attrs
-      ? Array.prototype.concat(target.attrs, attrs).filter(Boolean)
+    isTargetStyledComp && ((target: any): IStyledComponent).attrs
```
probablyup, line 206: I would love another option than doing `((target: any): IStyledComponent)` everywhere if you have a suggestion @kitten. Flow doesn't seem to be smart enough to infer things correctly.

packages/styled-components/src/models/StyledComponent.js
```diff
@@ -238,20 +230,23 @@ export default function createStyledComponent(
     isTargetStyledComp ? ((target: Object).componentStyle: ComponentStyle) : undefined
   );
 
+  // statically styled-components don't need to build an execution context object,
+  // and shouldn't be increasing the number of class names
+  const isStatic = componentStyle.isStatic && attrs.length === 0;
```
probablyup, line 235: Hoisted this up to the definition layer rather than runtime layer so we can use it for future optimizations...

packages/styled-components/src/models/StyledComponent.js
kitten: I'm not sure; you can add a `%checks` directive directly but I think that only works if we remove the intermediate `isTargetStyledComp` variable or if we just make a larger `if` statement branch for when the function returns true

kitten (approved):
🙌

PR #3267, opened 2020-09-08
Fix regex breaking on IE in hyphenateStyleName

### LINKED ISSUE #3266: Syntax error in regular expression on IE11
## Environment

IE11

## Reproduction

Syntax error appears instantly when opening page using styled-components@5.2.0 in IE11.

`Syntax error in regular expression`

Most probably caused by dd6b3028f91dbbf1c026f7ee58d1ec1bea76b765.

`/([A-Z])/g` works but `new RegExp(/([A-Z])/, 'g');` breaks.

N1cc3:
Fixes #3266

### DIFF
--- a/packages/styled-components/src/utils/hyphenateStyleName.js
+++ b/packages/styled-components/src/utils/hyphenateStyleName.js
@@ -6,7 +6,7 @@
  */
 
 const uppercaseCheck = /([A-Z])/;
-const uppercasePattern = new RegExp(uppercaseCheck, 'g');
+const uppercasePattern = /([A-Z])/g;
 const msPattern = /^ms-/;
 const prefixAndLowerCase = (char: string): string => `-${char.toLowerCase()}`;
 


packages/styled-components/src/utils/hyphenateStyleName.js
```diff
@@ -27,7 +26,7 @@ const prefixAndLowerCase = (char: string): string => `-${char.toLowerCase()}`;
  * @return {string}
  */
 export default function hyphenateStyleName(string: string): string {
-  return uppercaseCheck.test(string)
+  return uppercasePattern.test(string)
```
kitten, line 29: I suppose this is an accidental change? This purposefully uses a non-global regex to avoid the `lastIndex` being moved around. So let's restore `uppercaseCheck` here and only change `uppercasePattern` to a literal :)

N1cc3: Oh I see now! :) Amended fix resulting in 813329a.

kitten (approved):

MarkAPhillips:
Please advise when this fix will be released?

frosas:
@MarkAPhillips this was released a couple of weeks ago as part of 5.2.1 (https://github.com/styled-components/styled-components/commit/683b668d9255d990d67f3a74f29f4ce42034f5fb)

Issue #3270, opened 2020-09-14
Sandbox is not working


strozw:
I'm forking and working another issue, but sandbox is not working and display error.

## Environment

```
## System:
 - OS: macOS 10.15.6
 - CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
 - Memory: 18.03 GB / 64.00 GB
 - Shell: 5.8 - /usr/local/bin/zsh
## Binaries:
 - Node: 12.14.1 - ~/.anyenv/envs/nodenv/versions/12.14.1/bin/node
 - Yarn: 1.22.4 - ~/.anyenv/envs/nodenv/versions/12.14.1/bin/yarn
 - npm: 6.13.4 - ~/.anyenv/envs/nodenv/versions/12.14.1/bin/npm
 - Watchman: 4.9.0 - /usr/local/bin/watchman
```

## Steps to reproduce

It's reference from [CONTRIBUTING.md](https://github.com/styled-components/styled-components/blob/master/CONTRIBUTING.md).
```
yarn install
cd ./packages/sandbox
yarn install
yarn start
```

## Expected Behavior

open `http://localhost:3000`, It display sandbox and work.

## Actual Behavior

open `http://localhost:3000`, It display bellow error.

```
Error: __SERVER__ is not defined
    at createError (~/styled-components/node_modules/micro/lib/index.js:20:14)
    at module.exports (~/styled-components/packages/sandbox/server/index.js:11:11)
```

show server log on terminal, It display bellow error.

```
  Error: __SERVER__ is not defined

  - index.js:20 createError
    [styled-components]/[micro]/lib/index.js:20:14

  - index.js:11 module.exports
    ~/styled-components/packages/sandbox/server/index.js:11:11

  - task_queues.js:94 processTicksAndRejections
    internal/process/task_queues.js:94:5
```

PR #3271, opened 2020-09-14
Fix sandbox error displaying "__SERVER__ is not defined"

strozw:
It's PR for the following sandbox issue.
https://github.com/styled-components/styled-components/issues/3270


probablyup (approved):
Thanks!

PR #3289, opened 2020-09-25
fix: Allow `DISABLE_SPEEDY` to be set to `false`

fastfedora:
The old initialization of `DISABLE_SPEEDY` meant that the only way it could be set to a falsey value was if `NODE_ENV` was set to `production` and no other setting was set. Setting `SC_DISABLE_SPEEDY` or the environment variables to false would be overridden by `NODE_ENV`.

The new initialization ensures that if one of these options is set to false, its value is respected.

This change should only affect the computed value of `DISABLE_SPEEDY` if one of the settings has the value `false` or `’false’`. Otherwise it should have the same truthy or falsey value as before.

This basically fixes the expected behavior of setting one of the configuration variables to `false` to set `DISABLE_SPEEDY`  to false.

Note that aims to fix the same problem fixed by PR #2951, but does so without any unexpected breaking changes.

### DIFF
--- a/packages/styled-components/src/test/constants.test.js
+++ b/packages/styled-components/src/test/constants.test.js
@@ -76,6 +76,7 @@ describe('constants', () => {
     afterEach(() => {
       process.env.NODE_ENV = 'test';
       delete process.env.DISABLE_SPEEDY;
+      delete window.SC_DISABLE_SPEEDY;
     });
 
     it('should be false in production NODE_ENV when SC_DISABLE_SPEEDY is not set', () => {
@@ -83,6 +84,12 @@ describe('constants', () => {
     });
 
     it('should be false in production NODE_ENV when window.SC_DISABLE_SPEEDY is set to false', () => {
+      window.SC_DISABLE_SPEEDY = false;
+      renderAndExpect(false, '');
+    });
+
+    it('should be false in development NODE_ENV when window.SC_DISABLE_SPEEDY is set to false', () => {
+      process.env.NODE_ENV = 'development';
       window.SC_DISABLE_SPEEDY = false;
       renderAndExpect(false, '');
     });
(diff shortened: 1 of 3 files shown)


kitten:
So, this is awkward, but I believe we got a PR for this before, and had to reject it. I didn't get the chance to create an alternative one because I haven't slated a lot of time for styled-components for now; but tl;dr this PR presents the same problem.

Helper functions, anonymous functions, and other structures that pull apart `process.env` may not work in all environments, bundles, and minifiers, as those often just evaluate pieces of code as text.

So any change here must use `process.env` without any indirections for better or for worse.

I am aware that in some misconfigured environments, in development, `process.env` will just be evaluated to a large inline object literal; but either way in others `process.env.X` will be evaluated as is and must not be split up

So once that change is made we can go ahead with this change ✌️ cc @probablyup

fastfedora:
Okay, to clarify, use `process.env.REACT_APP_SC_DISABLE_SPEEDY` directly instead of `process.env[name]`, correct?

That's actually how I originally wrote it, but it looked so gnarly, I decided to clean it up with a function. I can easily merge all that back into one long statement.

fastfedora:
@kitten I inlined all the `process.env` references. It looks a bit ugly, but should work. Let me know if you need anything else done.

kitten (approved):
Not a fan of how it looks either but this is sure to work with any `process.env` evaluation 😅

packages/styled-components/src/test/constants.test.js
```diff
@@ -87,6 +88,12 @@ describe('constants', () => {
       renderAndExpect(false, '');
     });
 
+    it('should be false in development NODE_ENV when window.SC_DISABLE_SPEEDY is set to false', () => {
+      process.env.NODE_ENV = 'development';
+      window.SC_DISABLE_SPEEDY = false;
+      renderAndExpect(false, '');
```
tigerabrodi, line 94: Hey 👋, was wondering what `renderAndExpect` does? Many thanks, this is just out of curiosity 😄

probablyup: https://github.com/styled-components/styled-components/pull/3289/files#diff-8ee68021d59f52aef5b5d6fbd8a082e4R12-R28

probablyup (approved):
@fastfedora could you add a changelog entry?

PR #3285, opened 2020-09-24
Fix module resolve error in benchmarks package

strozw:
I've showing following error in case execute `yarn run benchmark` at master branch.

<details>
<summary>Showing Errors</summary>

```
 | ERROR in ./src/implementations/styled-components/Box.js
 | Module not found: Error: Can't resolve 'styled-components' in '/Users/strozw/ghq/github.com/strozw/styled-components/packages/benchmarks/src/implementations/styled-components'
 |  @ ./src/implementations/styled-components/Box.js 13:0-39 41:15-21
 |  @ ./src/implementations/styled-components/index.js
 |  @ ./src/implementations sync index\.js$
 |  @ ./src/impl.js
 |  @ ./src/index.js
 |
 | ERROR in ./src/implementations/styled-components/View.js
 | Module not found: Error: Can't resolve 'styled-components' in '/Users/strozw/ghq/github.com/strozw/styled-components/packages/benchmarks/src/implementations/styled-components'
 |  @ ./src/implementations/styled-components/View.js 13:0-39 14:15-21
 |  @ ./src/implementations/styled-components/index.js
 |  @ ./src/implementations sync index\.js$
 |  @ ./src/impl.js
 |  @ ./src/index.js
 |
 | ERROR in ./src/implementations/styled-components/Dot.js
 | Module not found: Error: Can't resolve 'styled-components' in '/Users/strozw/ghq/github.com/strozw/styled-components/packages/benchmarks/src/implementations/styled-components'
 |  @ ./src/implementations/styled-components/Dot.js 13:0-39 15:15-21
 |  @ ./src/implementations/styled-components/index.js
 |  @ ./src/implementations sync index\.js$
 |  @ ./src/impl.js
 |  @ ./src/index.js
 |
 | ERROR in ./src/implementations/styled-components-object/Box.js
 | Module not found: Error: Can't resolve 'styled-components' in '/Users/strozw/ghq/github.com/strozw/styled-components/packages/benchmarks/src/implementations/styled-components-object'
 |  @ ./src/implementations/styled-components-object/Box.js 1:0-39 29:15-21
 |  @ ./src/implementations/styled-components-object/index.js
 |  @ ./src/implementations sync index\.js$
 |  @ ./src/impl.js
 |  @ ./src/index.js
 |
 | ERROR in ./src/implementations/styled-components-object/Dot.js
[...]

PR #3295, opened 2020-09-29
Improve GroupedTag#insertRules performance in speedy disabled

strozw:
If components have many style rules, GroupedTag#insertRules bring many reflow (layout) in speedy disabled.
This PR improve it by using DocumentFragment.

In the sample below, you can see the difference in CRA behavior when using this PR branch and v5.2.
https://github.com/strozw/sample-many-rules-styled-components

You can also check by loading the following json on the Performance tab of chrome devtools.

* [v5_2-performance.json.txt](https://github.com/strozw/styled-components/files/5300755/v5_2-performance.json.txt)
* [fixed-profile.json.txt](https://github.com/strozw/styled-components/files/5300762/fixed-profile.json.txt)

Or you can check this problem in [this codesandobx](https://codesandbox.io/s/has-many-rules-styled-componetns-cra-9lg1b?file=/src/App.js)

### DIFF
--- a/packages/styled-components/src/sheet/Tag.js
+++ b/packages/styled-components/src/sheet/Tag.js
@@ -15,6 +15,24 @@ export const makeTag = ({ isServer, useCSSOMInjection, target }: SheetOptions): 
   }
 };
 
+function insertRulesIntoTag(
+  startRuleIndex: number,
+  rules: string[],
+  tag: Tag
+): number {
+  let inserted = 0
+  let ruleIndex = startRuleIndex
+
+  for (let i = 0, l = rules.length; i < l; i++) {
+    if (tag.insertRule(ruleIndex, rules[i])) {
+      inserted++
+      ruleIndex++;
+    }
+  }
+
+  return inserted
+}
+
 export class CSSOMTag implements Tag {
   element: HTMLStyleElement;
 
@@ -40,6 +58,10 @@ export class CSSOMTag implements Tag {
     } catch (_error) {
       return false;
     }
+  }
+
+  insertRules(startRuleIndex: number, rules: string[]) {
+    return insertRulesIntoTag(startRuleIndex, rules, this)
   }
 
   deleteRule(index: number): void {
@@ -84,6 +106,23 @@ export class TextTag implements Tag {
     }
   }
 
+  insertRules(startRuleIndex: number, rules: string[]): number {
+    if (startRuleIndex > this.length || startRuleIndex < 0) {
+      return 0;
+    }
+    
+    const fragment = document.createDocumentFragment();
+    for (let i = 0, l = rules.length; i < l; i++) {
+      const node = document.createTextNode(rules[i]);
+      fragment.appendChild(node);
+    }
+
+    const refNode = this.nodes[startRuleIndex];
+    this.element.insertBefore(fragment, refNode || null);
+    this.length += rules.length;
+    return rules.length;
+  }
+
   deleteRule(index: number): void {
     this.element.removeChild(this.nodes[index]);
     this.length--;
@@ -117,6 +156,10 @@ export class VirtualTag implements Tag {
     } else {
       return false;
     }
+  }
+
+  insertRules(startRuleIndex: number, rules: string[]) {
+    return insertRulesIntoTag(startRuleIndex, rules, this)
   }
 
   deleteRule(index: number): void {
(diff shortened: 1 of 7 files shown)


kitten:
This is a great change! It'd be cool to get some numbers from our benchmark so we can have reliable comparisons, instead of observational numbers from those single recordings.

The thing I'm most concerned about however is adding more code for the non-speedy mode which is really just a degraded more for development that is also now mostly obsolete. So I'm not sure whether we should even keep it for long going forward. There are some minor improvements we can make in the PR, but it's mostly good.
But before jumping on a review, can we try to get some more numbers please, and @probablyup, shall we have a quick discussion on where we are with non-speedy?

strozw:
@kitten 
> It'd be cool to get some numbers from our benchmark so we can have reliable comparisons

Indeed. I wanted to use `@storybook/benchmark`, but it was broken, so [I fixed it first](https://github.com/styled-components/styled-components/pull/3285).

@kitten, @probablyup  

> @kitten yeah given that CSSOM is richly processed in Chromium now I don't think we need texttag mode anymore at all

Yeah i think so too 👍. using CSSOM (speedy mode) is more fast 🚀 and, If this can be use in development mode, Developer Experiment will be more comfortable.

This PR is for that is current styled component is working non speedy mode in `NODE_ENV=development`. 

However, [previous PR merge](https://github.com/styled-components/styled-components/pull/3289) will enable `DISABLED_SPEEDY = false`, so if you plan to remove the TextTag in future, I think it's ok to close this PR. 👌 

Is it OK to close this PR?

probablyup:
If we remove TextTag it’ll be in v6, is this a big enough issue that it should be fixed in v5?

kitten:
@probablyup @strozw I'm fine with reviewing and merging it since we still have more patches coming for v5 for sure, so let's get this in, since it's a nice quality of life feature?

probablyup:
Sounds good, @strozw could you add a changelog entry?

packages/styled-components/src/sheet/Tag.js
```diff
@@ -84,6 +108,29 @@ export class TextTag implements Tag {
...
+
+      const elementRefNode = this.nodes[startRuleIndex]
+      inserted = fragmentNodes.length
+
+      this.element.insertBefore(fragment, elementRefNode)
+      this.length += inserted;
+    }
+
+    return inserted
+  }
```
kitten, line 132:
```suggestion
  insertRules(startRuleIndex: number, rules: string[]): number {
    if (startRuleIndex > this.length || startRuleIndex < 0) {
      return 0;
    }
    
    const fragment = document.createDocumentFragment();
    for (let i = 0, l = rules.length; i < l; i++) {
      const node = document.createTextNode(rules[i]);
      fragment.appendChild(node);
    }

    const refNode = this.nodes[startRuleIndex];
    this.element.insertBefore(fragment, refNode || null);
    this.length += rules.length;
    return rules.length;
  }
```

So instead of typing out some random suggestions here, I'm just demonstrating some things in code in form a suggested change 😅 

- `insertBefore` is unnecessary in the fragment loop since we're always appending in order
- `this.length` and the return value can be increased in one go
- add missing `|| null` default to `insertBefore` call at the end

packages/styled-components/src/sheet/Tag.js
```diff
@@ -15,14 +15,36 @@ export const makeTag = ({ isServer, useCSSOMInjection, target }: SheetOptions):
   }
 };
 
-export class CSSOMTag implements Tag {
+class BaseTag {
+  insertRule(_index: number, _rule: string) {
```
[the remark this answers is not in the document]
strozw: Thank you for your nice advice. I thought BaseTag wasn't cool, so it was helpful.
I fixed it with the following commit, so please give me your opinion.
[`15b89ad` (#3295)](https://github.com/styled-components/styled-components/pull/3295/commits/15b89adb3d4f2b836679ed1e8a66f60e00e6cdf5)

packages/styled-components/src/sheet/Tag.js
strozw: thanks! I'd adopted.

strozw:
@probablyup @kitten 

I'm sorry. I missed the following note.

> The master branch is under development of the upcoming v6 major version of styled-components. For changes targeting v5, please point your PRs at the legacy-v5 branch.

Therefore, I changed the target branch of PR to legacy-v5.

kitten:
I believe, looking at it this violates our expectations around specificity and ordering. I believe at the time I must've thought this was about the `TextTag` or non-speedy/CSSOM tag, but I think I was mistaken.

PR #3296, opened 2020-09-30
Update ISSUE_TEMPLATE.md

LironEr:
When running `npx envinfo` command with the flag `clipboard` I get this error:
```
*** Clipboard option removed - use clipboardy or clipboard-cli directly ***
```

So I just replaced the flag with `clipboard-cli` in order to copy the output to my clipboard.


kitten:
@probablyup I remember us receiving g a PR to add this but I can't remember this ever influencing our bug hunts beyond quickly seeing what version things are installed at, so I'd love to reduce the template here and maybe even remove this, since the system info is much too detailed 😅

probablyup:
Yeah, pretty much all we need to hunt a bug is a sandbox

Issue #3178, opened 2020-06-16
Nested `attrs` result in the wrong prop value when one prop depends on another


exogen:
## Environment

`npx envinfo --system --binaries --npmPackages styled-components,babel-plugin-styled-components --markdown --clipboard`

```
## System:
 - OS: macOS 10.15.5
 - CPU: (8) x64 Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
 - Memory: 611.90 MB / 16.00 GB
 - Shell: 3.2.57 - /bin/bash
## Binaries:
 - Node: 10.13.0 - ~/.nvm/versions/node/v10.13.0/bin/node
 - Yarn: 1.22.4 - /usr/local/bin/yarn
 - npm: 6.10.2 - ~/.nvm/versions/node/v10.13.0/bin/npm
## npmPackages:
 - styled-components: ^5.1.1 => 5.1.1
```

## Reproduction

https://codesandbox.io/s/nostalgic-dhawan-y2jiu?file=/src/App.js

## Steps to reproduce

Use `attrs` in a nested way. Let's start with this component:

```js
const Foo = styled.button.attrs(props => ({
  size: props.size || 10
})`
  font-size: ${props => props.size}px;
`;
```

Then `<Foo>` will render a button with font-size 10. `<Foo size={20}>` will render a button with font-size 20. So far so good!

Now let's create another styled-component from that:

```js
const Bar = styled(Foo).attrs({
  size: 20
})`
  color: red;
`;
```

Now rendering `<Bar>` will render a button with font-size 20. Still all good!

Now let's update Foo to default one of its *other* props to depend on the `size` prop:

```js
const Foo = styled.button.attrs(props => ({
  size: props.size || 10,
  children: props.children || `size: ${props.size}`
})`
  font-size: ${props => props.size}px;
`;
```

Now something strange happens. `<Bar>` doesn't pass `children`, but does pass `size: 20` – so we expect Foo's `children` default to run and see `size: 20`. But when we render `<Bar>` we see:

```
size: undefined
```

It seems as if somehow the `children` default from Foo was applied *before* Bar's `attrs` were taken into consideration.

Rendering `<Bar size={20}>` does give `children` the correct value. The result is that rendering `<Bar>` which is defined as `styled(Foo).attrs({ size: 20 })` differs from rendering `<Foo size={20}>`.

[...]

PR #3299, opened 2020-10-05
added clarification with link to docs on how several layers of .attrs are applied

vzaidman:
if anybody would debug why some attrs are not as expected for them, they will see a link to docs that explains it.

here is an example of a bug opened on it before:
https://github.com/styled-components/styled-components/issues/3178


kitten:
Adding comments after the fact is a little unusual and we can indeed see the order in a completely different place already: https://github.com/styled-components/styled-components/blob/958fee73e2a79e47e693f12ae8878e97b3b914d8/packages/styled-components/src/models/StyledComponent.js#L200-L204

If we improved the comments for some of our code I'd actually just rewrite most of it, ha 😆 So since that's not being done here and the docs are the canonical way to get this information, I'll close this, Sorry

PR #3303, opened 2020-10-09
minor improvements

anatoliy8493:


packages/styled-components/src/macro/index.js
```diff
@@ -6,12 +6,15 @@ import babelPlugin from 'babel-plugin-styled-components';
 
 function styledComponentsMacro({ references, state, babel: { types: t }, config = {} }) {
   const program = state.file.path;
+  const refNames = Object.keys(references);
```
lifeiscontent, line 10: @anatoliy8493 pretty sure Object.keys results in another loop internally.

anatoliy8493: reverted that hunk of edits

packages/styled-components/src/constructors/createGlobalStyle.js
```diff
@@ -14,6 +14,8 @@ declare var __SERVER__: boolean;
 
 type GlobalStyleComponentPropsType = Object;
 
+const isEnvDevelopment = process.env.NODE_ENV !== 'production';
```
lifeiscontent, line 17:
```suggestion
const isNotProduction = process.env.NODE_ENV !== 'production';
```

```suggestion
const isDevelopment = process.env.NODE_ENV === 'development';
```

packages/styled-components/src/constructors/styled.js
```diff
@@ -8,8 +8,8 @@ import type { Target } from '../types';
 const styled = (tag: Target) => constructWithOptions(StyledComponent, tag);
 
 // Shorthands for all valid HTML Elements
-domElements.forEach(domElement => {
-  styled[domElement] = styled(domElement);
-});
+for (let i = 0; i < domElements.length; i += 1) {
+  styled[domElements[i]] = styled(domElements[i]);
+}
```
lifeiscontent, line 13: @anatoliy8493 if you use a for loop you need to use `hasOwnProperty` to check that it belongs to the object and not the prototype. forEach is better here imo.

packages/styled-components/src/constructors/test/styled.test.js
```diff
@@ -4,8 +4,8 @@ import domElements from '../../utils/domElements';
 
 describe('styled', () => {
   it('should have all valid HTML5 elements defined as properties', () => {
-    domElements.forEach(domElement => {
-      expect(styled[domElement]).toBeTruthy();
-    });
+    for (let i = 0; i < domElements.length; i += 1) {
+      expect(styled[domElements[i]]).toBeTruthy();
+    }
```
lifeiscontent, line 9: @anatoliy8493 same as above.

packages/styled-components/src/models/StyledComponent.js
```diff
@@ -30,6 +30,7 @@ import { useStyleSheet, useStylis } from './StyleSheetManager';
 import { ThemeContext } from './ThemeProvider';
 
 const identifiers = {};
+const isEnvDevelopment = process.env.NODE_ENV !== 'production';
```
lifeiscontent, line 33: isEnvDevelopment is misleading, pick one:

```suggestion
const isNotProduction = process.env.NODE_ENV !== 'production';
```

```suggestion
const isDevelopment = process.env.NODE_ENV === 'development';
```

packages/styled-components/src/models/StyledNativeComponent.js
```diff
@@ -39,8 +39,8 @@ function useResolvedAttrs<Config>(theme: any = EMPTY_OBJECT, props: Config, attr
   const context = { ...props, theme };
   const resolvedAttrs = {};
 
-  attrs.forEach(attrDef => {
-    let resolvedAttrDef = attrDef;
+  for (let i = 0; i < attrs.length; i += 1) {
```
lifeiscontent, line 42: @anatoliy8493 if you use a for loop you need to use `hasOwnProperty` to check that it belongs to the object and not the prototype. forEach is better here imo.

lifeiscontent (changes requested):

packages/styled-components/src/constructors/styled.js
anatoliy8493: nope, it's needed in `for (var key in obj)`, not in `for | forEach`

packages/styled-components/src/constructors/test/styled.test.js
anatoliy8493: the same as previous

packages/styled-components/src/models/StyledNativeComponent.js
anatoliy8493: the same as previous

packages/styled-components/src/models/StyledComponent.js
anatoliy8493: in case if this app has more than 2 environments, anyway I've fixed it

packages/styled-components/src/constructors/createGlobalStyle.js
anatoliy8493: fixed

kitten (commented):
Sorry, but these are inconsequential changes that we'd usually do as part of a larger effort (of which there aren't too many) or as part of other changes as we'd go.

The reason why I'm pointing this out is that there are too many minor small nits and changes I'd have to comment on before this would get merged, e.g. there's a minification aspect to not changing `process.env.NODE_ENV` inline or extracting it into a variable, which would be subject to build tool tests. (which we'll also refactor in the future as part of build pipeline changes)

The `renderStyles` change is pretty inconsequential to gzip/minification sizes pretty much. Some changes in tests aren't important. Some addition of the eslint comments are either not necessary as they're not from our eslint config or are only necessary because the files have been touched.

Sorry! But keeping the PRs list low is currently what I'm aiming for to get newer changes ready, since we may also be considering a rewrite.

Issue #3297, opened 2020-10-02
Breaks toHaveStyle assertions in v5.2.0


gugiserman:
## Environment

```
## System:
 - OS: macOS Mojave 10.14.6
 - CPU: (8) x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
 - Memory: 1.61 GB / 16.00 GB
 - Shell: 3.2.57 - /bin/bash
## Binaries:
 - Node: 12.18.3 - ~/.nvm/versions/node/v12.18.3/bin/node
 - Yarn: 1.22.5 - ~/.yarn/bin/yarn
 - npm: 6.14.6 - ~/.nvm/versions/node/v12.18.3/bin/npm
 - Watchman: 4.9.0 - /usr/local/bin/watchman
## npmPackages:
 - babel-plugin-styled-components: ^1.10.7 => 1.11.1 
 - styled-components: ^5.1.1 => 5.2.0 
```

## Steps to reproduce
Install v5.2.0 and try to test a React component that renders a StyledComponent with `.toHaveStyle()` from `@testing-library/jest-dom/extend-expect`.

e.g.:

```js
import '@testing-library/jest-dom/extend-expect'
import React from 'react'
import { render } from '@testing-library/react'

it('renders with pointer cursor', () => {
  const { container } = render(<MyComponentWithStyled />)
  expect(container.firstChild).toHaveStyle('cursor: pointer')
})
```

## Expected Behavior

Asserts element has style rule.

## Actual Behavior

Always fail.

## Notes

I upgraded all dependencies today and a lot of tests broke.

I first went to `jest-dom` repository and found a similar issue (https://github.com/testing-library/jest-dom/issues/295#issuecomment-702520302) but after upgrading dependency by dependency I noticed it's actually only when I upgrade `styled-components` to `v5.2.0` that it starts to fail.

barucAlmaguer:
Same problem here (I raised the issue at `jest-dom` library [jest-dom#295](https://github.com/testing-library/jest-dom/issues/295)).

I fixed it temporarily by downgrading to `styled-components@5.1.1`

yannickcornaille:
This problem also occurs in tests using getComputedStyle(). As @barucAlmaguer I fixed it by downgrading to styled-components@5.1.1

Lazyuki:
So it seems that `toHaveStyle` was working up until `5.2.0-test.7`, but `test.8` broke it and started giving the error

[...]

PR #3308, opened 2020-10-13
fix styles not being added when using toHaveStyle()

Lazyuki:
Fixes: #3297 

I checked out every commit to find what caused this, and it seems that [this commit](https://github.com/styled-components/styled-components/commit/2214485f9cf5f0e0ca6b7caffd93351b2e4130d3) and [this commit](https://github.com/styled-components/styled-components/commit/4351e2b6da2d82f1167a73fe1f4aff23ba264178) were the culprit. It seems like having `window` as either `false` or `undefined` makes `jest-dom`'s `toHaveStyle` function fail by not rendering any styles at all. I imagine this has something to do with how jest interacts with the window object.

@probablyup I don't exactly know the reason behind these commits so I hope this doesn't break things elsewhere. I tested the build locally and I can confirm that this will fix `toHaveStyle`

### DIFF
--- a/packages/styled-components/rollup.config.js
+++ b/packages/styled-components/rollup.config.js
@@ -119,7 +119,6 @@ const serverConfig = {
   ],
   plugins: configBase.plugins.concat(
     replace({
-      window: undefined,
       __SERVER__: JSON.stringify(true),
     }),
     minifierPlugin


Lazyuki:
Hi @probablyup @kitten sorry to bother you but this is preventing us and I believe quite a few people from upgrading to 5.2.0 since the tests cannot be run with it.

probablyup:
@Lazyuki you'll need to switch your base branch to legacy-v5

Lazyuki:
Done.

probablyup (approved):
Looking at the code I don't think this will break anything, so long as people don't have a `window` object in their SSR environment that has an `HTMLElement` property. That's the heuristic we use for our `IS_BROWSER` check.

christemple:
> Looking at the code I don't think this will break anything, so long as people don't have a `window` object in their SSR environment that has an `HTMLElement` property. That's the heuristic we use for our `IS_BROWSER` check.

I might be stupid here, but doesn't JSDOM replicate the window in a node environment for jest tests by default?
I just added a console log to my Jest config and it says it does exist:
```js
console.log('HTMLElement in window: ', 'HTMLElement' in window);
=> HTMLElement in window:  true
```
Is this unwanted behaviour @probablyup?

Only reason I'm asking is because my tests are now failing as a result of this change (`v5.2.1`) using react testing library + jest

PR #3337, opened 2020-11-25
Add react-is package as a dependency

araujogui:
React-is is not a dev dependency. React-is tests the arbitrary values at runtime, so it does not make much sense to install react-is as a dev dependency.

### DIFF
--- a/packages/styled-components/package.json
+++ b/packages/styled-components/package.json
@@ -69,6 +69,7 @@
     "babel-plugin-styled-components": ">= 1",
     "css-to-react-native": "^3.0.0",
     "hoist-non-react-statics": "^3.0.0",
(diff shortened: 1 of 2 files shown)


Dennissiq (approved):

kitten (commented):
It's currently a peer dependency, so either we change that, which we'd discuss and wouldn't be subject to a PR quite yet, since at that point I'd also want to remove things like `shallowequal` and discuss either the removal of this completely or bundle the part we need.

But also this isn't correct in that it moves it to `"dependencies"` and doesn't remove it from `"peerDependencies"`, but even then we unfortunately can't just accept small PRs that need further discussions around other changes that'd go along with this

PR #3353, opened 2020-12-16
[bugfix] [v5.2.1] Array.prototype.includes is not compatible with IE 11

gatsbimantico:
COMPLEX_SELECTOR_PREFIX.includes wasn't transpiled,
causing issues in Internet Explorer 11 due lack of compatibility.

The fix is just replacing it for indexOf, in the meantime, the solution is to roll back to v5.2.0

----

**CAUSE**

The transpilation done with babel is not transpiling deep enough, as other Array.prototype.includes are being correctly transpiled, this single one (maybe because of being in the middle of an if statement or something fancy like that) seems to not be transpiled.

**EFFECT**

IE 11 is not compatible with this property of the array object, producing an error.

**TEST**

Compile the project and search for ".includes" in the minified file.

**FIX**

Replacing ".includes" for ".indexOf" for compatibility

**PREVENTION**

Consider adding some bundle validation as part of the Pull Request validation, making sure your bundle transpiles correctly for the browsers supported before merging into master.


probablyup:
@gatsbimantico this needs to be applied to the `legacy-v5` branch. v6 will not support IE

gatsbimantico:
Repeated from legacy-v5 https://github.com/styled-components/styled-components/pull/3397
Closing this PR

Issue #3256, opened 2020-09-04
Error: 'typeOf' is not exported by node_modules/react-is/index.js, imported by node_modules/styled-components/dist/styled-components.esm.js


cieldon32:
my technology stack: rollup + styled-components
when I run build,  I got the error:
Error: 'typeOf' is not exported by node_modules/react-is/index.js, imported by node_modules/styled-components/dist/styled-components.esm.js

and I run "npm list react-is"
@ywfe/ywd@0.0.1 /Users/xiayun/yw/ywd
├─┬ @storybook/addon-actions@6.0.21
│ └─┬ prop-types@15.7.2
│   └── react-is@16.13.1  deduped
├─┬ @types/jest@25.2.3
│ ├─┬ jest-diff@25.5.0
│ │ └─┬ pretty-format@25.5.0
│ │   └── react-is@16.13.1  deduped
│ └─┬ pretty-format@25.5.0
│   └── react-is@16.13.1  deduped
├─┬ antd@4.6.2
│ ├─┬ rc-collapse@2.0.0
│ │ └── react-is@16.13.1  deduped
│ └─┬ rc-util@5.2.1
│   └── react-is@16.13.1  deduped
├─┬ jest@26.4.2
│ └─┬ @jest/core@26.4.2
│   └─┬ jest-config@26.4.2
│     └─┬ pretty-format@26.4.2
│       └── react-is@16.13.1  deduped
├── react-is@16.13.1 
├─┬ styled-components@5.1.1
│ └─┬ hoist-non-react-statics@3.3.2
│   └── react-is@16.13.1  deduped
└─┬ ts-jest@26.3.0
  └─┬ @types/jest@26.0.13
    └─┬ pretty-format@25.5.0
      └── react-is@16.13.1  deduped

my package.json:
`
{
  "name": "@ywfe/ywd",
  "version": "0.0.1",
  "description": "ui framework base on antd",
  "main": "lib/ywd.cjs.js",
  "module": "lib/ywd.esm.js",
  "scripts": {
    "build": "NODE_ENV=production rollup -c",
    "build:type": "tsc -p ./tsconfig.type.json",
    "publish:npm": "npm publish --access public",
    "dev": "rollup -c -w",
    "lint": "NODE_ENV=development eslint . --cache --ext .js,.jsx,.ts,.tsx",
    "lint-fix": "yarn --silent lint --fix; exit 0",
    "test": "jest --coverage",
    "storybook": "start-storybook -p 6006",
[...]

PR #3367, opened 2021-01-06
Fix ESM bundle relying on named imports from react-is

### LINKED ISSUE #3366: Incorrect usage of react-is from ESM bundles
Hiya! There was a previous issue about this (#3256), but it was closed with instructions on how to configure the user's bundler to work around the issue.

I've done some spelunking and found the root cause here:  `styled-components` is trying to use named imports with a CommonJS package. This does sometimes work in Node or a configured bundler, however `react-is` is a package that uses conditional commonjs re-exporting techniques that cause it to fail.

The solution here is actually deceptively simple: `styled-components.*.esm.js` should use a _default_ import for `react-is`, not named imports. Or it could go the extra mile and do interop:

#### Current [styled-components.esm.js](https://unpkg.com/browse/styled-components@5.2.1/dist/styled-components.esm.js) (incorrect):

```js
import{typeOf as e,isElement as t,isValidElementType as n}from"react-is";
```

#### Fixed version:

```js
import reactIs from"react-is";
var { e: typeOf, t: isElement, n: isValidElementType } = reactIs;
```

#### Fancy interop version:

```js
import * as reactIs from"react-is";
var { e: typeOf, t: isElement, n: isValidElementType } = reactIs.default || reactIs;
```

This issue is caused by a configuration mix-up in styled-component's Rollup config: the `rollup-plugin-commonjs` `namedExports` option is used for `react-is`, however this is never applied because the module is marked as external (and thus never processed by the plugin).

I can submit a PR to fix this if there's a willingness to merge!

developit:
This PR fix `styled-components`' ESM bundles so that they don't rely on named imports `react-is`.
From a correctness standpoint, it's generally best to avoid relying on named export inference.
More importantly though, `react-is` doesn't support inference because it conditionally defines `module.exports` based on build-time constants.

The solution I've implemented in this PR is essentially applying the recommendation from #3256 (passing `namedExports` to the commonjs plugin). The difference is that it is applied to the styled-components bundles themselves, rather than requiring everyone include the workaround in their own config.

With this change, the ESM bundles now import `react-is` in a way that works in environments with even the most basic CommonJS-ESM interop:

```js
import e from 'react-is';
const { typeOf, isElement, etc } = e;
```

Fixes #3366.

### DIFF
--- a/packages/styled-components/rollup.config.js
+++ b/packages/styled-components/rollup.config.js
@@ -28,7 +28,44 @@ const esm = {
 const getCJS = override => ({ ...cjs, ...override });
 const getESM = override => ({ ...esm, ...override });
 
+/**
+ * Support external dependencies on CommonJS modules without relying on named export inference
+ * (works similarly to more recent versions of Rollup's commonjs plugin)
+ */
+const cjsExternalPlugin = ({ namedExports }, isExternal) => ({
+  name: 'cjs-external',
+  options(opts) {
+    const external = (isExternal = opts.external || []);
+    if (typeof isExternal !== 'function') {
+      isExternal = id => external.some(e => id.match(e));
+    }
+    opts.external = id => !(id in namedExports) && isExternal(id);
+  },
+  resolveId(id, importer) {
+    if (importer === id + '?cjs') {
+      return { id, external: isExternal(id) };
+    }
+    if (id in namedExports) {
+      return { id: id + '?cjs', external: false };
+    }
+  },
+  load(id) {
+    if (id.endsWith('?cjs')) {
+      id = id.slice(0, -4);
+      return `
+        import $ from ${JSON.stringify(id)};
+        export const ${namedExports[id].map(v => `${v}=$.${v}`)};
+      `;
+    }
+  },
+});
+
 const commonPlugins = [
+  cjsExternalPlugin({
+    namedExports: {
+      'react-is': ['isElement', 'isValidElementType', 'ForwardRef', 'typeOf'],
+    },
+  }),
   flow({
     // needed for sourcemaps to be properly generated
     pretty: true,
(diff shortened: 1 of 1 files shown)


kitten:
@probablyup Is it worth it to finally drop `react-is` as we've discussed a while ago? I don't see any code that is explicitly relying on it anymore per se.

There's one section for development warning purposes that isn't effective anymore, i.e. it looks flawed since it checks for an element (`typeof x === 'object'`) while the check is for functions: https://github.com/styled-components/styled-components/blob/a808f40156e62db5c92b06d362e200e5c172a897/packages/styled-components/src/utils/flatten.js#L70-L77

The other usage checks for the input type to `constructWithOptions` and essentially just does `typeof type === 'string' || typeof type === 'function'`: https://github.com/styled-components/styled-components/blob/72a6f642c3dedbcee91a2b5d76f96b93f345b86f/packages/styled-components/src/constructors/constructWithOptions.js#L14-L16

It may be effective in theory since we're trying to include `React.memo` and `React.forwardRef` but `React` itself would eventually throw an error about these kinds of mistakes anyway, and I don't see the cost of `react-is` and the problems it causes as worth it in relation anymore 😅

packages/styled-components/rollup.config.js
```diff
@@ -28,7 +28,44 @@ const esm = {
 const getCJS = override => ({ ...cjs, ...override });
 const getESM = override => ({ ...esm, ...override });
 
+/**
+ * Support external dependencies on CommonJS modules without relying on named export inference
+ * (works similarly to more recent versions of Rollup's commonjs plugin)
```
probablyup, line 33: >  * (works similarly to more recent versions of Rollup's commonjs plugin)

Should we just update the CJS plugin?

developit: You could, but that would require updating Rollup 2 major versions forward, which is a pretty major change (everything broke when I tried).

probablyup:
Dropped this dep in https://github.com/styled-components/styled-components/commit/8165cbe994f6f749236244f6f7017c2f0b9afcfe :)

Issue #3154, opened 2020-05-27
Move `react-is` to regular `dependencies`


JalilArfaoui:
Not all React application have to use `react-is`, so I don’t think developer should have to manually manage this dependency in final application …

I suggest that it can be moved to a regular dependency instead of a peerDependency.

probablyup:
The goal was primarily to encourage deduplication at the package manager level, which peerDependencies helps to enable. If we set a pretty permissive range though it should be fine.

JalilArfaoui:
I’m closing this now that https://github.com/styled-components/styled-components/pull/3371 has been done !

antoinewolff:
I reopen this since react-is is still in peer dependencies in the version 5. Can we please move it to regular dependencies in v5 ?

nickmccurdy:
Note that it's possible to have a package in `dependencies` **and* `peerDependencies` to provide a default version while still allowed it to be overriden.

quantizor:
Closing — this issue is stale (2+ years without resolution) and relates to an older version or has been superseded by v6.x changes. If this is still reproducible on the latest v6.x release, please open a new issue with a minimal reproduction.

PR #3371, opened 2021-01-08
Remove react-is from styled-components

kitten:
Resolves #3367 #3366 #3154

## Summary

Previously we've been known to include checks that go above and beyond to give meaningful output in nonsensical / non-standard situations. Included in one of them is our use of `react-is`, which used to be much more impactful (with `forwardRef` checks), but now boils down to checks whether an input is unexpected with appropriate warnings.

The impact that `react-is` is having on users in certain environments and the confusion it causes isn't worth it anymore in my estimation, and all checks can either be removed without replacement, since they're very besides what is stated in the docs as normal usage, or are easily replaced.

## Changes

- Remove `react-is` from all dependencies, `peerDependencies`, and `rollup.config.js`
- Replace check for elements as output of non-StyledComponents / pure function components in `flatten.js`
- Remove check from `constructWithOptions` (i.e. `styled(tag)`) that includes `isValidElementType`
- Replace check from `isPlainObject` for elements


packages/benchmarks/package.json
```diff
@@ -23,7 +23,6 @@
     "react-jss": "^8.6.1",
     "react-native-web": "^0.11.4",
     "reactxp": "^1.3.3",
-    "styled-components": "link:../styled-components",
```
probablyup, line 26: Why did this get removed @kitten?

packages/styled-components/src/utils/flatten.js
```diff
@@ -64,20 +62,9 @@ export default function flatten(
...
-          `${getComponentName(
-            chunk
-          )} is not a styled component and cannot be referred to via component selector. See https://www.styled-components.com/docs/advanced#referring-to-other-components for more details.`
-        );
-      }
-
-      return flatten(result, executionContext, styleSheet, stylisInstance);
-    } else return chunk;
+    return isStatelessFunction(chunk) && executionContext
+      ? flatten(chunk(executionContext), executionContext, styleSheet, stylisInstance)
```
kitten, line 66: _Note:_ This is the only "purposeful regression" that is reasonably unexpected here, since it removes a very obvious warning from the library. I'd argue it's not really a useful one, since it'll fail in obvious ways, but since interpolations can now return objects there isn't really an alternative for this. We could reuse the check from `isPlainObject` here though if we wanted to bring it back later.

packages/styled-components/src/constructors/constructWithOptions.js
```diff
@@ -11,7 +10,11 @@ export default function constructWithOptions(
   tag: Target,
   options: Object = EMPTY_OBJECT
 ) {
-  if (!isValidElementType(tag)) {
+  // We trust that the tag is a valid component as long as it isn't nullish
+  // Typically the tag here is a string or function (i.e. class or pure function component)
+  // However a component may also be an object if it uses another utility, e.g. React.memo
+  // React will output an appropriate warning however if the `tag` isn't valid
+  if (tag == null) {
```
probablyup, line 17: `!tag` should work just as well here and is smaller

packages/styled-components/src/utils/flatten.js
mxstbr: > since it'll fail in obvious ways

I don't feel like this is an obvious fail... :confused:

packages/benchmarks/package.json
kitten: I explicitly removed it, since it's not needed in a monorepo. Yarn Workspaces will create a symlink for `styled-components` (the package; not the repo) in the root `node_modules` folder.

packages/styled-components/src/utils/flatten.js
```diff
@@ -64,20 +62,9 @@ export default function flatten(
...
   if (isFunction(chunk)) {
-    if (isStatelessFunction(chunk) && executionContext) {
-      const result = chunk(executionContext);
-
-      if (process.env.NODE_ENV !== 'production' && isElement(result)) {
-        // eslint-disable-next-line no-console
-        console.warn(
-          `${getComponentName(
-            chunk
-          )} is not a styled component and cannot be referred to via component selector. See https://www.styled-components.com/docs/advanced#referring-to-other-components for more details.`
```
probablyup, line 75: I'm not sure I agree with fully removing this check. We have a util for detecting styled components that could be used here to get a similar result.

packages/styled-components/src/utils/isPlainObject.js
```diff
@@ -1,8 +1,8 @@
 // @flow
-import { typeOf } from 'react-is';
 
 export default (x: any): boolean =>
   x !== null &&
   typeof x === 'object' &&
   (x.toString ? x.toString() : Object.prototype.toString.call(x)) === '[object Object]' &&
-  !typeOf(x);
+  /* check for reasonable markers that the object isn't an element for react & preact/compat */
+  !(x.$$typeof && (x._owner || (x.__v && x.__o)));
```
probablyup, line 8: This needs tests

packages/styled-components/src/constructors/constructWithOptions.js
kitten: oap true

kitten:
The check replacement for `isElement` has been replaced in `flatten.js` by making a less specific check (`isPlainObject`) more specific by checking for all allowed values when `typeof result === 'object'`

packages/styled-components/src/utils/test/flatten.test.js
```diff
@@ -128,23 +128,6 @@ describe('flatten', () => {
     expect(flatten(['foo', func], { bool: false })).toEqual(['foo', 'static', 'baz']);
   });
 
-  it('throws if trying to interpolate a normal React component', () => {
```
probablyup, line 131: This test should stick around

probablyup (approved):
Approved pending the final comment

probablyup:
@kitten oddly this change increased bundle size by 1kB... it looks like `hoist-non-react-statics` still uses it so removing the rollup optimization actually makes things heavier

kitten:
@probablyup Interesting! Maybe we'll have to have a look at whether `hoist-non-react-statics` is trying to be "too accurate" compared to what we're using it for.

packages/styled-components/src/utils/isPlainObject.js
```diff
@@ -1,8 +1,8 @@
 // @flow
-import { typeOf } from 'react-is';
 
 export default (x: any): boolean =>
   x !== null &&
   typeof x === 'object' &&
   (x.toString ? x.toString() : Object.prototype.toString.call(x)) === '[object Object]' &&
-  !typeOf(x);
+  /* check for reasonable markers that the object isn't an element for react & preact/compat */
+  !(x.$$typeof && ('_owner' in x) || x.__v);
```
developit, line 8: This avoids relying on internal/undocumented properties:

```suggestion
  !('props' in x && (x.$$typeof || x.constructor === undefined));
```

developit (approved):
LGTM, with the one suggestion about a more reliable object brand check.

probablyup:
thanks @developit!

Kitton:
any chance to get this merged?

jasonwilliams (approved):

jasonwilliams:
Sadly i don't know why the build is failing

PR #3393, opened 2021-02-02
PR: Implement CSP nonce handling like in css-in-js library

aplavsa:
I have implemented a client side nonce support that was requested in this issue #2363 . I did not use the approach that was suggested in the issue, but rather I have implemented a css-in-js library approach.

Basically, when there is a csp meta tag in the head tag of the document like this:
```html
<meta property="csp-nonce" content="some nonce" />
```
this nonce gets added to the style tag of **styled-components** library.

Basically, I have recreated this approach on https://cssinjs.org/csp/?v=v10.5.1 

It would be nice if you could merge this feature into the version v5 of your library and possibly also in the future version of your library.

BR

Andrija Plavsa


JonShort:
This would be a great addition for clientside-rendered apps! 🎉 

I shimmed your work into a sandbox to test it out, and I wasn't able to get it to work with `textContent`; using `content` works fine though 😄

```diff
-      return cspNonceMetaTag && cspNonceMetaTag.textContent;
+      return cspNonceMetaTag && cspNonceMetaTag.content;
```

Heading will be red if the nonce is working:
https://codesandbox.io/s/shim-styled-components-nonce-u3c70?file=/src/App.js

aplavsa:
Yes, you are right. Actually, I used .content in my version of library which I use but when I tried to do a flow check, it was complaining about .content property so I've set it up to .textContent. I will fix it by ignoring the flow rule I have problem with and that should do the trick. But thanks for pointing this out to me 😉

aplavsa:
I have committed all the latest changes and it should work now. 🚀

TheoMer:
@aplavsa How do I access the correct version with the committed changes? The current version 5.2.1, which I have, does not have it.

aplavsa:
You can't until they merge it with the v5 branch and do a new release as a new version. Until then, you can clone my branch from this pull request and build it on your own.

bbenezech:
Please, this patch would make a big difference for non-webpack users.
The current solution `__webpack_nonce__` is `webpack` only.

aplavsa:
I have merged the latest legacy-v5 branch into the PR branch. Can you now try to merge into the upstream project?

kitten:
See: https://github.com/styled-components/styled-components/issues/2363#issuecomment-895464410

Issue #3399, opened 2021-02-09
Not being able to style components wrapped in forwardRef


andormade:
First of all I'm not 100% sure if this is a bug in _styled-components_, but my problems started with updating to v5. Well, more specifically when this PR was released: #2934 

It introduces a new dependency called _hoist-non-react-static_ - which I already had in my project, but it was an older version: 2.5.5. For some reason styled-components or yarn accepted it as a compatible version, which caused a lot of weird bugs like not being able to style components wrapped in forwardRef.

After installing version 3.3.0 from _hoist-non-react-static_ the bugs disappeared.

I would suggest to replace the caret with a greater or equal sign in the package.json, because according to the [docs](https://www.npmjs.com/package/hoist-non-react-statics#compatible-react-versions) of _hoist-non-react-static_ it's crucial to have at least version 3 to have support for forwardRef. What do you think?

kitten:
Resolved by https://github.com/styled-components/styled-components/commit/8165cbe994f6f749236244f6f7017c2f0b9afcfe

PR #3400, opened 2021-02-09
Enforce greater or equal version of hoist-non-react-statics

andormade:
Fixes issue #3399


probablyup:
@andormade thanks, will merge this. You might want to make the same PR against `legacy-v5` branch as v6 is currently unreleased.

PR #3422, opened 2021-03-06
feat: macro can customize importModuleName

agriffis:
If `styled` should come from a wrapper instead of directly from
styled-components, this enables users of the macro to customize the
import by configuring in `packages.json`:

    "babelMacros": {
      "styledComponents": {
        "importModuleName": "my-styled-components"
      }
    }

Previously similar functionality landed in
babel-plugin-styled-components so users of the plugin could use
a wrapper at that level, but it wasn't available to users of the macro.
This builds on that work by passing the customized module name through
to the plugin. Consequently we also update the dependency to ensure the
plugin supports `topLevelImportPaths`

See https://github.com/styled-components/babel-plugin-styled-components/commit/325167bb570362adfb5ff78aa84c460adeae32ac

Fixes https://github.com/gregberge/xstyled/issues/44

### DIFF
--- a/packages/styled-components/package.json
+++ b/packages/styled-components/package.json
@@ -66,7 +66,7 @@
     "@babel/traverse": "^7.4.5",
     "@emotion/is-prop-valid": "^0.8.8",
     "@emotion/unitless": "^0.7.4",
-    "babel-plugin-styled-components": ">= 1",
+    "babel-plugin-styled-components": ">= 1.12.0",
     "css-to-react-native": "^3.0.0",
     "hoist-non-react-statics": "^3.0.0",
     "shallowequal": "^1.1.0",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 - Upgrade to stylis v4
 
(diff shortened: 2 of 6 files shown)


packages/styled-components/package.json
```diff
@@ -66,7 +66,7 @@
     "@babel/traverse": "^7.4.5",
     "@emotion/is-prop-valid": "^0.8.8",
     "@emotion/unitless": "^0.7.4",
-    "babel-plugin-styled-components": ">= 1",
+    "babel-plugin-styled-components": "^1.12.0",
```
probablyup, line 69: I'd prefer to continue using the `>=` notation so we don't have to update the library here if v2 comes out that doesn't contain incompatible changes with styled-components core

agriffis: np, will force push that change

agriffis:
@probablyup That should be all set now. I force-pushed twice because I hadn't run yarn the first time, so `yarn.lock` wasn't updated.

Would you mind cherry-picking this to `legacy-v5`? Or do you prefer another PR?

probablyup:
> @probablyup That should be all set now. I force-pushed twice because I hadn't run yarn the first time, so `yarn.lock` wasn't updated.
> 
> Would you mind cherry-picking this to `legacy-v5`? Or do you prefer another PR?

Another PR please

probablyup (approved):

agriffis:
> Another PR please

See #3432, thanks!

PR #3080, opened 2020-03-30
Add support for animations in objects

dvingo:
This PR adds support for animations in styled objects.
Example usage:

```jsx
const animation = keyframes({
  from: { opacity: 0 },
  to: { opacity: 1 }
})

const Animated = styled.div({ animation: css`${animation} 3s` })

const Animated2 = styled.div(({time = 1}) => ({
  animation: css`${animation} ${time}s infinite` 
}))
```
This addresses issue:
https://github.com/styled-components/styled-components/issues/2561

The implementation detects arrays in value position during `objToCssArray` expansion
and mutually recurses into `flatten` to further transform nested forms of styles.

---
update:
After seeing the feedback here (and a similar implementation):
https://github.com/styled-components/styled-components/pull/2612

I updated this implementation to add a flag on arrays returned by `css` to determine if we should expand the nested styles within `objToCssArray`.

PR #3426, opened 2021-03-10
Support use of css tagged template inside style objects

roginfarrer:

```js
const animation = keyframes({
  from: { opacity: 0 },
  to: { opacity: 1 }
})

}))
```

This addresses issue:
#2561

### DIFF
--- a/packages/styled-components/src/constructors/css.js
+++ b/packages/styled-components/src/constructors/css.js
@@ -6,10 +6,22 @@ import isFunction from '../utils/isFunction';
 import flatten from '../utils/flatten';
 import type { Interpolation, RuleSet, Styles } from '../types';
 
+/**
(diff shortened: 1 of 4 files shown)


roginfarrer:
Hi @probablyup, I saw that you looked at the original PR. Is there anything about this implementation that you think should change?

roginfarrer:
Just wanted to bump this PR since it's been a few weeks. @probablyup Sorry to ping you if you're not the right person, just trying to resolve this (even if this doesn't ultimately get added to the library). Let me know if I should reach out to someone else.

roginfarrer:
Since this PR is aimed at `main`, which was just converted to TS, I'm closing and reopening this PR aimed at `legacy-v5`

roginfarrer:
Reopened in #3649

PR #3436, opened 2021-03-19
feat: pass elementToBeCreated to shouldForwardProp

agriffis:
Currently it's not possible to replicate styled-components' internal logic in `shouldForwardProp`, because it doesn't have all the information. It only gets the prop `key`, not `elementToBeCreated`, and so it can't properly decide in a generic manner whether to forward props.

This PR upgrades `shouldForwardProp` to include `elementToBeCreated` as a third parameter. This should be backward-compatible with existing code because it only passed two parameters previously.

This is especially for when a styled component is called with `as`, because that makes the decision of whether to forward props more dynamic than is currently possible to express.

This PR should probably also update some documentation somewhere. I can make an additional commit if it otherwise looks good to y'all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 - Upgrade to stylis v4
 
-- Enable users of the babel macro to customize the styled-components import with `importModuleName` (see [#3422](https://github.com/styled-components/styled-components/pull/3422))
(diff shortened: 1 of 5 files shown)


probablyup:
Makes sense to me, can you add a changelog entry?

agriffis:
@probablyup I just force-pushed with the changelog entry.

I actually replaced the babel macro changelog entry which doesn't need to be duplicated between 5.2.2 and the unreleased section.

probablyup (approved):

Issue #3437, opened 2021-03-22
Provide `styled` as a named export


JacobLey:
Currently `styled` is only available as the default export: https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/index.js#L5

In transpiled/commonJS environments that works as expected, but breaks server-side ESM usage of styled components.

NodeJS does not support CommonJS modules have _both_ a default and a named export: https://nodejs.org/api/esm.html#esm_import_statements

So for styled components to work server side instead of:
```
import styled from 'styled-components';
```

We must write:
```
import styledWrapper from 'styled-components';
styled = styledWrapper.default;
```
Which is confusing, and not a viable option for code that should run in both browser and server.

Adding `styled` as a named export allows environment agnostic access:
```
import { styled } from 'styled-components';
```

Happy to open a PR, should be a one liner.
But wanted to explain reasoning as well

kitten:
We already have an ESM build so I don't see why the solution would be to provide a redundant export rather than to add the `"exports"` field to the `package.json` 🤔

JacobLey:
The `exports` field IMHO is great for things like restricting what files can be imported, and changing behavior based on node/browser environments.
While it does support ESM vs CJS importing, that would actually break things further (in my case).

My use case is I use styled components in a NextJS environment. NextJS uses CJS internally, so there is nothing I can do to force them to import ESM version. My code however is pure ESM (I have a custom Webpack which stops NextJS from its usual CJS transpiling), which would result in an ESM import.
That would mean two different versions of styled-components would get imported into every project, with all the headaches that come along with that.

Being able to always import the CJS version is the best way to ensure I only get one version of styled-components, but then run into the issue described in this ticket

brandonryan:
This is also a problem when trying to import /native

kitten:
@JacobLey I think you misunderstood me, sorry for the confusion. Default exports are a thing. Next.js uses Webpack and does enable support of ESM, so I'm not sure what's going on here. However, adding a redundant export won't resolve the underlying issue you're probably experiencing here. But without details of the underlying issue, I can't really pin down what's going on here.

So, in other words, given that I don't think it makes sense to do this, and without another course of action, I'll have to close this, sorry!

JacobLey:
Next.js does not use ESM. It uses webpack to transform ESM compliant code into CJS. Which is nice if you go all in on NextJS + CJS, but that is not my case. I want to continue using _actual_ ESM where possible.

[...]

PR #3438, opened 2021-03-22
feat: Add styled as named export

JacobLey:
closes https://github.com/styled-components/styled-components/issues/3437

Added to the files listed as inputs [here](https://github.com/styled-components/styled-components/blob/master/packages/styled-components/rollup.config.js)

With exception to:
* standalone - it seems intentional there is only one default export and everything else is attached as a name?
* macro - this seems like a babel-only file, which is swapped for the real `styled-components` at build time?


samarmohan (approved):

samarmohan (changes requested):
fix conflicts first

JacobLey:
Fixed merge conflicts

samarmohan (approved):
nice

JacobLey:
Just a bit, running things like `import('./dist/styled-components.cjs.js').then(x => console.log(x.default.default === x.styled)) // true` locally.
And that worked as expected.

I tried installing the package locally with my current project that was having the related issue and started getting `Invalid hook call. Hooks can only be called inside of the body of a function component` errors. (This is a NextJS project if relevant)

So I tried using the `main` branch and ended up getting the same errors. So I don't think the error is related to my changes, but I don't really have great test proof my changes work as expected (generally speaking my changes are small so I _expect_ them to work, but looking for solid proof)

Is this a known/expected issue? What is best way to test locally?

JacobLey:
Also I see this project is moving more over to typescript, which is great. But looks like that is still for internal usage for now?

So once this is merged I will followup with a PR to https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/styled-components to declare the named export

samarmohan:
Your error that you get, `Invalid hook call. Hooks can only be called inside of the body of a function component `, happens a lot. Check your hooks and stuff. That’s not a styled components issue.

JacobLey:
I only get this issue when installing my local version of styled components though. So I suspect I am installing it wrong, what is the proper way to install it locally?

I think the issue stems from styled-components using its own instance of React (which only exists in a locally-installed dev environment) rather than my version of React. So the overlap causes the issue.

kitten:
See comments on original issue. Sorry for closing!

PR #3452, opened 2021-04-06
Fix - add textPath svg element

liorEhrlich:
The svg element textPath was missing from the list of dom elements. 
Added it to the domElements, and to v4's domElements.


probablyup:
@liorEhrlich thanks, could you change this PR to target the `legacy-v5` branch and add a changelog entry?

liorEhrlich:
@probablyup Thanks for reviewing my PR.
I've added a changelog entry, but had some issues changing the parent branch of the feature branch.
Moved the code to: https://github.com/styled-components/styled-components/pull/3461

PR #3469, opened 2021-04-21
Support use of css tagged template inside style objects

roginfarrer:

```js
const animation = keyframes({
  from: { opacity: 0 },
  to: { opacity: 1 }
})

}))
```

This addresses issue:
#2561


roginfarrer:
@probablyup let me know if you think this should just be slated for v6, happy to reopen a PR targeted at the main branch

probablyup:
@roginfarrer if you wouldn't mind opening a PR against main as well that would be great

probablyup (approved):
Ace, thank you

PR #3475, opened 2021-04-27
feat: allow using multiple style tags for rule updates

s0ber:
First of all, huge thanks for an amazing library.

We've encountered the performance issue caused by a Chrome quirk when maintaining the rules inside single `style` tag is causing unnecessary style recalculations (comparing to adding and removing many granular `style` tags). This issue is very noticable on large pages with many nodes.

Hopefully Chrome team would fix an issue some day. But for now it will be great if user would have an option to chose the strategy.

### DIFF
--- a/packages/sandbox/src/AppWithManySheets.js
+++ b/packages/sandbox/src/AppWithManySheets.js
@@ -1,0 +1,51 @@
+import React, { useState, memo } from 'react';
+import styled, { StyleSheetManager } from 'styled-components';
+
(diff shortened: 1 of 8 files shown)


kitten:
I'd love to actually make this a default since it could fit quite nicely into the existing sheet abstraction with prior code ✌️

I suppose it'd be useful to investigate when a performance cliff occurs in the size of the sheet and what the maximum number of rules is so we could tweak this a bit more

s0ber:
@kitten Thank you a lot for taking a look!
I have reopened this PR from our organization account so that other developers from our team would have access to the code. Also, I've added the small screencast that visualizes the problem on youtube:
https://www.youtube.com/watch?v=jxCWDhwG2o0 - here it is.

Closing this one. Please let us know what we should do to help moving this fix forward!

PR #3477, opened 2021-04-29
feat: allow using multiple style tags for rule updates

s0ber:
First of all, huge thanks for an amazing library.

We've encountered the performance issue caused by a Chrome quirk when maintaining the rules inside single style tag is causing unnecessary style recalculations (comparing to adding and removing many granular style tags). This issue is very pronounced on large pages with many nodes.

Hopefully Chrome team would fix an issue some day. But for now it will be great if user would have an option to chose the strategy.

**A small screencast where I explain the problem:**
https://www.youtube.com/watch?v=jxCWDhwG2o0

### DIFF
--- a/packages/sandbox/src/AppWithManySheets.js
+++ b/packages/sandbox/src/AppWithManySheets.js
@@ -1,0 +1,51 @@
+import React, { useState, memo } from 'react';
+import styled, { StyleSheetManager } from 'styled-components';
+
+const colors = ["bda779", "5b2ae8", "f3912d", "85120b", "92414d", "c3866e", "2c8bda", "e1fbb", "2ce471", "ed91c6", "e38f2f", "d31016", "176b1e", "3be33d", "b5c51a", "c496c5", "b5013a", "b0e764", "fa0f7b", "32e9de", "e81de3", "260e57", "7c2114", "892e29", "5acf1d", "ffa963", "677387", "96e65f", "729a2c", "513689"]
+
+const Header = styled.div`
+  height: 50px;
+  padding: 30px;
+`
+
+const ColoredComponents = colors.reduce((components, color) => {
+  const component = styled.div`
+    min-height: 20px;
+    padding: 30px;
+    background-color: #${color};
+  `
+  return { ...components, [color]: memo(component) }
+}, {})
+
+const ComponentsList = memo(({ children, reversed }) => (
+  <>
+    {(reversed ? colors.reverse() : colors).map(color => {
+      const ColoredComponent = ColoredComponents[color];
+
+      return <ColoredComponent key={color}>{children}</ColoredComponent>
+    })}
+  </>
+));
+
+const App = () => {
+  const [shouldShowHeader, toggleHeader] = useState(false)
+
+  return (
+    <StyleSheetManager useMultipleStyles>
+      <>
+        <button type='button' onClick={() => toggleHeader(!shouldShowHeader)}>
+          Toggle Header
+        </button>
+        <br />
+        <ComponentsList>
+          {shouldShowHeader && <Header>I AM HEADER</Header>}
+          <ComponentsList reversed>
+            <ComponentsList reversed />
+          </ComponentsList>
+        </ComponentsList>
+      </>
+    </StyleSheetManager>
+  )
+};
+
+export default App;
--- a/packages/sandbox/src/browser.js
+++ b/packages/sandbox/src/browser.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import { hydrate } from 'react-dom';
 
-import App from './App';
+import App from './AppWithManySheets';
 
 const render = Component => {
   hydrate(<Component />, document.querySelector('#react-root'));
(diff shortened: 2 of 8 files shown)


probablyup:
@s0ber hey there! we're not adding anything new to v5 at this point, but I'll add this to my list to consider for v6

kitten:
This is really interesting. Just got around to having a look! Also, thanks for submitting that video, since that really hammers the point home. What I'd say here however is that creating a tag per group is extremely taxing. In other words, it'll likely create the same problem but moving the cost someplace else. The reason for that is that while we'd then limit the impact of a total amount of rules, we'd be adding the cost of inserting a tag for each group. Since with each created component we have a new group that does add an excessive amount of stress in another place of our implementation, if a high amount of components is created (which isn't unlikely). It's also potentially wasteful given that we don't _need_ that many tags.

My suggestion is that we may have to look into this more. I'd like to highlight that I said we don't _need_ that many tags without knowing how many rules are too many. I'd actually opt for the least amount of tags possible rather than even a conservative amount. Hence, we may want to simply put some numbers here. How many rules have to be inserted until Chrome recalculates a new rule for longer than expected? If it's a performance cliff then it'll be relatively easy to avoid with other means, i.e. without creating a tag per component.

s0ber:
@kitten Thank you for taking a look! I cannot agree more, many tags look very ugly. Also they are not really required because in Safari and in Firefox both strategies provide same good performance. It is a Chrome-specific quirk that theorhetically could be fixed at any time. We discovered it almost by accident and we would never though that the performance issue would come from there.

But I want to add a few words in defence of the "many tags" strategy:
1) This is the strategy that is used by the JSS library:
https://github.com/cssinjs/jss
We are using jss library for quite a while and never had any issues with it. Also it was the main styling engine for the https://github.com/mui-org/material-ui project for quite some time.
2) We already using the patch from this PR in our project for some time and it works good!

Considering those facts:
- the Chrome quirk could be fixed at any moment,
- the patch for "multiple tags" strategy is pretty simple and the code is even simpler than for keeping everything under single tag
- not many apps would suffer from the issue, only then ones with very heavy UIs (the problem is pronounced here the most).

I think that it's not worth investing time into some complex solution that could become excessive at any moment. IMHO `styled-components` should just provide an option to "opt-in" for using alternative strategy. And it could be easily removed once Chrome will fix their behavior.

kitten:
I'm afraid issues like these stick around for much longer than is often expected though 😅 Chrome also exhibits a unique behaviour in relation to any assets referred to in CSS. While CSSOM should prevent it from re-evaluating assets when new rules are inserted Chrome often does so anyway. And quirks like these stick around for a while.

I think our approach is already complex enough though to warrant the extra look. The reason why I think it's worth it is that we may be missing something if we only circumstantially fix the problem and I'm also not too opposed to refactoring the sheet in the future and pulling it into a separate package.

I'd also add that it's not quite just an implementation detail to have multiple style tags. We consciously chose to go back down to a single one and we have to be careful to consider what happens on the flip-side when we increase this again.

Hence, I'm very much inclined to say that I'd love to know more about this performance cliff first and how it relates to other timings. If it's indeed a performance cliff then we can probably attach an exact cost and number to it. This would also help if we were to report it to the Chromium team.

s0ber:
@kitten I think I get the picture now! The lib is mature enough and there are reasons why things implemented the way they are. So it's up to you folks! My personal suggestion is to provide both options, the best one and alternative one that could be used at own risk 😉 

Anyway, the correlation that I've noticed so far is that it recalculates styles for all the rules inside the `style` tag. We can just check the "Elements affected" property in the profiler and see how it changes. So far the more rules there are inside the style tag the more elements are affected and the more time it takes to recalculate styles. It seems there is no real cliff there, just a linear corelation. Also it is the same both for CSSOM modifications and for the text strings.

kitten:
That may actually be related to the assets problem then 👀 I suppose if we find out why it'd still recalculate all styles then we'd find out what's going on in general, so I'll dig through some old issues to see whether we had a Chromium report for that.

Otherwise, if it's linear then you're definitely right. At that point in Chrome we'd just want to balance the cost of adding tags with the cost of recalculating all styles.

kitten:
This is really interesting, but in looking into this I've realised that this issue is quite deep. Together with the issue of "reloading assets" it warrants more attention. Also, creating new style tags per component isn't quite solving this as we pay a fixed cost of adding style tags, which in my tests was still significantly higher than acceptable. I've submitted a Chromium issue, but I'm not really getting my hopes up that it'll be addressed: https://bugs.chromium.org/p/chromium/issues/detail?id=1216451

s0ber:
@kitten Thank you a lot for propagating this to chromium team! Also if you'll have a chance can you please share the problems of injecting additional style tags? Maybe just throw some links to the old issues. Just so that we and any other wondering people will be aware. In our project we have a few hundreds of tags on the page :D

kitten:
@s0ber You can modify the test case that's attached to the issue to see some results. What I've observed is that we'd be recalculating styles a lot less but the time is then spent on the new style tag, if one is inserted. So for styles in the same tag, the issue still applies, and for new tags that seems to take longer than expected (1/3 of the time? still rather slow looking IMO) Anyway, this seems like there's really a foundational WebKit performance issue at work.

PR #3497, opened 2021-05-18
Support use of css tagged template inside style objects (v6)

roginfarrer:
See #3469 for v5.

```js
const animation = keyframes({
  from: { opacity: 0 },
  to: { opacity: 1 }
})

}))
```

This addresses issue:
#2561

### DIFF
--- a/packages/styled-components/src/constructors/css.ts
+++ b/packages/styled-components/src/constructors/css.ts
@@ -5,6 +5,18 @@ import interleave from '../utils/interleave';
 import isFunction from '../utils/isFunction';
 import isPlainObject from '../utils/isPlainObject';
 
+/**
+ * Used when flattening object styles to determine if we should
+ * expand an array of styles.
+ */
+const addTag = (arg: ReturnType<typeof flatten> & { isCss?: boolean }) => {
+  if (Array.isArray(arg)) {
+    // eslint-disable-next-line no-param-reassign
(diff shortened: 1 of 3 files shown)


probablyup (approved):
LGTM, not sure why the tests haven't run yet

packages/styled-components/src/constructors/css.ts
```diff
@@ -5,14 +5,28 @@ import interleave from '../utils/interleave';
...
+/**
+ * Used when flattening object styles to determine if we should
+ * expand an array of styles.
+ */
+const addTag = (arg: ReturnType<typeof flatten> & { isCss?: boolean }) => {
+  if (Array.isArray(arg)) {
+    // eslint-disable-next-line no-param-reassign
+    arg.isCss = true;
+  }
+  return arg;
```
juboba, line 17: Is this mutation required? or is it possible to:

```suggestion
  return Array.isArray(arg) ? { ...arg, isCss: true} : arg;
```

roginfarrer:
I pushed an empty commit, and that got the tests running

packages/styled-components/src/constructors/css.ts
probablyup: In this case I think the OP's code is probably a little more performant since it isn't introducing additional objects into a hot path

Issue #2974, opened 2020-01-20
[Micro Frontend] Duplicate class name if there is multiple styled-components


soulwu:
I use [qiankun](https://github.com/umijs/qiankun) to make my micro-frontend app.
I use `styled-components` for css-in-js.

If I mount multiple sub apps at the same time, and these sub apps use styled-components respectivly, the generated class name would be conflict.

Is there a way to config different seed to generate class name?

probablyup:
Do you have a live example of this? The odds of a styling collision are very low unless there are multiple instances of styled-components on the page at once and both are rendering the same components in separate subtrees.

casperle:
@soulwu Maybe the this comment can help you https://github.com/styled-components/styled-components/issues/1789#issuecomment-577704969

We have similar issue, but with an SSR setup which looks like this:
![image](https://user-images.githubusercontent.com/8900497/72994597-ab659880-3df7-11ea-9779-7f1a4faedae4.png)

The client calls the orchestrator which calls two different endpoints where react components (fragments) are rendered. The orchestrator puts it together and sends it to the client where every fragment gets hydrated. Both fragments are wrapped into `ThemeProvider`.
On the client side we have only one instance of the `styled-components` because no warning is provided.
With this setup we still get the following error:
![image](https://user-images.githubusercontent.com/8900497/72995314-fcc25780-3df8-11ea-9e7f-4c0da9b28852.png)

Also babel plugin is used with SSR setup.

Do you have any idea what can cause this?

soulwu:
> @soulwu Maybe the this comment can help you [#1789 (comment)](https://github.com/styled-components/styled-components/issues/1789#issuecomment-577704969)
> 
> We have similar issue, but with an SSR setup which looks like this:
> ![image](https://user-images.githubusercontent.com/8900497/72994597-ab659880-3df7-11ea-9779-7f1a4faedae4.png)
> 
> The client calls the orchestrator which calls two different endpoints where react components (fragments) are rendered. The orchestrator puts it together and sends it to the client where every fragment gets hydrated. Both fragments are wrapped into `ThemeProvider`.
> On the client side we have only one instance of the `styled-components` because no warning is provided.
> With this setup we still get the following error:
> ![image](https://user-images.githubusercontent.com/8900497/72995314-fcc25780-3df8-11ea-9e7f-4c0da9b28852.png)
> 
> Also babel plugin is used with SSR setup.
> 
> Do you have any idea what can cause this?

[...]

PR #3503, opened 2021-05-26
feat: Add prefix to all hash class to avoid collision when needed

rafael-cineis-ifood:
## What?
Created an environment variable `SC_CLASS_PREFIX` to add a prefix to all hash classes generated, including the ones that contain the style itself to avoid unexpected class collision.

## Why?
To avoid classes collision when we have multiple styled-components loaded at the same time, usually it happens when working with a micro-frontend structure and we have 2 micro-frontend loading the styled-components.

There is the option to use the offical babel plugin to add the [namespace option](https://styled-components.com/docs/tooling#namespace), but as the plugin only add the prefix in the `sc-???` class, it will not shield us from class collision between libraries used in common by multiple micro-frontends because usually the other classes (the ones with the style itself) are the ones that collides.

Based on the description above and the issue described here: https://github.com/styled-components/babel-plugin-styled-components/issues/308, I would like to propose an option to add a prefix to those classes that contain the css and with that shield the styled-components generated classes from colliding with something external when we set this variable.

## How?
Update the function responsible to generate all the class hashes and concatenate the env variable, this variable would be possible to be set by `REACT_APP_SC_CLASS_PREFIX` or `SC_CLASS_PREFIX`, if they aren't set we will use and empty string which will not affect anything.

## Testing
Change the file `packages/sandbox/server/webpack.config.js` by adding the option `SC_CLASS_PREFIX: 'custom-class-prefix_',` inside the webpack EnvironmentPlugin.

### Before
image

### After
By adding the environment variable:
```javascript
new webpack.EnvironmentPlugin({
  // This will provide default value for NODE_ENV, unless defined otherwise
  // more info: https://webpack.js.org/plugins/environment-plugin/#usage-with-default-values
  NODE_ENV: 'development',
  SC_CLASS_PREFIX: 'custom-class-prefix_',
}),
```

And running the sandbox, we can see the hash classes with the prefix provided in the env variable:
image

## References
Special thanks to @justnewbee for contributing it in https://github.com/styled-components/styled-components/pull/3225, I'm trying to reopen this issue with more tests and talking about why the babel plugin isn't working for us.

### DIFF
--- a/packages/styled-components/src/utils/generateAlphabeticName.js
+++ b/packages/styled-components/src/utils/generateAlphabeticName.js
@@ -23,5 +23,7 @@ export default function generateAlphabeticName(code: number): string {
     name = getAlphabeticChar(x % charsLength) + name;
   }
 
-  return SC_CLASS_PREFIX + ((getAlphabeticChar(x % charsLength) + name).replace(AD_REPLACER_R, '$1-$2'));
+  const prefix = SC_CLASS_PREFIX ? `${SC_CLASS_PREFIX}_` : '';
+
+  return prefix + ((getAlphabeticChar(x % charsLength) + name).replace(AD_REPLACER_R, '$1-$2'));
 }
--- a/packages/styled-components/src/utils/generateAlphabeticName.js
+++ b/packages/styled-components/src/utils/generateAlphabeticName.js
@@ -1,6 +1,8 @@
 // @flow
 /* eslint-disable no-bitwise */
 
+import { SC_CLASS_PREFIX } from "../constants";
+
 const AD_REPLACER_R = /(a)(d)/gi;
 
 /* This is the "capacity" of our alphabet i.e. 2x26 for all letters plus their capitalised
@@ -21,5 +23,5 @@ export default function generateAlphabeticName(code: number): string {
     name = getAlphabeticChar(x % charsLength) + name;
   }
 
-  return (getAlphabeticChar(x % charsLength) + name).replace(AD_REPLACER_R, '$1-$2');
+  return SC_CLASS_PREFIX + ((getAlphabeticChar(x % charsLength) + name).replace(AD_REPLACER_R, '$1-$2'));
 }
--- a/packages/styled-components/src/utils/test/generateAlphabeticName.test.js
+++ b/packages/styled-components/src/utils/test/generateAlphabeticName.test.js
@@ -24,7 +24,7 @@ describe('generateAlphabeticName', () => {
       jest.resetModules();
     });
 
-    it.only('should concat SC_CLASS_PREFIX with generated alphabetic names', () => {
+    it('should concat SC_CLASS_PREFIX with generated alphabetic names', () => {
       jest.doMock('../../constants', () => ({ SC_CLASS_PREFIX: 'custom-class-prefix_' }))
       const generateAlphabeticName = require('../generateAlphabeticName');
 
--- a/packages/styled-components/src/utils/test/generateAlphabeticName.test.js
+++ b/packages/styled-components/src/utils/test/generateAlphabeticName.test.js
@@ -25,7 +25,7 @@ describe('generateAlphabeticName', () => {
     });
 
     it('should concat SC_CLASS_PREFIX with generated alphabetic names', () => {
-      jest.doMock('../../constants', () => ({ SC_CLASS_PREFIX: 'custom-class-prefix_' }))
+      jest.doMock('../../constants', () => ({ SC_CLASS_PREFIX: 'custom-class-prefix' }))
       const generateAlphabeticName = require('../generateAlphabeticName');
 
       expect(generateAlphabeticName(1000000000)).toEqual('custom-class-prefix_cGNYzm');
--- a/packages/styled-components/src/utils/test/generateAlphabeticName.test.js
+++ b/packages/styled-components/src/utils/test/generateAlphabeticName.test.js
@@ -18,4 +18,18 @@ describe('generateAlphabeticName', () => {
     expect(generateAlphabeticName(2733)).toMatchInlineSnapshot(`"a-D"`);
     expect(generateAlphabeticName(7390035)).toMatchInlineSnapshot(`"a-Da-d"`);
   });
+
+  describe('with SC_CLASS_PREFIX', () => {
+    beforeEach(() => {
+      jest.resetModules();
+    });
+
+    it.only('should concat SC_CLASS_PREFIX with generated alphabetic names', () => {
+      jest.doMock('../../constants', () => ({ SC_CLASS_PREFIX: 'custom-class-prefix_' }))
+      const generateAlphabeticName = require('../generateAlphabeticName');
+
+      expect(generateAlphabeticName(1000000000)).toEqual('custom-class-prefix_cGNYzm');
+      expect(generateAlphabeticName(2000000000)).toEqual('custom-class-prefix_fnBWYy');
+    });
+  })
 });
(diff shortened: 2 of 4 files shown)


packages/styled-components/src/utils/test/generateAlphabeticName.test.js
```diff
@@ -18,4 +18,18 @@ describe('generateAlphabeticName', () => {
     expect(generateAlphabeticName(2733)).toMatchInlineSnapshot(`"a-D"`);
     expect(generateAlphabeticName(7390035)).toMatchInlineSnapshot(`"a-Da-d"`);
   });
+
+  describe('with SC_CLASS_PREFIX', () => {
+    beforeEach(() => {
+      jest.resetModules();
+    });
+
+    it.only('should concat SC_CLASS_PREFIX with generated alphabetic names', () => {
```
guimello, line 27: I guess this is a development leftover, `only`.
👀

rafael-cineis-ifood: Oh, my bad, thanks for the review :smile:

kassi:
Hey @rafael-cineis-ifood  and @guimello : What's the status on this?
We would like to see this being merged since we would like to use it#s functionality rather than having some hacks doing something similar in our environment only.
Anything we can do to push this forward?

guimello:
> Hey @rafael-cineis-ifood and @guimello : What's the status on this?
> We would like to see this being merged since we would like to use it's functionality rather than having some hacks doing something similar in our environment only.
> Anything we can do to push this forward?

I think we are just waiting for maintainers to weigh in

rafael-cineis-ifood:
> Hey @rafael-cineis-ifood and @guimello : What's the status on this?
> We would like to see this being merged since we would like to use it's functionality rather than having some hacks doing something similar in our environment only.
> Anything we can do to push this forward?

As @guimello mentioned, we are just waiting for the review from any maintainers, I already tried contacting them but no response. We are creating a fork right now to handle it for our micro-frontends.

kassi:
Ok, thanks for the update. Sounds reasonable. We may be using it as well in the same way then.

rafael-cineis-ifood:
> Ok, thanks for the update. Sounds reasonable. We may be using it as well in the same way then.
> Do you plan a public fork or private one?

Right now we are going with a private fork with hope that this feature got reviewed and somehow handled properly inside styled-components official solution, we don't want to keep this fork for too long or we will need to start to keep it up-to-date and it could be a pain.

taranek (approved):

c0d3-k1ra:
Hey there,

Any idea when this can be merged? We are also facing the same issue where styles have collision.

rafael-cineis-ifood:
Hi @probablyup, I saw that you reviewed some other PRs related to this theme, could you have a look at it please?

kbalus (approved):

kitten:
I'm currently inclined not to merge this unfortunately. There are several other options that can be combined to combat conflicts, in case the library isn't shared/federated in an app, which I've written down here: https://github.com/styled-components/styled-components/issues/2974#issuecomment-895467259

> However, this is a hash based on component IDs and the contents of the CSS string, so collisions will be extremely unlikely given the namespace option, read: as unlikely as any other hash collision with 32 bits if no mistake is made in our hashing logic.

Technically, the hash classes that are used for styling hence shouldn't conflict. If they do conflict (which is something I think we had once for a past release, then their entryopy is likely bugged), then it's more likely to be something that needs to be fixed in the library. In other words, it'd be mostly a visual change at the cost of adding yet another environment variable.

To also mention the methods to actually do address micro frontends that aren't federating styled-components as a dependency:
https://github.com/styled-components/styled-components/issues/2974#issuecomment-895301213

> So, I will stand by this, but either you share styled-components like many microfrontend builds share react/react-dom across a given page.
> Or, you can use multiple versions of styled-components on the same page by altering process.env.SC_ATTR per build. I haven't tested this in a while, but it should still work just fine.
> Also, to summarise, the babel-plugin namespace option can now also be used to prevent hash collisions: https://styled-components.com/docs/tooling#namespace
> Since in all cases a combination of these three solutions work, I'll only leave this issue open for a bit and will lock it after, since it's gotten too long. I appreciate if any other issues are highlighted here, but I'm fairly certain at this point that hunting for "the big fix" isn't really a replacement for the above.

But what's going to be the most likely option here is using `namespace`, potentially with `SC_ATTR`, and the hash-based classes shouldn't conflict.

packages/styled-components/src/utils/generateAlphabeticName.js
```diff
@@ -21,5 +23,7 @@ export default function generateAlphabeticName(code: number): string {
     name = getAlphabeticChar(x % charsLength) + name;
   }
 
-  return (getAlphabeticChar(x % charsLength) + name).replace(AD_REPLACER_R, '$1-$2');
+  const prefix = SC_CLASS_PREFIX ? `${SC_CLASS_PREFIX}_` : '';
```
jonasantonelli, line 26:
```suggestion
  const prefix = SC_CLASS_PREFIX ? `${SC_CLASS_PREFIX}-` : '';
```

What do you think of using a hyphen, like it's done with the "sc-" prefix?
`sc-pFZIQ`

rafael-cineis-ifood:
> I'm currently inclined not to merge this unfortunately. There are several other options that can be combined to combat conflicts, in case the library isn't shared/federated in an app, which I've written down here: [#2974 (comment)](https://github.com/styled-components/styled-components/issues/2974#issuecomment-895467259)
> 
> > However, this is a hash based on component IDs and the contents of the CSS string, so collisions will be extremely unlikely given the namespace option, read: as unlikely as any other hash collision with 32 bits if no mistake is made in our hashing logic.
> 
> Technically, the hash classes that are used for styling hence shouldn't conflict. If they do conflict (which is something I think we had once for a past release, then their entryopy is likely bugged), then it's more likely to be something that needs to be fixed in the library. In other words, it'd be mostly a visual change at the cost of adding yet another environment variable.
> 
> To also mention the methods to actually do address micro frontends that aren't federating styled-components as a dependency:
> [#2974 (comment)](https://github.com/styled-components/styled-components/issues/2974#issuecomment-895301213)
> 
> > So, I will stand by this, but either you share styled-components like many microfrontend builds share react/react-dom across a given page.
> > Or, you can use multiple versions of styled-components on the same page by altering process.env.SC_ATTR per build. I haven't tested this in a while, but it should still work just fine.
> > Also, to summarise, the babel-plugin namespace option can now also be used to prevent hash collisions: https://styled-components.com/docs/tooling#namespace
> > Since in all cases a combination of these three solutions work, I'll only leave this issue open for a bit and will lock it after, since it's gotten too long. I appreciate if any other issues are highlighted here, but I'm fairly certain at this point that hunting for "the big fix" isn't really a replacement for the above.
> 
> But what's going to be the most likely option here is using `namespace`, potentially with `SC_ATTR`, and the hash-based classes shouldn't conflict.

Hi @kitten , sorry for the late response... I tried using the `namespace` only as also the `namespace + SC_ATTR` but the hashes continue conflicting, if I click on the classes I could see that it was inside the `<script />` tag with the `SC_ATTR` from each application, but as they were loaded in the same container application, it caused this issue.

But I agree that probably the best idea would be to have the styled-components shared between the applications via a CDN or module federation.

packages/styled-components/src/utils/generateAlphabeticName.js
rafael-cineis-ifood: It's following the same pattern that the [namespace option](https://styled-components.com/docs/tooling#namespace) is using, it adds the prefix with an underscore after it

jonasantonelli: Great, make sense for me

PR #3504, opened 2021-05-27
feat: allow PlatformColor and DynamicColor

birkir:
Now you can pass DynamicColor and PlatformColor as a string.

```tsx
const MyView = styled.View`
  color: '${DynamicColor({ light: 'red', dark: 'blue' })}';
  background-color: '${PlatformColor(['label'])}';
`

### DIFF
--- a/packages/styled-components/src/models/InlineStyle.ts
+++ b/packages/styled-components/src/models/InlineStyle.ts
@@ -1,4 +1,5 @@
 import transformDeclPairs from 'css-to-react-native';
+import { DynamicColorIOS, PlatformColor } from 'react-native';
 import { parse } from 'postcss';
 import {
   ExtensibleObject,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/styled-components/src/models/InlineStyle.ts [hunk 1]
```diff
@@ -1,4 +1,5 @@
 import transformDeclPairs from 'css-to-react-native';
+import { DynamicColorIOS, PlatformColor } from 'react-native';
```
probablyup, line 2: Does this require a major version bump? Not sure how long these exports have been around for

kitten:
Closing due to inacitivity. I think we could consider _changing_ the approach of dynamic styling in the implementation if we do split off the package to avoid complexity / differences in the CSS pipeline to the main package, see #1617

PR #3510, opened 2021-06-03
fix(component-style): classname collision due to static style rules not hashed

garryyao:
## Environment
All

## Reproduction
Running the classname generating logic in v5 will produce the same hash even if the ruleset is different.

```
//  const name = generateName(phash(this.baseHash, cssStatic.length) >>> 0);

// component ruleset 1
const name1 = generateName(phash(12345678, "width:100px;") >>> 0);

// component ruleset 2 
const name2 = generateName(phash(12345678, "width:10px;") >>> 0);

expect(name1).not.toEqual(name2);

```

## Steps to reproduce
N/A

## Expected Behavior
Same component id with different ruleset should generate different classNames.

## Actual Behavior
Same component id with different ruleset currently will result in classname collision.

### DIFF
--- a/packages/styled-components/src/models/ComponentStyle.js
+++ b/packages/styled-components/src/models/ComponentStyle.js
@@ -64,7 +64,7 @@ export default class ComponentStyle {
         names.push(this.staticRulesId);
       } else {
         const cssStatic = flatten(this.rules, executionContext, styleSheet, stylis).join('');
-        const name = generateName(phash(this.baseHash, cssStatic.length) >>> 0);
+        const name = generateName(phash(this.baseHash, cssStatic) >>> 0);
 
         if (!styleSheet.hasNameForId(componentId, name)) {
           const cssStaticFormatted = stylis(cssStatic, `.${name}`, undefined, componentId);


packages/styled-components/src/models/ComponentStyle.js
```diff
@@ -64,7 +64,7 @@ export default class ComponentStyle {
         names.push(this.staticRulesId);
       } else {
         const cssStatic = flatten(this.rules, executionContext, styleSheet, stylis).join('');
-        const name = generateName(phash(this.baseHash, cssStatic.length) >>> 0);
+        const name = generateName(phash(this.baseHash, cssStatic) >>> 0);
```
kitten, line 67: This isn't quite going to work. Yes, it does increase the likelihood of getting a different `name`, however, this is static rules and it should be sufficient to go by `componentId` (which `this.baseHash` represents) as that's unique per component. The issue here is that it ignores an underlying problem. Inserted rules are ordered by component group, i.e. each group has a number that determines where the rule is inserted. If we have two components here with different IDs then this assumption already fails with `StyleSheet.registerId(componentId);` and we won't get the desired rule ordering.

I suppose given this line the question is whether we'd even ever see duplicate `componentId`s: https://github.com/styled-components/styled-components/blob/98cff2f501aca4bee73bac54e601e33edda8bf6a/packages/styled-components/src/models/StyledComponent.js#L37

garryyao: Hi Kitten


In our test case two components on the page happened to have an id collision (probably due to unset component displayName which is common pitfall) in which case we shall rely on the the style rules to generate a different classname.

The idea of not just relying on this.baseHash isn't from me. It's in this repo's main branch code:
https://github.com/styled-components/styled-components/blob/main/packages/styled-components/src/models/ComponentStyle.ts#L71

kitten: Yes, what I'm pointing out however is that regardless of what that line looks like now, this change doesn't achieve what it aims to do.

I'm also not quite sure how it was introduced but it may have been accidental, cc @probablyup https://github.com/styled-components/styled-components/commit/971b57d1807b833aef9d5813fdb88de1875c2936#diff-df880830a1185d5137aec7d5978f9634b7deb787d6340d49ab04ed09214e8156R71

garryyao: I could try to produce a test case to see if it works if helpful, surprisingly this change didn't break any unit tests in my local run 😄

kitten:
Closing due to inactivity; See comments.

PR #3535, opened 2021-07-17
Updated the peer-dependencies of styled-components to latest react an…

Anishpras:
When I installed the package styled components through npx there was a warning regarding the peer dependencies of react and react-dom. So, I updated the package.json file under the packages folder and styled components to the latest stable release of react and react-dom 17.0.2 under peerDependencies and devDependencies section. As it was updated 3 years ago.


sachingvit:
@Anishpras  When you updated and re-installed the package one yarn.lock file was also generated, can you chcekin that file to PR?

kitten:
This isn't really necessary. What you've basically changed is saying "react must be at least at version `17.0.2`", which doesn't really make sense as it doesn't. It can be used with the version we've specified in the peer dependency, so this isn't an upgrade per se. The previous peer range said `>= 16.8.0`, which means "any version _after and including_ 16.8.0". Basically, this isn't `^16.8.0`, so the range already allows for `17.x.x` and beyond.

I don't know what warning it is you got but without it I can't really advice on that other than saying that your peer version is in-range 😅

PR #3551, opened 2021-08-09
Fix forced server-side mode not triggering global styles

### LINKED ISSUE #3369: Collecting global styles does not work in browser
## System:
 - OS: macOS 10.15.7
 - CPU: (8) x64 Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
 - Memory: 287.80 MB / 32.00 GB
 - Shell: 5.7.1 - /bin/zsh
## Binaries:
 - Node: 12.18.1 - /usr/local/bin/node
 - Yarn: 1.22.0 - ~/code/hex-inc/hex/node_modules/.bin/yarn
 - npm: 6.14.5 - /usr/local/bin/npm

In our app, we're constructing iframe contents as a string and relying heavily on styled-components' theming functionality. For this reason, we need to render the style tags from a styled component to a string. Per this issue: https://github.com/styled-components/styled-components/issues/2660 the SSR rendering functionality should be exactly what we need, and indeed for a while that did work, following the docs here: https://styled-components.com/docs/advanced#server-side-rendering. 

However, this recently started failing for us, with the styled components failing to render with the following error: 

```Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://fb.me/react-uselayouteffect-ssr for common fixes.```

Upon inspection, this seems to be due to this line, which is only hit for global styles based out of the browser: https://github.com/styled-components/styled-components/blob/72a6f642c3dedbcee91a2b5d76f96b93f345b86f/packages/styled-components/src/constructors/createGlobalStyle.js#L60

Right now there's no good way around this, since the `__SERVER__` flag is set at compile time. I know this is intended to be server side functionality, but there are a number of reasons you might want to render HTML to a string on the client side (including one in the original ticket), so it would be ideal if this could be fixed in the core library.

## Reproduction

Minimal repro: https://codesandbox.io/s/styled-components-base-forked-ilm86?file=/index.js

## Expected Behavior

`renderToString` should contain a stringified styles tag with the correct styles

## Actual Behavior

`renderToString` returns the empty string

### LINKED ISSUE #3506: Unable to collect style while doing "ssr in the browser"
`sheet.collectStyle` do not collect global style when called inside a browser environment.

[codesandbox for reproduction](https://codesandbox.io/s/styled-component-global-ssr-up2z5?file=/src/index.tsx)

The issue seems to be related this test:
https://github.com/styled-components/styled-components/blob/25d5a87c6bb5d22b1157c2355bd971e434ec2093/packages/styled-components/src/constructors/createGlobalStyle.ts#L52

It checks if the environment is node. Which is irrelevant since we might want to use ReactDOM.renderToString in a browser environment. It should instead check if the render was started from ReactDOM.renderToString.

I was thinking of a solution where the context used for collecting style ( [here](https://github.com/styled-components/styled-components/blob/30dab74acedfd26d227eebccdcd18c92a1b3bd9b/packages/styled-components/src/models/ServerStyleSheet.tsx#L42) ) exposes a flag to indicate that useLayoutEffect should not be used. Will that work ?

I understand that this issue represent an almost comically edgy case. Do you think a fix is worth it ?

### PR DESCRIPTION
**Pull request by kitten**
Resolve #3506
Resolve #3369

`createGlobalStyle` components skip the synchronous insertion of styles
in the browser, which is based on a global variable. However, sheets can
be forced into server-side mode, which means we need to instead adhere
to the client-side sheet's settings.

### DIFF
--- a/packages/styled-components/src/constructors/createGlobalStyle.ts
+++ b/packages/styled-components/src/constructors/createGlobalStyle.ts
@@ -51,13 +51,15 @@ export default function createGlobalStyle(
       renderStyles(instance, props, styleSheet, theme, stylis);
     }
 
-    // eslint-disable-next-line react-hooks/rules-of-hooks
-    useLayoutEffect(() => {
-      if (!styleSheet.server) {
-        renderStyles(instance, props, styleSheet, theme, stylis);
-        return () => globalStyle.removeStyles(instance, styleSheet);
-      }
-    }, [instance, props, styleSheet, theme, stylis]);
+    if (!__SERVER__) {
+      // eslint-disable-next-line react-hooks/rules-of-hooks
+      useLayoutEffect(() => {
+        if (!styleSheet.server) {
+          renderStyles(instance, props, styleSheet, theme, stylis);
+          return () => globalStyle.removeStyles(instance, styleSheet);
+        }
+      }, [instance, props, styleSheet, theme, stylis]);
+    }
 
     return null;
   };
--- a/packages/styled-components/src/constructors/createGlobalStyle.ts
+++ b/packages/styled-components/src/constructors/createGlobalStyle.ts
@@ -10,8 +10,6 @@ import determineTheme from '../utils/determineTheme';
 import generateComponentId from '../utils/generateComponentId';
 import css from './css';
 
-declare const __SERVER__: boolean;
-
 export default function createGlobalStyle(
   strings: Styles,
   ...interpolations: Array<Interpolation>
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by tonycassara** [comment]
Any chance there's a test release of this somewhere? I'm having a similar issue and would love to know if I can wait for this to be released or if I need to downgrade to 4.4.1 😞

**Comment by minisemi** [comment]
I still get the "Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format..." warnings from styled-components.browser.esm.js in v 5.3.1 when using createGlobalStyle for SSR. Using createGlobalStyle works, but the warnings are taking up the whole server log. Downgraded to 4.4.1 which got rid of the warnings.

**Comment by yhaiovyi-tripactions** [comment]
@kitten hey, from my understanding of that code while it fixes rendering issue, it would still throw that warning. Can't we just put `useLayoutEffect` inside of the condition?

**Comment by plerma** [comment]
Hey there! I landed here after updating `styled-components` from `5.3.0` to `5.3.11` and then facing 1 issue in our `createGlobalStyle`, we have the following test:
```
...

describe('global-styles', () => {
  it('should match snapshot', () => {
    renderWithTheme(<GlobalStyle />);
    expect(document.head).toMatchSnapshot();
  });
});
```

Which was working as expected having the styles on the Jest snapshot:
```
// Jest Snapshot v1, https://goo.gl/fbAQLP

exports[`global-styles should match snapshot 1`] = `
<head>
  <style
    data-styled="active"
    data-styled-version="5.3.0"
  >
    ... styles ...
  </style>
</head>
`;
```

But after updating, the styles are not loaded:
```
// Jest Snapshot v1, https://goo.gl/fbAQLP

exports[`global-styles should match snapshot 1`] = `
<head>
  <style
    data-styled="active"
    data-styled-version="5.3.11"
  / >
</head>
`;
```

We're using `jsdom` as our test environment and `react-testing-library` to render the component.

Is there a way to "force" the GlobalStyles (`createGlobalStyle`) to be rendered?

cc @kitten

Issue #3333, opened 2020-11-19
styled-components v6


kitten:
> **What is this?** This is a list of discussions and questions that need answering before we start a v6 complete rewrite and effort. This is for our own notes and transparency but for now I'll **lock** this issue for a week until we answer some of these questions myself, after which this issue will be **unlocked** and open for discussion and other members of the CSS-in-JS community that we'll talk to to add their notes and opinions.
> So this issue isn't yet (or at all) for general discussion or consumption and is just reflecting our thought process for the time being. More umbrella issues will follow.

### What happens to `@font-face`?

**Context:** There's a Chrome quirk that reloads fonts and images on each change to the `CSSStyleSheet` (which is definitely unintended) which can cause a flash of fonts when no cache header has been added. It's also possible that this is already fixed in newer versions. _Investigation needed._

Nothing, special behaviour could still remain dropped, although it's possible we could have a workaround that directly mutates the inserted `@font-face` at-rule to swap out the URL with something else in development to make this issue less apparent.

### What happens to `@import`?

Nothing, special behaviour could still remain dropped, although a warning would be welcome.

### Do we support IE11 in v6?

It'd be great to drop IE11 and this may just now be possible since people could stop upgrading when they need IE11 support since v4/v5 have been relatively stable (as long as we still patch bugs)

### Do we start with a rewrite?

**Current Answer:** A TypeScript rewrite would be welcome. We can still auto-generate typings for Flow and Flow in general is "falling out of popularity." Meanwhile a TypeScript rewrite is just ergonomically nicer for ourselves.

Currently we believe that a rewrite to get rid of old legacy code paths and questionable implementation details is the way forward.

### What happens to the style system?

We can split out our style management system into a separate non-React subpackage that is usable by other libraries so that other style systems can better cooperate with styled-components.

### Is our specificity approach ripe for an upgrade?

In the current systems we have an elaborate system that ensures that style ordering is predictable and solid because of composition. This is _really important_, but hinders optimisations for static styles. It's possible that we may be able to come up with more or better solutions for this. _Investigation needed._

### What about our theming system?

Even in a TypeScript rewrite this will be an issue as no type can be spread out via the `ThemeProvider`. So either we need a factory here or another system that distributes the theme type better. _Investigation needed._

**This list is still incomplete**

OKNoah:
[...]

PR #3555, opened 2021-08-09
Prevent ServerStyleSheet from emitting empty tags 

### LINKED ISSUE #3291: Empty style tags in streaming SSR output | Breaks in IE11
_**Context**_: Universal React application with styled-components running on Hapi server, streaming SSR enabled.

**_Problem_**: `"Unable to get property 'split' of undefined or null reference"` at `style.innerHTML.split('/!sc/\n')` in `rehydrateSheetFromTag()`

**_RCA_**: Looks like some of the style tags that are generated on server in streaming output are empty and they end up inside a SVG element (I think that's because they are always placed at top a chunk?). Now the problem is IE11 returns `undefined` for innerHTML of a style tag which is inside SVG and that throws an exception at `style.innerHTML.split('/!sc/\n')` in rehydrateSheetFromTag(). There are other empty styles tags who's parent node is not SVG and IE11 would just return innerHTML as `""`, and in turn `innerHTML.split('/!sc/\n')` would just return empty array, so no problem there. 

Screen Shot 2020-09-26 at 11 36 38 PM

**_Questions_**: 

1. So I am trying to figure out why would there be an empty style tag generated on server? anyone here with internal of knowledge of styled-components know in what circumstances that can happen? Does that it indicate much larger issue underneath from application standpoint?
2. If those empty style tags are expected in some scenario, can we add a null check for innerHTML before we try to split it? I can send a PR up for this.
3. Is my understanding correct about why a style tag would end up inside a SVG? I can confirm this doesn't happen for all pages, only for certain pages with varying amount of data.

### System:
 - OS: macOS 10.15.3
 - CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
 - Memory: 73.68 MB / 16.00 GB
 - Shell: 5.7.1 - /bin/zsh
### Binaries:
 - Node: 12.17.0 - ~/.nvm/versions/node/v12.17.0/bin/node
 - Yarn: 1.22.4 - ~/.yarn/bin/yarn
 - npm: 6.14.4 - ~/.nvm/versions/node/v12.17.0/bin/npm
### npmPackages:
 - babel-plugin-styled-components: ^1.10.7 => 1.10.7
 - styled-components: ^5.0.1 => 5.1.1

kitten:
Resolve #3291

_Note:_ There's no need to port this to `main` i.e. v6, as we won't be supporting IE11 for it as discussed a while ago
(For anyone following along, check #3333)

### DIFF
--- a/packages/styled-components/src/sheet/Rehydration.js
+++ b/packages/styled-components/src/sheet/Rehydration.js
@@ -18,7 +18,7 @@ export const outputSheet = (sheet: Sheet) => {
 
     const names = sheet.names.get(id);
     const rules = tag.getGroup(group);
-    if (names === undefined || rules.length === 0) continue;
+    if (!names || !rules || !names.size) continue;
 
     const selector = `${SC_ATTR}.g${group}[id="${id}"]`;
 
--- a/packages/styled-components/src/sheet/Rehydration.js
+++ b/packages/styled-components/src/sheet/Rehydration.js
@@ -52,7 +52,7 @@ const rehydrateNamesFromContent = (sheet: Sheet, id: string, content: string) =>
 };
 
 const rehydrateSheetFromTag = (sheet: Sheet, style: HTMLStyleElement) => {
-  const parts = style.innerHTML.split(SPLITTER);
+  const parts = (style.innerHTML || '').split(SPLITTER);
   const rules: string[] = [];
 
   for (let i = 0, l = parts.length; i < l; i++) {
--- a/packages/styled-components/src/models/ServerStyleSheet.js
+++ b/packages/styled-components/src/models/ServerStyleSheet.js
@@ -25,6 +25,8 @@ export default class ServerStyleSheet {
 
   _emitSheetCSS = (): string => {
     const css = this.instance.toString();
+    if (!css) return '';
+
     const nonce = getNonce();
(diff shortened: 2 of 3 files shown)


packages/styled-components/src/sheet/Rehydration.js
```diff
@@ -52,7 +52,7 @@ const rehydrateNamesFromContent = (sheet: Sheet, id: string, content: string) =>
 };
 
 const rehydrateSheetFromTag = (sheet: Sheet, style: HTMLStyleElement) => {
-  const parts = style.innerHTML.split(SPLITTER);
+  const parts = (style.innerHTML || '').split(SPLITTER);
```
probablyup, line 55: This change seems unrelated?

probablyup (approved):

packages/styled-components/src/sheet/Rehydration.js
kitten: it isn't really super unrelated. The original issue said that the main problem is that this returns undefined for empty elements in IE11. Basically, I'm just afraid that IE11 users could still experience the bug itself when some HTML snippet is generated using an old version of the library. Basically, just wanted to cover this on both ends

probablyup: I think `.innerHTML` will always return an empty string though

PR #3563, opened 2021-08-21
Fix SSR collisions caused by static style collisions and group ID minimums

### LINKED ISSUE #3094: Extends function ignoring child style with server side rendering
I tried upgrading from version 4 to version 5, but it broke all extends features in our application. We do server side rendering with Rails 5. I have look for other issues but have not found anything that could help me so I'm reaching out here to see if this may be a bug.

## System:
 - OS: macOS Mojave 10.14.6
 - CPU: (8) x64 Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz
 - Memory: 358.81 MB / 16.00 GB
 - Shell: 5.3 - /bin/zsh
## Binaries:
 - Node: 10.8.0 - /usr/local/bin/node
 - Yarn: 1.19.2 - ~/Documents/rtb-trader/ui/node_modules/.bin/yarn
 - npm: 6.2.0 - /usr/local/bin/npm
## npmPackages:
 - babel-plugin-styled-components: ^1.10.0 => 1.10.0 
 - styled-components: ^5.1.0 => 5.1.0

## Reproduction

I have this really simple code snippet that isn't working.

```
import React from 'react';
import { render } from 'react-dom';

const Button = styled.button`
  color: blue;
  font-size: 1em;
  margin: 1em;
  padding: 0.25em 1em;
  border: 2px solid blue;
  border-radius: 3px;
`;

const RedButton = styled(Button)`
  color: red;
  border-color: red;
`;

const root = document.getElementById('app');
if (root) {
  render(
    <div>
      <Button>Blue Button</Button>
      <RedButton>Red Button</RedButton>
    </div>
    , root
  );
}
```

## Expected Behavior

For the styles to be extended

## Actual Behavior
image

### LINKED ISSUE #3321: SSR: class order is not persisted
SSR + styled-components:
```
const A = styled.div`
  position: relative;
`
const B = styled(A)`
  position: absolute;
`
```
It happens that element B has `position: relative` (the order of generated classes is wrong) from time to time. It only happens when using SSR and not on all elements... pretty weird.... Has anyone exprerienced the same isssue?
I can create proper issue using the bug template if necessary.

### LINKED ISSUE #3411: Override Styles not working correctly with Server Side Rendering
Sometimes when doing server side rendering (I am using Gatsby, with gatsby styled component plugin) overriding styles is not always consistent  and the css specificity randomly changes.

Fore example I have 2 components

```
const ComponentOne = styled.div`
     background: green;
`;

// Second component based on ComponentOne  overriding the existing style 

const ComponentTwo = styled(ComponentOne)`
    background: red;
    padding: 0;
`;
```
When running on devserver  the ComponentTwo is correctly red, when running SSR the specificity is inverted and the component is actually green. For now I fixed increasing the specificity on all overrides

```
const ComponentTwo = styled(ComponentOne)`
  &&&{
     background: red;
   }
    padding: 0;
`;
```

but it's not the cleanest solution and might lead to other bugs. 

Is there another way to be sure that the specificity is maintained across client and server side rendering?

### LINKED ISSUE #3482: 5.2.x and 5.3.x: difference in SSR and locally when you extend another component
## System:
 - OS: macOS 11.2.3
 - CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
 - Memory: 4.94 GB / 64.00 GB
 - Shell: 5.8 - /bin/zsh
## Binaries:
 - Node: 14.16.1 - ~/.nvm/versions/node/v14.16.1/bin/node
 - Yarn: 1.22.10 - ~/git/panter/veloplus/veloplus-shop/veloplus-shop-fe/node_modules/.bin/yarn
 - npm: 6.14.12 - ~/.nvm/versions/node/v14.16.1/bin/npm
## npmPackages:
 - babel-plugin-styled-components: ^1.10.0 => 1.12.0
 - styled-components: ^5.3.0 => 5.3.0

## Reproduction

given some component:

```

const Foo = styled.div`
   border: 1px solid red;
`
```

now you extend this component:

```

const Bar = styled(Foo)`
    border: 1px solid green;
`

```

Now you mount `<Foo />` on a page, load the page. You see a red border as expected

If you mount now `<Bar />` without reload it will still have a red border!

But if you reload the page and SSR `<Bar />` it will have a green border as expected.

Reason is that the order of the classes change depending and as both have the same specificity, the order matters.

The Bug was introduced with 5.2.0, so downgrading to 5.1.1 fixes this inconsitency.

## Steps to reproduce

see above

## Expected Behavior

`styled(X)` should always have a higher specificity than just X

## Actual Behavior

`styled(X)` has the same specificity as X

### LINKED ISSUE #3522: Next.js 11 issue with styled components
1. ## System:
 - OS: macOS 11.4
 - CPU: (8) arm64 Apple M1
 - Memory: 93.88 MB / 8.00 GB
 - Shell: 3.2.57 - /bin/bash
## Binaries:
 - Node: 15.14.0 - ~/.nvm/versions/node/v15.14.0/bin/node
 - Yarn: 1.22.10 - ~/.nvm/versions/node/v15.14.0/bin/yarn
 - npm: 7.7.6 - ~/.nvm/versions/node/v15.14.0/bin/npm

2. ## System:
 - OS: Linux 5.4 Ubuntu 20.04.2 LTS (Focal Fossa)
 - CPU: (2) x64 AMD EPYC Processor (with IBPB)
 - Memory: 87.25 MB / 1.89 GB
 - Container: Yes
 - Shell: 5.0.17 - /bin/bash
## Binaries:
 - Node: 15.14.0 - ~/.nvm/versions/node/v15.14.0/bin/node
 - Yarn: 1.22.10 - ~/.nvm/versions/node/v15.14.0/bin/yarn
 - npm: 7.7.6 - ~/.nvm/versions/node/v15.14.0/bin/npm

I as a next js newbie started developing my first app last week. (next.js 11 was released in just a couple of days). As I read styled components works stably with next.js and I decided to use that. The problems started after the development of the 2nd and 3rd pages. The symptoms are very strange, so I will describe all my experiments.

I deployed via docker image - https://nextjs.org/docs/deployment#docker-image

1. I made the first page - everything works well.
2. I added a second page. Everything works fine for me on macos in dev mode and in build / start mode. But when an image is built on ubuntu, the styles are replaced in some elements. I couldn't solve the problem in any way and as a quick solution I moved the application to vercel.app. This solved the problem, but it turned out to be temporary.
3. I added a 3rd page to the site. Then, in general, all the styles were mixed and the elements were not displayed correctly.

Then I already checked that the next.js 11 release was only a few weeks ago and it is not stable, rolled back next.jx to version 10 and everything works well.

I used https://styled-components.com/docs/advanced#nextjs to install.
I also tried adding "styled-componets: ssr" to .babelrc, it doesn't help.
I can't figure out if the problem appears after the amount of css code written or pages generated, that's why it's very strange.

Working example on next.js 10: https://captain-ukraine-frontend-8ack12i0r-erive.vercel.app/
Issue example on next.js 11: https://captain-ukraine-frontend-glhcs4uly-erive.vercel.app/
Pages:
- /casting/  ('кастинг' in menu)
- /casting/1/ ('кастинг' in menu)
- /contacts/ ('трипільська пошта' in menu)

### LINKED ISSUE #3561: Hash collision in CSS rule hashing
## Environment

System:
 - OS: macOS 11.5
 - CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
 - Memory: 2.19 GB / 32.00 GB
 - Shell: 5.8 - /bin/zsh

Binaries:
 - Node: 14.16.1 - ~/.nvm/versions/node/v14.16.1/bin/node
 - Yarn: 1.22.10 - /usr/local/bin/yarn
 - npm: 6.14.12 - ~/.nvm/versions/node/v14.16.1/bin/npm

npmPackages:
 - babel-plugin-styled-components: ^1.13.2 => 1.12.0 
 - styled-components: ^5.3.0 => 5.3.0 

## Steps to reproduce

Very hard to reproduce in isolation. We tried to narrow it down in a client project, but it was only visible in a full environment with nextjs and SSR at play.

## Expected Behavior

Elements created via styled components should receive the CSS rules they were created with.

## Actual Behavior

In our setup a styled component suddenly receives CSS from another component. In our case it happened when we hydrate an SSR'ed HTML in the client. Upon further inspection it turns out that the same hash is already present in the ruleset collection. This leads to the wrong className being added to the current element. The source of this problem is a fault in the hashing code.

The actual hashing function that is shipped in the npm package for `5.3.0` contains an error due to which most of the hashing code is bypassed. The source code looks like this:

```js
// simplified:
generateAndInjectStyles(...) {
  //...snip
  const cssStatic = "color: red" // example
  // Note that the second parameter is a string of the CSS
  const name = generateName(phash(this.baseHash, cssStatic) >>> 0);
}

const phash = (h, x) => {
  let i = x.length;

  while (i) {
    h = (h * 33) ^ x.charCodeAt(--i);
  }

  return h;
};
```

And this is the generated minified code from the above:

```js
var t = ke(this.rules, e, a, n).join("")
  // Second parameter should be a string, namely just `t` and not `t.length`
  , o = W(q(this.baseHash, t.length) >>> 0);

// q is phash()
var q = function(e, a) {
  // Parameter "a" will always be a number. Therefore the test expression
  // inside the for construct will ALWAYS be falsy. In effect the actual
  // hashing logic is skipped and we'll just return parameter "e" as is.
  for (var n = a.length; n; )
    e = 33 * e ^ a.charCodeAt(--n);
  return e
}
```

So in essence all the hashnig function is doing is this: 

```js
const phash = (h, x) => {
  return h;
};
```

The bug is here:

```diff
  var t = ke(this.rules, e, a, n).join("")
-   , o = W(q(this.baseHash, t.length) >>> 0);
+   , o = W(q(this.baseHash, t) >>> 0);
```

In summary the current hashing is bypassed and the hash is solely based on the concatenated CSS string length. This can easily lead to collisions.

The good news is that the issue is fixed in the current `main` branch. We weren't able to narrow it down to a particular commit as the git history seems to be composed of detached commits. Due to that `git bisect` doesn't work on this repo.

The fix can be verified by building the source in `main` and comparing the generated files in the `dist/` folder (just search for `>>>`, the hashing function is close by).

From my observation this seems to be the cause for these issues:

- https://github.com/styled-components/styled-components/issues/3482
- https://github.com/styled-components/styled-components/issues/3522

kitten:
Resolve #3561
Resolve #3482
Resolve #3411
Resolve #3321
Resolve #3094
Resolve #3522

## Summary

Server-side rendering collisions were increasingly common in seemingly random cases, which was caused by a previous change that allowed new groups to have lower specificity and group IDs than all rehydrated groups. (See #3233)

Additionally, the hash was applied incorrectly (compare to `main` branch), where static styles would use `cssStatic.length` rather than `cssStatic`. This was previously broken due to changes in the hashing function. Overall however, this was causing issues when component IDs collided in hashes already, since `cssStatic.length` wouldn't do anything to alter the hash provided. We could still use `cssStatic.length.toString(16)` or an equivalent input, however, I've made this change to align with `main` for now, since the difference didn't make sense anyway.

## Set of changes

- Revert #3233 but preserve change in `getGroupForId`
- Fix typo in `ComponentStyle` relating to `phash(x, cssStatic.length)`

### DIFF
--- a/packages/styled-components/src/models/ComponentStyle.js
+++ b/packages/styled-components/src/models/ComponentStyle.js
@@ -64,7 +64,7 @@ export default class ComponentStyle {
         names.push(this.staticRulesId);
       } else {
         const cssStatic = flatten(this.rules, executionContext, styleSheet, stylis).join('');
-        const name = generateName(phash(this.baseHash, cssStatic.length) >>> 0);
+        const name = generateName(phash(this.baseHash, cssStatic) >>> 0);
 
         if (!styleSheet.hasNameForId(componentId, name)) {
           const cssStaticFormatted = stylis(cssStatic, `.${name}`, undefined, componentId);
--- a/packages/styled-components/src/sheet/GroupIDAllocator.js
+++ b/packages/styled-components/src/sheet/GroupIDAllocator.js
@@ -42,6 +42,10 @@ export const getIdForGroup = (group: number): void | string => {
 };
 
 export const setGroupForId = (id: string, group: number) => {
+  if (group >= nextFreeGroup) {
+    nextFreeGroup = group + 1;
+  }
+
   groupIDRegister.set(id, group);
   reverseRegister.set(group, id);
 };


kitten:
An initial **prerelease** version to verify this fix has been released under a tag:

| Package | Version |
| -------- | -------- |
| `styled-components@pr3563` | `5.3.1-pr3563` |

This will automatically enable this PR's patch.

probablyup (approved):
LGTM

Issue #3567, opened 2021-08-24
ServerStyleSheet.interleaveWithNodeStream inserts style elements as children of SVG elements


keeganstreet:
`ServerStyleSheet.interleaveWithNodeStream` inserts `<style>` elements as child elements of `<svg>` elements, which leads to styles being HTML encoded during hydration.

This is the cause of https://github.com/styled-components/styled-components/issues/3516

| Serverside render | After hydration |
| - | - |
| ![image](https://user-images.githubusercontent.com/513363/130601208-331e6958-6342-46d1-8dfa-f9cccbbd4874.png) | ![image](https://user-images.githubusercontent.com/513363/130601152-081110b9-62f8-4d83-abb2-eff26cbd90fd.png) |

**HTML:**

![image](https://user-images.githubusercontent.com/513363/130599973-ee1d973c-eec9-49a9-8b0d-de968661ba04.png)

## Environment

- Chrome
- Firefox
- Safari

## Reproduction

https://github.com/keeganstreet/styled-components-streaming-example

To run:

```
npm i
npm run build
npm start
```

## Steps to reproduce

Open http://localhost:3000/

## Expected Behavior

Rule `.kxnGBH > *{ text-decoration: underline; }` is intact and text in the box is underlined.

## Actual Behavior

Rule is encoded as `.kxnGBH &gt; *{ text-decoration: underline; }` and text in the box is not underlined.

keeganstreet:
I think the reason the `>` is encoded to `&gt;` is because `rehydrateSheetFromTag` uses `innerHTML`:

https://github.com/styled-components/styled-components/blob/fde4e44cc9bff71d8e2ed09330a8511d4ef3806f/packages/styled-components/src/sheet/Rehydration.ts#L52-L53

> **Element.innerHTML**
> Note: If a ... node has a child text node that includes the characters `(&)`, `(<)`, or `(>)`, `innerHTML` returns these characters as the HTML entities `"&amp;"`, `"&lt;"` and `"&gt;"` respectively. `Use Node.textContent` to get a raw copy of these text nodes' contents.
> https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML

Could `rehydrateSheetFromTag` be changed to `style.textContent` rather than `style.innerHTML`?

PR #3568, opened 2021-08-25
Read from textContent rather than innerHTML during rehydrate

keeganstreet:
Resolves https://github.com/styled-components/styled-components/issues/3567

Sorry about the lack of tests. I tried adding the following test but it passes even without this change, so it's not particularly useful. I think that `innerHTML` behaves differently in actual browsers compared with the environment where the Jest tests are run (JSDOM?)

**packages/styled-components/src/sheet/test/Rehydration.test.js**

```js
  it('supports > characters in selectors when style is inside another element', () => {
    document.body.innerHTML = `
      <svg>
        <style ${SC_ATTR} ${SC_ATTR_VERSION}="${SC_VERSION}">
          .b > * {}/*!sc*/
          ${SC_ATTR}.g22[id="idB"]{content:"nameB,"}/*!sc*/
        </style>
      </svg>
    `;
    const sheet = new StyleSheet({ isServer: true });
    rehydrateSheet(sheet);
    expect(sheet.getTag().getGroup(22)).toBe('.b > * {}/*!sc*/\n');
  });
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 ## Unreleased
 
+- Rehydrate stylesheets from `textContent` rather than `innerHTML` to fix issue of selectors sometimes being HTML encoded (see [#3567](https://github.com/styled-components/styled-components/issues/3567)).
+
 - Use `class` instead of `className` for unidentified DOM elements (better support for styling web components)
 
 - Upgrade to stylis v4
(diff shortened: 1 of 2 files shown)


probablyup:
I want to say we use `innerHTML` because `textContent` actually had buggy behavior in IE. @kitten do you remember?

kitten:
@probablyup I don't quite remember anymore. We could just retest it. As long as we use `textContent` though, we should be fine. However, `innerText` is usually the one that's inconsistent and to avoid.

probablyup (approved):
Thanks!

probablyup:
@keeganstreet Mind adding a quick changelog entry before I merge?

keeganstreet:
@probablyup sure, done 👍 

Would you like another PR to apply this to the legacy-v5 branch?

probablyup:
> @probablyup sure, done 👍
> 
> Would you like another PR to apply this to the legacy-v5 branch?

Sure, go ahead.

PR #3586, opened 2021-09-21
feat: CSS variable injection, cssvar helper, media query theme pattern for cssvar

probablyup:
When using `<ThemeProvider>`, css variables synthesized from the given theme
will be automatically injected into the page. The developer can then decide to
consume the variables via the new `cssvar` helper.

In tandem with this change, I have introduced a special key to the theme object
definition (`MQ_THEME_KEY`) that allows for media-query based adaptation of theme. Note: this only
applies to `cssvar` usage, `useTheme` and such will not adapt the retrieved value based
on media queries.

The primary use case for this functionality is easier SSR support for things like dark mode.

An example of usage can be found in changed files for this PR relating to "sandbox".

Other benefits to this pattern include fewer dynamic classes since CSS variable-based styling is emitted statically.

### DIFF
--- a/packages/sandbox/package.json
+++ b/packages/sandbox/package.json
@@ -14,7 +14,8 @@
   "dependencies": {
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
-    "react-live": "^2.3.0"
+    "react-live": "^2.3.0",
+    "styled-components": "link:../styled-components"
   },
   "engines": {
     "node": ">= 12"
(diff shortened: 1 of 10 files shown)


gregberge:
> Can you elaborate on that? My thought is by injecting once and using cssvars in the styles themselves it should actually be significantly faster than the current process of evaluating theme accessor functions for every component?

Problem is that xstyled contains a huge theme. Injecting everything represent a big insert at boot time and kills perf. Maybe we should use a system with "$" prefix to specify variables that should be injected as CSS properties. Also it could be natural and we would not need any `cssvar('foo')`, just `p => p.$foo` and that's it.

If you want to understand the problem, just load the xstyled theme in your actual implementation, it will freeze at initial mount.

PR #3602, opened 2021-10-14
Maintain stable style prop references in React Native

bolasblack:
related #3149, #3020


probablyup:
@bolasblack I'm not super familiar with react-native, would this introduce any performance or memory-related penalties? Just wondering in general if you've done perf testing with this change you can share.

zhigang1992:
IMHO, a useMemo would not introduce any visible performance or memory-related penalties.

In cases of a super long list, the benefit of this PR however is drastic.

w/o this each render will trigger on every single cell even with `React.memo`, simply because the style passed down was always a new array, even though of it is always the same.

probablyup:
Superceded by #3636, thank you for contributing!

PR #3603, opened 2021-10-15
fix: return stable style component with simple array equal compare

zhigang1992:
Basically same thing as https://github.com/styled-components/styled-components/pull/3602

But targeting legacy-v5 branch

### DIFF
--- a/packages/styled-components/src/models/StyledNativeComponent.js
+++ b/packages/styled-components/src/models/StyledNativeComponent.js
@@ -23,6 +23,7 @@ class StyledNativeComponent extends Component<*, *> {
   root: ?Object;
 
   attrs = {};
+  previousStyle: ?any;
 
   render() {
     return (
@@ -69,6 +70,12 @@ class StyledNativeComponent extends Component<*, *> {
             return [generatedStyles].concat(style(state))
           }
           : [generatedStyles].concat(style);
+          if (isArrayEqual(propsForElement.style, this.previousStyle)) {
+            propsForElement.style = this.previousStyle
+          } else {
+            this.previousStyle = propsForElement.style
+          }
+          this.previousStyle = propsForElement.style
           propsForElement.testID = testID;
 
           if (forwardedRef) propsForElement.ref = forwardedRef;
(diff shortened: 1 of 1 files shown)


probablyup:
@zhigang1992 I meant can the arrays themselves be just checked for referential equality without having to evaluate every array item

zhigang1992:
I think I misunderstood your point.

We can't because the line right above these changes

```
          propsForElement.style = typeof style === 'function' ?
          (state) => {
             return [generatedStyles].concat(style(state))
           }
           : [generatedStyles].concat(style);
```

So the `propsForElement.style` will always be a new reference

packages/styled-components/src/models/StyledNativeComponent.js
```diff
@@ -244,3 +251,10 @@ export default (InlineStyle: Function) => {
 
   return createStyledNativeComponent;
 };
+
+function isArrayEqual(a: any, b: any) {
```
nik72619c, line 255: @zhigang1992 Any specific reason for not adding the array type in the function argument itself? 🤔

packages/styled-components/src/models/StyledNativeComponent.js
```diff
@@ -69,6 +70,12 @@ class StyledNativeComponent extends Component<*, *> {
             return [generatedStyles].concat(style(state))
           }
           : [generatedStyles].concat(style);
+          if (isArrayEqual(propsForElement.style, this.previousStyle)) {
```
nik72619c, line 73: Would recommend if we could add a small comment here specifying why we are doing it.

nik72619c (commented):
@zhigang1992 Left a few comments !

packages/styled-components/src/models/StyledNativeComponent.js
zhigang1992: Because `previousStyle` is typed `any?`

packages/styled-components/src/models/StyledNativeComponent.js
zhigang1992: Added

zhigang1992:
sry, I realized I'm not familiar with flow syntax, can't get around updating it to be more strong type. Will close this for now.

PR #3636, opened 2021-11-30
Stabilize style prop for React Native

jbaudanza:
This is somewhat similar to https://github.com/styled-components/styled-components/pull/3602 but I took a different take.

Since `generatedStyles` already seems stable, I just allow it to be passed directly to the `propsForElement` object.

The style prop will still be unstable in the case that the caller passes in some inline stables, but I think this should cover most use cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 - Rename `masterSheet` in internals to `mainSheet`
 
+- Stabilize the identity of the style prop that is passed into components. This will help in styling PureComponent or React.memo components.
+
 ## [v5.3.0] - 2021-05-04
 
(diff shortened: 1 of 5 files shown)


probablyup:
@jbaudanza I think I'd still want this to be wrapped in a memo

jbaudanza:
> @jbaudanza I think I'd still want this to be wrapped in a memo

OK. Would you like me to add useMemo to this PR? Or would you rather merge https://github.com/styled-components/styled-components/pull/3602? Either one would be a big help to React-native users.

probablyup:
> > @jbaudanza I think I'd still want this to be wrapped in a memo
> 
> OK. Would you like me to add useMemo to this PR? Or would you rather merge #3602? Either one would be a big help to React-native users.

Feel free to add the `useMemo` wrapper to this PR, I generally like your change to use fewer arrays where possible 👍🏼 

Can you also add a changelog entry?

jbaudanza:
OK! useMemo wrapper added and changelog updated. Thanks for the feedback!

probablyup (approved):
Thank you!

I've been a bit slow about v6 but plan on cutting an initial release during my winter break

PR #3647, opened 2021-12-16
Pin babel-plugin-styled-components to v1

### LINKED ISSUE #3635: styled-components/macro doesn't work in latest version of package
Version 5.2.1 and above

### LINKED ISSUE #3645: display names no longer generated in babel-plugin-styled-components v2+using macro import
When upgrading npm or yarn package lock dependencies, babel-plugin-styled-components@2.0.2 is installed. But babel-plugin-styled-components doesn't generate `displayNames` after version `1.13.3`. 

And it looks like styled-components takes a dependency on babel-plugin-styled-components>=1.12.0 which resolves to 2.0.2.
Screen Shot 2021-12-16 at 12 30 55 PM
Screen Shot 2021-12-16 at 12 31 20 PM

## Environment

```
## System:
 - OS: macOS 12.1
 - CPU: (8) x64 Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz
 - Memory: 5.66 GB / 48.00 GB
 - Shell: 5.8 - /bin/zsh
## Binaries:
 - Node: 16.13.1 - ~/.nvm/versions/node/v16.13.1/bin/node
 - Yarn: 1.22.17 - ~/.nvm/versions/node/v16.13.1/bin/yarn
 - npm: 8.1.2 - ~/.nvm/versions/node/v16.13.1/bin/npm
## npmPackages:
 - styled-components: 5.3.3 => 5.3.3
 ```
## Reproduction
Reproducible in codesandbox below and in screenshots above

https://codesandbox.io/s/styled-components-base-forked-e7x8k?file=/index.js

## Steps to reproduce
Take a direct dependency or indirect dependency on babel-plugin-styled-components@2 or greater. Use a macro import e.g.

```
import styled from 'styled-components/macro;
```

## Expected Behavior

displayName should work whether v1.13.3 or v2+ of babel-plugin-styled-components is used with macro.

## Actual Behavior

displayName doesn't work after v2

esetnik:
Resolves #3645 
Resolves #3635 

by limiting dependency resolution to a compatible version of babel-plugin-styled-components

### DIFF
--- a/packages/styled-components/package.json
+++ b/packages/styled-components/package.json
@@ -68,7 +68,7 @@
     "@emotion/is-prop-valid": "^1.1.0",
     "@emotion/stylis": "^0.8.4",
     "@emotion/unitless": "^0.7.4",
-    "babel-plugin-styled-components": ">= 1.12.0",
+    "babel-plugin-styled-components": "^1.12.0",
     "css-to-react-native": "^3.0.0",
     "hoist-non-react-statics": "^3.0.0",
     "shallowequal": "^1.1.0",
--- a/packages/styled-components/src/macro/test/__snapshots__/macro.test.js.snap
+++ b/packages/styled-components/src/macro/test/__snapshots__/macro.test.js.snap
@@ -111,8 +111,9 @@ css\`
       ↓ ↓ ↓ ↓ ↓ ↓
 
 import { css as _css } from \\"styled-components\\";
-
-_css([\\"color:\\", \\";\\"], props => props.whiteColor ? 'white' : 'black');
+_css\`
+  color: \${props => props.whiteColor ? 'white' : 'black'};
+\`;
 "
 `;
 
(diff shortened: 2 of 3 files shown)


nik72619c (commented):
@esetnik Curious to know the rationale behind the changes in this file 🙂

esetnik:
> Hey @esetnik 👋🏼 , Curious to know the rationale behind the changes in the snap file 🙂

I think the snap file needed to be updated because the snapshots were generated using babel-plugin-styled-components 1.10.2 which generated different output in some cases.

nik72619c (commented):
got it. LGTM 👍🏼

rockwotj:
If you have time I'd love to understand why the tests here still pass with v2, but your setup is not.

probablyup:
There's something else at play, v2 of the babel plugin did was introduce a new syntax for a particular option field.

esetnik:
@probablyup why was this closed as it clearly resolves the issue as demonstrated in the codesandbox reproduction demo I provided in #3645

probablyup:
Read my comment

esetnik:
> 

I did and I don't understand the comment. We have multiple users (including me) reporting that the only way to get displayNames working is pinning to v1. I also provided a simple reproduction on codesandbox demonstrating that it's broken in v2 and working in v1. Additionally we have a report from another user with the same issue with a proposed identification of the bug in v2 [here](https://github.com/styled-components/styled-components/issues/3645#issuecomment-1002633474). What else should we provide in order to get this merged?

PR #3654, opened 2021-12-29
Use regex for topLevelImportPaths option

rockwotj:
In the 2.x.x version of babel-plugin-styled-components, topLevelImportPath requires a regex, so we escape the literal here.

### DIFF
--- a/packages/styled-components/package.json
+++ b/packages/styled-components/package.json
@@ -68,7 +68,7 @@
     "@emotion/is-prop-valid": "^1.1.0",
     "@emotion/stylis": "^0.8.4",
     "@emotion/unitless": "^0.7.4",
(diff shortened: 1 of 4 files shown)


agriffis:
I don't think this is a good idea. I'd rather [apply the anchors in the plugin](https://github.com/styled-components/babel-plugin-styled-components/pull/368) than run around adding them to callers of the plugin.

If that PR gets merged, resulting in plugin 3.0, then the dep here could change to `^1.12.0 || ^3`

If for some reason we go with this approach instead, then beware the anchors are backwards so it will never match. I don't know if this is somehow passing tests, but if so, it needs a new test.

rockwotj:
I'm closing this for now, I'm not sure how this macro is working at all, and I've run out of time to debug

PR #3670, opened 2022-01-21
Add withTheme flag to fix folding up for themed components

### LINKED ISSUE #3659: styled(withTheme(styled())) pattern results in undefined theme prop since v4
## Environment

"react": "17.0.2",
"react-dom": "17.0.2",
"styled-components": "5.3.3"

## Reproduction

https://codesandbox.io/s/styled-withtheme-styled-pattern-bug-nyk7f?file=/src/index.js

## Steps to reproduce

1. Define a component that takes theme as a prop, then wrap it in withTheme(styled())`` 
2. Restyle the component from step 1 via styled()``, essentially ending up with styled(withTheme(styled(Component)``))``
3. Render the restyled component in a component tree that is wrapped in ThemeProvider that provides a theme object

## Expected Behavior

The theme prop passed to the component from step 1 should not be undefined.

## Actual Behavior

The theme prop passed to the component from step 1 is undefined, since v4 of styled components. It works properly in v3.

## Additional Notes

The code sandbox also shows that wrapping the component from step 1 in a passthrough component seems to be a workaround for this issue, but that has some serious drawbacks and shouldn't be required to use styled-components in this way (I'm happy to explain further if necessary).

I reported this bug previously in https://github.com/styled-components/styled-components/issues/3137 but that issue also mentioned similar issues with react-redux connect(), and didn't get any response. I'm limiting this report to problems within styled-components on its own in the hopes that it will get more attention. Both of these issues are problematic for our production app, as they are preventing us from cleanly upgrading beyond v3.

EduardAkhmetshin:
Fixes #3659 
I added a flag to detect if a component was themed via withTheme HOC. When styled again, a styled component is folded up and doesn't get a theme from the HOC anymore. I fixed this by using the flag and passing the theme to the folded up component again.

### DIFF
--- a/packages/styled-components/src/hoc/withTheme.js
+++ b/packages/styled-components/src/hoc/withTheme.js
@@ -35,6 +35,12 @@ export default (Component: AbstractComponent<*, *>) => {
   hoistStatics(WithTheme, Component);
 
   WithTheme.displayName = `WithTheme(${getComponentName(Component)})`;
+  
+  // The underlying component is expected to get a theme as a prop.
+  // However, if the underlying component is folded up (see StyledComponent.js),
+  // we cannot tell that it needs a theme as a prop without a mark.
+  // So we mark WithTheme with withTheme flag.
+  WithTheme.withTheme = true;
 
   return WithTheme;
 };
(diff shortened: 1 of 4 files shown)


probablyup:
@EduardAkhmetshin thanks for the PR, we aren't accepting any new functionality for v5 but you're welcome to make this change to current `main` branch

jtheis85:
@EduardAkhmetshin any chance you could make a new PR for the current version? Thanks!

Issue #3656, opened 2022-01-02
Add / update a community channel or medium in contribution guidelines for discussion


nik72619c:
Currently, the [contribution guidelines](https://github.com/styled-components/styled-components/blob/e932038fffa1400c4ca68d04010b7d3f319298cb/CONTRIBUTING.md#how-can-we-help-you-get-comfortable-contributing) point contributors towards [spectrum](https://spectrum.chat/styled-components) for asking questions about contribution or asking mentors to guide them on the same. 

Since the spectrum chat is now read-only and might not suffice for the community to discuss `styled-components` questions for help or codebase, I would recommend we decide on a medium and make it official inside our styled-components documentation as well as our contribution guidelines. 

**Possible suggestions**
- [Discord](discord.gg/hfGUrbrxaU)
-[ Stack overflow](https://stackoverflow.com/questions/ask?tags=styled-components&newreg=70fdbe3a6a574ddebccd599eb75b12f4) (this is what we have right now but is not enforced via documentation, website, or contribution guidelines)
- [Github discussions](https://docs.github.com/en/discussions) (recommended since it requires less inter-application switch and keeps GitHub as the SSOT)
- [Slack](https://slack.com/intl/en-in/)

**Related links**
- [Posted the same question on stack overflow](https://stackoverflow.com/questions/70548644/is-spectrum-outdated-for-styled-components-contribution-and-discussion)
- [Github issue for adding discord as community channel](https://github.com/styled-components/styled-components-website/issues/764)

nik72619c:
cc: @probablyup ☝🏼

kitten:
Agreed; It's currently really hard to coordinate closing issues that have come up before or are questions that can be answered quickly. 

Stack Overflow is IMHO a last resort but currently what I point at in issue response templates, but I'd prefer a more topical destination. That said; there's simply not enough of a community yet to do so. I'm in favour of promoting and pointing to Discord more often to make it the primary destination and grow it even more. It could also be useful for coordinating new contributors in the hopes that we can get more traction on changes we've discussed but that the core team doesn't have time for.

I'm not too keen on opening GH Discussions for this repository, since it keeps questions out of issues but generally doesn't do so at a 100% success rate anyway, and isn't more likely to receive more credible replies apart from core team members 

So, I'd suggest to link to Discord consistently, use it in issue reply templates, add an issue template link to it which comes up before opening issues, and generally attempt at making it the first destination for contributors and users ✌️

[...]

PR #3682, opened 2022-02-03
Remove reference to Spectrum support channel

fatso83:
After Github purchased Spectrum all forums are now read-only, including [SC's](https://spectrum.chat/styled-components?tab=posts). Let's remove the reference to starting threads on this closed forum :smile: 

P.S. I was looking for a forum to _discuss_ pros/cons on importing theme vs getting it from the provider, which is a non-definitive question type that is out of scope for StackOverflow (i.e. will be closed). Having a new forum for such things would be great. I suggest enabling Github Discussions (which is probably povered by Spectrum code anyway).h


nik72619c:
Seems like this relates to #3656

nik72619c (commented):
LGTM 👍🏼

PR #3686, opened 2022-02-09
Replace mixinDeep with a cycle-safe implementation

### LINKED ISSUE #3576: `mixinDeep` causes infinite recursion when `function.name-polyfill` is used.
## Reproduction
The [function.name-polyfill](https://www.npmjs.com/package/function.name-polyfill) polyfill causes infinite recursion in Styled components that use the `mixinDeep util`.

With this polyfill included, the [mixinDeep](https://github.com/styled-components/styled-components/blob/30dab74acedfd26d227eebccdcd18c92a1b3bd9b/packages/styled-components/src/utils/mixinDeep.ts) method ends up blowing the stack. 

This happens because the polyfill introduces a `Function.prototype._name` prototype method that itself has a `_name` property which is a function (and on and on it goes).

It feels like there are 2 potential issues in the `mixinDeep` implementation that cause this.  The main issue is that `for (const key in obj) {` is used for iteration without the `hasOwnProperty` check.  This doesn't feel right.  It feels like `for..of` should be used instead.  The second potential issue is that `isObject` returns `true` for functions.  Coupled together these 2 methods cause the infinite recursion.  

## Steps to reproduce
Include https://www.npmjs.com/package/function.name-polyfill in your code.  Trigger `mixinDeep` with an object that has a method defined.

## Expected Behavior
Code should work properly

## Actual Behavior
Application crashes due to stack overflow

9still:
Fixes #3576

Since #2838, styled-components has been using a different (smaller) implementation of deep merge that doesn't play well with cyclical objects & particularly objects with cyclical prototypes.  In particular, this causes styled-components to blow the stack when used with some Function prototype extensions.

This PR replaces the mixinDeep implementation with a custom one that's closer to the [previously used library](https://github.com/mesqueeb/merge-anything) & is most similar in usage to [_.merge](https://lodash.com/docs/4.17.15#merge).  It keeps the tests unchanged, but adds a new one that explicitly ensures that cyclical objects don't blow the stack (try it with the old implementation if you want to see what happens in the current version).

See discussion in #3576 for further detail.

### DIFF
--- a/packages/styled-components/src/utils/mixinDeep.ts
+++ b/packages/styled-components/src/utils/mixinDeep.ts
@@ -1,60 +1,33 @@
 import { ExtensibleObject } from '../types';
+import isPlainObject from './isPlainObject'
 
-/**
-  mixin-deep; https://github.com/jonschlinkert/mixin-deep
-  Inlined such that it will be consistently transpiled to an IE-compatible syntax.
-
-  The MIT License (MIT)
-
-  Copyright (c) 2014-present, Jon Schlinkert.
-
-  Permission is hereby granted, free of charge, to any person obtaining a copy
-  of this software and associated documentation files (the "Software"), to deal
-  in the Software without restriction, including without limitation the rights
-  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-  copies of the Software, and to permit persons to whom the Software is
-  furnished to do so, subject to the following conditions:
-
-  The above copyright notice and this permission notice shall be included in
-  all copies or substantial portions of the Software.
-
-  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-  THE SOFTWARE.
-*/
-const isObject = (val: any) => {
-  return (
-    typeof val === 'function' || (typeof val === 'object' && val !== null && !Array.isArray(val))
-  );
-};
-
-const isValidKey = (key: string) => {
-  return key !== '__proto__' && key !== 'constructor' && key !== 'prototype';
-};
-
-function mixin(target: ExtensibleObject, val: any, key: string) {
-  const obj = target[key];
-  if (isObject(val) && isObject(obj)) {
-    mixinDeep(obj, val);
-  } else {
-    target[key] = val;
-  }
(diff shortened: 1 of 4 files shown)


packages/styled-components/src/utils/mixinDeep.ts
```diff
@@ -1,60 +1,33 @@
...
+function isRecursible(obj: any) {
+  return isPlainObject(obj) || Array.isArray(obj);
+}
 
-const isValidKey = (key: string) => {
-  return key !== '__proto__' && key !== 'constructor' && key !== 'prototype';
-};
+function mixinRecursively(target: any, source: any) {
+  const isRecursive = isRecursible(target);
+  for (const [key, sourceVal] of Object.entries(source)) {
```
kitten, line 10: I'd love to avoid the use of `for-of` and `Object.entries` here. Simply because it's not a rather performant variant and it's imho a bit of unnecessary uses of two non-ES primitives. There's a simpler way here though 👀 

```js
function mixinRecursively(target: any, source: any) {
  if (Array.isArray(target)) {
    for (let key = 0; key < target.length; key++) {
      target[key] = mixinRecursively(target[key], source[key]);
    }
    return target
  } else if (isPlainObject(target)) {
    for (const key in source) {
      target[key] = mixinRecursively(target[key], source[key]);
    }
    return target
  } else if (source != null) {
    return source
  } else {
    return target
  }
}
```

_Note:_ We don't need a `hasOwnProperty` check here I realised, since it's a plain object.

I did notice that we're assigning and mutating here, but seems like we did that before, so 🤷 :+1:
Also, I may have forgotten a case, but simply pointing out here that we can avoid `for-of` and `Object.entries` here entirely without drawbacks :v:

9still: Thanks for the suggestion @kitten! I refactored my original implementation to simplify & stop using `Object.entries` and `for..of` as you suggested.   Unfortunately couldn't use your implementation as is.  I added some tests in my latest commit that fail in your implementation (see below).    Also, while testing with your code I realized that `isPlainObject` was actually broken & was returning `true` for some objects with a custom prototype.  I updated the PR and took a stab at fixing that. 

Here is the test output with your implementation:

```
  ● should shallow merge properties if target is not a POJO or array

    expect(received).toMatchInlineSnapshot(snapshot)

    Snapshot name: `should shallow merge properties if target is not a POJO or array 1`

    - Snapshot  - 6
    + Received  + 0

      MyObj {
    -   "a": Object {
    -     "b": 2,
    -     "d": Object {
    -       "f": "f",
    -     },
    -   },
        "bar": 6,
        "foo": 5,
      }

      47 |   const obj2 = { a: { b: 2, d: { f: 'f' } } };
      48 |
    > 49 |   expect(mixinDeep(obj1, obj2)).toMatchInlineSnapshot(`
         |                                 ^
      50 | MyObj {
      51 |   "a": Object {
      52 |     "b": 2,

      at Object.<anonymous> (packages/styled-components/src/utils/test/mixinDeep.test.ts:49:33)

  ● should properly mixin arrays

    expect(received).toEqual(expected) // deep equality

    - Expected  - 5
    + Received  + 0

      Array [
        12,
    -   22,
    -   Object {
    -     "abc": 123,
    -   },
    -   true,
      ]

      158 |     60,
      159 |   ]);
    > 160 |   expect(mixinDeep([10], [11, 21, { abc: 123 }, true], [12, 22])).toEqual([
          |                                                                   ^
      161 |     12,
      162 |     22,
      163 |     { abc: 123 },

      at Object.<anonymous> (packages/styled-components/src/utils/test/mixinDeep.test.ts:160:67)
```

probablyup (approved):
Thank you!

PR #3446, opened 2021-03-31
Fix `ReferenceError: window is undefined` Server-Side-Rendering

tillhainbach:
fix #3444 by wrapping with `typeof window !== 'undefined'`

Issues was introduced by v5.2.2

Edit: 
Patch for 15b07960718a4fcd0d15ff407a4dee271e8a0c3e

Changed target branch to legacy-v5 (I hope that's the correct one..)

PR #3415, opened 2021-03-01
use window.__webpack_nonce__ to avoid overwriting by react-scripts

torgeirsk:
Fixes #3414

Issue #3414, opened 2021-03-01
`__webpack_nonce__` is not picked up in CRA builds


torgeirsk:
## Environment

`npx envinfo --system --binaries --npmPackages styled-components,babel-plugin-styled-components --markdown --clipboard`

## System:
 - OS: macOS 10.15.7
 - CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
 - Memory: 1.46 GB / 16.00 GB
 - Shell: 5.8 - /usr/local/bin/zsh
## Binaries:
 - Node: 15.10.0 - ~/.nvm/versions/node/v15.10.0/bin/node
 - Yarn: 1.22.10 - ~/.nvm/versions/node/v15.10.0/bin/yarn
 - npm: 7.5.3 - ~/.nvm/versions/node/v15.10.0/bin/npm
## npmPackages:
 - styled-components: ^5.2.1 => 5.2.1

## Reproduction

https://codesandbox.io/s/styled-components-base-gc367

## Steps to reproduce
1. create an empty cra and add styled-components
2. set `window.__webpack_nonce__ = "secret" ` anywhere
3. yarn start or yarn build

## Expected Behavior
`<style>` tags should have a `nonce` attribute

## Actual Behavior 
`<style>` tags dont have a `nonce` attribute

Confirmed that replacing `__webpack_nonce__` with `window.__webpack_nonce__` fixes the problem. I can submit a PR if desired.

kitten:
Potentially resolved given the PR? 

Please don't forget to be a good netizen on GitHub and revisit issues, especially ones that have been resolved and need to be closed, or ones that have received replies ❤️

PR #3755, opened 2022-06-14
Fix/nonce in ssr

### LINKED ISSUE #3571: The `nonce` attr is not added to the `<style>` tag during SSR since v5.2.1+
### System:
 - OS: macOS 11.5.2
 - CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
 - Memory: 83.88 MB / 16.00 GB
 - Shell: 5.8 - /usr/local/bin/zsh
### Binaries:
 - Node: 16.7.0 - ~/.nvm/versions/node/v16.7.0/bin/node
 - npm: 7.20.3 - ~/.nvm/versions/node/v16.7.0/bin/npm
### npmPackages:
 - styled-components: ^5.3.1 => 5.3.1

## Steps to reproduce

This is affecting a project I'm working on, but unfortunately that project is not easy to set up. But it's easy to confirm this issue by looking in styled-components' generated `dist` folder. Any of the generated `js` files (`styled-components.cjs.js` is the one I'm using) reveals the problem. [This function](https://github.com/styled-components/styled-components/blob/25d5a87c6bb5d22b1157c2355bd971e434ec2093/packages/styled-components/src/utils/nonce.ts#L3-L5):

```typescript
export default function getNonce() {
  return typeof __webpack_nonce__ !== 'undefined' ? __webpack_nonce__ : null;
}
```

...is transpiled into:

```javascript
function(){return"undefined"!=typeof window&&void 0!==window.__webpack_nonce__?window.__webpack_nonce__:null}
```

Specifically, the problem is that `__webpack_nonce__` becomes `window.__webpack_nonce__`, and `window` does not exist server-side. So the returned nonce is `null` instead of the actual nonce.

The end result is that the browser rejects the style tag (due to CSP), and you get a page of unstyled content.

If it helps, I believe the bug was introduced by https://github.com/styled-components/styled-components/pull/3308. I can confirm that the bug does not exist in v5.2.0, but it does exist in v5.2.1 and every 5.x.x version since.

## Expected Behavior

When `__webpack_nonce__` is set on the global object, the `<style>` tag generated by server-side rendering looks like this:

```html
<style nonce="T3e7m/F76YzluqD0I1VF3Q==" data-styled="true" data-styled-version="5.3.1">
```

## Actual Behavior

Whether or not `__webpack_nonce__` is set on the global object, the `<style>` tag generated by server-side rendering looks like this:

```html
<style data-styled="true" data-styled-version="5.3.1">
```

samimakicc:
Fixes #3571
Reverts #3415 and #3446

#3415 was a fix for issue #3414 but I believe the issue wasn't in styled-components but maybe just lack of documentation on how `__webpack_nonce__` should be used. And #3446 was just a fix on top of #3415.

`__webpack_nonce__` is a magic variable and webpack does some special handling for it. To my understanding it should be set at the beginning of entry point and used directly like `__webpack_nonce__ = 'secret'` instead of through window (which doesn't even exist on server).

### DIFF
--- a/packages/benchmarks/src/implementations/styled-components-v4/styled-components-v4.esm.js
+++ b/packages/benchmarks/src/implementations/styled-components-v4/styled-components-v4.esm.js
@@ -397,7 +397,7 @@ function stringifyRules(rules, selector, prefix) {
 /* eslint-disable camelcase, no-undef */
 
 var getNonce = function() {
-  return typeof window.__webpack_nonce__ !== 'undefined' ? window.__webpack_nonce__ : null;
+  return typeof __webpack_nonce__ !== 'undefined' ? __webpack_nonce__ : null;
 };
 
 //
--- a/packages/styled-components/src/utils/nonce.js
+++ b/packages/styled-components/src/utils/nonce.js
@@ -1,10 +1,10 @@
 // @flow
 /* eslint-disable camelcase, no-undef */
 
-declare var window: { __webpack_nonce__: string };
+declare var __webpack_nonce__: string;
 
 const getNonce = () => {
-  return typeof window.__webpack_nonce__ !== 'undefined' ? window.__webpack_nonce__ : null;
+  return typeof __webpack_nonce__ !== 'undefined' ? __webpack_nonce__ : null;
 };
 
 export default getNonce;
--- a/packages/styled-components/src/utils/nonce.js
+++ b/packages/styled-components/src/utils/nonce.js
@@ -4,12 +4,7 @@
 declare var window: { __webpack_nonce__: string };
 
 const getNonce = () => {
-
-  return typeof window !== 'undefined'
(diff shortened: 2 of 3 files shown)


3imed-jaberi (commented):
sgtm +1

latema:
Any chance to get this reviewed and merged? For now I have patched this myself in my project with `patch-package`, but it would be nice to have this fix in styled-components v5.

agriffis:
@probablyup I don't use webpack personally, but this looks good at a glance, and seems to match more closely to what's in v6.

probablyup (approved):
LGTM, I agree. I had it closer to this implementation originally and someone sent a PR with the current impl saying the other didn't work, but it's entirely possible that was also user error.

PR #3786, opened 2022-08-16
Fix sandbox SSR, project config and contributing instructions

janicduplessis:
This fixes a few things in the sandbox project, to make testing changes easier.

- SSR is currently broken. We need to add some code for nextjs to add the styled-components style tag. This is based on https://github.com/vercel/next.js/blob/canary/examples/with-styled-components/pages/_document.tsx.
- The project doesn't link against the version of styled-components in the repo. This uses link: scheme to make sure it does.
- Update the CONTRIBUTING instructions as the current ones are missing a few things.

Tested SSR by disabling JS and making sure the sandbox looks the same as with JS enabled and using "View Source".


probablyup:
These changes made it in at some point since the PR was open, but thank you!

PR #3787, opened 2022-08-16
Fix speedy rule insertion when css contains multiple rules

janicduplessis:
When using CSSOM insertion any css that contains more than one rule will cause it to fail. This is because the `insertRule` api requires only one rule. To fix this we can have a custom stylis serializer that returns an array of rules instead of the default that concatenate them together.

For example this will break:

```js
styled.div`
  color: red;
  @media (min-height: 200px) {
    color: blue;
  }
`
```

since it will try to insert the following rule, which is actually 2 rules:

```
.a {color: red;} @media (min-height: 200px) {.a {color:blue;}}
```

This can be easily reproduced in the sandbox example. To make sure CSSOM is used change `DISABLE_SPEEDY` to false in `constants.ts` and run the sandbox example. You can see some styles are missing, for example the font.

Without speedy (ok):

image

With speedy (broken):

image

After this fix, it renders the same with and without speedy.

### DIFF
--- a/packages/styled-components/src/hoc/withTheme.spec.tsx
+++ b/packages/styled-components/src/hoc/withTheme.spec.tsx
@@ -49,9 +49,9 @@ describe('withTheme', () => {
       </ThemeProvider>
     );
     expect(wrapper.toJSON()).toMatchInlineSnapshot(`
-<span>
-  {"color":"red"}
-</span>
-`);
+      <span>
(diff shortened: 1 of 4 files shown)


probablyup:
Thanks for catching this!

probablyup (approved):
Great work!

PR #3796, opened 2022-09-01
only minify standalone prod build

henryqdineen:
I briefly mentioned this on [twitter](https://twitter.com/henryqdineen/status/1565175665468530688) but we prefer, when possible for third party libraries to distribute unminified source to allow us to minify it ourselves. It seems to me that only the `.min.js` was meant to be minified so I removed the terser plugin from the others.


probablyup (approved):
Seems OK

PR #3799, opened 2022-09-05
fix(types): allow interpolating false

agriffis:
Allow `${bool && styles}` instead of requiring `${bool ? styles : null}`


probablyup (approved):
I had originally taken this out because I figured it'd be better to keep things to string-equivalents but yeah it seems like this pattern is in use everywhere so better to keep it. Thanks

PR #3801, opened 2022-09-10
Disallow mix of $as and as

agriffis:
This is a bit opinionated, but I think there's no reason anybody should be passing both, and one benefit we get from types is avoiding those kinds of mistakes.


probablyup (approved):
Fair enough

PR #3806, opened 2022-09-11
kill base extensible object

agriffis:
Only last two commits matter, the rest are from my previous PRs (it's just getting too complicated to do everything as separated changes on main)

See commit messages and comment at https://github.com/styled-components/styled-components/issues/3800#issuecomment-1242989090


agriffis:
Closing in favor of #3809 which I'll rebase

PR #3809, opened 2022-09-13
collected fixes for v6 typescript

agriffis:
This fixes a bunch, but not all, of the problems reported in #3800 

The problems fixed:

- [styled components don't emit an error on missing props](https://github.com/styled-components/styled-components/issues/3800#issuecomment-1242746314)
- [`styled(Custom).attrs()` breaks typing of properties](https://github.com/styled-components/styled-components/issues/3800#issuecomment-1242760145)

The changes:

- kill `BaseExtensibleObject` (and use `Dict<any>` explicitly in places that need that kind of thing, where `type Dict<T> = {[k: string]: T}`)
- rename `ExtensibleObject` to `ExecutionProps` since it's essentially a partial `ExecutionContext`
- change lots of `<Props = unknown>` to `<Props extends object>` which allows TS to check and infer a bit more
- allow the `attrs` call to infer or optionally be explicitly typed:

```ts
// infers {device: string}
const X = styled.div.attrs({device: 'mobile'})``

// explicit constraint
const Y = styled.div.attrs<{device: 'desktop' | 'mobile'}>(props => ({
  device: 'mobile',
  ...props,
}))``
```

- improve typing on `shouldForwardProp` since it always receives `elementToBeCreated`

I'm confident these changes don't solve all the problems, but I think it's a good stepping stone. With this merged, I think we should be able to muddle through the remaining issues.

### DIFF
--- a/packages/styled-components/src/types.ts
+++ b/packages/styled-components/src/types.ts
@@ -119,7 +119,7 @@ export interface CommonStatics<R extends Runtime, Props> {
   withComponent: any;
 }
 
-export interface IStyledStatics<R extends Runtime, OuterProps extends {}>
+export interface IStyledStatics<R extends Runtime, OuterProps extends object>
   extends CommonStatics<R, OuterProps> {
   componentStyle: R extends 'web' ? ComponentStyle : never;
   // this is here because we want the uppermost displayName retained in a folding scenario
@@ -136,7 +136,7 @@ export interface IStyledStatics<R extends Runtime, OuterProps extends {}>
 type PolymorphicComponentProps<
   R extends Runtime,
   ActualComponent extends StyledTarget<R>,
-  PropsToBeInjectedIntoActualComponent extends {},
+  PropsToBeInjectedIntoActualComponent extends object,
   ActualComponentProps = ActualComponent extends KnownTarget
     ? React.ComponentPropsWithRef<ActualComponent>
     : {}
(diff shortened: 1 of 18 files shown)


TmLev:
Would love to try it out, is it possible to release this as `beta-X`?

agriffis:
@TmLev You should be able to clone my fork, build with `yarn build` and then use `yarn link` to use it in your project.

I say "should" because I've done this with pnpm rather than yarn, but using pnpm requires a few more steps, especially some peer deps that need to be installed directly for the build to run. In theory that shouldn't be needed for yarn.

packages/styled-components/src/types.ts
```diff
@@ -67,13 +65,16 @@ export type Interpolation<Props> =
   | IStyledComponent<'web', any, any>
   | Interpolation<Props>[];
 
-export type Attrs<Props> =
-  | (ExtensibleObject & Props)
+export type Attrs<Props extends {}> =
+  | (ExecutionProps & Props)
```
joshjg, line 69: I think this should be:
```suggestion
  | Partial<ExecutionProps & Props>
```

probablyup:
@agriffis apologies, it looks like we've accumulated some merge conflicts. If you want to close your other PRs and just maintain this one, that's fine

agriffis:
> 

Sounds good. I'll rebase this and sort through the conflicts. I have a busy weekend and week though, so it might take me a few days.

packages/styled-components/src/types.ts
agriffis: @joshjg You're probably right, but this stuff can be a doozy to think about. Would you be able to provide an example of before/after behavior to aid my understanding?

agriffis:
@probablyup This is rebased now. IMHO best path forward is to merge this then revisit the remaining problems. I'd especially like to see some tests for the types, rather than relying on user feedback.

probablyup:
There's some good stuff in here, but the types are noticeably slower to evaluate in this branch. Not quite sure why... perhaps because of all the behind the scenes object merging?

agriffis:
Hard to say, but since the previous verson didn't actually _check_ the props passed to styled components, that could account for some of it :stuck_out_tongue_winking_eye:

probablyup:
> Hard to say, but since the previous verson didn't actually _check_ the props passed to styled components, that could account for some of it 😜

Hahahah touché. I’ll look at it more later, started reviewing this morning.

probablyup:
Going to merge and continue iterating against `main`

Issue #3813, opened 2022-09-21
Styled Components v6 types breaks react-docgen-typescript prop inference 


DavidReinberger:
Styled Components v6 now have their own type definitions, unfortunately it breaks props inference through react-docgen-typescript.

I've posted this also to the react-docgen-typescript repository however I'm not sure where is the right place, I'll be happy to look in to it more if someone can point me to the right direction.

### v6
**System:**

 - OS: macOS 12.5
 - CPU: (10) arm64 Apple M1 Pro
 - Memory: 108.22 MB / 32.00 GB
 - Shell: 5.8.1 - /bin/zsh

**Binaries:**
 - Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
 - Yarn: 1.22.19 - ~/.nvm/versions/node/v16.16.0/bin/yarn
 - npm: 8.11.0 - ~/.nvm/versions/node/v16.16.0/bin/npm

**npmPackages:**
 - styled-components: ^6.0.0-beta.2 => 6.0.0-beta.2 

### v5

**System:**
 - OS: macOS 12.5
 - CPU: (10) arm64 Apple M1 Pro
 - Memory: 99.06 MB / 32.00 GB
 - Shell: 5.8.1 - /bin/zsh

**Binaries:**
 - Node: 16.16.0 - ~/.nvm/versions/node/v16.16.0/bin/node
 - Yarn: 1.22.19 - ~/.nvm/versions/node/v16.16.0/bin/yarn
 - npm: 8.11.0 - ~/.nvm/versions/node/v16.16.0/bin/npm

**npmPackages:**
 - styled-components: ^5.3.5 => 5.3.5

## Reproduction

Here is a repository with v5 - https://github.com/DavidReinberger/storybook-typescript-react-doc-gen-styled-components-5, where it works flawless (even with default exports).
Here is a repository with v6 - https://github.com/DavidReinberger/storybook-typescript-react-doc-gen-styled-components-6

In both repositories, run `npm install && npm run storybook`

## Steps to reproduce
1. Open `ButtonStyled` story
2. Only `as` and `forwardedAs` props are visible3. 

## Expected Behavior
All props are inferred as with version 5.

## Actual Behavior
No props are inferred.

DavidReinberger:
Fixed by #3815

macko911:
It looks like there was a regression and this doesn't work anymore in `6.1.1`
Here's the updated repo, steps to reproduce are the same as in initial post.
[...]

PR #3815, opened 2022-09-21
Pass ExpectedProps to PolymorphicComponentProps (v6)

DavidReinberger:
Passes `ExpectedProps` to `PolymorphicComponent` as last argument to fix type inference with `react-docgen` and probably other tools. There is probably some problem with the inference from `ActualComponent`. Keeps inference from `styled(Component)` notation.

### DIFF
--- a/packages/styled-components/src/types.ts
+++ b/packages/styled-components/src/types.ts
@@ -164,13 +164,15 @@ interface PolymorphicComponent<
     props: PolymorphicComponentProps<
       R,
       ActualComponent,
-      ExpectedProps & PropsToBeInjectedIntoActualComponent
+      ExpectedProps & PropsToBeInjectedIntoActualComponent,
+      ExpectedProps
     >
   ): React.ReactElement<
     PolymorphicComponentProps<
       R,
       ActualComponent,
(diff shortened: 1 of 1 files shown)


joshjg (commented):
I think this would break polymorphism, since the `ActualComponentProps` type parameter will no longer be derived from `ActualComponent`

agriffis:
@DavidReinberger @joshjg Let's retest/revisit this after #3809 is merged.

DavidReinberger:
@agriffis I can see the PR is merged 🥳 I've been poking around some more and just can't get the hang of it 😮‍💨... Nevertheless as much as I understand the problem is that `PolymorhicComponentProps` does not correctly return all the props.

Whenever I either add the `ActualProps` to the generic call or remove `keyof ActualComponentProps` from the `Omit` within `PolymorphicComponentProps` it starts working 🤔...

Since I'm bit lost in this whole problem I won't be rebasing the PR and can be closed.

agriffis:
@DavidReinberger It's not easy to get the hang of it. I'm holding on by my fingernails.

You can go ahead and close this. We'll track the issue in https://github.com/styled-components/styled-components/issues/3813

I use Storybook too and I'm quite interested to see this fixed.

PR #3817, opened 2022-09-25
Refactor else blok is unnecesary

ta1m1kam:
## Summary
Some codes do not need else block because it has been returned ahead in if block.
It is able to eliminate the complexity of the code (although it is slightly).

> If an if block always executes a return statement, the subsequent else block is unnecessary.

## How did you test this change?
I run all test cases and confirmed NOT change the logic of code execution.

### DIFF
--- a/packages/benchmarks/src/implementations/inline-styles/View.js
+++ b/packages/benchmarks/src/implementations/inline-styles/View.js
@@ -2,11 +2,7 @@
 import React from 'react';
 
 const compose = (s1, s2) => {
-  if (s1 && s2) {
-    return { ...s1, ...s2 };
-  } else {
-    return s1 || s2;
-  }
+  return s1 && s2 ? { ...s1, ...s2 } : s1 || s2;
 };
 
 class View extends React.Component {
--- a/packages/benchmarks/src/implementations/inline-styles/View.js
+++ b/packages/benchmarks/src/implementations/inline-styles/View.js
@@ -26,7 +26,7 @@ const viewStyle = {
   position: 'relative',
   // fix flexbox bugs
   minHeight: 0,
-  minWidth: 0
+  minWidth: 0,
 };
 
 export default View;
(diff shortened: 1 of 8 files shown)


packages/benchmarks/src/implementations/inline-styles/View.js
```diff
@@ -2,11 +2,7 @@
 import React from 'react';
 
 const compose = (s1, s2) => {
-  if (s1 && s2) {
-    return { ...s1, ...s2 };
-  } else {
-    return s1 || s2;
-  }
+  return s1 && s2 ? { ...s1, ...s2 } : s1 || s2;
```
Yaronglp, line 5: What is your opinion about adding parenthesis to the variables that are being checked?
for example:
`(s1 && s2) ? { ...s1, ...s2 } : (s1 || s2);`
I think it would be more clear, for the variables that are being tested :)

packages/styled-components/src/utils/error.ts
```diff
@@ -31,10 +31,11 @@ export default function throwStyledComponentsError(
 ) {
   if (process.env.NODE_ENV === 'production') {
     return new Error(
-      `An error occurred. See https://github.com/styled-components/styled-components/blob/main/packages/styled-components/src/utils/errors.md#${code} for more information.${interpolations.length > 0 ? ` Args: ${interpolations.join(', ')}` : ''
+      `An error occurred. See https://github.com/styled-components/styled-components/blob/main/packages/styled-components/src/utils/errors.md#${code} for more information.${
+        interpolations.length > 0 ? ` Args: ${interpolations.join(', ')}` : ''
```
Yaronglp, line 35: Why do we have to check for the array's length?
if we have an empty array and we would call the join method, it would produce an empty string.
if the array will be with at least one item, the join will also perform the job we are looking for.
My suggestion is to call only:
`Args: ${interpolations.join(', ')}`
What do you say?

PR #3822, opened 2022-10-02
chore: correctly link local styled-components to sandbox

alexandernanberg:
- Link local styled-components to sandbox, to easily test changes
- Fix sandbox tsconfig
- Remove unused yarn.lock
- Remove unused React import


packages/sandbox/package.json
```diff
@@ -9,13 +9,12 @@
     "start": "next start"
   },
   "devDependencies": {
-    "babel-plugin-styled-components": "^2.0.6",
     "next": "^12.2.0"
   },
   "dependencies": {
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
-    "styled-components": "*"
```
probablyup, line 18: This shouldn't be required since we're using yarn workspaces...

probablyup (approved):

packages/sandbox/package.json
alexandernanberg: `*` is only supported in Yarn Berry AFAIK. Would you like a PR upgrading to that or maybe `pnpm` (which I prefer)?

PR #3825, opened 2022-10-06
fix: missing polymorphic props

agriffis:
This fixes the problem where `ZProps` is missing `as` in the example below:

```
const Z = styled.div``
type ZProps = ComponentProps<typeof Z>
```

See https://github.com/styled-components/styled-components/issues/3800#issuecomment-1268523022 for the original report against beta.3

### DIFF
--- a/packages/styled-components/src/types.ts
+++ b/packages/styled-components/src/types.ts
@@ -148,11 +148,17 @@ type PolymorphicComponentProps<R extends Runtime, E extends StyledTarget<R>, P e
 } & P &
   (E extends KnownTarget ? Omit<React.ComponentProps<E>, keyof P | 'as'> : object);
 
+/**
+ * Remove the function call signature, keeping the additional properties.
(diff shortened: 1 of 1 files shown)


agriffis:
Actually this doesn't work quite yet

```tsx
const X = styled.div``
const a = <X as="div" /> // fine
const b = <X as="span" /> // error
```

agriffis:
It's the right spot to poke, though

joshjg:
I have a change locally which resolves this, I think. Will try to open a PR later if I get a chance

agriffis:
I amended this and it seems to work now. Is your solution similar, @joshjg?

joshjg:
@agriffis This looks good, and my changes ran into heavy conflicts. I added an overload signature to `PolymorphicComponent` which forbids the `as` prop, but it seems like that might not be necessary anymore.

agriffis:
Closing in favor of #3828

PR #3826, opened 2022-10-07
fix: interpolation typing

agriffis:
This seems to fix the interpolation issue mentioned at https://github.com/styled-components/styled-components/issues/3800#issuecomment-1268523022


agriffis:
Actually I'll just close this in favor of #3828

PR #3833, opened 2022-10-12
Fix the return type for `css` tag

garbles:
At present the following yields a type error because the return type of `css` is inferred weirdly from a composition of several functions.

```ts
import { css } from 'styled-components';

const child = css``;
const parent = css`
  ${child}
`;
```

In the above, the type of `child` is the following which is unable to be nested inside of another css string (i.e. `parent`).

```ts
string | number | false | Keyframes | IStyledComponent<"web", any, any> | TemplateStringsArray | StyledObject<Props> | StyleFunction<Props> | RuleSet<Props> & { isCss?: boolean | undefined; }) | null | undefined;
```

This PR fixes the problem by:
- Explicitly defining the return type of `css` as `RuleSet<Props>` and then...
- Updating the `flatten` function to always return a `RuleSet`.

It seems like this might be what is already desired internally as I was able to remove type casting of `css(...) as RuleSet<X>` in several places.

### DIFF
--- a/packages/styled-components/src/constructors/constructWithOptions.ts
+++ b/packages/styled-components/src/constructors/constructWithOptions.ts
@@ -4,7 +4,6 @@ import {
   IStyledComponent,
   IStyledComponentFactory,
   KnownTarget,
-  RuleSet,
   Runtime,
   StyledOptions,
   StyledTarget,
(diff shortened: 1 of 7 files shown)


agriffis:
This looks great. IMHO we merge #3828 then rebase this, since that one adds tsc coverage to tests.

probablyup:
Thank you, this went in with #3828

PR #3834, opened 2022-10-13
fix: un-exclude tests from tsc

agriffis:
This is a minimal set of changes to un-exclude tests from TypeScript coverage. There are very few changes in actual code, opting instead for `@ts-expect-error TODO` comments in the tests so we can merge and iterate more quickly.


agriffis:
This was pulled by way of #3828

PR #3837, opened 2022-10-17
chore: add node 18 to test matrix

alexandernanberg:
**What has changed**
- Add Node 18 to test matrix
- Upgrade actions and simplify config with `cache` from `actions/setup-node`


agriffis:
LGTM :+1:

probablyup (approved):
Thank you! I'm actually going to drop 14 at this point as well

PR #3839, opened 2022-10-18
fix: don't send undefined props to DOM

agriffis:
Prior to v6 this built a new object `propsForElement` with only the forwarded props. Since v6, it changed to mutating `context` in place, see 0b8605ab20a2c1886e4ceea7b8b7f559d820511e

The new approach keeps the props but sets them undefined. This makes React emit errors in dev and test. Delete them from the context instead.

Additionally check and delete any undefined values in context, for example if `.attrs()` sets some to undefined, they should be removed at this point.


probablyup (approved):
Ty!

PR #3840, opened 2022-10-19
test: kill disused validAttr test

agriffis:
We stopped calling validAttr in 6baaea48, and that commit also removed the dep on `@emotion/is-prop-valid`


probablyup (approved):
Thanks!

PR #3844, opened 2022-10-21
refactor(types): satisfy attrs scenarios with updated types, fix strict "as" type resolution

probablyup:


packages/styled-components/src/types.ts
```diff
@@ -154,7 +154,7 @@ type PolymorphicComponentProps<R extends Runtime, E extends StyledTarget<R>, P e
   as?: E;
   theme?: DefaultTheme;
 } & P &
-  (E extends KnownTarget ? Omit<React.ComponentProps<E>, keyof P | 'as'> : object);
+  (E extends KnownTarget ? Omit<React.ComponentProps<E>, keyof P> : object);
```
probablyup, line 157: @agriffis this might fix the issue you were seeing in your sandbox, not sure

agriffis:
@probablyup This is looking great. But it doesn't seem to fix the `as` problem yet:

image

Here is the snippet for cut-n-paste:

```tsx
describe('foo', () => {
  const X = (props: { as?: 'div' | 'button' }) => null;
  const StyledX = styled(X)``;
  return (
    <>
      {/* GOOD: X doesn't accept section */}
      <X as="section" />
      {/* BAD: StyledX should accept section */}
      <StyledX as="section" />
    </>
  );
});

```

PR #3365, opened 2021-01-05
Ensure components for react-native pass testID

gauthamchandra:
When developing for react native, components are sometimes marked with
the `testID` prop. This prop is functionally similar to the
`data-testid` attached for react web components.

Unfortunately, styled components for react native didn't pass this prop
down, making it difficult for react native test frameworks like Detox to
test.

This change fixes that so when a `testID` prop is specified for a styled
component for react-native, it is passed down to the corresponding
native component it represents.

**NOTE**: Since this only affects `react-native`, it **ONLY** makes a change to `StyledNativeComponent`

**NOTE2**: I was hoping to use this change in the next minor version (`5.3.0`) which is why I branched off `legacy-v5`. Hopefully that's alright.

For context:

* [React Native Docs on `testID`](https://reactnative.dev/docs/view#testid)
* [The original issue that tried to start this](https://github.com/styled-components/styled-components/pull/1698)
* [Detox `by.id`Matcher that utilizes `testID`](https://github.com/wix/Detox/blob/master/docs/APIRef.Matchers.md#byidid)
* [React Testing Library's use via `byTestId`](https://callstack.github.io/react-native-testing-library/docs/api-queries#bytestid)

PR #3857, opened 2022-11-02
fix: (React Native) passing testID as attrs property

ku8ar:
PR that broke down the logic: https://github.com/styled-components/styled-components/pull/3365/files

Currently, adding testID as an attribute in `attrs` does not work, because testID is always overwritten with regular prop as `undefined`.

This change adds the ability to specify testID with `.attrs()`, as was possible before version 5.2.2

```
const MyComponent = styled.View.attrs({ testID: 'something' })`` // such code does not work
```


probablyup (approved):
Ty!

PR #3874, opened 2022-11-26
Generate more consistent class names between client and server

iiroj:
This commit fixes the _"Warning: Prop `className` did not match"_ error emitted during React DOM hydration. The reason for this was the the internal tracker for component identifiers was incremented for every component construction during server-side compilation, but client-side it happened only for the components in the current view, since JS code is typically code-splitted for the browser.

The fix is to take a different approach for generating component ids entirely. Instead of only looking at the component's displayName, its parent displayName and the running increment, the actual style rule set is also used for generating the hash part of the id. The running increment is still there but only used for components with exactly the same hash. This happens with identical styles and display names. When this collision happens, the generated class name is incremented with a separate literal counter like `sc-hash`, `sc-hash-1`, ...

----

Here's a fancy screenshot of a React Remix project I tested this on. Remix builds with esbuild so I couldn't use the babel plugin to pre-generate component ids.

| Before | After |
|:-------|:------|
|  Screenshot 2022-11-26 at 14 42 13 | Screenshot 2022-11-26 at 14 43 34 |


iiroj:
I checked that the browser unit tests pass, and the feature seems to work nicely in my local project. However, I don't seem to have the proper setup for the React Native tests...

packages/styled-components/src/models/StyledComponent.ts
```diff
@@ -29,24 +31,40 @@ import isTag from '../utils/isTag';
 import joinStrings from '../utils/joinStrings';
 import merge from '../utils/mixinDeep';
 import ComponentStyle from './ComponentStyle';
+import Keyframes from './Keyframes';
 import { useStyleSheet, useStylis } from './StyleSheetManager';
 import { ThemeContext } from './ThemeProvider';
 
-const identifiers: { [key: string]: number } = {};
+/** Shared context for component identifiers, preventing duplicate ids */
+const COMPONENT_IDENTIFIERS: Record<string, Record<string, number>> = {}
```
iiroj, line 39: The new identifier collection shape is like this:

```js
{
  'sc': {
    'hash': 0
  }
}
```

where `'sc'` is the (default) namespace, `'hash'` is the calculated hash, and `0` is the increment meaning how many times this exact hash has been found.

The resulting class name from this setup would be `'sc-hash'`, since zero increment is left out. If there was a collision, the second class name would be `'sc-hash-1'`.

packages/styled-components/src/models/StyledComponent.ts
```diff
@@ -29,24 +31,40 @@ import isTag from '../utils/isTag';
...
-function generateId(displayName?: string, parentComponentId?: string): string {
-  const name = typeof displayName !== 'string' ? 'sc' : escape(displayName);
-  // Ensure that no displayName can lead to duplicate componentIds
-  identifiers[name] = (identifiers[name] || 0) + 1;
+function generateId<OuterProps extends object>(rules: RuleSet<OuterProps>, displayName?: string, parentComponentId?: string): string {
+  const namespace = typeof displayName !== 'string' ? 'sc' : escape(displayName);
+
+  /** Keyframes cannot be interpolated without stylesheet */
+  const isNotKeyframe = (interpolation: Interpolation<OuterProps>) => !(interpolation instanceof Keyframes)
+  const styleSheet = flatten(rules).filter(isNotKeyframe).join('')
```
iiroj, line 48: Some tests were failing due to "_It seems you are interpolating a keyframe declaration (%s) into an untagged string_"; at first glance I couldn't find a way to use the `css` helper here, so I chose to filter any keyframes out from the hashing. If all the other CSS rules would be identical between two components, the `increment` number would handle it.

packages/styled-components/src/models/StyledComponent.ts
```diff
@@ -29,24 +31,40 @@ import isTag from '../utils/isTag';
...
+
+  // SC_VERSION gives us isolation between multiple runtimes on the page at once
+  // this is improved further with use of the babel plugin "namespace" feature
+  const componentId = `${namespace}-${generateComponentId(SC_VERSION + namespace + styleSheet)}`;
```
iiroj, line 52: I removed the `name` from here, since styles should already be enough for hashing... not sure if it should be put back. Or maybe the `namespace` could be removed too, since it's prefixed anyway.

packages/styled-components/src/models/StyledComponent.ts
```diff
@@ -29,24 +31,40 @@ import isTag from '../utils/isTag';
...
+
+  let increment = 0
+  if (namespace in COMPONENT_IDENTIFIERS && componentId in COMPONENT_IDENTIFIERS[namespace]) {
+    increment = COMPONENT_IDENTIFIERS[namespace][componentId]++
+  } else {
+    COMPONENT_IDENTIFIERS[namespace] = {
+      ...COMPONENT_IDENTIFIERS[namespace],
+      [componentId]: increment
+    }
+  }
```
iiroj, line 62: This increments the `increment`, and creates the default `0` value for first contact.

packages/styled-components/src/models/StyledComponent.ts
```diff
@@ -29,24 +31,40 @@ import isTag from '../utils/isTag';
...
 
-  const componentId = `${name}-${generateComponentId(
-    // SC_VERSION gives us isolation between multiple runtimes on the page at once
-    // this is improved further with use of the babel plugin "namespace" feature
-    SC_VERSION + name + identifiers[name]
-  )}`;
+  // Ensure that no displayName can lead to duplicate componentIds
+  const postfix = increment > 0 ? `-${increment}` : ''
```
iiroj, line 65: I thought the `-0` postfix would be a bit redundant, so opted to only add it starting from `-1`.

packages/styled-components/src/models/test/StyleSheetManager.test.tsx
```diff
@@ -358,12 +358,12 @@ describe('StyleSheetManager', () => {
...
-            >
-              Foo
-            </div>
-        `);
+      <div
+        className="sc-a-10 b"
+      >
+        Foo
+      </div>
+    `);
```
iiroj, line 366: We can nicely see the incremented class names working in unit tests, since a single test file will have the same collection, and tests components create the same component multiple times.

iiroj:
This does seem to some performance implications due to hashing all the styles, but I can't really tell if it's relevant or not:

| Before | After |
|:-------|:------|
| Screenshot 2022-11-26 at 15 11 05 | Screenshot 2022-11-26 at 15 15 13 |

packages/styled-components/src/models/StyledComponent.ts
iiroj: Maybe a bit confusing, index starts from zero so one means there's a collision.

packages/styled-components/src/models/StyledComponent.ts
iiroj: Hmm, maybe using `JSON.stringify()` here would produce similar output while being faster.

iiroj:
Here are updated benchmark results with the `JSON.stringify()` version:

| Before this PR | After |
|:----|:----|
| Screenshot 2022-11-26 at 16 29 40 | Screenshot 2022-11-26 at 16 21 48 |

packages/styled-components/src/models/StyledComponent.ts
iiroj: This filtering is no longer necessary with `JSON.stringify()`.

packages/styled-components/src/models/StyledComponent.ts
iiroj: Actually, might be more robust to use `name` instead of `namespace` here.

iiroj: Oops, my own self-review confusion; don't need `namespace` in hash since it's prefixed anyway.

iiroj: Fixed!

probablyup:
Not sure what's going on here but it's likely an issue with your environmental setup or using conditional logic relying on `window` in your styles somehow. If you follow React idioms things just work, I promise you.

iiroj:
Hello,

I don't really agree with your assessment. For as long as I've used styled-components, the class names haven't been consistent between client and server without special setup. This is the main purpose of the `babel-plugin-styled-components`:

> consistently hashed component classNames between environments (a must for server-side rendering)

https://github.com/styled-components/babel-plugin-styled-components

It's also mentioned in the styled-components docs directly:

> In order to reliably perform server side rendering and have the client side bundle pick up without issues, you'll need to use our [babel plugin](https://styled-components.com/docs/tooling#babel-plugin). It prevents checksum mismatches by adding a deterministic ID to each styled component. Refer to the [tooling documentation](https://styled-components.com/docs/tooling#serverside-rendering) for more information.

What this PR tries to solve is actully  consistent class names without having to use the babel plugin at all.

I don't think needing babel is a "React idiom".

probablyup:
Let’s put it this way, it’s not a problem we are willing to solve in the library itself due to the performance impacts you mentioned. We’ve had a babel implementation for years to handle this in a performant way, and even SWC has added support by porting our babel plugin to their infrastructure.

Esbuild will need to do the same if it wants to be as relevant as other compilers; many other libraries take a similar approach to adding behaviors/features via compiler plugins. 

SSR is an edge case that we support, React itself hasn’t had proper primitives to support SSR until recently.

PR #3900, opened 2022-12-23
breaking refactor: remove withComponent API

probablyup:
"as" prop works just as well; if you need to encode it at the component
level, use `attrs({ as: formerWithComponentValue })`


kachkaev:
To those who is here from search: when updating from v5 to v6, you need to do this:

```diff
-const MyComponentB = styled(MyComponentA.withComponent("div"))`
+const MyComponentB = styled(MyComponentA).attrs({ as: "div" })`
   margin: 0;
   padding:0;
 `;
```

PR #3907, opened 2023-01-03
fix: prevent crash when process.env is not defined

eMerzh:
there are times when process.env is not defined, 

let's prevent breaking at least...

related to https://github.com/cypress-io/cypress/issues/21434

### DIFF
--- a/packages/styled-components/src/constants.ts
+++ b/packages/styled-components/src/constants.ts
@@ -2,7 +2,7 @@ declare let SC_DISABLE_SPEEDY: boolean | null | undefined;
 declare let __VERSION__: string;
 
 export const SC_ATTR: string =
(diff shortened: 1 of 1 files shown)


probablyup:
I think you’d need to add the same fix to the other variables as well?

eMerzh:
not sure which one you would have protected? 
it seems to be good like that, but i can ammend if needed

Entkenntnis:
Thank you for working on this.

I our use case the variable `process` is set to `{}`, so `env` is undefined. To be safe, the other variables should be checked as well. The error might not occur for everybody because some other checks will handle it before

probablyup (approved):
Thank you!

suhas1294:
I have raised a PR for legacy-v5 branch as well. Kindly review it @probablyup 
cc: @santala 
Reference -  https://github.com/styled-components/styled-components/pull/3957

PR #3908, opened 2023-01-03
Fixes issue when using an ampersand (&) to scope css to a parent selector alongside the new namespace feature

marmite22:
The instructions on raising an issue said to consider making a PR with a failing test so here is one:

The issue this tests for is that when using an ampersand (&) to scope css to a parent selector and also using the new namespace feature added in the latest beta version, the resulting css has the parent selector before the namespace selector.

I took a stab at resolving this but while I was able to get the original issue solved, it breaks down with more complex use cases (as seen in the 2nd test)

### DIFF
--- a/packages/styled-components/src/utils/stylis.ts
+++ b/packages/styled-components/src/utils/stylis.ts
@@ -91,17 +91,25 @@ export default function createStylisInstance(
     }
 
     middlewares.push(selfReferenceReplacementPlugin, stringify);
+    let compiled = compile(prefix || selector ? `${prefix} ${selector} { ${flatCSS} }` : flatCSS);
 
-    return serialize(
-      compile(
-        options.namespace || prefix || selector
-          ? `${
-              options.namespace ? options.namespace + ' ' : ''
-            }${prefix} ${selector} { ${flatCSS} }`
-          : flatCSS
-      ),
+    if(options.namespace) {
+      compiled = compiled.map(rule=> {
+        // add the namespace to the start
+        rule.value = `${options.namespace} ${rule.value}`;
+        // add the namespace after each comma for subsequent selectors.
+        rule.value = rule.value.replaceAll(',', `,${options.namespace} `)
+        rule.props = rule.props.map(prop => {
+          return `${options.namespace} ${prop}`
+        });
+        return rule;
+      })
+    }
+    const serialized = serialize(
+      compiled,
       middleware(middlewares)
     );
+    return serialized;
   };
 
   stringifyRules.hash = plugins.length
--- a/packages/styled-components/src/utils/stylis.ts
+++ b/packages/styled-components/src/utils/stylis.ts
@@ -20,6 +20,29 @@ function serialize(children: Element[], callback: Middleware): string[] {
   return children.map((c, i) => callback(c, i, children, callback)).filter(Boolean) as string[];
 }
 
+/**
+ * Takes an element and recurses through it's rules added the namespace to the start of each selector.
+ * Takes into account media queries by recursing through child rules if they are present.
+ */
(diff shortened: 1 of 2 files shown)


marmite22:
Is anyone (@probablyup ?) able to verify the output of my 2nd test please? I've, perhaps wrongly, assumed that the outputted CSS should target the sc-* class but if I replicate the same code I get different results in v4 vs. v5. In fact, v5 actually switches between the sc-* class and the other generated class.

Given the following styled component
```
const StyledComponent = styled.div`
  color: blue;
  .child &,
  .child2 & {
    color: green;
  }
`;
```

### v4.4.1

```
.gMjrFY{color:blue;}
.child .gMjrFY,.child2 .gMjrFY{color:green;}
```
### v5.3.6

```
.QcRiW{color:blue;}
.child .sc-bcXHqe,.child2 .QcRiW{color:green;}
```

So why does v5 target the sc-* class where v4 didn't? Is it correct? If it is correct then my fix for the issue is working and it is the test that is wrong.

marmite22:
I can spend some time in my next sprint at work to look at this but I need to know the answer to my question above about the sc- classes. 

I'll happily work more on this if anyone has any feedback?

probablyup:
The relevant documentation is here: https://styled-components.com/docs/basics#pseudoelements-pseudoselectors-and-nesting

marmite22:
Looking at that documentation
image
I can see the same issue is present (the sc-* class is being targeted instead of the other one).

If you are happy that the generated css targets those classes then I will update the test to pass and this can then be merged I think.

In fact, I think I will raise a completely separate issue specifically about it.

marmite22:
I have now fixed the tests I've added so they are passing and raised a separate issue here https://github.com/styled-components/styled-components/issues/3923

I think this PR now resolves https://github.com/styled-components/styled-components/issues/3909

packages/styled-components/src/utils/stylis.ts
```diff
@@ -91,17 +91,22 @@ export default function createStylisInstance(
...
+        rule.value = `${options.namespace} ${rule.value}`;
+        // add the namespace after each comma for subsequent selectors.
+        rule.value = rule.value.replaceAll(',', `,${options.namespace} `);
+        rule.props = rule.props.map(prop => {
+          return `${options.namespace} ${prop}`;
+        });
+        return rule;
+      });
+    }
+    const serialized = serialize(compiled, middleware(middlewares));
```
probablyup, line 108: Can you drop this variable that isn’t reused? Looks good otherwise.

packages/styled-components/src/utils/stylis.ts
```diff
@@ -91,17 +91,22 @@ export default function createStylisInstance(
...
-          : flatCSS
-      ),
-      middleware(middlewares)
-    );
+    let compiled = compile(prefix || selector ? `${prefix} ${selector} { ${flatCSS} }` : flatCSS);
+
+    if (options.namespace) {
+      compiled = compiled.map(rule => {
+        // add the namespace to the start
+        rule.value = `${options.namespace} ${rule.value}`;
```
probablyup, line 99: Actually does this work appropriately with media queries? I’m not sure if we have a test for this explicitly

packages/styled-components/src/utils/stylis.ts
marmite22: done

packages/styled-components/src/utils/stylis.ts
marmite22: I have added an extra test for a media query inside a namespaced selector. I had to write a recursive function to go down through the rules and ensure that the namespace was added inside the media query rather than before it. 

I think it works now!

probablyup (approved):
Thanks, looks good!

PR #3944, opened 2023-02-06
Fix to recently added test that wasn't running

marmite22:
Fixed a test that was a) slightly broken and b) not actually running because it was mistakenly defined outside the describe block


probablyup (approved):
Thank you!

PR #3945, opened 2023-02-06
namespace option on StyleSheetManager better match babel-plugin-styled-components-css-namespace

marmite22:
I've modified the behaviour of the namespace option on StyleSheetManager so that it doesn't prefix global stye rules added via createGlobalStyle.

This more closely matches the behaviour in the `babel-plugin-styled-components-css-namespace` which was the previous way to namespace/isolate styles.

I think it's better to do things this way around rather than providing some option to stop the global styles getting prefixed. This way they are never prefixed and if you want to prefix them it would just be a case of manually adding your prefix to any global styles you write.

### DIFF
--- a/packages/styled-components/src/constructors/test/createGlobalStyle.test.tsx
+++ b/packages/styled-components/src/constructors/test/createGlobalStyle.test.tsx
@@ -561,4 +561,28 @@ describe(`createGlobalStyle`, () => {
       }"
     `);
   });
+
+  it(`namespaced StyleSheetManager does not namespace global styles`, () => {
+    const { render } = context;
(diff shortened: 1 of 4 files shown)


probablyup:
This actually doesn't match the design goal of the feature. The object is to prefix _all_ styles for a subtree. This is commonly used when deploying micro-frontends for isolation or as a global precedence boost when deploying a 3rd party widget on a destination site.

We could add additional configuration to support your case, but I do think it's currently working as expected.

quantizor:
Closing — stale namespace option PR with no activity for 3 years.

PR #3957, opened 2023-03-01
fix: prevent crash when process.env is not defined

suhas1294:
image
Was Getting error when process.env was tried to access. So have provided a null check.
Note that these changes are intentionally targeted for legacy-v5 branch since these fixes were already merged to ongoing beta v6 branch. reference - https://github.com/styled-components/styled-components/pull/3907

All tests are passing locally when done **yarn test**

PS : above reference link is not properly formatted, but still it would work.


probablyup (approved):
Thanks!

PR #3959, opened 2023-03-03
chore(sandbox): migrate to nextjs 14

quantizor:


quantizor:
Need to rebase this now that the dependency versions in trunk are pinned to exact versions.

quantizor:
Closing — sandbox has since been updated beyond Next 14.

Issue #3269, opened 2020-09-13
styled-components v5.2.0 not working in React + Rollup + TypeScript build


rpivo:
Here is a link to repo where the issue is reproducible: https://github.com/rpivo/styled-components-v5.2.0-issue

To reproduce in that repo:
- run `npm install` and check that styled-components v5.2.0 gets installed.
- run `npm run build`
- run `npm run serve`
- see that the styled-component from src > components > App > index.tsx does not have the styles written in that file.

That repo will work with v5.1.1. To get it to work with 5.1.1:
- delete the generated package-lock.json file if it's there
- delete node_modules folder if it's there
- delete all js files inside dist if they're there
- in package.json, change the dependency `"styled-components": "^5.1.1"` to `"styled-components": "5.1.1"`
- rerun `npm install`
- rerun `npm run build`
- run `npm run serve`
- see that the styled-component now works in the browser (you may need to empty cache and do a hard reload)

kitten:
Can you add some details please regarding what you're seeing in your output before I take a look at the reproduction?

raksooo:
I've run into this issue in a TypeScript project as well. After upgrading to `5.2.0` the styles are missing from the page. Here are two screenshots of the inspector when running the repo @rpivo posted, the first is with `5.1.1` and the second one is `5.2.0` and the difference is that the `<style>` tag is missing in the second one:

<img width="495" alt="Screen Shot 2020-09-17 at 22 53 01" src="https://user-images.githubusercontent.com/3668602/93527310-f4df9680-f938-11ea-8477-11e960d7dfd3.png">
<img width="364" alt="Screen Shot 2020-09-17 at 22 51 13" src="https://user-images.githubusercontent.com/3668602/93527316-f741f080-f938-11ea-9890-c43decb98e46.png">

matthew-nichols-westtel:
I ran into this issue today.

After lots of investigating and looking at diffs it looks like @rollup/plugin-node-resolve does not use the browser property inside package.json by default.

If I set "browser: true" to @rollup/plugin-node-resolve's options styles work again in 5.2.0.

rpivo:
@matthew-nichols-westtel that worked. Thanks

Bilb:
@matthew-nichols-westtel can you give me a bit more details about how you solved this ? I think I have the same issue but I cannot make it work. I tried to add this at the end of my projects `package.json`

```json

  "rollup": {
    "browser": true
  }
```

[...]

PR #3961, opened 2023-03-06
Remove "exports" entries

SilviaPag95:
Fixes the issue raised in https://github.com/styled-components/styled-components/issues/3800#issuecomment-1453355949

### DIFF
--- a/packages/styled-components/package.json
+++ b/packages/styled-components/package.json
@@ -11,17 +11,6 @@
     "./dist/styled-components.cjs.js": "./dist/styled-components.browser.cjs.js"
   },
   "sideEffects": false,
-  "exports": {
-    "types": "./dist/index.d.ts",
-    "node": {
-      "require": "./dist/styled-components.cjs.js",
-      "import": "./dist/styled-components.esm.js",
-      "default": "./dist/styled-components.cjs.js"
-    },
-    "require": "./dist/styled-components.browser.cjs.js",
-    "import": "./dist/styled-components.browser.esm.js",
-    "default": "./dist/styled-components.browser.cjs.js"
-  },
   "scripts": {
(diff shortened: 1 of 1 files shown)


probablyup (approved):
Thanks!

SilviaPag95:
@probablyup it seems this fix led to another issue. I use styled components inside a library (with rollup) to export them and after this change (when I upgrade to any version after v6.0.0-beta.12) no styles are applied to any of the styled components. The classnames are properly generated.

probablyup:
I think it's probably a different issue. Do you perhaps have two versions of s-c installed?

SilviaPag95:
I see this when I check for duplication with `$ yarn list --pattern styled-components`

In the repo that uses the library with styled-components:
```
├─ babel-plugin-styled-components@2.1.3
└─ styled-components@5.3.11
```

In the repo that generates the library:

```
├─ babel-plugin-styled-components@2.1.3
├─ styled-components@6.0.0-rc.2-4007
└─ typescript-plugin-styled-components@2.0.0
```

If I try to force a resolution for s-c in the repo that uses the library with styled components I get:

```
```
+-- react-drag-drop-files@2.3.9
| `-- UNMET DEPENDENCY styled-components@^5.3.0
`-- styled-components@6.0.0-rc.2-4007
```

probablyup:
Yup that'd be it. You can also force one version of s-c used via webpack resolve alias if you don't want to mess with dependencies.

SilviaPag95:
Unfortunately even when I add the resolution to the latest 6.0.0-rc.2 version the styles are not being applied at all. I'm at a loss

SilviaPag95:
Turns out this solves the issue: https://github.com/styled-components/styled-components/issues/3269#issuecomment-694562245

PR #3964, opened 2023-03-10
[v5] Make styled components work when the Object prototype is frozen

jportner:
Backport of #3963 for the v5 branch!


probablyup (approved):
Awesome thanks :)

PR #3965, opened 2023-03-10
A mechanism to use css variable

behnammodi:
This PR related to Support CSS Varaible cause:
- prevent any recalculation by `styled-components` for dynamic value
- generates just 1 style rule by relying on CSS variables as a tool to ship value from component to the rule
- by the way, there is no breaking change

## How can we use it?

To identify which prop is used for CSS variable purposes we can use `$$` prefix, but it is not required.

```jsx
const Comp = styled.div<{ $$fontSize?: string }>`      
  font-size: ${{ var: 'fontSize', compute: props => props.$$fontSize || "1rm" }};
`;
```

Usage:
```jsx
<Comp $$fontSize="2rm" />
```

Then our component will be something like this:
``` jsx
<div className="sc-a b" style={{ "--fontSize": 2rm" }} />,
```

and it just generates a style rule even if we pass any dynamic values:
```css
.b {
  font-size: var(--fontSize);
}
```


behnammodi:
@mxstbr @probablyup Could you please share your opinion on this PR?

probablyup:
I wonder if synthesizing css variables dynamically to avoid re-injecting styles would have some perf wins… like making it more automatic than this method so you don’t have to manually define anything. 🤔

behnammodi:
```
You don’t have to manually define anything. 
```
Then  we have to move logic from `styles` to `component`:

Like this:
```js
const Comp = styled.div`      
  font-size: var(--fontSize);
`;

<Comp style={{'--fontSize': props.$$fontSize || "1rm" }} />
```



```
Interpolations in styles lead to style branching which injects a lot of extra CSS
```
Exactly, but this PR can prevent any branching cause it relays on CSS variables and does not generate any other branch for dynamic values. please check this section:

https://github.com/styled-components/styled-components/blob/7f00c7fbaf04e1563021a6e0751c8e082b39d453/packages/styled-components/src/models/StyledComponent.ts#L160

packages/styled-components/src/utils/injectVar.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export default function injectVar<Props extends object>(
+  chunk: Interpolation<Props>[]
+): Interpolation<Props>[] {
+  const result: Interpolation<Props>[] = [];
+
+  for (let i = 0, len = chunk.length; i < len; i += 1) {
+    const next = chunk[i];
+    if (isVar(next)) {
+      result[i - 1] = result[i - 1] + 'var(--' + next.var + ')';
```
aspirisen, line 19: There is one thing to think about - css variables accept default value, here you have no possibility to pass default value. However, I do not have ideas how to do that properly.

behnammodi: We always can have a default value in another way, please have a look at this example:

```js
const Comp = styled.div<{ $$fontSize?: string }>`      
  font-size: ${{ var: 'fontSize', compute: props => props.$$fontSize || "1rm" }};
`;
```

Here `1rm` is the default value as programmatically

packages/styled-components/src/models/StyledComponent.ts
```diff
@@ -154,6 +157,14 @@ function useStyledComponentImpl<Target extends WebTarget, Props extends Executio
     }
   }
 
+  propsForElement.style = varRules?.reduce(
```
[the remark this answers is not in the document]
behnammodi: Do you have any idea how?

behnammodi:
@mxstbr any idea or feedback?

quantizor:
Closing — stale CSS variable proposal with no activity for 3 years.

Issue #3912, opened 2023-01-07
Feature Request: Issue with attrs.className ordering


radenkovic:
Hi folks,

I am using `styled-components` with tailwind to generate pre-defined components with some classes, however there is this unexpected behavior where instance className is **prepended** instead of **appended**. This prevents from overriding the css, IMO it makes more sense that instances override attrs (which are default values).

Fix should be trivial, if someone can point me to part of the code for `attrs` I can submit the PR with fix myself.

## Steps to reproduce

```js
// Define a styled component
const Btn = styled.button.attrs({ className: 'default-class' })``;

// In react component, omitted for clarity
<Btn className="appended-class">Test</Btn>
```

## Expected Behavior

pay attention the to order of the classes:

```
<button class="default-class appended-class">Test</button>
```

## Actual Behavior
```
<button class="appended-class default-class">Test</button>
```

probablyup:
Here’s the relevant code: https://github.com/styled-components/styled-components/blob/d21a7af6435336f10b7e8d891eef253e1ad30051/packages/styled-components/src/models/StyledComponent.ts#L151

Jittoqb:
hey @probablyup  can you just mention which lines are to be exchanged in the concat operation

Jittoqb:
hey @probablyup can you just review the PR ?

probablyup:
This should be fixed in latest beta

EricWVGG:
When you say "latest beta," is that in reference to v5 or v6? I just encountered this again, I'm on 5.3.9

probablyup:
v6

PR #3969, opened 2023-03-14
Appending classnames

Jittoqb:
When using tailwind to generate pre-defined components with some classes,  there is this unexpected behavior where instance className is prepended instead of appended. This prevents from overriding the css. This solves #3912 . Fixed by changing the order of concat and some tests.

### DIFF
--- a/packages/styled-components/src/models/StyledComponent.ts
+++ b/packages/styled-components/src/models/StyledComponent.ts
@@ -12,7 +12,7 @@ import type {
   OmitNever,
   RuleSet,
   StyledOptions,
-  WebTarget,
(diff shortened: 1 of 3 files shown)


radenkovic (commented):
looks good to me 👁️ 👁️

probablyup:
Can't do it this way because it violates attrs application order: https://styled-components.com/docs/basics#overriding-attrs

probablyup:
I have an alternate fix that I think will take care of your use case

Jittojoyes98:
> I have an alternate fix that I think will take care of your use case

Fine

PR #3972, opened 2023-03-26
chore: remove --clipboard option in envinfo

ImBIOS:
clipboard argument in envinfo isn't supported anymore, we should directly use clipboardly or clipboard-cli.


probablyup (approved):
Thanks!

PR #3975, opened 2023-03-29
fix: Do not add px to unitless CSS variables

Lazyuki:
Currently, the `addUnitIfNeeded` function is adding `px` to CSS variables that could potentially be used unitless. i.e. `--tooltip-z-index` or `--font-weight-bold`. It should ignore adding `px` if it's used for CSS variables since we can't know what they are used for.


probablyup (approved):
Makes a ton of sense, thank you. Could you also make the same PR against `main`?

PR #3977, opened 2023-04-03
Fixes an issue where keyframes are declared inside a namespaced StyleSheetManager

marmite22:
This fixes an issue that I spotted when declaring keyframs inside a styled component inside a StyleSheetManager with a namespace

the outputted code was adding the namespace to the start of the keyframes e.g. the `.parent` seen below

```@keyframes dvUGTh{.parent 0%{transform:rotate(0deg);}.parent 100%{transform:rotate(360deg);}}```

### DIFF
--- a/packages/styled-components/src/constructors/test/keyframes.test.tsx
+++ b/packages/styled-components/src/constructors/test/keyframes.test.tsx
@@ -375,4 +375,35 @@ describe('keyframes', () => {
       }"
     `);
   });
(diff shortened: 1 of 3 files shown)


marmite22:
> I'm not sure I understand why this is failing the tests so spectacularly

Ok I fixed this by moving the test

probablyup (approved):
Thanks, makes sense to me

Issue #3918, opened 2023-01-12
Feature Request: Advanced polymorphic components


aspirisen:
`as` prop is a powerful feature that allows to implement "mixins" approach. However, current implementation has limitations to unleash its abilities. There are some suggestions:

1. Make "as" to be forwarded by default
```
const Text = styled.span``;
const Link = styled(Text)``;
const Button = styled.button``;

export default function App() {
  return <Button as={Link}>Click me</Button>;
}
```
In such situations I can hardly imagine the case where you do not need styles (or behavior) from all nested components. If you really need to replace component we can add a prop for that like `replacedAs` or `$as`. 

2. Allow array in `as` prop
Sometimes you want to bind together several components, to create one single component:
```
const Text = styled.span``;
const Link = styled(Text)``;
const Button = styled.button``;

export default function App() {
  return <Button as={[Link, NextLink]}>Click me</Button>;
}
```
In this example you receive both your `Link` component styles and at the same time you get Link behavior from nextjs.

3. Add `mix` prop
This prop behaves the same like `as` but without replacing current component. It is used when you want to add behavior of other components
```
const Box = styled.div``;
const Flex = styled.div``;
const Button = styled.button``;

export default function App() {
  return <Button mix={[Box, Flex]}>Click me</Button>;
}
```
In this example we will finally have `button` tag but styles from `Box` and `Flex` will be applied. In fact it is pretty similar to `as` prop:
```
  return <Button as={[Box, Flex, Button]}>Click me</Button>;
```
Except that `Button` component will be called twice

probablyup:
I know what you’re getting at, but have questions. Do other component props get passed to all layers? What happens if one of the “as” prop targets is not a styled component?

aspirisen:
The same as we have now with `forwardedAs` :
```
function SomeComponent(props: PropsWithChildren<{ fullName: string }>) {
  return <div>Hello, {props.fullName}</div>;
}

const Main = styled.main``;

const StyledSomeComponent = styled(SomeComponent)`
  color: red;
`;

export default function App() {
  return (
    <StyledSomeComponent forwardedAs={Main} fullName="fullName">
      Click me
    </StyledSomeComponent>
  );
}
```

There will be no effect.

Or we could handle that on type level - all items in the array must be `PolymorphicComponent` except the last one.

PR #3979, opened 2023-04-04
Allow array in "as" prop and mix other component

aspirisen:
> This is a rough implementation to demonstrate the idea

Related issue with more detailed explanation https://github.com/styled-components/styled-components/issues/3918

* Ability to pass several components into `as` prop. That will allow you to combine behavior of several components and simplify JSX nesting structure
* Ability to `mix` other component. That will allow you to add styles from other components but current one will remain as is

The most complicated part is typescript types, there is a problem with string tags in mix prop.

### DIFF
--- a/packages/styled-components/src/test/types.tsx
+++ b/packages/styled-components/src/test/types.tsx
@@ -49,7 +49,7 @@ const Component3 = () => {
   return (
     // Wrapper should not be influenced by Button's types
     <Wrapper>
-      <Button mix={['video']} foo loop src="sd">
+      <Button mix={['video' as const]} foo loop>
         test
       </Button>
     </Wrapper>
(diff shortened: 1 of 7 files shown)


packages/styled-components/src/test/types.tsx
```diff
@@ -38,7 +38,18 @@ const Component2 = () => {
...
+      </Button>
+    </Wrapper>
+  );
+};
+
+const Component3 = () => {
+  return (
+    // Wrapper should not be influenced by Button's types
+    <Wrapper>
+      <Button mix={['video' as const]} foo loop>
```
aspirisen, line 52: For some reason typescript do not treat 'video' as const string

kaaboaye: To have const string you can also mark entire array as const otherwise it will be `string[]` by default

aspirisen: I think I should just remove strings in `mix` since there is no sense to mix plain tags.

quantizor:
Closing — stale for 3 years, no plans to support array in `as` prop.

PR #3982, opened 2023-04-07
Fix `css` prop typings

ahutchings:
Fix the `css` prop typings to resolve TypeScript errors when using the prop (described in https://github.com/styled-components/styled-components/issues/3800#issuecomment-1498409379) and added type tests for a few common usages.

I also included some related improvements:
* Reworded and updated the `CSSProp` docblock
  * Updated the example usage to augment `Attributes` rather than `DOMAttributes`. This makes it possible to use `css` with custom components whose props don't extend DOMAttributes, and matches [the example usage in DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/35f7bccf0d312292ea79feac97d3b3274d34158c/types/styled-components/index.d.ts#L431).
  * Changed the description to be more oriented around the `css` prop rather than styled-components-related module augmentations in general (especially `DefaultTheme`).
* Export the `RuleSet` type, which could be useful when building functions that accept or return RuleSets (asked about in [this discussion topic](https://github.com/orgs/styled-components/discussions/3915)).

### DIFF
--- a/packages/styled-components/src/index.ts
+++ b/packages/styled-components/src/index.ts
@@ -11,6 +11,7 @@ export {
   IStyledStatics,
   PolymorphicComponent,
   PolymorphicComponentProps,
+  RuleSet,
   Runtime,
   StyledObject,
   StyledOptions,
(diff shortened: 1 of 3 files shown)


probablyup (approved):
🔥 thank you!!

PR #3985, opened 2023-04-09
chore: yarn 3

probablyup:


probablyup:
A bit stuck on the test failure... can't reproduce it locally even with the exact same node version...

probablyup:
This fail is so weird. The babel output is correct in CI but incorrect locally...

.github/workflows/test.yml
```diff
@@ -1,48 +1,41 @@
...
       - uses: actions/checkout@v3
 
-      - name: Setup node
-        uses: actions/setup-node@v3
+      - uses: actions/setup-node@v3
         with:
-          node-version: 18
-          cache: "yarn"
+          node-version-file: .nvmrc
+          cache: yarn
```
wojtekmaj, line 19: This option does not work with modern Yarn unfortunately. Consider:

```yml
      - name: Cache .yarn/cache
        uses: actions/cache@v3
        env:
          cache-name: yarn-cache
        with:
          path: .yarn/cache
          key: ${{ runner.os }}-${{ env.cache-name }}-${{ hashFiles('**/yarn.lock') }}
          restore-keys: |
            ${{ runner.os }}-${{ env.cache-name }}
```

.yarn/releases/yarn-3.5.0.cjs
wojtekmaj, line 1: This should not be needed with Corepack, which you already appear to use in package.json

.yarnrc.yml
```diff
@@ -0,0 +1,7 @@
+nodeLinker: node-modules
+
+plugins:
+  - path: .yarn/plugins/@yarnpkg/plugin-interactive-tools.cjs
+    spec: "@yarnpkg/plugin-interactive-tools"
+
+yarnPath: .yarn/releases/yarn-3.5.0.cjs
```
wojtekmaj, line 7: This should not be needed with Corepack, which you already appear to use in package.json

.github/workflows/test.yml
```diff
@@ -1,48 +1,41 @@
...
 
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
 
-      - name: Setup node
-        uses: actions/setup-node@v3
+      - uses: actions/setup-node@v3
```
wojtekmaj, line 16: After setting up Node, consider:

```yml
      - name: Enable Corepack
        run: corepack enable
```

to enable Yarn 3 without the need of having legacy Yarn v1 installed.

PR #3987, opened 2023-04-13
fix: Do not add px to CSS variables (main branch)

Lazyuki:
Replay of #3975 against the main branch


probablyup (approved):
Sweet, thanks!

PR #3993, opened 2023-04-20
Enable exactOptionalPropertyTypes check

aspirisen:
Here I enabled [exactOptionalPropertyTypes](https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes) check in `tsconfig`.

Also, I added `| undefined` to most optional fields (marked with `:?`)

If your project has `exactOptionalPropertyTypes` enabled you will face issues when you want to pass `undefined` in some conditions.

For example currently here
```
<StyleSheetManager sheet={condition ? styledComponentsStyleSheet.instance : undefined}>
  {children}
</StyleSheetManager>
```
You will have an error
```
  Types of property 'sheet' are incompatible.
    Type 'StyleSheet | undefined' is not assignable to type 'StyleSheet'.
      Type 'undefined' is not assignable to type 'StyleSheet'.ts(2375)
```


probablyup:
Is that because of a setting in your project vs code in styled components? I don’t like the extra verbosity…

aspirisen:
Yes, if you add have `exactOptionalPropertyTypes` you will get TS errors from styled-components for no actual reason

probablyup:
Why wouldn’t you use `skipLibCheck`? I don’t think it’s responsible to require all 3p libraries to conform to your own project’s TS settings.

aspirisen:
It will not work because as I understand this flag just disables errors in `d.ts` files theirselves

https://codesandbox.io/s/serene-julien-8mlg1q?file=/src/App.tsx

bcole808:
I think we should merge this and enable `exactOptionalPropertyTypes` in styled-components. 

It's not just about code style; without this enabled, it's possible to write code that has different behaviors at runtime depending on if a value is set to `undefined`, or not set at all. [Here's an explanation of it from the TS 4.4 release notes. ](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#exact-optional-property-types)  

Without this, projects using `exactOptionalPropertyTypes` with styled-components will see the error above which is essentially a warning that styled-components code may not have accounted for the difference in behavior between a value being `undefined` and not being provided at all. 

Even though it feels a lot more verbose to have to add `| undefined` all over the place, it is worth it to enable this in JS libraries because it makes it clear to Typescript that the library code has accounted for how to handle the two situations.

probablyup:
Alright, I think I've adequately warmed up to this. Can we get the conflicts sorted and I'll merge it?

aspirisen:
@probablyup hey, I updated this PR

probablyup (approved):
Thank you @aspirisen!

PR #4011, opened 2023-05-15
Fix perf issue with large union types

justinbhopper:
This fixes a significant performance issue that I encountered when styling components that contained props that had very large union types.  

Repro can be found here: https://github.com/justinbhopper/styled-component-v6-perf
Without fix: ~12 sec
With fix: ~0.4 sec

The solution was to stop using Omit on the Props.  The[ type definitions from v5 didn't seem to need Omit](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/styled-components/index.d.ts#L41), so I felt this was probably acceptable.  

I wanted to add a test to benchmarks for this use case, but I got errors when running the commands mentioned in the README.md.  I'll be happy to add some tests if I can get help with those errors.

cc: @probablyup


probablyup:
Ooh, thank you! I'll try to set aside time to review it today

justinbhopper:
@probablyup I believe I've figured out the issue going on, and this may help solve some other use cases being reported in v6.

The Omit happening was causing a huge performance issue, but also was masking the following problem:

The `Styled` function has a `Props` generic argument which is meant allow for users to explicitly pass in prop types, like so:
`const foo = styled.div<`**`{ blah: number}`**`>'...'`

However, TypeScript was _inferring_ this type based on what was passed in the styles.  Since other components can be referenced within the Interpolation array, it was inheriting the props of those components mentioned within the styles.

The fix was to force TypeScript to stop type argument inference from happening using a custom `NoInfer<T>` helper type (found [here](https://stackoverflow.com/questions/56687668/a-way-to-disable-type-argument-inference-in-generics)).  

image

This ensures that Props defaults to `object` unless the argument is explicitly passed in.

probablyup (approved):
Thank you so much for this, I struggled for weeks with it

PR #4012, opened 2023-05-16
Version 6 release candidate typescript issues - failing tests

aliceHendicott:
This is a PR to add failing tests for some typescript issues found in the v6 release candidate.

Issue comment describing the bugs - https://github.com/styled-components/styled-components/issues/3800#issuecomment-1548941843
Code sandbox demonstrating the problems - https://codesandbox.io/s/styled-components-v6-examples-gu95sw?file=/src/App.tsx:2062-2149

### DIFF
--- a/packages/styled-components/src/constructors/constructWithOptions.ts
+++ b/packages/styled-components/src/constructors/constructWithOptions.ts
@@ -6,6 +6,7 @@ import {
   IStyledComponent,
   IStyledComponentFactory,
   KnownTarget,
+  NoInfer,
   Runtime,
   StyledOptions,
   StyledTarget,
@@ -43,9 +44,6 @@ type PropsSatisfiedByAttrs<
 > = Omit<Props, keyof Result> &
   OptionalIntersection<Props, Result> &
   Partial<Omit<Result, keyof Props | 'as'>>;
-
-// Prevents TypeScript from inferring generic argument
-type NoInfer<T> = [T][T extends any ? 0 : never];
 
 export interface Styled<
   R extends Runtime,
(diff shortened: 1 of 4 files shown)


probablyup:
Thank you

probablyup:
I pushed a variety of fixes... the one I'm currently stuck on is `attrs` object-style requiring the whole type vs allowing a partial. If you try to make `Props` partial for that syntax it breaks a bunch of things in the attrs spec file.

probablyup:
@justinbhopper if you have any ideas for this last failing test I'd appreciate it, you seem to know your way around TS :)

justinbhopper:
@probablyup I actually took a stab at this a few days ago, but struggled because I felt I wasn't properly understanding your intended approach for how you structured `ExecutionProps`.  I got to a point where I was ripping up too much in an effort to fix it, so I backed off.

If you have any insights you can give me behind what approach you originally had in mind, I can better help.  Otherwise, it _might_ result in several unintended changes as I rip too much apart out of a shear misunderstanding.

I feel like I could figure it out especially having the v5 type definitions as a reference guide.  Let me know your thoughts before I proceed.

probablyup:
> If you have any insights you can give me behind what approach you originally had in mind, I can better help. Otherwise, it might result in several unintended changes as I rip too much apart out of a shear misunderstanding.

Sure! The tricky part is if `as` is provided we want to infer it such that we can figure out props for the new target. Otherwise the object form of `attrs` is just meant to be a partial of any prop the component could conceivably take.

justinbhopper:
@probablyup Thanks!  I'll take a stab at it 👍🏻

probablyup:
I have a feeling we might need to go for TS 5 and use the new const modifier for some of our generics

justinbhopper:
Added these test cases (plus more) and fixes to #4014

PR #4014, opened 2023-05-19
Version 6 release candidate typescript issues - failing tests

justinbhopper:
This PR addresses the issues relating to the attrs function.  

- Adds test cases mentioned from PR #4012
- Adds additional test cases I discovered were failing, relating to attrs and general expectations regarding the `as` prop
- Adds the perf test case to ensure we don't re-introduce the performance test issue relating to large union types again in the future.

Closes #4012


packages/styled-components/src/types.ts
```diff
@@ -36,6 +36,11 @@ export interface StyledOptions<R extends Runtime, Props extends object> {
 
 export type Dict<T> = { [key: string]: T };
 
+// Allows for jsx ignored attributes such as data-* and aria-*
+export interface HyphenatedAttributes {
+  [key: `${string}-${string}`]: unknown;
```
probablyup, line 41: Is there a way to limit this more to only known aria attributes? `data-${string}` can stay since that's an open field

probablyup:
Thank you so much for doing this work again!

packages/styled-components/src/types.ts
probablyup: It looks like you could pull in `React.AriaAttributes`

justinbhopper: Done

packages/styled-components/src/test/types.tsx
```diff
@@ -149,15 +150,153 @@ const DivWithProps = styled.div<{ waz: number }>`
...
+
+const AttrFunctionRequiredTest1 = styled(DivWithRequiredProps).attrs(props => ({
+  // Should not have to provide foo within attrs
+  bar: "hello",
+  // Should allow additional props
+  newProp: 42,
+  // Should allow hyphenated props
+  "data-foo": 42,
+  // Props should have inherited props
+  foo2: props.foo?.toPrecision(0),
```
probablyup, line 289: Allowing `foo2` is a bug, we don't want to allow unknown props to be returned from `attrs`.

justinbhopper: I noticed v5 type defs allowed it, so I thought maybe it was a feature.  I'll update it to restrict it so only unknown data attributes are allowed.

justinbhopper: @probablyup I've tried a bunch of ways to prevent unknown props within the attrs, but it doesn't play nicely with TypeScript's normal behavior of generic argument duck checking.  

I should note, this sort of thing isn't typically checked in other simple scenarios when it comes to return types, so perhaps we shouldn't expect TS to error on this situation.  It may lead to some unexpected side effects to have it be _too_ strict.

For example:
```ts
interface MyProps {
  foo: number;
}

type MyPropsFactory = () => MyProps;

// Error: Object literal may only specify known properties, and 'bar' does not exist in type 'MyProps'
const props: MyProps = { foo: 42, bar: 52 };

// No error
const propFactory: MyPropsFactory = () => ({ foo: 42, bar: 52 });
```

probablyup: > perhaps we shouldn't expect TS to error on this situation

The problem is lack of this is what leads to subtle errors due to typos. If you misspell something by a character it'll still take the prop and not alert you that it's buggy.

justinbhopper: Alrighty!  I managed to add these restrictions, but boy was it difficult.  Like, really difficult.

One tradeoff to note -- I think it slowed down the compilation a bit.  `tsc` of the styled-components folder went from 8sec to 12sec.  Not a huge amount, but just felt like I would note it.

packages/styled-components/src/constructors/constructWithOptions.ts
```diff
@@ -46,6 +48,15 @@ type PropsSatisfiedByAttrs<
   OptionalIntersection<Props, Result> &
   Partial<Omit<Result, keyof Props | 'as'>>;
 
+/**
+ * Rejects an attr factory function argument (T) if it returns any properties not defined in Props.
+ */
+type StrictAttrFactory<T, Props> = T extends ((props: ExecutionContext & Props) => infer TResult)
```
probablyup, line 54: Very clever.

probablyup (approved):
Trust me, I know how insane these types are. I really appreciate your work on this. And thank you @aliceHendicott for contributing the failing tests that informed this work. 🙇🏼

justinbhopper:
My pleasure -- it's the least I can do for everything that styled-components has done for me and my team ❤️

Issue #4019, opened 2023-05-25
Version 6.0.0-rc.2 regressions


DavidReinberger:
After installing the latest release candidate there are some regressions in term of typing results.

### System:
 - OS: macOS 12.6.5
 - CPU: (10) arm64 Apple M1 Pro
 - Memory: 168.86 MB / 32.00 GB
 - Shell: 5.8.1 - /bin/zsh
 
### Binaries:
 - Node: 18.13.0 - ~/.nvm/versions/node/v18.13.0/bin/node
 - Yarn: 1.22.19 - ~/.nvm/versions/node/v18.13.0/bin/yarn
 - npm: 9.4.2 - ~/.nvm/versions/node/v18.13.0/bin/npm

## Regressions

1. `ref` and HTMLAttributes inference is broken
     - `ref` typing is `any`
     - while using `as`, `ref` is not a valid prop
     - using `as="button"` on a `div` does not correctly infer `onClick` for instance
     - I've identified [this change](https://github.com/styled-components/styled-components/compare/v6.0.0-rc.1...v6.0.0-rc.2#diff-adc775db811d749ad367a5fd61454d85f453f549d088fd97ea1575c3eb3290f4R161-R163) by @justinbhopper as the cause. Rolling back the change solves the issue.

2. Prop inference from `css<props>` is not working on built components.

```typescript

const style = css<{prop?: boolean}>``;

const Box = styled.div`
    ${style}
`

const Usage = () => (
    <Box prop>something</Box>
)
```

In this case `prop` throws 'property does not exists on type...`.  Because the types are not generated correctly when built. Had the fix and lost it down the road 😩... 

3. The new `attrs` implementation by @justinbhopper breaks `css<Props>` with `.attrs` usage and HTML props within `.attrs` factory
    - `attrs` function props are not typed. In the below example `$textVariant` is `any`,
    ```typescript
    styled.label.attrs<TextProps>(({ $textVariant = 'footnote' }) => ({$textVariant}))
    ```
    - I've isolated the above problem to [this line](https://github.com/styled-components/styled-components/commit/d31b393c05947563e804b83b109b845d5fc00220#diff-330e887239a86c15b1f5d462583b1ac0f48a8cac1b2a12b0ea1dc015dc75757fR98) and fixed it by replacing the default typing to `((props: ExecutionContext & Props) => Props)`
    - The `css<Props>` problem, is caused by the same line and fixed by the same replacement
[...]

PR #4022, opened 2023-05-26
Add failing tests to fix regressions concerning type inference from component props

DonnyVerduijn:
Added failing tests to fix regressions in v6.0.0-rc.2
#4019


probablyup:
Folded into https://github.com/styled-components/styled-components/pull/4028

PR #4027, opened 2023-05-30
Add failing TS tests for `forwardedAs` and css prop

aliceHendicott:
Add failing type tests in `v6.0.0-rc.2` for the following issues:

1. `forwardedAs` prop is currently erroring, stating it does not exist.

Screen Shot 2023-05-30 at 2 08 51 pm

2. `css` prop does not accept a simple string (as per docs example - https://styled-components.com/docs/api#css-prop)
    - This works if the string is passed as part of an array but not as just a string

Screen Shot 2023-05-30 at 2 10 32 pm

### DIFF
--- a/packages/styled-components/src/test/types.tsx
+++ b/packages/styled-components/src/test/types.tsx
@@ -84,6 +84,8 @@ declare module 'react' {
   `}
 />;
 
+<Button css="padding: 0.5em 1em;" />;
+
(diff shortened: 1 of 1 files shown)


probablyup:
Folded into [ed47cc9](https://github.com/styled-components/styled-components/commit/ed47cc988a84f3ce42f578515a29254fbd42a077)

PR #4028, opened 2023-05-30
Regression Fixes

### LINKED ISSUE #4019: Version 6.0.0-rc.2 regressions

justinbhopper:
This is a large overhaul of the `IStyledComponent` interface.  I took some techniques were taken from the DefinitelyTyped v5 type definitions as a way to solve inference and ambiguity problems that the current definitions are suffering from.

Some notable things:

1. Performance is greatly improved.  Compilation of the repo went from 8sec to 3sec.
2. ~It is not possible to disallow unknown props in `attrs()` (both the factory and non-factory overload).  The reason is because we depend on TS to infer the values they provide in order to know which attrs become optional.  I was not able to make TS both infer and enforce restrictions on the type, and any attempt to do so just caused it to resort to the `any` type on the props.~

Fixes #4019, #4022, and #4027


packages/styled-components/src/test/types.tsx
```diff
@@ -266,12 +284,53 @@ const StyledDiv = styled.div``;
...
+// bar should be typed boolean
+const StyledCustomComponentWithAttrs = styled(
+  CustomComponentWithRequiredProps
+).attrs((props) => ({
+  foo: props.foo === 'black' ? 'red' : 'black',
+  hello: (world) => props.hello(world),
+  bar: !props.bar,
+}));
+
+const StyledCustomComponentWithBadAttrs = styled(
```
DonnyVerduijn, line 325: this needs a @ts-expect-error comment

justinbhopper: Fixed this, and also updated the negative test case examples

DonnyVerduijn:
LGTM

DonnyVerduijn (commented):
The regressions in v6.0.0-rc.1 have been resolved in this PR. From giving it a quick look everything looks okay to me. I'm not aware of any performance degradation in terms of type inference, but i didn't investigate.

probablyup:
Looking at this locally and I see some TS errors in `types.tsx`

probablyup (approved):
Awesome

PR #4033, opened 2023-06-05
ensure useTheme hook returns theme object

nksfrank:
Right now the return type for `useTheme: DefaultTheme | undefined` cause the usage of the hook to be extremely inconvenient where it might be `undefined`, using the hook need to verify the themes existence inside every component.

```typescript
export default () => {
  const theme: DefaultTheme | undefined = useTheme();

  return (
    <div>
      {theme.foo.bar} // 👈 this gives an error because theme could be undefined.
    <div>
  );
}
```

My suggestion is to add a check, inside the hook, that the theme is set using a `ThemeProvider`, otherwise throw an error for using the hook without a `ThemeProvider`.

### DIFF
--- a/packages/styled-components/src/models/ThemeProvider.tsx
+++ b/packages/styled-components/src/models/ThemeProvider.tsx
@@ -66,11 +66,20 @@ function mergeTheme(theme: ThemeArgument, outerTheme?: DefaultTheme): DefaultThe
   return outerTheme ? { ...outerTheme, ...theme } : theme;
 }
 
+/**
+ * Returns the current theme (as provided by the closest ancestor `ThemeProvider`.)
+ *
+ * If no `ThemeProvider` is found, the function will error. If you need access to the theme in an
+ * uncertain composition scenario, `React.useContext(ThemeContext)` will not emit an error if there
+ * is no theme.
+ */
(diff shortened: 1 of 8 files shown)


robinclaes:
We encountered the same as we moved from @types to v6, I like this suggestion.

packages/styled-components/src/models/ThemeProvider.tsx
```diff
@@ -66,8 +66,12 @@ function mergeTheme(theme: ThemeArgument, outerTheme?: DefaultTheme): DefaultThe
   return outerTheme ? { ...outerTheme, ...theme } : theme;
 }
 
-export function useTheme(): DefaultTheme | undefined {
-  return useContext(ThemeContext);
+export function useTheme(): DefaultTheme {
+  const theme = useContext(ThemeContext);
+  if(!theme) {
+    throw Error('can not use useTheme outside a ThemeProvider');
```
probablyup, line 72: We have a `StyledError` class that should be used here for consistency. Can you also make sure you're applying prettier formatting please?

nksfrank: I will update the PR with the suggested changes

probablyup (approved):

PR #4042, opened 2023-06-15
Compatibility with frozen Object prototype, part 2

jportner:
## Background

This PR is a follow-on to #3963. To summarize:

* Freezing the global Object prototype is one of the best ways to defend against Prototype Pollution attacks
* Doing this changes all properties of the Object prototype to become non-writable
* JavaScript does not allow you to use the "=" assignment operator to "shadow" any inherited, non-writable object properties -- in particular, the "toString" property
* The correct way to do this is to use the `Object.defineProperty()` syntax instead

## Description

I wasn't aware that this package uses Babel to transform ES6 class methods to object assignment.

For example, take this class:

```ts
class Foo {
  constructor() {}
  toString = () => 'bar';
}
```

If you have frozen your global Object prototype, that ES6 class method syntax would still work just fine.

However, this project uses Babel to transpile its code, which changes the ES6 class to a plain function:

```ts
var Foo = function Foo() {
  this.toString = function () {
    return 'bar';
  };
};
```

So, if you have frozen your global Object prototype, and you try to stringify an instance of `Foo`, it will not work.

The solution here is the same as in #3963. This PR has two commits:

1. 84fb41bd8cb1b951f2f25a79bd66caacfd560e78 -- add tests that are passing
2. efba71241826685948716d0aefb3e6ca62725372 -- change existing ES6 classes that define "toString" methods to use the `Object.defineProperty()` syntax instead

I also have another PR ready with the same changes for the legacy-v5 branch: #4043

### DIFF
--- a/packages/styled-components/src/models/Keyframes.ts
+++ b/packages/styled-components/src/models/Keyframes.ts
@@ -1,6 +1,7 @@
 import StyleSheet from '../sheet';
 import { Keyframes as KeyframesType, Stringifier } from '../types';
 import styledError from '../utils/error';
+import { setToString } from '../utils/setToString';
 import { mainStylis } from './StyleSheetManager';
 
 export default class Keyframes implements KeyframesType {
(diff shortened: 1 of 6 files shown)


probablyup (approved):
Thank you! Learned something :)

PR #4069, opened 2023-06-30
fix: StyleSheetManager must accept undefined props

gineika:
If `exactOptionalPropertyTypes` is enabled in `tsconfig.json`, this code should work:
```js
import { type ReactElement, type ReactNode } from 'react'
import { StyleSheetManager, type ServerStyleSheet } from 'styled-components'

export const Test = (
  { children, sheet }: {
    readonly children: ReactNode
    readonly sheet?: undefined | ServerStyleSheet['instance'] // defined only on server side
  },
): ReactElement => {
  return (
    <StyleSheetManager
      sheet={sheet}
    >
      {children}
    </StyleSheetManager>
  )
}
```

v5 types:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b384d9e82713dc8400a1289eeb9be7f2fc467390/types/styled-components/index.d.ts#L400-L407


probablyup (approved):
Thanks

PR #4078, opened 2023-07-06
fix: StyleSheetManager missing Stylis types

### LINKED ISSUE #4051: Cannot find namespace 'stylis' / Vite
Hi,
I'm trying to configure Styled Components V6 with Vite and Storybook and during the build, I get the below error:
```
node_modules/styled-components/dist/models/StyleSheetManager.d.ts:58:21 - error TS2503: Cannot find namespace 'stylis'.
58     stylisPlugins?: stylis.Middleware[];
```

Any solution? Thanks!

MattIPv4:
Hopefully, this should resolve #4051

### DIFF
--- a/packages/styled-components/src/models/StyleSheetManager.tsx
+++ b/packages/styled-components/src/models/StyleSheetManager.tsx
@@ -1,5 +1,6 @@
 import React, { useContext, useEffect, useMemo, useState } from 'react';
 import shallowequal from 'shallowequal';
+import type stylis from 'stylis';
 import StyleSheet from '../sheet';
 import { ShouldForwardProp, Stringifier } from '../types';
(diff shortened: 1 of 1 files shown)


packages/styled-components/src/models/StyleSheetManager.tsx
```diff
@@ -1,5 +1,6 @@
 import React, { useContext, useEffect, useMemo, useState } from 'react';
 import shallowequal from 'shallowequal';
+import type stylis from 'stylis';
```
MattIPv4, line 3: I'm also open to switching this to `import * as stylis from 'stylis';` to match the other imports in the codebase (it seems type-only imports aren't used much elsewhere?)

probablyup (approved):
Thank you!

Issue #4085, opened 2023-07-10
Event handlers are no longer typed


tim-field:
## System:
 - OS: macOS 13.4.1
 - CPU: (8) arm64 Apple M1
 - Memory: 87.84 MB / 16.00 GB
 - Shell: 3.6.0 - /opt/homebrew/bin/fish
## Binaries:
 - Node: 18.16.1 - ~/.asdf/installs/nodejs/18.16.1/bin/node
 - Yarn: 1.22.19 - /opt/homebrew/bin/yarn
 - npm: 9.5.1 - ~/.asdf/plugins/nodejs/shims/npm
 - Watchman: 2023.05.22.00 - /opt/homebrew/bin/watchman
## npmPackages:
 - babel-plugin-styled-components: 2.1.4 => 2.1.4 
 - styled-components: 6.0.3 => 6.0.3 

## Regression

Event handles are no longer typed after v6 upgrade,  [There is an issue that mentions this](https://github.com/styled-components/styled-components/issues/4019#issuecomment-1562866203), and suggests its been fixed but I'm on 6.0.3 and am still seeing this issue. 

## Reproduction
```TS
const NameField = styled.input``
const r = (
  <NameField
    defaultValue="test"
    onChange={ev => { // <-- ev is typed as any
      console.log(ev.target)
    }}
  />
)
```

Issue #4052, opened 2023-06-25
Typescript type checking error about component callback parameter


Zousdie:
## System:
 - OS: Linux 5.15 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
 - CPU: (2) x64 AMD EPYC
 - Memory: 552.62 MB / 2.01 GB
 - Container: Yes
 - Shell: 5.8 - /bin/zsh
## Binaries:
 - Node: 18.15.0 - /usr/local/bin/node
 - Yarn: 1.22.19 - /usr/local/bin/yarn
 - npm: 9.5.0 - /usr/local/bin/npm
## npmPackages:
 - styled-components: 6.0.0-rc.6 => 6.0.0-rc.6 

## Reproduction
https://codesandbox.io/p/sandbox/zen-jepsen-6nhsdm

I'm working on styled-component upgrade, I find an issue about typescript and I created a simple demo about it:
```typescript
import * as React from 'react';
import styled from 'styled-components';

const Test = styled.div`
  color: red;
`;

type TOnClick = React.ComponentProps<typeof Test>['onClick'];

const x: NonNullable<TOnClick> = (e) => {
  e.stopPropagation();
};

const App = () => {
  return (
    <>
      <Test onClick={x} />
      <Test
        onClick={(e) => {
          e.stopPropagation();
        }}
      />
    </>
  );
};

export default App;
```
Typescript cannot correctly infer the param type of the `onClick` of the second Test component. And this example is working in version 5.

## Steps to reproduce
```
> yarn exec tsc

yarn exec v1.22.19
src/App.tsx:19:19 - error TS7006: Parameter 'e' implicitly has an 'any' type.

19         onClick={(e) => {
                     ~


Found 1 error in src/App.tsx:19

error Command failed.
Exit code: 2
Command: tsc
```
<img width="667" alt="image" src="https://github.com/styled-components/styled-components/assets/20764411/c5ecbe17-5fe8-4897-bbd8-29c3c73d8ecb">

## Expected Behavior
Typescript should automatically infer the parameter type of the component's callback.

## Actual Behavior
tsc error

paavohuhtala:
I can confirm this happens with all(?) built-in DOM callbacks in 6.0.0-rc.6. In previous versions TSC is able to infer the types of the callbacks with no issue, but with the latest RC you need to specify the types by hand. Interestingly TypeScript in VS Code is still able to autocomplete the fields from `event`, but the actual type check fails afterwards.

[...]

PR #4086, opened 2023-07-11
Fix untyped event handler callbacks 

ziolekjj:
Fixes https://github.com/styled-components/styled-components/issues/4085 and https://github.com/styled-components/styled-components/issues/4052
Let me know what you think :)

### DIFF
--- a/packages/styled-components/src/types.ts
+++ b/packages/styled-components/src/types.ts
@@ -22,6 +22,8 @@ export type BaseObject = {};
 // from https://stackoverflow.com/a/69852402
 export type OmitNever<T> = { [K in keyof T as T[K] extends never ? never : K]: T[K] };
 
+type Distributive<T> = T extends any ? T : never;
+
 type FastOmit<T extends object, U extends string | number | symbol> = {
   [K in keyof T as K extends U ? never : K]: T[K];
 };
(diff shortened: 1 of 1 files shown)


probablyup:
How does it work?

ziolekjj:
From what I inspected here, it seems like type `PolymorphicComponent` has some issues with type inference. 
For instance:  

```ts
const ButtonEventTest = styled.button``;

const ButtonEventTestExample = () => (
  <ButtonEventTest onClick={e => { 
    // Parameter 'e' implicitly has an 'any' type.   
    }} 
  />
);  

const ButtonEventTestExampleWorkaround = () => (
  <ButtonEventTest<void> onClick={e => { 
      // e: React.MouseEvent<HTMLButtonElement, MouseEvent>   
     }}
  />
);  

const ButtonEventTestExampleWorkaround = () => (
  <ButtonEventTest<“a”> onClick={e => { 
    // e: React.MouseEvent<HTMLAnchorElement, MouseEvent>  
    }}
  />
); 
```
On the other hand, when I remove the Substitution part with `ForwardedAsTargetProps` and `AsTargetProps`, from the `PolymorphicComponentProps` type, the callback is inferred correctly (obviously when not using `as` and `forwardedAs`).  
ex.
```ts
export type PolymorphicComponentProps<
...
> = FastOmit<Substitute<BaseProps, {}>, keyof ExecutionProps> & FastOmit<...> & {...}
...

```

From what I see here, the `PolymorphicComponentProps` type can’t correctly default `AsTargetProps` and `ForwardedAsTargetProps` from the generic type declaration when inferring `AsTarget` and `ForwardedAsTarget`. By wrapping the `Omit` part using this distributive type, I can force the inference to work correctly.

Another approach, that could be applied, is to replace `Distributive<T>` with `NoInfer<T>` typings, that are already defined in the codebase, and following the same idea

probablyup (approved):
LGTM thank you!

PR #4097, opened 2023-07-20
Enahnce omit never

kerolossamir165:
// export type OmitNever<T> = { [K in keyof T as T[K] extends never ? never : K]: T[K] };

  if T == never type can not extends never , never is the empty union and it could lead to infinite weird behavior 
  like in is never in the TS src
  https://github.com/microsoft/TypeScript/blob/main/tests/cases/conformance/types/conditional/conditionalTypes1.ts#L212 
 
  if we change IsNever 
          type IsNever<T> = T extends never ? true : false;
           and try it  
                    type Test = IsNever<never>   // it will be never 
 
 so it could be like ts src  
export type OmitNever<T> = { [K in keyof T as [T[K]] extends [never] ? never : K]: T[K] };

### DIFF
--- a/packages/styled-components/src/types.ts
+++ b/packages/styled-components/src/types.ts
@@ -20,7 +20,20 @@ export type StyledComponentBrand = { readonly _sc: symbol };
 export type BaseObject = {};
 
 // from https://stackoverflow.com/a/69852402
-export type OmitNever<T> = { [K in keyof T as T[K] extends never ? never : K]: T[K] };
+// export type OmitNever<T> = { [K in keyof T as T[K] extends never ? never : K]: T[K] };
+/** 
+ * if T == never type can not extends never , never is the empty union and it could lead to infinite weird behavior 
+ * like in is never in the TS Docs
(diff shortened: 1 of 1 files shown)


packages/styled-components/src/types.ts
```diff
@@ -20,7 +20,20 @@ export type StyledComponentBrand = { readonly _sc: symbol };
...
+ * like in is never in the TS src
+ * https://github.com/microsoft/TypeScript/blob/main/tests/cases/conformance/types/conditional/conditionalTypes1.ts#L212 
+ * 
+ * if we change IsNever 
+ *          type IsNever<T> = T extends never ? true : false;
+ *          ant try it  
+ *                   type Test = IsNever<never>   // it will be never 
+ * */
+ 
+export type OmitNever<T> = { [K in keyof T as [T[K]] extends [never] ? never : K]: T[K] };
```
probablyup, line 35: Not sure I understand what this is doing, it seems to just be going from `never` to `never[]`?

kerolossamir165: I just figure out that the concepts got mixed up in my mind 
I thought it pass never as generic not as a value 
so it will not make a difference here  

Thanks for your review

Issue #4100, opened 2023-07-21
regression: createGlobalStyle does not remove styles on unmount


mdeschamps:
## System:
 - OS: macOS 13.4.1
 - CPU: (10) arm64 Apple M1 Max
 - Memory: 5.18 GB / 64.00 GB
 - Shell: 3.2.57 - /bin/bash
## Binaries:
 - Node: 18.1.0 - ~/.nvm/versions/node/v18.1.0/bin/node
 - npm: 9.7.2 - ~/.nvm/versions/node/v18.1.0/bin/npm
 - pnpm: 8.6.0 - ~/.nvm/versions/node/v18.1.0/bin/pnpm
## npmPackages:
 - babel-plugin-styled-components: ^2.1.4 => 2.1.4 
 - styled-components: ^6.0.4 => 6.0.4 

## Reproduction

In this change https://github.com/styled-components/styled-components/pull/3726/files#diff-b8702cbcaf39eaa95deb156bf4132fef9fd7096c4f87679cf656105070c38bc8R64 @probablyup changed it to conditionally use `useInsertionEffect` if it exists; but `useInsertionEffect` does not have a return type: https://react.dev/reference/react/useInsertionEffect#returns

After updating to v6 I started noticing that none of the global styles created by `createGlobalStyle` are removed after the component is unmounted.

I couldn't come up with a small reproducible case; in a codepen the global styles are removed; but in our app with hundreds of lines in the `<style/>` those global styles are clearly sticky.

## Expected Behavior

Global style is removed when component is unmounted

## Actual Behavior

Global styles remain in the stylesheet after component is unmounted

probablyup:
That's annoying! Would you be open to shooting over a quick PR to remove use of useInsertionEffect?

mdeschamps:
> That's annoying! Would you be open to shooting over a quick PR to remove use of useInsertionEffect?

sure thing... PR ^

PR #4101, opened 2023-07-21
Fix createGlobalStyle not removing styles on unmount

mdeschamps:
Fixes: https://github.com/styled-components/styled-components/issues/4100

Current implementation conditionally uses React18 `useInsertionEffect`; but useInsertionEffect does not have a return type: https://react.dev/reference/react/useInsertionEffect#returns and the global styles will not be removed when the GlobalStyle component is unmounted.

### DIFF
--- a/packages/styled-components/src/constructors/createGlobalStyle.ts
+++ b/packages/styled-components/src/constructors/createGlobalStyle.ts
@@ -49,8 +49,7 @@ export default function createGlobalStyle<Props extends object>(
     }
 
     if (!__SERVER__) {
-      // @ts-expect-error still using React 17 types for the time being
-      (React.useInsertionEffect || React.useLayoutEffect)(() => {
+      React.useLayoutEffect(() => {
         if (!ssc.styleSheet.server) {
           renderStyles(instance, props, ssc.styleSheet, theme, ssc.stylis);
           return () => globalStyle.removeStyles(instance, ssc.styleSheet);


probablyup (approved):
Thanks!

guillaumebrunerie:
This is pretty strange, `useInsertionEffect` was specifically designed to inject styles in CSS-in-JS libraries, because according to the React team it is a bad idea to do it in `useLayoutEffect`. Is there a bug in React?
And why does it matter that `useInsertionEffect` does not have a return type? The return type is ignored here anyway, and `useLayoutEffect` does not have a return type either.

mdeschamps:
@guillaumebrunerie sorry bad wording... I meant that the setup function while it is the same signature as `useEffect` which can return a `Destructor` cleanup function; it doesnt seem to fire when the component is unmounted. Just checked the documentation again and it mentions that a cleanup function can be returned; so it could a bug in React.

But after switching it back to `useLayoutEffect` our global styles were cleaned up correctly like on SC5 so that definitely caused the regression.

I wanted to add a test but that meant bumping up react to v18 which I guess that is desirable just yet.

guillaumebrunerie:
Could it be https://github.com/facebook/react/issues/26670?

mdeschamps:
And that could totally be it.. We do use React.lazy.. I even tried to repro that yesterday for you guys but I couldnt get it to NOT work 

https://codesandbox.io/s/styled-components-base-forked-9nszy4?file=/index.js

mdeschamps:
@guillaumebrunerie tweaked my code sample a bit more but I think I got it to reproduce the issue https://codesandbox.io/s/styled-components-base-forked-9nszy4?file=/index.js

probablyup:
Happy to reintroduce it at a later date once the mainline bug is resolved

Issue #4099, opened 2023-07-20
TS Error When Using `css` Interpolations in `styled(...)`


joealden:
## Environment
The `envinfo` command didn't work on my system, but I don't think it's too relevant here.

- `styled-components` version: `6.0.4`
- `typescript` version: `5.1.6`

## Reproduction

https://www.typescriptlang.org/play?jsx=4#code/JYWwDg9gTgLgBAbzgYwM6oDR1TAngGwFMATOAXzgDMoIQ4AiHAkgWmVsgDtDOZV6A3ACgheMITgBlPEWIAFGmFRwAvIjgASAG4BDKMB28AXNhj7OAc3LCh7TjmwALPSWnNlatKgA8b2QoglAD4AAyE4FAh8aBMNBDBFDyC4AAoEwNQAOm09A15VFTV6Yj0Aa3o4AH4GACN8HWRyuBN6AHdHYBhCegBKMmEQmzEJAGEOCG5eAKVVdWR69AA5HRBCSpMccwsBOC6cDbNgS2sROwcx8AmeGBMAJUIGmEyAMRHvC65r6dRktRSkeY6JYrQhYPbwMg9VTJBDhOBQQgwACuUE4cG8xGAWhQC1Qy1WKgQgOBqzIQQQ4LI3gA9JitEFhP0RIQAB6QWCRezwPwkD5XfJqJiyFJ8yYwHq+GQkb6hOFxVDOBHEHmoJmDIRAA

```tsx
import { css, styled } from "styled-components";

type StyledProps = { $variant: string };

const sharedStyles = css<StyledProps>`
  color: ${props => (props.$variant === "dark" ? "black" : "white")};
`;

type ComponentProps = { className?: string; test: string };

const Component: React.FC<ComponentProps> = ({ className, test }) => {
  return <div className={className}>{test}</div>;
};

export const StyledComponent = styled(Component)<StyledProps>`
  ${sharedStyles}; // this line shows a TS error
`;

```

## Steps to reproduce

1. Create a value to interpolate via the `css` function - this needs to be typed to require props.
2. Create a simple component that you can wrap with `styled`.
3. Wrap this component with `styled`, using the interpolation created above and it's types.
4. See TS error where interpolation is used.

FYI, this has been mentioned in https://github.com/styled-components/styled-components/issues/4062 too, but I thought it made sense for this to be a seperate issue.

## Expected Behavior

No TS errors (this worked before v6), or guidance on if there is some other way to do this now.

[...]

PR #4107, opened 2023-07-28
Fix Typescript issues with StyleFunctions and StyledObjects

bcole808:
This PR improves the types when a `StyleFunction` returns another `StyleFunction` or a `StyledObject` as the return value. This prevents type errors when component prop types are explicitly provided (i.e. `styled.div<Props>`)

Examples have been addd in `test/types.tsx` for three different test cases:

1. When a reusable `css` block is provided as an interpolation value
2. When a StyleFunction returns another StyleFunction
3. When a StyleFunction returns a StyledObject

This fixes https://github.com/styled-components/styled-components/issues/4099

### DIFF
--- a/packages/styled-components/src/test/types.tsx
+++ b/packages/styled-components/src/test/types.tsx
@@ -448,7 +448,7 @@ styled.div<PropsWithVariant>`
 /**
  * Styled object as function return value when prop types are provided
  */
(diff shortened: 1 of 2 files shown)


probablyup:
Thanks will review this today

probablyup (approved):
Thank you! This looks like a good fix and appreciate the tests :)

Issue #4090, opened 2023-07-14
Support for numeric values in <length> props using Style Objects in styled-components v6


bexsoft:
In previous version (v5.3.6), we had support to use numeric values without units in props like width, height, maxWidth, maxHeight and so on, now with v6 this is not supported anymore by getting the following error:

`Type '10' is not assignable to type 'MarginRight<0 | (string & {})> | undefined'.` 

This is a standard for React inline styling in React, According Documentation there is the following statement:

```html
React will automatically append a “px” suffix to certain numeric inline style properties. If you want to use units other than “px”, specify the value as a string with the desired unit. For example:

// Result style: '10px'
<div style={{ height: 10 }}>
  Hello World!
</div>

// Result style: '10%'
<div style={{ height: '10%' }}>
  Hello World!
</div>
```

This is our code
```es6
const SubItemsBox = styled.div(({ theme }) => ({
  position: "fixed",
  top: 0,
  left: 0,
  width: "100vw",
  height: "100vh",
  marginRight: 10,
  ...
});
```
Is this support removed on purpose? If so I think it should be documented as a breaking change

Thank you in advance.

probablyup:
I think this has to do with the types being pulled in from csstype... will need to look into it

juliuscalvium:
This is causing me upgrade issues, I'd really prefer to not refactor hundreds of files to convert numeric values to strings in order to use v6. Any updates on whether this is going to be fixed for numeric value use?

probablyup:
This is unintentional and should be fixed

probablyup:
My time is stretched pretty thin at the moment, if someone could look into how we'd modify or augment the underlying csstype type to allow raw numbers would appreciate it

juliuscalvium:
> My time is stretched pretty thin at the moment, if someone could look into how we'd modify or augment the underlying csstype type to allow raw numbers would appreciate it

Thank you. I'm afraid the typings are a bit too complicated for me to get my head around right now.

PR #4111, opened 2023-08-02
fix(types): allow number for height/width/etc (#4090)

drewbrend:
Fixes https://github.com/styled-components/styled-components/issues/4090

The `TLength` generic from `CSS.Properties` defaults to `TLength = 0 | (string & {})`, I verified that before my change I am able to provide `0`, but any other size needs to be a string.

With this changed to `number | (string & {})` I can provide any number or string.

I also added a test for this, providing the number as a height presented a type error before this fix but is resolved after.


juliusbangert (commented):
These changes fix the issue for me. I no longer have hundreds of ts errors related to the use of numbers for heights and widths etc for the object syntax native implementation of styled components.

probablyup (approved):
_Fantastic_ thank you @drewbrend!

bexsoft:
Thank you very much @drewbrend! =D

Issue #4062, opened 2023-06-28
missing types in v6


why-this-code-works:
After upgrading to v6 some types are missing:

<img width="994" alt="CleanShot 2023-06-28 at 16 26 14@2x" src="https://github.com/styled-components/styled-components/assets/97026035/3b023ce8-2bb4-4594-b87e-223a35125d89">

In our case missing types are: 

-   `FlattenSimpleInterpolation`
-  `FlattenInterpolation`
-  `ThemeProps`

AntonNiklasson:
It would be great to collect some resources on how to migrate from `@types/styled-components` to this new setup with "native" types. We're consuming a few of the types from that packages, and I'm not sure how to proceed with our migration.

guillaumebrunerie:
The type `StyledComponentProps` is also missing. I am not sure if there is another way to get a type of the props of a styled component?

```typescript
const SSearchField = styled.input`
  [...]
`;

const Search = (props: StyledComponentProps<"input", object, object, never>) => {
  return (
    <SSearchSection>
      <SLookingGlass><Icons.Search/></SLookingGlass>
      <SSearchField type="search" spellCheck={false} {...props}/>
    </SSearchSection>
  );
};
```

na-ji:
Another type missing in v6 is `ThemedStyledComponentsModule`. This allows to type your theme [without overriding styled-components type](https://github.com/styled-components/styled-components/issues/1589#issuecomment-459075819). This is extremely useful on a monorepo when you need to export multiple theme. It is also efficient when exporting a component library. In my case, it's not possible to upgrade my monorepo to v6 without a similar solution.

probablyup:
Just fyi there was no effort to sync repo types to definitelytyped's setup. A type migration guide would be great if someone wants to contribute one over on the website repo

jphawk:
Same, we can't update to v6, because the following types are missing: 

-   CSSObject
-   FlattenInterpolation
-   FlattenSimpleInterpolation
-   Interpolation
-   InterpolationFunction
-   SimpleInterpolation
-   ThemedStyledProps
-  StyledComponentBase
-  StyledComponent

A migration guide or at least a proper changelog is much needed 🙏

jstol:
[...]

PR #4117, opened 2023-08-07
Add missing types CSSProperties, CSSObject, CSSPseudos and CSSKeyframes

takurinton:
https://github.com/styled-components/styled-components/issues/4062

I found that styled-components has lost some types since v6. There is actually an issue.

I added some of them.

- CSSProperties
- CSSObject
- CSSPseudos
- CSSKeyframes

I tried to work on the other types, but they were too complicated, so I am dealing with what I can for now.

### DIFF
--- a/packages/styled-components/src/types.ts
+++ b/packages/styled-components/src/types.ts
@@ -237,15 +237,25 @@ export interface IInlineStyle<Props extends object> {
   generateStyleObject(executionContext: ExecutionContext & Props): object;
 }
 
-export type StyledObject<Props extends object> = CSS.Properties<number | (string & {})> & {
-  [key: string]:
-    | string
-    | number
-    | StyleFunction<Props>
-    | StyledObject<Props>
-    | RuleSet<any>
-    | undefined;
-};
+export type CSSProperties = CSS.Properties<number | (string & {})>;
+
+export type CSSPseudos = { [K in CSS.Pseudos]?: CSSObject };
+
+export type CSSKeyframes = object & { [key: string]: CSSObject };
+
+export type CSSObject = StyledObject<any>;
+
+export type StyledObject<Props extends object> = CSSProperties &
+  CSSPseudos & {
+    [key: string]:
+      | StyledObject<any> // StyledObject<any> instead of CSSObject. Because writing CSSObject directly results in a circularly references.
+      | string
+      | number
+      | StyleFunction<Props>
+      | StyledObject<Props>
+      | RuleSet<any>
+      | undefined;
+  };
 
 /**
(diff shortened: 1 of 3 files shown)


probablyup:
I think `CSSObject` is basically the same as `StyledObject`, maybe we just need to add what's missing to `StyledObject` and just alias it.

takurinton:
@probablyup 
I certainly thought so, so I fixed it here. What do you think?

https://github.com/styled-components/styled-components/pull/4117/commits/dea5bfd2d60ad05359b992c51c1b13dd3b1487a1

packages/styled-components/src/types.ts
```diff
@@ -230,15 +237,25 @@ export interface IInlineStyle<Props extends object> {
...
-export type StyledObject<Props extends object> = CSS.Properties<number | (string & {})> & {
-  [key: string]:
-    | string
-    | number
-    | StyleFunction<Props>
-    | StyledObject<Props>
-    | RuleSet<any>
-    | undefined;
-};
+export type CSSProperties = CSS.Properties<number | (string & {})>;
```
probablyup, line 240: Last thing is these types need to be exposed in the barrel files `src/index.ts` and `src/native/index.ts`

takurinton: Thanks, I completely missed that.

And I added it.
https://github.com/styled-components/styled-components/pull/4117/commits/765498df6975b20ae52b04ef6b3b77ac78098f91

probablyup (approved):
Awesome, thank you!

packages/styled-components/src/types.ts
```diff
@@ -230,15 +237,25 @@ export interface IInlineStyle<Props extends object> {
...
+  CSSPseudos & {
+    [key: string]:
+      | StyledObject<any> // StyledObject<any> instead of CSSObject. Because writing CSSObject directly results in a circularly references.
+      | string
+      | number
+      | StyleFunction<Props>
+      | StyledObject<Props>
+      | RuleSet<any>
+      | undefined;
+  };
```
pedroSoaresll, line 258: At lines 251 and 255 we have the same type but with different generics being passed. Is it ok?

probablyup: Ahh, that seems incorrect.

takurinton: @pedroSoaresll 
Thank you for your comment. It is fixed here.
https://github.com/styled-components/styled-components/pull/4140#discussion_r1302280083

takurinton: Furthermore, I found a duplicate type definition, so I'll fix it here. https://github.com/styled-components/styled-components/pull/4200

PR #4118, opened 2023-08-07
Run prettier format

takurinton:
I ran prettier locally and got a diff, so I commit it.

I think this should be prettier check in CI as well, what do you think?


probablyup (approved):

probablyup:
> I think this should be prettier check in CI as well, what do you think?

That or `pretty-quick` on precommit

takurinton:
@probablyup 
I think it is smart to check with CI.
I will make a PR later!!

PR #4125, opened 2023-08-10
Run prettier on CI

takurinton:
refs https://github.com/styled-components/styled-components/pull/4118#issuecomment-1672523534

I thought it would be better to run prettier on CI, so I did the following.

- update prettier v3
- run prettier
- run prettier on CI

### DIFF
--- a/packages/styled-components/.prettierignore
+++ b/packages/styled-components/.prettierignore
@@ -1,0 +1,1 @@
+src/utils/errors.js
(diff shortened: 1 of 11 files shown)


packages/styled-components/.prettierignore
```diff
@@ -0,0 +1 @@
+src/utils/errors.js
```
takurinton, line 1: I created a `.prettierignore` under `packages/styled-components`. 
This is to prevent the CI from failed.
https://github.com/styled-components/styled-components/actions/runs/5817983702/job/15773544427

takurinton:
@probablyup 
I tried to run prettier check on CI. What do you think about this?

probablyup (approved):
:shipit:

PR #4126, opened 2023-08-10
Add missing types AnyIfEmpty, ThemeProps,  ThemedStyledProps, StyledProps, and IntrinsicElementsKeys

takurinton:
refs #4062 #4117 

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/styled-components/index.d.ts#L37-L43

Add more missing types in addition to #4117.

### DIFF
--- a/packages/styled-components/src/index.ts
+++ b/packages/styled-components/src/index.ts
@@ -12,13 +12,13 @@ export {
   ExecutionProps,
   IStyledComponent,
   IStyledComponentFactory,
-  IStyledStatics,
-  PolymorphicComponent,
+  IStyledStatics, IntrinsicElementsKeys, PolymorphicComponent,
   PolymorphicComponentProps,
   RuleSet,
   Runtime,
   StyledObject,
-  StyledOptions,
-  WebTarget,
+  StyledOptions, StyledProps, ThemeProps,
+  ThemedStyledProps, WebTarget
 } from './types';
 export { styled as default, styled };
+
--- a/packages/styled-components/src/types.ts
+++ b/packages/styled-components/src/types.ts
@@ -285,3 +285,13 @@ export type CSSProp = Interpolation<any>;
 export type NoInfer<T> = [T][T extends any ? 0 : never];
 
 export type Substitute<A extends object, B extends object> = FastOmit<A, keyof B> & B;
+
+type AnyIfEmpty<T extends object> = keyof T extends never ? any : T;
+
+export interface ThemeProps<T> {
+  theme: T;
+}
+
+export type ThemedStyledProps<P, T> = P & ThemeProps<T>;
+export type StyledProps<P> = ThemedStyledProps<P, AnyIfEmpty<DefaultTheme>>;
+export type IntrinsicElementsKeys = keyof JSX.IntrinsicElements;
--- a/packages/styled-components/src/native/index.ts
+++ b/packages/styled-components/src/native/index.ts
@@ -110,12 +110,16 @@ export {
   IStyledComponent,
   IStyledComponentFactory,
   IStyledStatics,
+  IntrinsicElementsKeys,
   NativeTarget,
   PolymorphicComponent,
   PolymorphicComponentProps,
   Runtime,
   StyledObject,
   StyledOptions,
+  StyledProps,
+  ThemeProps,
+  ThemedStyledProps,
 } from '../types';
 export {
   ThemeConsumer,


takurinton:
As @grubersjoe has written, the absence of `ThemedStyledProps` can be sort of resolved with this method.
https://github.com/styled-components/styled-components/issues/4062#issuecomment-1630907333

I think it's also a strategy to write this in the migration guide.   
What do @probablyup think??

packages/styled-components/src/index.ts
```diff
@@ -12,13 +12,13 @@ export {
...
-  IStyledStatics,
-  PolymorphicComponent,
+  IStyledStatics, PolymorphicComponent,
   PolymorphicComponentProps,
   RuleSet,
   Runtime,
   StyledObject,
-  StyledOptions,
-  WebTarget,
+  StyledOptions, StyledProps, ThemeProps,
```
probablyup, line 20: This stacking is kind of weird?

packages/styled-components/src/types.ts
```diff
@@ -285,3 +285,12 @@ export type CSSProp = Interpolation<any>;
 export type NoInfer<T> = [T][T extends any ? 0 : never];
 
 export type Substitute<A extends object, B extends object> = FastOmit<A, keyof B> & B;
+
+type AnyIfEmpty<T extends object> = keyof T extends never ? any : T;
+
+export interface ThemeProps<T> {
```
probablyup, line 291: This could definitely be an alias to something else like `ExecutionContext`

probablyup: But it'd need on top of #4103 so we can have the first generic param to `ExecutionContext` be the Theme

takurinton: Should it be defined as an alias? Or should I add to the migration guide?

takurinton: Is it possible to change the direction of the branch?

takurinton: @probablyup 
#4103 is still WIP, so I felt it wouldn't be too late to merge this PR first. 
Until then, how about exporting the conventional `ThemeProps` once?

What do you think??

bcole808: When I upgraded from v5 to v6, I replaced `ThemeProps<T>` with `ExecutionContext` in my codebase. 

If we do decide to add this to ease upgrading that it should be aliased instead of declaring a new type. For example:

```ts
export interface ThemeProps extends ExecutionContext {}
```

packages/styled-components/src/types.ts
```diff
@@ -285,3 +285,12 @@ export type CSSProp = Interpolation<any>;
...
+  theme: T;
+}
+
+export type ThemedStyledProps<P, T> = P & ThemeProps<T>;
+export type StyledProps<P> = ThemedStyledProps<P, AnyIfEmpty<DefaultTheme>>;
```
bcole808, line 296: When I upgraded from v5 to v6, I was able to replace `ThemedStyledProps<P, T>` with `P & ExecutionContext` in my codebase. 

If we do want to add some aliases for the old v5 types to make upgrading easier, we could map them to the new types like this:

```ts
export type ThemedStyledProps<P> = P & ExecutionContext;
```

packages/styled-components/src/types.ts
takurinton: I've done the same in my codebase, but I'm afraid it won't accept generics.
I think I'll either wait for the ExecutionContext to accept generics in #4103, or temporarily include the code in its current state and include it later in #4103.

packages/styled-components/src/types.ts
takurinton: Sure, that would be fine, but it would lead to inconsistencies in the type arguments of `StyledProps`.


I think it's better to wait for this to be addressed.
https://github.com/styled-components/styled-components/pull/4103/files#diff-adc775db811d749ad367a5fd61454d85f453f549d088fd97ea1575c3eb3290f4R75-R76

(and, until then, the old-style types should be defined anyway.)

bcole808 (commented):
👍 This looks reasonable to me. Aliasing these old type names from v5 to the new equivalent types in v6 should make upgrading easier for folks.

packages/styled-components/src/types.ts
```diff
@@ -299,3 +299,8 @@ export type CSSProp = Interpolation<any>;
 export type NoInfer<T> = [T][T extends any ? 0 : never];
 
 export type Substitute<A extends object, B extends object> = FastOmit<A, keyof B> & B;
+
+export interface ThemeProps extends ExecutionContext {}
```
kaedenwile, line 303: ThemeProps was generic in @types/styled-components package: 

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/styled-components/index.d.ts#L25C1-L28C1

takurinton: Yes, it is.

As mentioned in https://github.com/styled-components/styled-components/pull/4126#discussion_r1290796829, it needs to be placed on top of #4103. Or merge this first and then follow.
However, in some cases it should be possible to replace it by using ExecutionContext.

Amerr:
Any blocks in this merging PR or is it still in review

mthorne-godaddy:
Any update on this?

Louis345:
any update on this?

mthorne-godaddy:
Talk about a blast from the past. I gave up on this. 😂 

But, any updates on this?

quantizor:
Closing — we decided in #4062 that v6 intentionally uses different type names. The migration mappings (RuleSet, ExecutionContext, etc.) are documented. Adding compatibility aliases increases surface area without clear benefit. Thanks for the effort!

PR #4127, opened 2023-08-10
Add other missing types

takurinton:
refs #4062 #4117 #4126

Add more missing types in addition to #4126.

I added some of them.

- FlattenInterpolation
- FlattenSimpleInterpolation
- InterpolationFunction
- SimpleInterpolation
- StyledComponentBase

I also added the types they depend on.

referenced here https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/styled-components/index.d.ts


packages/styled-components/src/types.ts
```diff
@@ -80,6 +80,92 @@ export interface StyleFunction<Props extends object> {
...
+  O extends object,
+  A extends keyof P,
+  P = React.ComponentPropsWithRef<
+    C extends IntrinsicElementsKeys | React.ComponentType<any> ? C : never
+  >
+> =
+  // Distribute unions early to avoid quadratic expansion
+  P extends any ? OmitU<ReactDefaultizedProps<C, P> & O, A> & Partial<PickU<P & O, A>> : never;
+
+export type StyledComponentProps<
```
probablyup, line 123: This is a bit overwhelming to add to the list of things to maintain

takurinton: I think so too (I thought it was difficult when I wrote it)

I'm having a hard time deciding how to maintain this complicated type definition, or considering alternatives, so I'd like some advice.
what do you think you should do

packages/styled-components/src/types.ts
```diff
@@ -92,7 +178,21 @@ export type Interpolation<Props extends object> =
   | Keyframes
   | StyledComponentBrand
   | RuleSet<Props>
-  | Interpolation<Props>[];
+  | Interpolation<Props>[]
+  | InterpolationFunction<Props>
+  | FlattenInterpolation<Props>;
+
+export type InterpolationFunction<P extends object> = (props: P) => Interpolation<P>;
```
bcole808, line 185: `InterpolationFunction` and other types here looks like they come from [the Definitely Typed definitions for v5](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/styled-components/index.d.ts#L106).

`InterpolationFunction` from v5 can be replaced with `StyleFunction` from v6. 

Maybe we can either write a migration guide from v5 to v6, or consider just aliasing some of the types that are 1 to 1 replacements? 

I don't think we should copy all of the types over from Definitely Typed because there are new, similar types defined in v6. It might get really confusing to maintain two very similar copies of the same type but with different names.

quantizor:
Closing — superseded by #4126.

Issue #4074, opened 2023-07-05
Impossible to style react-select component with styled-components


dogfootruler-kr:
## Environment
```
## System:
 - OS: macOS 13.4.1
 - CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
 - Memory: 17.73 MB / 16.00 GB
 - Shell: 5.9 - /bin/zsh
## Binaries:
 - Node: 18.16.0 - ~/.nvm/versions/node/v18.16.0/bin/node
 - Yarn: 1.22.19 - ~/.yarn/bin/yarn
 - npm: 9.5.1 - ~/.nvm/versions/node/v18.16.0/bin/npm
 - Watchman: 2023.07.03.00 - /usr/local/bin/watchman
## npmPackages:
 - babel-plugin-styled-components: ^1.13.2 => 1.13.3
 - styled-components: ^6.0.2 => 6.0.2
```

## Reproduction
https://codesandbox.io/s/styled-component-with-react-select-wsd87q

## Steps to reproduce
Styling a components.ValueContainer from [react-select](https://react-select.com/home) is impossible as it expects a `theme` prop that is removed by styled-components.

## Expected Behavior
Possible to style react-select

## Actual Behavior
Impossible to style react-select

pankajpatel:
I believe this is caused by the `theme` prop collision between SC and ReactSelect.

dogfootruler-kr:
That's correct @pankajpatel but is there any solution to this?

CyrilQuandalle:
Hello @dogfootruler-kr ,
Like my colleague @pankajpatel and you were saying there is indeed a prop collision conflict. Having the same issue after making some thinking on it I managed to found a solution :
 `export const DropdownBaseOption: FC<OptionProps<any>> = (props) => {
  const { OptionComponent } = props.selectProps;
  const { theme, ...otherProps } = props;
  return (
    <BaseOptionContainer selectTheme={theme} {...otherProps}>
      {OptionComponent ? <OptionComponent {...props} /> : <div>{props.children} </div>}
    </BaseOptionContainer>
  );
};

const BaseOptionContainer = styled(components.Option).attrs<{ selectTheme: Theme }>((props) => ({
  theme: props.selectTheme,
}))<{ selectTheme: Theme }>``;
`

dogfootruler-kr:
Hello @CyrilQuandalle and thank you for your example but sadly I still have an error when opening the select.

[...]

PR #4129, opened 2023-08-14
Fix don't omit theme at runtime

takurinton:
https://github.com/styled-components/styled-components/issues/4074

Include the `theme` in the  execution props.

Some UI component libraries require a theme at runtime. (e.g. react-select)
If you skip here, you will not be able to refer to the theme when overwriting the theme, resulting in a runtime error.

In fact, if you try using the codesandbox in the [issue comment](https://github.com/styled-components/styled-components/issues/4074#issuecomment-1633601407), A runtime error occurs because the theme cannot be referenced.

> Cannot read properties of undefined (reading 'spacing')

### DIFF
--- a/packages/styled-components/src/models/StyledComponent.ts
+++ b/packages/styled-components/src/models/StyledComponent.ts
@@ -143,7 +143,7 @@ function useStyledComponentImpl<Props extends object>(
     if (context[key] === undefined) {
       // Omit undefined values from props passed to wrapped element.
       // This enables using .attrs() to remove props, for example.
-    } else if (key[0] === '$' || key === 'as' || key === 'theme') {
+    } else if (key[0] === '$' || key === 'as') {
       // Omit transient props and execution props.
     } else if (key === 'forwardedAs') {
       propsForElement.as = context.forwardedAs;
--- a/packages/styled-components/src/models/test/StyleSheetManager.test.tsx
+++ b/packages/styled-components/src/models/test/StyleSheetManager.test.tsx
@@ -316,6 +316,7 @@ describe('StyleSheetManager', () => {
       <div
         bar={true}
         className="sc-a b"
+        theme={{}}
       >
         Foo
       </div>
@@ -381,12 +382,13 @@ describe('StyleSheetManager', () => {
     `);
 
     expect(wrapper.toJSON()).toMatchInlineSnapshot(`
-            <div
-              className="sc-a b"
-            >
-              Foo
-            </div>
-        `);
+      <div
+        className="sc-a b"
+        theme={{}}
+      >
+        Foo
+      </div>
+    `);
 
     act(() => {
       wrapper.update(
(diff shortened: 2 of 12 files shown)


packages/styled-components/src/models/StyledComponent.ts
```diff
@@ -143,7 +143,7 @@ function useStyledComponentImpl<Props extends object>(
     if (context[key] === undefined) {
       // Omit undefined values from props passed to wrapped element.
       // This enables using .attrs() to remove props, for example.
-    } else if (key[0] === '$' || key === 'as' || key === 'theme') {
```
probablyup, line 146: I'd rather not do this, can use `useTheme` to get the theme in your component if needed.

takurinton: Sorry, I didn't know how.
Could you please show me an example?

probablyup: https://styled-components.com/docs/api#usetheme

birmitt: @probablyup I think there is misunderstanding about the intention of this change. 
It is not about to acquire the theme via the theme prop or via the useTheme hook. It is not about the styled-components theming mechanism at all. 
It is about other libs relying on the "theme" prop which is now "killed" by the styled-components wrapper.

Many internal components of react-select passing "theme" props around. When you style them via styled-components, this prop is gone and we ran into a runtime error.
This was also mentioned in #4074.

So changing the line 146 to not remove the "theme" prop fixes that issue with react-select.

Dron007: @quantizor Could we please have this PR added or implement other change to fix the https://github.com/styled-components/styled-components/issues/4074. issue due to which it is impossible to use react-select with styled components?

quantizor: I see. Yeah we can fix this.

quantizor: The only thing I'm not sure about is what should happen if a theme prop is passed that is incompatible with the styled-components theme. The styles still need the context theme to work, whereas the theme being described here has nothing to do with styled-components.

quantizor: Maybe the theme should be a second argument to style interpolations instead of being lumped in with all props. That would disambiguate the two, but we wouldn't be able to remove the s-c theme from the props object until v7 since it would be a breaking change.

quantizor: Further complicating this, we allow the s-c theme to be overridden _via theme prop_ as a general feature...

Dron007: @quantizor Not sure that I understand in details what was changed in s-c v6 but it worked in v5. Here is the [sandbox](https://codesandbox.io/p/sandbox/codesandboxer-example-forked-l7s2ny?file=%2Fexample.tsx%3A18%2C6). If you downgrade s-c to  5.3.11 there is no any error.

quantizor:
I prepared an [alternate implementation](https://github.com/styled-components/styled-components/pull/4242) that will not leak the `theme` prop in normal scenarios. Passing `theme` always would be kind of a breaking change for v6 since the current behavior has been established, but using `attrs` to feed an override `theme` prop seems like a reasonable workaround.

AhmedMuhammedElsaid:
ThanQ

quantizor:
Closing — stale for 2+ years. If the theme omission bug is still present, please open a fresh PR against the current codebase.

Issue #4113, opened 2023-08-04
Regression in styled-components v6 when compiling in production mode


guillaumebrunerie:
## System:
 - OS: macOS 13.3.1
 - CPU: (4) x64 Intel(R) Core(TM) i5-8210Y CPU @ 1.60GHz
 - Memory: 22.09 MB / 16.00 GB
 - Shell: 5.9 - /bin/zsh
## Binaries:
 - Node: 20.2.0 - /usr/local/bin/node
 - npm: 9.6.6 - /usr/local/bin/npm
## npmPackages:
 - babel-plugin-styled-components: ^2.1.4 => 2.1.4
 - styled-components: ^6.0.7 => 6.0.7

## Reproduction

I have the following styled component:
```ts
const Thing = styled.div<{$side: "left" | "right"}>`
  position: absolute;
  width: 10px;
  height: 100%;
  cursor: ew-resize;
  left: ${props => props.$side == "left" && "-5px"};
  right: ${props => props.$side == "right" && "-5px"};
`;
```

It is used as follows
```ts
const Parent = () => {
  return (
    <div>
      <Thing $side="left"/>
      <Thing $side="right"/>
    </div>
  );
};
```

It is supposed to create two divs, the first one with style `left: -5px` and the second one with style `right: -5px`.

It worked fine with styled-components v5, and works fine in development mode with styled-components v6, but breaks in *production mode* with styled-components v6: they both get the same CSS class containing `left: -5px`, for some reason. I checked with the React DevTools and the second one does get the correct `$side: "right"`, so I have no clue what is happening.

I have now changed the left/right lines to
```ts
  ${props => props.$side == "left" && "left: -5px"}
  ${props => props.$side == "right" && "right: -5px"}
```
because the original version feels kind of weird (it relies on `left: ${props => false}` to be removed from the CSS), and it now works.

Is that a known regression?
I feel a bit worried about upgrading to styled-component v6 now, as production-only regressions are pretty hard to catch and to debug, and I don’t know if I use a similar pattern in other places in my project.

~I'd be happy to try to make a reproduction, but I’m not sure how to make the codesandbox compile in production mode?~
See below.

guillaumebrunerie:
Here is a local reproduction. Create a folder with the following four files, run `npm i`, and then compare how the page looks when running `npm run dev` (development version, correct) and `npm run prod` (production version, incorrect).

`package.json`
```json
{
  "name": "styled-components-issue",
  "version": "0.0",
[...]

PR #4132, opened 2023-08-18
Fix production mode inconsistent CSS rendering when dynamic property values are the same but property names are different

bcole808:
This PR fixes https://github.com/styled-components/styled-components/issues/4113

This problem occurs when dynamic interpolations in different places can return the same value. 

An example of how this problem can happen:

```ts
const MyComponent = styled.div<{ $variant: 'text' | 'background' }>`
      color: ${props => props.$variant == 'text' && 'red'};
      background-color: ${props => props.$variant == 'background' && 'red'};
`;
```

In the example above, there are two interpolation functions. Each returns the exact same value, but appears next to a different CSS property name. 

In production builds, the logic which determines if a new class name needs to be created fails to check for this situation and outputs the same CSS for both `<MyComponent $variant="text" />` and `<MyComponent $variant="background" />`.

In non-production builds, this problem does not occur.

### DIFF
--- a/packages/styled-components/src/test/basic.test.tsx
+++ b/packages/styled-components/src/test/basic.test.tsx
@@ -66,7 +66,7 @@ describe('basic', () => {
     `);
   });
 
-  it('should inject two different styles if the same compnoent is mounted with different props and css', () => {
+  it('should inject two different styles if the same component is mounted with different props and css', () => {
     const Comp = styled.div<{ $variant: 'text' | 'background' }>`
       color: ${props => props.$variant == 'text' && 'red'};
       background-color: ${props => props.$variant == 'background' && 'red'};
--- a/packages/styled-components/src/test/basic.test.tsx
+++ b/packages/styled-components/src/test/basic.test.tsx
@@ -66,6 +66,23 @@ describe('basic', () => {
     `);
   });
 
+  it('should inject two different styles if the same compnoent is mounted with different props and css', () => {
+    const Comp = styled.div<{ $variant: 'text' | 'background' }>`
+      color: ${props => props.$variant == 'text' && 'red'};
+      background-color: ${props => props.$variant == 'background' && 'red'};
+    `;
+    TestRenderer.create(<Comp $variant="text" />);
+    TestRenderer.create(<Comp $variant="background" />);
+    expect(getRenderedCSS()).toMatchInlineSnapshot(`
+      ".b {
+        color: red;
+      }
+      .c {
+        background-color: red;
+      }"
+    `);
+  });
+
   it('Should have the correct styled(component) displayName', () => {
     const CompWithoutName = () => (() => <div />) as React.FC<any>;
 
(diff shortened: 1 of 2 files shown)


guillaumebrunerie:
Nice! Looking at the log, the latest commit that changed this line was https://github.com/styled-components/styled-components/commit/d5fe2495b94a037234147bba6febfd8473dd86ac which removed the `+ i` from the two lines `dynamicHash = phash(dynamicHash, partString + i);`. Maybe that would fix it as well? This `+ i` was added in https://github.com/styled-components/styled-components/commit/fff5c1e4ab30ddf131dd6ff780f6a52ac480287d to fix an issue very similar to my bug report.

bcole808:
@guillaumebrunerie Good observation! 

After looking at those commits more closely, I think https://github.com/styled-components/styled-components/commit/d5fe2495b94a037234147bba6febfd8473dd86ac is the commit that introduced this bug. The `+ 1` for parts that are NOT a string was important for this use case. 

This bug happens when there is something dynamic (i.e. a function) in the CSS which results in the same hash but could change positions in the list of interpolations.  Removing the `if (process.env.NODE_ENV !== 'production')` conditional fixes the problem by accounting for parts of the CSS which are static strings, which accounts for when a dynamic interpolation changes position in the list of interpolations. 

### What about performance of a production build?

By removing `if (process.env.NODE_ENV !== 'production') `, **more** hashing happens in production builds. 

I ran the benchmarks briefly on my machine and observed a speed of **4.36 on main** compared to **4.43 on this PR**. This represents a decrease in production performance of about 1.6% (If someone else can also run the benchmarks that would be great)

### Is there a way we can solve this bug without impacting production performance?

Yes, A different way could be to put `+ i` back into the other check for when `partRule` is not a string.  The commit message on https://github.com/styled-components/styled-components/commit/fff5c1e4ab30ddf131dd6ff780f6a52ac480287d explains: 

> When the same interpolations swap places they still produce the same hash which can sometimes lead to the same hash.

```ts
else if (partRule) {
    const partString = joinStringArray(
      flatten(partRule, executionContext, styleSheet, stylis) as string[]
    );
    dynamicHash = phash(dynamicHash, partString + i); // <---- Add + i here
    css += partString;
}
```

This would **also** solve the problem. 

### So what are the options?

**Option 1: Remove `if (process.env.NODE_ENV !== 'production')`**

* This makes the behavior consistent between production and development modes at the cost of impacting production performance slightly. 

**Option 2: Add `+ i` back into the hash for non-strings**

* This fixes the issue without affecting performance, but arguably is harder to understand and maintain because it isn't obvious why `+ i` is necessary. 

@probablyup What do you think? What is the right balance here between production performance and understandability?

packages/styled-components/src/test/basic.test.tsx
```diff
@@ -66,6 +66,23 @@ describe('basic', () => {
     `);
   });
 
+  it('should inject two different styles if the same compnoent is mounted with different props and css', () => {
```
probablyup, line 69:
```suggestion
  it('should inject two different styles if the same component is mounted with different props and css', () => {
```

probablyup:
Performance is more important, let's go with option 2. Code comments appreciated :)

bcole808:
Thanks @probablyup, I've switched to option 2 and added a comment to explain it. The added unit test should also catch this regression in the future as well. 

PS: I'd love to be added as a contributor so I can push branches directly to the repo, and also help review PRs! It looks like there used to be a mechanism to auto-invite contributors, but it's no longer working. [I posted over here about it as well.](https://github.com/orgs/styled-components/discussions/4142)

probablyup:
One other option would be to derive from domElements.ts. That's actually
the list that s-c utilizes so would be even smaller than intrinsic elements
since not every html element can be styled

bcole808:
Not sure I follow what you're saying? I don't think this PR has anything to do with specific html elements, right?

probablyup:
Sorry this was meant for another PR

probablyup (approved):
Awesome, thanks for looking into this.

PR #4140, opened 2023-08-22
Add type exports for StyleFunction, Interpolation

bcole808:
Adds two more types to the main export, and also adjusts the type of `CSSObject` to be an exact alias of `StyledObject`

Follow up to https://github.com/styled-components/styled-components/pull/4117


packages/styled-components/src/types.ts
```diff
@@ -243,12 +243,12 @@ export type CSSPseudos = { [K in CSS.Pseudos]?: CSSObject };
...
 export type CSSKeyframes = object & { [key: string]: CSSObject };
 
-export type CSSObject = StyledObject<any>;
+export type CSSObject<Props extends object = BaseObject> = StyledObject<Props>;
 
-export type StyledObject<Props extends object> = CSSProperties &
+export type StyledObject<Props extends object = BaseObject> = CSSProperties &
   CSSPseudos & {
     [key: string]:
-      | StyledObject<any> // StyledObject<any> instead of CSSObject. Because writing CSSObject directly results in a circularly references.
```
bcole808, line 251: Hey @takurinton, can we change this back to `StyleObject<Props>`?  I think it should work as long as we are referencing `StyleObject` and not `CSSObject`, right? The circular reference would only occur if we referenced `CSSObject` here.

Or Is there another reason that this causes a problem?

takurinton: That's right, I think it's good!!

probablyup (approved):
Sorry for the delay, looks good on my end

PR #4149, opened 2023-08-30
fix: union type too complex

### LINKED ISSUE #4115: Expression produces a union type that is too complex to represent in v6
I've been trying to upgrade a few projects to styled-components 6 and have been running into some typescript errors. I've narrowed it down to an interaction between `@react-three/fiber`, `gatsby-plugin-image`, and `styled-components` that wasn't occurring in styled-components 5.

## System:
 - OS: macOS 13.0.1
 - CPU: (10) arm64 Apple M1 Pro
 - Memory: 71.77 MB / 16.00 GB
 - Shell: 5.8.1 - /bin/zsh
## Binaries:
 - Node: 18.17.0 - ~/programming/reform/testing-issue/node_modules/.bin/node
 - Yarn: 1.22.19 - ~/.yarn/bin/yarn
 - npm: 9.8.1 - ~/.nvm/versions/node/v18.16.1/bin/npm
## npmPackages:
 - babel-plugin-styled-components: ^2.1.4 => 2.1.4 
 - styled-components: ^6.0.7 => 6.0.7

## Reproduction

[github repository](https://github.com/RJWadley/styled-components-typescript-bug) and [codesandbox](https://githubbox.com/RJWadley/styled-components-typescript-bug)

## Steps to reproduce

1. Create a new gatsby project with `npm init gatsby`
2. In the setup process, install responsive images and styled components.
3. Install `@react-three/fiber`
4. Create a d.ts file with the following content:
```
import { ReactThreeFiber } from "@react-three/fiber";

declare global {
  namespace JSX {
    interface IntrinsicElements {
      anyFunNameYouWant: {};
    }
  }
}
```
5. Extend GatsbyImage with styled like so:
```
import { GatsbyImage } from "gatsby-plugin-image";

const Test = styled(GatsbyImage)<{ backgroundColor: string }>`
  background-color: ${(props) => props.backgroundColor}; // <-- Expression produces a union type that is too complex to represent. ts(2590)
`;
```

## Expected Behavior
No error in styled-components 5
No error in styled-components 6

## Actual Behavior
No error in styled-components 5
Typescript error in styled-components 6

RJWadley:
As libraries like `@react-three/fiber` add to JSX.IntrinsicElements, it can get too large to convert to a union in some scenarios. I'm using the tag name maps from `lib.dom.d.ts` instead. In an environment with no JSX augmentations, this results in the same union.

For example, in the case of `@react-three/fiber`, the original union looks something like this:
`'h1' | 'h2' | 'h3' | 'meshPhysicalMaterial' | 'meshStandardMaterial' | 'meshPhongMaterial' ...`

And the new one only includes standard HTML tag names:
`'h1' | 'h2' | 'h3' ...`

From what I can tell, this is what makes sense. (Any edge cases this would screw up?)

fixes #4115

### DIFF
--- a/packages/styled-components/src/base.ts
+++ b/packages/styled-components/src/base.ts
@@ -56,13 +56,9 @@ if (
 export * from './secretInternals';
 export { Attrs, DefaultTheme, ShouldForwardProp } from './types';
 export {
-  createGlobalStyle,
-  css,
-  isStyledComponent,
-  IStyleSheetManager,
   IStyleSheetContext,
+  IStyleSheetManager,
   IStylisContext,
-  keyframes,
   ServerStyleSheet,
   StyleSheetConsumer,
   StyleSheetContext,
@@ -70,6 +66,10 @@ export {
   ThemeConsumer,
   ThemeContext,
   ThemeProvider,
+  createGlobalStyle,
+  css,
+  isStyledComponent,
+  keyframes,
   useTheme,
   SC_VERSION as version,
   withTheme,
--- a/packages/styled-components/src/index.ts
+++ b/packages/styled-components/src/index.ts
@@ -21,6 +21,7 @@ export {
   StyleFunction,
   StyledObject,
   StyledOptions,
+  SupportedHTMLElements,
   WebTarget,
 } from './types';
 export { styled as default, styled };
--- a/packages/styled-components/src/types.ts
+++ b/packages/styled-components/src/types.ts
@@ -32,7 +32,7 @@ export type AnyComponent<P extends object = any> =
   | ExoticComponentWithDisplayName<P>
   | React.ComponentType<P>;
 
-export type KnownTarget = Exclude<keyof JSX.IntrinsicElements, 'symbol' | 'object'> | AnyComponent;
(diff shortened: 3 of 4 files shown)


probablyup:
> add to JSX.IntrinsicElements

People are augmenting that? 😲 oy vey

Going to look into this a little more deeply... do you think this built-in type could end up suffering from the same issue long-term if people decide to augment it instead of the higher-level React types?

probablyup:
Oddly I can't get this to work locally, it says `HTMLElementTagMap` can't be found. I tried upgrading TS to 5.x and adding "dom" to `lib` in `tsconfig.json` and neither made a difference...

RJWadley:
That's interesting. I've switched it to `React.ReactDOM`, which is probably a better option anyway since the React types are what's used elsewhere. The only difference is this also includes deprecated elements like `big`, which is fine.

In theory this could be extended, which could cause the same issue again, although the motivations to do that are a lot less compelling since adding to React.ReactDOM doesn't affect JSX types.

probablyup:
One other option would be to derive from domElements.ts. That's actually
the list that s-c utilizes so would be even smaller than intrinsic elements
since not every html element can be styled

That list can actually be shrunken down further too, I found some elements that aren't styleable the other day.

probablyup (approved):
Thanks!

PR #4150, opened 2023-08-31
Update typescript reference on issue template

chilled-capybara:
A proposed alteration to the issue template to reflect the native typescript support in v6.


probablyup (approved):
Yup I think that's right. Appreciated!

PR #4159, opened 2023-09-07
Prevent StyleSheetManager updating context on every render

keeganstreet:
In v6 the `StyleSheetManager` component returns the following:

```jsx
<StyleSheetContext.Provider
  value={
    {
      shouldForwardProp: props.shouldForwardProp,
      styleSheet: resolvedStyleSheet,
      stylis
    }
  }
>
```

Even though the `resolvedStyleSheet` and `stylis` are memoized, they are being wrapped in a new object on every render, which causes the context value to change, and Styled Components throughout the tree to be re-rendered.

Its only a problem when `StyleSheetManager` is re-rendered of course, but still, probably shouldn't happen.


probablyup (approved):
Thanks, makes sense :)

Issue #4165, opened 2023-09-14
6.0.8 breaks typing of custom props with React Native


pmk1c:
## System:
 - OS: macOS 13.5.2
 - CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
 - Memory: 449.48 MB / 16.00 GB
 - Shell: 5.9 - /bin/zsh
## Binaries:
 - Node: 18.16.1 - ~/.asdf/installs/nodejs/18.16.1/bin/node
 - npm: 9.5.1 - ~/.asdf/plugins/nodejs/shims/npm
 - pnpm: 8.7.4 - ~/.asdf/installs/nodejs/18.16.1/bin/pnpm
 - Watchman: 2023.08.28.00 - /usr/local/bin/watchman
## npmPackages:
 - styled-components: ^6.0.7 => 6.0.8 

## Reproduction
```tsx
import React from "react";
import styled from "styled-components/native";

interface Props {
  customProp?: string;
  secondProp?: string;
}

const StyledView = styled.View<Props>`
  ${(props) => {
    // correctly typed
    props.customProp;

    // correctly typed
    props.secondProp;

    return null;
  }}
`;

const Component = () => (
  <StyledView
    // No overload matches this call.
    // Property 'customProp' does not exist on type
    // 'IntrinsicAttributes & IntrinsicClassAttributes<View> & Readonly<ViewProps>'.
    customProp="test"
  />
);
```

## Expected Behavior
Using custom props with `styled-components/native` should not give a TypeScript error.

## Additional Info
Only the first custom prop of a component gives this error. The others do not.
When doing the same with `styled-components` (non-native) and a `div` instead of a `View` it works without a problem.

LucasPMM:
Same here, using expo 49

vm-agnislav:
The same. Reverted back to 6.0.7

centavosx:
I'm having the same problem with the latest version too.

cesarhilario:
I am using patch-package with this diff file https://gist.github.com/cam-shaw/fb8bb64870bb6889827491f235c01a63

krudos:
try 6.1.0

ademarsj:
Working on 6.1.0, thanks.

PR #4169, opened 2023-09-18
Fix: Do not add Target to return type for constructWithOptions

pmk1c:
resolves https://github.com/styled-components/styled-components/issues/4165

The addition of the Target type in 6.0.8 breaks prop-Typing for styled-components with React Native.

Instead of inferring static properties, these should be specified by the implementor. As far as I understand, there already is the OuterStatics generic for this use case.

### DIFF
--- a/packages/styled-components/src/constructors/constructWithOptions.ts
+++ b/packages/styled-components/src/constructors/constructWithOptions.ts
@@ -50,10 +50,7 @@ export interface Styled<
   <Props extends object = BaseObject, Statics extends object = BaseObject>(
     initialStyles: Styles<Substitute<OuterProps, NoInfer<Props>>>,
     ...interpolations: Interpolation<Substitute<OuterProps, NoInfer<Props>>>[]
-  ): IStyledComponent<R, Substitute<OuterProps, Props>> &
-    OuterStatics &
-    Statics &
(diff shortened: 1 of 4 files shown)


razvangeangu:
Nice

cam-shaw:
I was just about to put a PR up with this fix... should have checked the open ones first haha
Can confirm this resolves the issue.

packages/styled-components/src/constructors/constructWithOptions.ts
```diff
@@ -50,10 +50,7 @@ export interface Styled<
   <Props extends object = BaseObject, Statics extends object = BaseObject>(
     initialStyles: Styles<Substitute<OuterProps, NoInfer<Props>>>,
     ...interpolations: Interpolation<Substitute<OuterProps, NoInfer<Props>>>[]
-  ): IStyledComponent<R, Substitute<OuterProps, Props>> &
-    OuterStatics &
-    Statics &
-    (Target extends string ? {} : Target);
```
krudos, line 56: we could change this to 
`  (R extends 'web' ? (Target extends string ? {} : Target) : {});`

probablyup:
Ended up taking #4190, thank you for contributing!

PR #4190, opened 2023-10-11
fix ts warning

krudos:
since 6.0.8 release in react native typescript complains about the types


probablyup:
Thank you will review today

probablyup (approved):

Issue #4064, opened 2023-06-28
Lockfile diff is huge for v6


kachkaev:
## Environment

`npx envinfo --system --binaries --npmPackages styled-components,babel-plugin-styled-components --markdown`

```yaml
## System:
 - OS: macOS 13.4.1
 - CPU: (10) arm64 Apple M1 Max
 - Memory: 315.94 MB / 32.00 GB
 - Shell: 5.9 - /bin/zsh
## Binaries:
 - Node: 18.16.0 - ~/.nvm/versions/node/v18.16.0/bin/node
 - Yarn: 4.0.0-rc.46 - ~/.nvm/versions/node/v18.16.0/bin/yarn
 - npm: 9.5.1 - ~/.nvm/versions/node/v18.16.0/bin/npm
 - Watchman: 2023.06.12.00 - /opt/homebrew/bin/watchman 
```

## Reproduction

https://github.com/kachkaev/njt/pull/374
Next.js + TypeScript + Styled Components

## Steps to reproduce

1. Upgrade `styled-components` from `5.3.11` to `6.0.1`
2. See the [diff](https://github.com/kachkaev/njt/commit/0f2820ab887b943884443b5a1e716715f596751d) in `yarn.lock` (`pnpm-lock.yaml` etc.?)

## Expected Behavior

I have expected the lockfile to change insignificantly

## Actual Behavior

[![](https://github.com/styled-components/styled-components/assets/608862/2a8beaab-e600-47f5-a8bd-526ed8f9d023)](https://github.com/kachkaev/njt/commit/0f2820ab887b943884443b5a1e716715f596751d)

The number of app dependencies has increased from 502 to 705 (+40%). This can noticeably slow down local dev and CI in a project.

Can it be that some `devDependencies` have been accidentally added to `dependencies`? It’d be nice to see a smaller diff, it is otherwise a bit scary to upgrade because the diff feels off.

probablyup:
Likely related to all the new TS stuff if I had to guess. We didn't add much otherwise

timbset:
I think you don't need to add at least `@babel/core` and `@babel/cli` as dependencies. I found only one place where it is used for typings. So please add the `type` keyword into the `@babel/core` import construction (this says to compiler that it can cut it off from output runtime code) and more packages from dependencies to devDependencies. Once you compiled the lib, you don't need inner types anymore. You just export interfaces build by compiler, so the dependency is not necessary outside your development cycle.

probablyup:
[...]

PR #4193, opened 2023-10-14
chore: remove babel macro due to lack of usage

probablyup:
See #4064 for more information.


jojkos:
I'm updating libraries on a large codebase and just discovered that support for styled-components/macro has been removed. For us, having human-readable class names is crucial for debugging and understanding component styling at a glance.

I've reviewed [Issue #4064](https://github.com/styled-components/styled-components/issues/4064), but I didn’t see any mention of an alternative for achieving readable class names since the macro support was dropped. Has support for readable class names really been removed without a replacement, or is there another recommended approach?

Any guidance would be appreciated!

quantizor:
The alternative is just to use the babel plugin in a normal way via webpack or alternative bundler https://github.com/styled-components/babel-plugin-styled-components

Issue #4181, opened 2023-09-27
Styled Image gives warning when supplying borderColor prop in React Native


MosefAsad:
### Styled Components Version:
**6.0.7**

### React Native Version:
**0.72.4**

### Reproduction:

```
export const StyledImage = styled.Image({
  borderColor: '#FFF',
});
```

### Expected Behavior:
No console Error Warning.

### Actual Behavior:
 ```
ERROR  Warning: Failed prop type: Invalid props.style key `borderTopColor` supplied to `Image`.
Bad object: {
  "borderTopColor": "#fff",
  "borderRightColor": "#fff",
  "borderBottomColor": "#fff",
  "borderLeftColor": "#fff",
}
Valid keys: [
...
  "borderColor", ---> This is a valid key and supplying the same property to the Image component from RN works.
...
]

```

### Example of working standard code:
 ```
 <Image
    style={{
       borderColor: '#FFF',
     }}
     source={{ uri: someSource }}
  />
```

nazmeln:
+1

nazmeln:
@jacobp100 
Looks like it's related to this [PR](https://github.com/styled-components/styled-components/pull/267)

PR #4195, opened 2023-10-18
fix image unsupported border color style

nazmeln:
Hi,
According to the current RN [docs](https://reactnative.dev/docs/image-style-props#props), those styles `borderColor, borderWidth, borderStyle` can be only one instead of multiple.

Note: how I can test those changes?

Fixes https://github.com/styled-components/styled-components/issues/4181


probablyup:
Can you add a test or two? Thanks

nazmeln:
> Can you add a test or two? Thanks

@probablyup added

probablyup (approved):

Issue #3658, opened 2022-01-04
React 18 Streaming SSR


gshokanov:
Hi folks,

**React 18** will be rolled out in the near future and currently has an RC release that anyone can try out. I thought I would try the new streaming SSR mode with one of the projects I'm working on that uses styled-components and while working on an implementation I realised that styled-components doesn't have a compatible API yet.

Here's a [link to the sandbox](https://codesandbox.io/s/kind-sammet-j56ro) with the demonstration of the new streaming API by Dan Abramov. **React 18** exposes a new function for rendering on the server called `renderToPipeableStream`. As part of it's API it exposes a `pipe` function that expects a target stream (typically a response stream). The existing styled-components API `interleaveWithNodeStream` expects a stream exposed by React which makes the whole thing incompatible with each other.

I tinkered with the whole thing a bit and the solution seems to be simple - expose a new API that wraps `pipe` function provided by React, returns a new function with the same API that uses the same internal `Transformer` stream logic used in `interleaveWithNodeStream`. The version I came up with looks something like this:
```ts
type PipeFn = (destination: streamInternal.Writable) => void;

export default class ServerStyleSheet {
 // existing code here...

  interleaveWithPipe(pipe: PipeFn): PipeFn {
    this._guard();
    this.seal();

    return (destination) => {
      const transformStream = this._getNodeTransformStream();

      transformStream.pipe(destination);
      pipe(transformStream);
    };
  }
}
```
`_getNodeTransformStream` creates the same stream as currently seen in `interleaveWithNodeStream`.

I got the whole thing working together and would be glad to contribute this as a PR, however I run into an interesting issue while working on it.

It seems that there's currently a difference in handling SSR on the `main` branch and on the `legacy-v5` branch which I believe is the latest stable release. The difference pretty much comes down to this [particular commit](https://github.com/styled-components/styled-components/commit/b9e60f7efb66c0710002f0e133e42af1c7deb497). React seems to be emitting rendered HTML in very small chunks, sometimes even without closing the current tag. Here's what I am talking about:
```html
<div

data-kit
="
progress
"

role
="
progressbar
"
<!-- the rest goes on -->
```
[...]

PR #4213, opened 2023-11-12
Fix interleaveWithNodeStream streaming flow to be compatible with React 18 api

ignatiqq:
As mentioned here - https://github.com/styled-components/styled-components/issues/3658

New React "renderToPipeable" stream api. It decides how much HTML to send and it’s not a fact that it can be valid, for example, the middle of the svg path is a large base64 image or just long text.

So in this pr I suggest you a solution to work compatible with the new react streaming api.

As far as we have react < 18 in the project we can't test it now.

### DIFF
--- a/packages/styled-components/src/constants.ts
+++ b/packages/styled-components/src/constants.ts
@@ -9,7 +9,7 @@ export const SC_ATTR: string =
 
 export const SC_ATTR_ACTIVE = 'active';
 export const SC_ATTR_VERSION = 'data-styled-version';
-export const SC_VERSION = 'JEST_MOCK_VERSION';
+export const SC_VERSION = __VERSION__;
 export const SPLITTER = '/*!sc*/\n';
 
 export const IS_BROWSER = typeof window !== 'undefined' && 'HTMLElement' in window;
--- a/packages/styled-components/src/constants.ts
+++ b/packages/styled-components/src/constants.ts
@@ -9,7 +9,7 @@ export const SC_ATTR: string =
 
 export const SC_ATTR_ACTIVE = 'active';
 export const SC_ATTR_VERSION = 'data-styled-version';
-export const SC_VERSION = __VERSION__;
+export const SC_VERSION = 'JEST_MOCK_VERSION';
 export const SPLITTER = '/*!sc*/\n';
 
 export const IS_BROWSER = typeof window !== 'undefined' && 'HTMLElement' in window;
--- a/packages/styled-components/src/models/ServerStyleSheet.tsx
+++ b/packages/styled-components/src/models/ServerStyleSheet.tsx
@@ -6,13 +6,14 @@ import { IS_BROWSER, SC_ATTR, SC_ATTR_VERSION, SC_VERSION } from '../constants';
 import StyleSheet from '../sheet';
 import styledError from '../utils/error';
 import getNonce from '../utils/nonce';
-import StyleSheetManager from './StyleSheetManager';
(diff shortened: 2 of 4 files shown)


quantizor:
Thank you for working on this, I have no practical test bed for streaming testing at this point but I'd be happy to post a test release under a tag for people to try and report back

quantizor:
Published under `styled-components@react-18-streaming-prototype` for those interested in testing

ignatiqq:
Oh, i really forgot about this pr. It's really unstable now. 
But if you don't mind i can continue work on it in the next month. 
The main problem now is that it will trigger hydration issue without any scripts which will move styles outside the root react component. We should somehow give users way to control how should it work.

quantizor:
Of course! That's why it's a test build 👍

burnsdy:
Hi @ignatiqq,

I wrote a Transform stream implementation for handling server-side rendering of styled-components based on this PR. I noticed a couple issues with this PR due to updating the React 16 streaming approach to work with React 18 streaming:

- Original `CLOSING_TAG` regex is invalid
  - This regular expression only checks if the chunk begins with a closing tag (which works for React 16 streaming), not if the closing tag exists anywhere in the chunk
- Search for end of closing tag position is incorrect
  - Current approach looks for first '>' character in the chunk, which may be part of a starting tag rather than a closing tag
- Need a `final` method to flush the styles buffer of any remaining style tags before stream closure

I can submit my code as a separate PR, but these are the only three required changes to patch the bugs I noticed.

ignatiqq:
@burnsdy You are right. This code is buggy. But the main reason i do not push any commits here. Becuase i find problem which i cant solve by myself here. The problem related to SSR error during new react 18 stream api. It means React will start hydrate page partially and we cant remove all "style" tags from server.

After reading the instructions from React maintainers and paying attention to other CSS in JS libraries, the only solution that I could come up with at the moment is to use a script that moves the “style” tag behind the “root” component. But this solution is not perfect and has many trade-offs, such as: performance (both in the case of multiple tag styles, if there are not many "Suspense" on the page, and if there are many "Suspense" borders on the page, we will need to execute the script after processing the css every time and process it using the sc.rehydrate function)
@quantizor What do you think about that?

hamidrezahanafi:
> @burnsdy You are right. This code is buggy. But the main reason i do not push any commits here. Becuase i find problem which i cant solve by myself here. The problem related to SSR error during new react 18 stream api. It means React will start hydrate page partially and we cant remove all "style" tags from server.
> 
> After reading the instructions from React maintainers and paying attention to other CSS in JS libraries, the only solution that I could come up with at the moment is to use a script that moves the “style” tag behind the “root” component. But this solution is not perfect and has many trade-offs, such as: performance (both in the case of multiple tag styles, if there are not many "Suspense" on the page, and if there are many "Suspense" borders on the page, we will need to execute the script after processing the css every time and process it using the sc.rehydrate function) @quantizor What do you think about that?

Do you have any solution for this yet @ignatiqq?
Would you mind sharing your code that moves streamed (after first hydration) to top here?

hamidrezahanafi:
@ignatiqq @quantizor https://github.com/styled-components/styled-components/pull/4327 should fix the empty styles and streaming seems to be mostly broken because of those
We still have rehydration issue on chunk that streamed in after initial hydration

quantizor:
Closing — the interleaveWithNodeStream API has been superseded by RSC-based streaming in v6.3.x.

PR #4214, opened 2023-11-14
Upgrading babel dependencies to latest to fix CVE-2023-45133

JesseObrien:
As noted in https://nvd.nist.gov/vuln/detail/CVE-2023-45133 @babel/traverse needs to be updated to avoid the CVE. This PR bumps all versions of babel and plugins to the latest versions to fix this.


quantizor (approved):
Thank you!

PR #4230, opened 2023-12-15
Expose FastOmit to improve emit performance

### LINKED ISSUE #4229: Emit performance is very slow due to FastOmit
Emitted declaration files end up very large because HTML props are explicitly listed instead of referencing the `FastOmit` type.  This causes a huge performance loss due to larger `.d.ts` files needing to be written, as well as a much larger file size.

I myself noticed my builds taking 2x-3x longer due to the shear amount of additional writes to disk having to happen. 

## npmPackages:
 - styled-components: 6.0.2 => 6.1.1

## Steps to reproduce
```ts
export const MyStyledDialog = styled(Dialog)``;
```
Open the emitted `.d.ts` declaration file.

Example repo: 
https://github.com/justinbhopper/styled-component-v6-perf/tree/emit-perf-bug

## Expected Behavior
Emitted `.d.ts` file references `FastOmit` and the original prop types:
```ts
export declare const MyStyledDialog: IStyledComponent<
  "web",
  FastOmit<React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement>, never>
>
  & ((props: DialogProps) => JSX.Element);
````
## Actual Behavior
The emitted `.d.ts` file has materialized the resulting type to avoid referencing `FastOmit` since it wasn't exported.  This results in a much larger file and much longer write times.
```ts
export declare const MyStyledDialog: IStyledComponent<
  "web",
  {
      foo: number;
      defaultChecked?: boolean | undefined;
      defaultValue?: string | number | readonly string[] | undefined;
      suppressContentEditableWarning?: boolean | undefined;
      suppressHydrationWarning?: boolean | undefined;
      ... 254 more ...
      onTransitionEndCapture?: React.TransitionEventHandler<HTMLDivElement> | undefined;
  }
>
  & ((props: DialogProps) => JSX.Element);
```

justinbhopper:
Fixes #4229

Emitted declaration files end up very large because HTML props are explicitly listed instead of referencing the `FastOmit` type. This causes a huge performance loss due to larger `.d.ts` files needing to be written, as well as a much larger file size.

I myself noticed my builds taking 2x-3x longer due to the shear amount of additional writes to disk having to happen.

The bug happens because `FastOmit` is not exposed, so TypeScript ends up explicitly defining the resulting type instead.  Exposing `FastOmit` fixes the issue.

### DIFF
--- a/packages/styled-components/src/types.ts
+++ b/packages/styled-components/src/types.ts
@@ -23,7 +23,7 @@ export type BaseObject = {};
 // from https://stackoverflow.com/a/69852402
 export type OmitNever<T> = { [K in keyof T as T[K] extends never ? never : K]: T[K] };
 
-type FastOmit<T extends object, U extends string | number | symbol> = {
+export type FastOmit<T extends object, U extends string | number | symbol> = {
   [K in keyof T as K extends U ? never : K]: T[K];
 };
 


quantizor (approved):
Thanks 👍🏼

Issue #4104, opened 2023-07-26
v6 TS Missing static properties when component is wrapped by styled-component


oskarwrobel:
TS does not see static properties when component is wrapped by styled-component.

## Reproduction

```jsx
function MyComponent({ className }: { className?: string }) {
  return <div>My component</div>
}
MyComponent.SomeStaticData = "Foo bar";

const StyledMyComponent = styled(MyComponent)`
  color: blue;
`;

return (
  <>{StyledMyComponent.SomeStaticData}</>
);
```
> Property 'SomeStaticData' does not exist on type 'IStyledComponent<"web", typeof MyComponent, { className?: string | undefined; } & object> & object'.ts(2339)

Example: https://codesandbox.io/s/keen-euler-gxn4zj?file=/src/App.tsx
(CodeSandbox supports styled-components 6.0.0-beta.14 however it works exactly the same for 6.0.5)

## Steps to reproduce
Wrap a component that contains static properties.

## Expected Behavior
Static properties should remain accessible.

## Actual Behavior
TS does not see static properties.

oskarwrobel:
Fixed by @MartinPELCAT in https://github.com/styled-components/styled-components/pull/4141.
Thanks!

Issue #4225, opened 2023-12-04
Storybook broken from >=6.0.8 because React.ComponentProps cannot derive all props


alex2:
## System:
 - OS: macOS 14.1.1
 - CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
 - Memory: 3.95 GB / 64.00 GB
 - Shell: 5.9 - /bin/zsh
## Binaries:
 - Node: 18.16.0 - ~/.nvm/versions/node/v18.16.0/bin/node
 - Yarn: 1.22.19 - ~/.yarn/bin/yarn
 - npm: 9.5.1 - ~/.nvm/versions/node/v18.16.0/bin/npm
 - pnpm: 8.6.6 - ~/.nvm/versions/node/v18.16.0/bin/pnpm
 - bun: 1.0.15 - ~/.bun/bin/bun
## npmPackages:
 - styled-components: latest => 6.1.1 

## Reproduction

```tsx
import styled from "styled-components";

const A = styled.div<{ foo?: string }>``;

const B = styled(A)<{ bar?: string }>``;

// typescript has no issue with the "bar" prop and it is typed correctly
const C = () => <B foo="one" bar="two" />

// Typescript error: "Property 'bar' does not exist on type ..."
// This works as expected on v6.0.7 but is broken in any version since
type T = React.ComponentProps<typeof B>["bar"]
```

https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcARFDvmQNwBQokscqMAngDZYAmcRJ5Vpx4BaXCUgA7LJJipadOuMms4AQTgBeFuy7cAdN2AA3ADwBvPhAgB+AFw6owSQHM4AXwB8AA2-0lECrwAEJaOsLcABRqAJQWcABGyFD2js5uXr7+yqoAwmGRMVqecKahBNaaZIFYZInJVTAA7hB1APSeim1tcAAiEFjokhDwqAAWEE0ocOxgg7hOYPBYUMRQcM5wxgBs+gAM+gDsdLNYcAAqYdh4MPq5EjWyAArEYKimpxAEcMGeANpkJJQMgAXSAA

## Steps to reproduce

Using styled-components v6.0.8 or above, up to `@ latest` results in props from wrapped types (including those introduced by `attrs` not being picked up by `React.ComponentProps`.

[...]

Issue #4174, opened 2023-09-22
Nesting styled components with styled() and adding prop types does not work correctly


fabb:
### System:
 - OS: macOS 13.5.2
 - CPU: (10) arm64 Apple M1 Max
 - Memory: 49.19 MB / 64.00 GB
 - Shell: 5.9 - /bin/zsh
### Binaries:
 - Node: 18.17.0 - ~/.nvm/versions/node/v18.17.0/bin/node
 - Yarn: 1.22.19 - ~/.yarn/bin/yarn
 - npm: 9.6.7 - ~/.nvm/versions/node/v18.17.0/bin/npm
### npmPackages:
 - babel-plugin-styled-components: 2.1.4 => 2.1.4 
 - styled-components: 6.0.8 => 6.0.8 
 
## Reproduction

https://codesandbox.io/s/styled-components-base-forked-j8f2nt?file=/index.tsx

## Steps to reproduce

```
const Component1 = styled.a<{ prop1?: boolean }>``

const Component2 = styled(Component1)<{ prop2?: boolean }>``

const Component3 = styled(Component2)<{ prop3?: boolean }>``

const Component4 = () => <Component3 prop1={true} prop2={true} /> // compiles fine

const Component5 = () => <Component3 prop1={true} prop2={true} prop3={true} /> // compile error, prop2 does not exist
```

## Expected Behavior

Should compile.

## Actual Behavior

Component5 does not compile, `prop2` cannot be found.

krudos:
i am getting same error with 6.0.8 . with 6.0.7 types are working

seanmcquaid:
Seeing the same as well!

krudos:
i think this is related to
https://github.com/styled-components/styled-components/issues/4165

ViniDevBR:
6.0.8 is broken with types... USE 6.0.7

krudos:
i think this got broken on this commit
https://github.com/styled-components/styled-components/commit/6ee6c23994b9ac41eba06b112a0f6670e41de757

krudos:
i created this PR
https://github.com/styled-components/styled-components/pull/4190

krudos:
try 6.1.0

MarineP31:
I still got the same issue in 6.1.0 too

macko911:
I've added failing tests to help fix this in https://github.com/styled-components/styled-components/pull/4219

PR #4236, opened 2023-12-20
Fix nested styled with additional prop types (with fix)

justinbhopper:
Fixes ts errors due to additional prop types not being included (#4225, #4174), but I am unsure if it is the proper approach.  

Should we only omit the constructor instead?

The issue specifically began here when support for static properties was added #4141 in order to fix issue #4104.


quantizor:
This fix seems reasonable and I'm not seeing any obvious complications from it... let's give it a try.

quantizor (approved):

PR #4238, opened 2023-12-22
fix: pin dependencies to exact versions prior to bun migration

quantizor:


dubzzz:
Is the pinning just for the time of the migration?

Actually pinning tends to imply duplicated dependencies on user's side and thus higher bundle size. For instance if one user pulls `@emotion/unitless` on his own either directly or indirectly there is a huge chance that it might pull `0.8.1` and not `0.8.0` for various reasons (including security). Why would you not allow `0.8.1`? (mostly asking to know the reasons behind that change, I'm not challenging the poetntial need for it)

quantizor:
@dubzzz it's meant to be accompanied by https://github.com/styled-components/styled-components/pull/4243

I'll ship it soon, found a corresponding bug in npm that I was hoping would be fixed soon.

dubzzz:
awesome ❤️

dumistoklus:
@quantizor 

> I'll ship it soon, found a corresponding bug in npm that I was hoping would be fixed soon.
This solution doesn't work for deduplication in yarn 3.2.0.

PR #4242, opened 2023-12-27
fix: allow passing theme prop to child if using attrs

quantizor:
Fixed the specific scenario documented in #4074.


AhmedMuhammedElsaid (commented):
I needed it really, Thanks

PR #4243, opened 2023-12-27
chore: add all missing peer dependency statements

### LINKED ISSUE #4226: Getting error "ReferenceError: Property 'CssSyntaxError' doesn't exist" while using styled-component in React native repack project.
## Environment

"react": "18.2.0",
"react-native": "0.72.7",
"styled-components": "^6.1.1",
"@callstack/repack": "^3.6.0"

## Description

I'm trying to use styled-component in react native repack project but getting "Property 'CssSyntaxError' doesn't exist react native" while running the project.

Repack project sample:
https://github.com/callstack/repack

Screenshot 2023-12-05 at 8 44 52 AM

Screenshot_1701746204

##LocalModule.tsx

````import React from 'react';
import { Text } from 'react-native';
import styled from 'styled-components/native';

const LocalModule = () => {
    return (
        <Container>
            <Text >React Native</Text>
        </Container>
    );
};

const Container = styled.View`
      flex: 1;
      background-color: white;
      align-items: center;
      justify-content: center;
    `;


export default LocalModule;
```

Thanks in advance.

quantizor:
This allows for better deduping in client projects.

Might fix #4226


packages/styled-components/package.json
```diff
@@ -62,8 +62,45 @@
     "tslib": "2.5.0"
   },
   "peerDependencies": {
-    "react": ">= 16.8.0",
-    "react-dom": ">= 16.8.0"
+    "@emotion/is-prop-valid": "^1.2.1",
```
brunocabral88, line 65: I don't think this change will do what you think it will do they it stands:
- Listing packages as a `peerDependency` means styled-components expect the consuming application to have downloaded those dependencies beforehand, which I believe is not the case here (I might be wrong though 😄)

To achieve what you want (better dedupe on the consuming apps side) you can simply change the dependencies listed in the `dependencies` array to use the range syntax like you used here. NPM/Yarn will be smart and reuse the dependencies versions already present on the consuming side if there's a match for that range..

quantizor: That's why they're defined in both peer and regular dependencies, with peerDependenciesMeta declaring all but "react" are optional

brunocabral88: But since the `dependencies` array still defines fixed versions, it'll overwrite the peerDependenciesMeta, won't it? 🤔

quantizor: Tested the explicit dependency + permissive peer dep with a bunch of different pms:

Worked (installed only the higher version): yarn, pnpm
Didn't work (installed both versions): npm, bun

The tricky thing is with `react` the range is really wide (16.8 and up basically.) I could move things out of peerDependencies into devDependencies, but basically no pm automatically installs peerDependencies out of the box so users would get hit with an error to add the dependency explicitly. The goal of this was to ensure the relevant libraries are available _at least_ at the version we specify, with peerDependencies meant to pull in a greater applicable version if defined in the client repository.

quantizor: I [opened an issue](https://github.com/npm/cli/issues/7106) against npm, I think yarn and pnpm behavior is correct.

brunocabral88: Hmm, interesting one.. I expected PMs to make `dependencies` take precedence over `peerDependencies` as, to me, `peerDependencies` was purely informational to let consuming apps know that you expect something to be there.. However, that doesn't seem to be the behavior the accepted RFC you linked describes..

Even if they fix npm, it'll take time for people to adopt the version with the fix.. 
IMHO, if I consume a library that has a dependency: `"A": "^1.1.0"`, I would think that library was integration-tested with that package with that version, although anything in the range might work.. 

> The goal of this was to ensure the relevant libraries are available at least at the version we specify

You just precisely described (in my understanding) the `dependencies` array when used with ranges :D

brunocabral88: I imported `styled-components@6.1.5-rc.0` into the styled-components-website and ran `yarn`.. it did not take into account the peerDependencies array (that specifies `postcss: ^8.4.31` - should've resolved postcss to 8.4.32)..

image

When I removed the `peerDependencies` and set the `dependencies` to be a range (`postcss: ^8.4.0` for testing purposes), it resolved postcss to 8.4.32
image

quantizor: It's more obvious in the repo I linked in the bug ticket since that repo has postcss saved already at a higher version

kachkaev:
I'm afraid that this change is breaking. Example: https://github.com/kachkaev/njt/pull/549

quantizor:
I added optional status to all of them except react itself, is the build tooling not taking that into account?

quantizor:
It's odd because I tested it in pnpm and resolution worked as expected... will roll the change back.

kachkaev:
Thanks for reverting! App build in https://github.com/kachkaev/njt/pull/549 worked after upgrading from 6.1.7 to 6.1.8 👍

PR #4282, opened 2024-03-22
Update flatten.ts

iegik:
To avoid situation where provided object do not have prototype we should use Object.hasOwn.

Test1:

`objToCssArray(Object.create(null))`

Test2:

```
const obj = Object.create(null);
obj.color = 'red';
objToCssArray(obj)
```

Test3:

```
const obj = {
  hasOwnProperty: () => { /*...*/ return true },
}
objToCssArray(obj)
```


quantizor:
This change unfortunately regresses hot path performance by a decent amount 
Screenshot 2024-04-30 at 5 35 35 PM

iegik:
I think, that in case, when it runs on the server side (NextJS SSR) - will be better to choose more secure option :)

PR #4288, opened 2024-04-07
fix: improve types for .attrs()

### LINKED ISSUE #4076: TS Not inferring correct type from attrs - v6
## Environment

`## System:
 - OS: macOS 13.3.1
 - CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
 - Memory: 186.73 MB / 16.00 GB
 - Shell: 5.9 - /bin/zsh
## Binaries:
 - Node: 16.14.0 - ~/.nvm/versions/node/v16.14.0/bin/node
 - npm: 9.2.0 - ~/.nvm/versions/node/v16.14.0/bin/npm
## npmPackages:
 - babel-plugin-styled-components: 2.1.3 => 2.1.3 
 - styled-components: 6.0.2 => 6.0.2 `

## Reproduction

v6 not inferring required props from attrs

## Steps to reproduce
```
const Test = styled.div.attrs<{ foo: number, bar: number }>({ foo: 1 })`

`

<Test bar={2} />
```

You get this error:
`No overload matches this call.
  Overload 1 of 2, '(props: PolymorphicComponentProps<"web", FastOmit<Substitute<Substitute<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, Omit<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "ref"> & { ...; }>, { ...; }>, never>, void, void, {}, {}>): Element', gave the following error.
    Property 'foo' is missing in type '{ bar: number; }' but required in type 'FastOmit<Substitute<FastOmit<Substitute<Substitute<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, Omit<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "ref"> & { ...; }>, { ...; }>, never>, FastOmit<...>>, keyof ExecutionProps>'.`

## Expected Behavior
Only bar is required in tsx file.

## Actual Behavior
Both foo and bar are required.

### LINKED ISSUE #4138: v6 .attrs() types are broken
## Environment
**styled-components:** 6.0.7
**react:** 18.2.0
**react-dom:** 18.2.0
**typescript:** 5.0.2

## Reproduction
https://codesandbox.io/s/styled-component-playground-forked-9fvr9m?file=/src/index.tsx
```
import type { FC } from "react";
import { createRoot } from "react-dom/client";

type ContainerProps = {
  color: string;
};
const Container: FC<ContainerProps> = ({ color }) => {
  return <div style={{ color }} />;
};

const StyledContainer = styled(Container).attrs({ color: 'red' })``;

function App() {
  return <StyledContainer />; // TS Error: "color" props is missing
}

const root = createRoot(document.getElementById("root")!);
root.render(<App />);

```

## Expected Behavior
No TS error, just like in styled-components v5

## Actual Behavior
"color" prop is missing, although it was binded inside `.attrs(...)` call and should not be required

### LINKED ISSUE #4183: styled-component does not inherit props passed through attrs method in react-native
## System:
 - OS: macOS 13.5.2
 - CPU: (10) arm64 Apple M2 Pro
 - Memory: 266.34 MB / 16.00 GB
 - Shell: 5.9 - /bin/zsh
## Binaries:
 - Node: 18.18.0 - ~/.nvm/versions/node/v18.18.0/bin/node
 - Yarn: 1.22.19 - ~/Desktop/Developer/mobile-app/node_modules/.bin/yarn
 - npm: 9.8.1 - ~/.nvm/versions/node/v18.18.0/bin/npm
 - Watchman: 2023.09.04.00 - /opt/homebrew/bin/watchman
## npmPackages:
 - babel-plugin-styled-components: ^2.1.4 => 2.1.4
 - styled-components: 6.0.7 => 6.0.7

## Environment
"dependencies": {
  ...,
  "react": "18.2.0",
  "react-is": "18.2.0",
  "react-native": "0.72.5",
  "styled-components": "6.0.7",
  "stylis": "^4.0.0"
},
"devDependencies": {
  ...,
  "@types/react": "^18.0.24",
  "@typescript-eslint/eslint-plugin": "^5.54.1",
  "@typescript-eslint/parser": "^5.54.1",
  "@typescript-eslint/typescript-estree": "^5.54.1",
  "babel-plugin-styled-components": "^2.1.4",
},
resolutions: {
  "styled-components": "^6"
}

## Reproduction

given the following component, typescript recognise *source* and *resizeMode* as valid props

`
const BlurImage = styled.Image.attrs({
  source: blur,
  resizeMode: 'stretch'
})'
  width: 100%;
  height: 100%;
';
`

## Expected Behavior

Typescript doesn't show any error

## Actual Behavior

image

uhyo:
closes #4076
closes #4138
closes #4183

Hello, the above issue is blocking us from upgrading to v6, so here is a possible fix.

## Improvement

The improvement brought by this PR can be seen in this example:

```tsx
  const DivWithRequiredFooBar = styled.div<{ foo: number; bar: string }>``;

  // foo is provided, so it becomes optional
  const DivWithRequiredBar = styled(DivWithRequiredFooBar).attrs({ foo: 42 })`
    margin; ${props => props.foo * 10}px;
  `;
  // @ts-expect-error must provide bar
  <DivWithRequiredBar />;
  // OK (NEW)
  <DivWithRequiredBar bar="3" />;
  // OK. Can still provide foo if we want
  <DivWithRequiredBar foo={3} bar="3" />;
  // @ts-expect-error foo must be a number
  <DivWithRequiredBar foo="3" bar="3" />;
```

The part marked as `(NEW)` is made possible by this PR. Once `foo` is provided through the `.attrs` call, the resulting component (`DivWithRequiredBar`) no longer requires `foo` as a prop.

## Implementation

To achieve this, the `Styled` interface now has distinction between Outer props and Inner props:

```tsx
  const DivWithRequiredBar = styled(DivWithRequiredFooBar).attrs({ foo: 42 })`
    // ↓ `props` here is the inner props which have undergone the defaulting,
    // so `props.foo` is always supplied (not optional)
    margin; ${props => props.foo * 10}px;
  `;
  // User of the resulting component sees the outer props,
  // so foo is optional here
  <DivWithRequiredBar bar="3" />;
```

Thus the `Styled` interface now has new `InnerProps` type attribute. It is usually same as `OuterProps`, but the updated `attrs()` types can make them diverge.

### DIFF
--- a/packages/styled-components/src/constructors/constructWithOptions.ts
+++ b/packages/styled-components/src/constructors/constructWithOptions.ts
@@ -41,15 +41,26 @@ type AttrsTarget<
     : FallbackTarget
   : FallbackTarget;
 
+/**
+ * Extract non-optional fields from given object type.
+ */
+type RequiredFields<T> = Pick<
+  T,
+  {
+    [K in keyof T]-?: undefined extends T[K] ? never : K;
+  }[keyof T]
+>;
+
 export interface Styled<
   R extends Runtime,
   Target extends StyledTarget<R>,
   OuterProps extends object,
   OuterStatics extends object = BaseObject,
+  InnerProps extends object = OuterProps,
 > {
   <Props extends object = BaseObject, Statics extends object = BaseObject>(
-    initialStyles: Styles<Substitute<OuterProps, NoInfer<Props>>>,
-    ...interpolations: Interpolation<Substitute<OuterProps, NoInfer<Props>>>[]
+    initialStyles: Styles<Substitute<InnerProps, NoInfer<Props>>>,
+    ...interpolations: Interpolation<Substitute<InnerProps, NoInfer<Props>>>[]
   ): IStyledComponent<R, Substitute<OuterProps, Props>> &
     OuterStatics &
     Statics &
@@ -72,10 +83,16 @@ export interface Styled<
     PrivateResolvedTarget extends KnownTarget
       ? Substitute<
           Substitute<OuterProps, React.ComponentPropsWithRef<PrivateResolvedTarget>>,
-          Props
+          Props & Partial<RequiredFields<PrivateAttrsArg>>
         >
       : PrivateMergedProps,
-    OuterStatics
+    OuterStatics,
+    PrivateResolvedTarget extends KnownTarget
+      ? Substitute<
+          Substitute<OuterProps, React.ComponentPropsWithRef<PrivateResolvedTarget>>,
+          Props
+        >
+      : PrivateMergedProps
   >;
 
   withConfig: (config: StyledOptions<R, OuterProps>) => Styled<R, Target, OuterProps, OuterStatics>;
--- a/packages/styled-components/src/constructors/constructWithOptions.ts
+++ b/packages/styled-components/src/constructors/constructWithOptions.ts
@@ -105,11 +105,12 @@ export default function constructWithOptions<
     ? React.ComponentPropsWithRef<Target>
     : BaseObject,
   OuterStatics extends object = BaseObject,
+  InnerProps extends object = OuterProps,
 >(
   componentConstructor: IStyledComponentFactory<R, StyledTarget<R>, object, any>,
   tag: StyledTarget<R>,
   options: StyledOptions<R, OuterProps> = EMPTY_OBJECT
-): Styled<R, Target, OuterProps, OuterStatics> {
+): Styled<R, Target, OuterProps, OuterStatics, InnerProps> {
   /**
    * We trust that the tag is a valid component as long as it isn't
    * falsish. Typically the tag here is a string or function (i.e.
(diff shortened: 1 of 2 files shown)


quantizor (approved):
This looks really good!

AntonNiklasson:
@quantizor I'm trying to run `6.1.9`, but it seems like that version contains an older version of the code. Was this change actually included in the published code?

AntonNiklasson:
I'm sorry for the rude tone in my previous comment @uhyo. That didn't come out the way I wanted it to. I really appreciate all the hard you've put into this!

I've been trying to reproduce it, I'll open a new issue if I find something. What I'm seeing right now is just that the props I explicitly set in `.attrs()` isn't affecting the type in the final component.

uhyo:
No problem! Anyways I see that several new issues are already out there regarding the `6.1.9` release. I'm so sorry for this additional maintenance burden. I'm going to check all of them.

AntonNiklasson:
@quantizor I saw you reverted the changes here. Is there another initiative to solve the typing for `.attrs`? It would be so nice to get it fixed, we’re stuck on v5 still.

quantizor:
Will revisit later as a test release. I'm not keen to keep putting out problematic typing mainline releases...

AntonNiklasson:
That makes sense! Perhaps it would be worth it to invest a bit into a types testing setup? I noticed there’s nothing in place when I played with #4306.

quantizor:
There are a few type test files in the repo

quantizor:
The issue is more about scale. This repo might be fine but a large one with hundreds or thousands of files will have entirely different compile performance...

TFRadicalImaging:
Still not working. Why this is closed?

uhyo:
@TFRadicalImaging Because this fix broke stuff and had to be reverted 😇 Sorry for confusion.

Issue #4299, opened 2024-05-02
error TS4023 when compiling a library that uses "styled-components": "^6.1.9"


akkadaya:
>error TS4023: Exported variable 'StyledComponent1' has or is using name 'IStyledComponentBase' from external module ".../node_modules/styled-components/dist/types" but cannot be named.

I could compile the library after exporting the `IStyledComponentBase` interface in `node_modules/styled-components/dist/types.d.ts`

Here is the link to the interface in this repo:
https://github.com/styled-components/styled-components/blob/f688bef70237f314f1656e3c12c70725d29d2d5b/packages/styled-components/src/types.ts#L223

Am I doing something wrong? or is it a bug in `styled-components` that needs to be fixed?

TomasMoratoPerezPorro:
I'm having an issue that coul be related, after upgrading to Styled 6 i get the following error in all my styled components:

```
export const Relative = styled(Div)<PositionProps>`
	position: relative;
	z-index: ${({ zIndex }) => zIndex};
`;
```

`Exported variable 'Relative' has or is using name 'IStyledComponentBase' from external module "C:/.../project/node_modules/styled-components/dist/types" but cannot be named.ts(4023)`

Any idea of how could it be fixed ?

quantizor:
Fix should be here I think, though I don't understand why TS should require an internal type to be exposed https://github.com/styled-components/styled-components/releases/tag/v6.1.10

akkadaya:
Thank you @quantizor 

When you hover over a component created by Styled Components, it infers the type `IStyledComponentBase` 
<img width="964" alt="image" src="https://github.com/styled-components/styled-components/assets/1834445/356ff965-0a5f-423c-9938-1c0445860694">

But I believe it was meant to be `IStyledComponent` instead
<img width="950" alt="image" src="https://github.com/styled-components/styled-components/assets/1834445/66758cd6-12a0-42df-87ff-54e5c8c066f6">

quantizor:
Closing — fixed in a previous release.

PR #4300, opened 2024-05-02
Export IStyledComponentBase interface

akkadaya:
This is a fix for #4299


AntoineKM (commented):
awesome @quantizor should take a look at it

quantizor (approved):
I feel like this shouldn't be necessary but I will take it to fix the issue

AntoineKM:
it actually fixed my error, thanks @akkadaya & @quantizor !

PR #4306, opened 2024-05-07
add a failing test case for .attrs()

AntonNiklasson:
We're running in to some issues in our codebase on `6.1.9`. I managed to reproduce a minimal case that fails with the improved types from #4288.

@uhyo could you take a look at this? Perhaps that gives you a clue on how to fix it?


AntonNiklasson:
@quantizor I've attached a failing test case here, but I understand we can't just merge that without a fix. I'm afraid I'm not skilled enough to fix it myself though. I'm not sure how to proceed here, this is probably the main blocker for us to migrate to v6.

PR #4309, opened 2024-05-11
Allow Passing Shadow Root as Sheet Target

joealden:
In `@types/styled-components`, the `target` prop of `StyleSheetManager` allows passing a `ShadowRoot` (see [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/86953d1148f231c8f1c534e4a718ac4fa7888767/types/styled-components/index.d.ts#L390-L397)).

As I'm not familiar with this codebase, I started from the change to `IStyleSheetManager`, then fixed any type errors that surfaced. @quantizor please let me know if you see any problems with what I've done here, or if you know of a reason why the `ShadowRoot` option wasn't carried over into v6!

### DIFF
--- a/packages/styled-components/src/models/StyleSheetManager.tsx
+++ b/packages/styled-components/src/models/StyleSheetManager.tsx
@@ -75,7 +75,7 @@ export type IStyleSheetManager = React.PropsWithChildren<{
   /**
    * Provide an alternate DOM node to host generated styles; useful for iframes.
    */
-  target?: undefined | HTMLElement;
+  target?: undefined | HTMLElement | ShadowRoot;
 }>;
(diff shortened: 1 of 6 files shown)


quantizor (approved):
Cool! Would it make sense to create an `InsertionTarget` type so we can DRY this up a bit?

joealden:
@quantizor sounds good to me! Do you think it should live in `/src/types.ts` or `src/sheet/types.ts`?

quantizor:
> @quantizor sounds good to me! Do you think it should live in `/src/types.ts` or `/src/sheet/types.ts`?

src/types is good :)

Issue #4254, opened 2024-01-19
Bug: DefaultTheme type allows any key, even when defining explicit type for theme


narwold:
## Environment

```
❯ npx envinfo --system --binaries --npmPackages styled-components,babel-plugin-styled-components --markdown
Need to install the following packages:
envinfo@7.11.0
Ok to proceed? (y) y

## System:
 - OS: macOS 13.5.2
 - CPU: (10) arm64 Apple M1 Max
 - Memory: 49.94 MB / 32.00 GB
 - Shell: 5.9 - /bin/zsh
## Binaries:
 - Node: 20.10.0 - ~/.local/share/mise/installs/node/20.10.0/bin/node
 - npm: 10.2.3 - ~/.local/share/mise/installs/node/20.10.0/bin/npm
 - pnpm: 7.33.6 - ~/.local/share/mise/installs/node/20.10.0/bin/pnpm
 - Watchman: 2023.11.20.00 - /opt/homebrew/bin/watchman
## npmPackages:
 - styled-components: ^6.1.8 => 6.1.8
```

## Reproduction

The previous DefinitelyTyped types used an empty object for DefaultTheme, which had the effect of allowing any keys until DefaultTheme was extended via https://styled-components.com/docs/api#create-a-declarations-file, at which time the theme ended up being strictly typed, and references to unknown keys threw an error in TS.

In the new v6 types, an index is being added (`[key: string]: any;`) that explicitly allows any key. There does not appear to be any way to remove/override this.

## Steps to reproduce

1. In https://codesandbox.io/p/sandbox/reverent-sara-hckdzp?file=%2Fsrc%2FApp.tsx%3A7%2C2, note that using styled-components v6, TS does not complain about the non-existing key.
2. In https://codesandbox.io/p/sandbox/reverent-sara-forked-kfd7wm?file=%2Fsrc%2FApp.tsx%3A8%2C60, note that using v5 with DefinitelyTyped, TS DOES throw an error with the missing key. Note that you may not see the error until forking the CodeSandbox due to their read-only mode until doing so.

## Expected Behavior

TS throws an error with unknown keys on the theme, if I have specified an explicit type for the theme via https://styled-components.com/docs/api#create-a-declarations-file

## Actual Behavior

TS allows any key I want to reference, which does not enforce type safety on the theme.

rakestto:
[...]

Issue #4247, opened 2024-01-08
Overriding DefaultTheme with custom ThemeType only affects ThemeProvider and not styled components themselves.


nat-foo:
## Environment

`npx envinfo --system --binaries --npmPackages styled-components,babel-plugin-styled-components --markdown`

## System:
 - OS: macOS 13.4
 - CPU: (10) arm64 Apple M2 Pro
 - Memory: 1.23 GB / 16.00 GB
 - Shell: 5.9 - /bin/zsh
## Binaries:
 - Node: 20.2.0 - ~/.nvm/versions/node/v20.2.0/bin/node
 - Yarn: 3.6.1 - ~/.nvm/versions/node/v20.2.0/bin/yarn
 - npm: 9.6.6 - ~/.nvm/versions/node/v20.2.0/bin/npm
 - pnpm: 8.11.0 - ~/.nvm/versions/node/v20.2.0/bin/pnpm
## npmPackages:
 - styled-components: ^6.1.6 => 6.1.6 

## Reproduction

[Code Sandbox Reproduction](https://codesandbox.io/p/sandbox/styled-components-base-forked-7rv2hw)

## Steps to reproduce

1. Setup a standard CRA app and install styled-components (tested with pnpm).
2. Paste this code into a file:

```
import styled, { ThemeProvider } from "styled-components"

type ThemeType = {
  test: string
  debug: string
}

declare module "styled-components" {
  export interface DefaultTheme extends ThemeType {}
}

export const CustomText = styled.p`
  // Does not raise type error
  ${(props) => props.theme.anythingiwantgoeshere.text.body.fontFamily.etc}
`

export const App = () => {
  return (
    <ThemeProvider
      theme={{
        text: {
          body: {
            fontFamily: `Raises type error: Type '{ text: { body: { fontFamily: string; }; }; }' is not assignable to type 'ThemeArgument'.
              Type '{ text: { body: { fontFamily: string; }; }; }' is missing the following properties from
              type 'DefaultTheme': test, debug`,
          },
        },
      }}
    />
  )
}
```

3. The first usage in CustomText will not result in a type error, but the second usage in ThemeProvider will.

## Expected Behavior

Overriding the DefaultTheme interface in the way the docs provide (and in the type declaration for DefaultTheme itself) should result in being able to access the custom theme type when using `styled.div`${(props) => props.theme.etc}`.

## Actual Behavior

It does not. Only the ThemeProvider seems to benefit from the custom typing.

Thanks for your time!

[...]

PR #4317, opened 2024-05-22
Tighten `DefaultTheme` type definition

oscarvz:
Introduces a type that gets extended by the `DefaultTheme` interface.

As per v6 the default theme was loosely typed where any key/value pair was valid because the `any` type. This issue is resolved with this updated definition!

Related issues
- https://github.com/styled-components/styled-components/issues/4247
- https://github.com/styled-components/styled-components/issues/4254
- https://github.com/styled-components/styled-components/issues/4370


packages/styled-components/src/models/ThemeProvider.tsx
```diff
@@ -2,6 +2,10 @@ import React, { useContext, useMemo } from 'react';
 import styledError from '../utils/error';
 import isFunction from '../utils/isFunction';
 
+// Helper type for the `DefaultTheme` interface that enforces an object type & exclusively allows
+// for typed keys.
+type DefaultThemeAsObject<T = object> = Record<keyof T, any>;
```
quantizor, line 7: Interesting, haven't seen this pattern before

quantizor:
I think this is potentially a breaking change, probably can't land it safely in the current major version

brokenmass:
@quantizor  doesn't look like this should be a breaking change as it's actually an undocumented regression

joshuaRuff:
Can we get this merged? This fixes my issue.

iantbarker:
Yes this fixed my [issue](https://github.com/styled-components/styled-components/issues/4370) as well

brokenmass:
> I think this is potentially a breaking change, probably can't land it safely in the current major version

Any chance you can get this merged ?

Lonli-Lokli:
Why it was not merged? Types are completely useless now with the v6! It cannot break anything - it will just report existing bugs in our code!

Lonli-Lokli:
I decided to patch it myself with `patch-package` npm package

brokenmass:
> I decided to patch it myself locally with `patch-package` npm package

We are doing the same with yarn patch. 

The fact that this has not yet been merged is extremely surprising to me…

brokenmass:
thanks !

eMerzh:
Sorry to interupt here 👋 .. but it seems a bit more involved that just a "fix" no ? 
like it's probably for the better, but then there are other needed changes
even the [doc](https://styled-components.com/docs/api#themeprovider) is now wrong if i understand correctly

```jsx
  <ThemeProvider theme={{ bg: 'white', color: 'mediumseagreen' }}>
    <Box>
      I'm mediumseagreen with a white background!
      <ThemeProvider theme={outerTheme => ({...outerTheme, bg: 'black'})}>
        <Box>I'm mediumseagreen with a black background!</Box>
      </ThemeProvider>
    </Box>
  </ThemeProvider>
```
here outerTheme might be undefined, so the returned theme wouldn't match the theme itself

in our code, it makes a lot of red 😓  ...

brokenmass:
The examples focus on the js apis and they do not extend DefaultTheme interface or care about types as its not the purpose of the example. 
Typescript support is documented in its own page https://styled-components.com/docs/api#typescript , page that, before this pr was merged, was incorrect as typing DefaultTheme was a de facto no-op.

Lonli-Lokli:
@eMerzh do you have your own reproduction? Not smth from documentation?

eMerzh:
@Lonli-Lokli , 

yes i can provide some examples i got ... 

where i have mostly : 
``` ts
const defaultTheme  =  {
 colors: {
   primary: "red",
   primaryHover: "pink",
   secondary: "green"
   primaryHover: "dark-green"
 }
}
```

```ts
declare module 'styled-components' {
	export interface DefaultTheme extends CustomTheme {} // custom theme beeing the typeof defaultTheme above 
}
```

then in components; i got stuff like 

```tsx
<ThemeProvider theme={{ colors: { primary: "yellow", primaryHover: "brown" } }}>
	<Button primary>Open</Button>
</ThemeProvider>
```

looking at the doc it seems i need to either give the whole theme in the provider here, or use the current one (what i prefer) but then i need to change it  to use the callback... which might give undefined as param

brokenmass:
if you are sure the inner ThemeProvider is always rendered inside another one the nyou can use a non-null type assertions:
```ts
<ThemeProvider theme={({theme}) => ({ ...theme!, colors: { ...theme!.colors, primary: "yellow", primaryHover: "brown" } })}>
  <Button primary>Open</Button>
</ThemeProvider>
```

if you are not sure you can use a default
```ts
<ThemeProvider theme={({theme = defaultTheme}) => ({ ...theme, colors: { ...theme.colors, primary: "yellow", primaryHover: "brown" } })}>
  <Button primary>Open</Button>
</ThemeProvider>
```
or explicitly assert the presence of a theme:
```ts
<ThemeProvider theme={({theme }) => {
  assert.ok(theme, 'this theme provider must be rendered inside a parent one');
  return { ...theme, colors: { ...theme.colors, primary: "yellow", primaryHover: "brown" } }
}}>
  <Button primary>Open</Button>
</ThemeProvider>
```

eMerzh:
Hey, ok i see, but you'll need to ! the first spread as well ... and its a significative change 
and i'm not sure why the return type here is  a Full them object because it does a mergeTheme with the outer one in the provider
i have the feeling the the themeprovider need to accept a partial  theme object (or a partial as return of the func) 
what do you think?

lorenzo-octoenergy:
Posting the solution to **_my problem_** incase it helps someone else too. My issue was frankly ridiculous: I was using a type instead of an interface 😓 It worked fine up until 6.1.16, but 6.1.17 with the tighter DefaultTheme threw errors everywhere that I was accessing properties of my custom theme, e.g. `theme.colors.B300` etc.

```ts
// styled.d.ts

import 'styled-components';
import { CustomTheme } from '../types/custom-theme';

declare module 'styled-components' {
    // before
    // export type DefaultTheme = CustomTheme;

    // after
    export interface DefaultTheme extends CustomTheme {};
}
```

My eslint started complaining about the empty object:
```
An interface declaring no members is equivalent to its supertype.
eslint@typescript-eslint/no-empty-object-type
```
I could have silenced it with:
```ts
/* eslint-disable @typescript-eslint/no-empty-interface */
export interface DefaultTheme extends TakoTheme {}
/* eslint-enable @typescript-eslint/no-empty-interface */
```

But instead I chose to add a placeholder property, so the final result is:
```ts
import 'styled-components';
import { CustomTheme } from '../types/custom-theme';

declare module 'styled-components' {
    export interface DefaultTheme extends CustomTheme {
        _never?: never; // Ensures the interface is not empty but has no impact
    }
}
```

PR #4322, opened 2024-06-03
feat: Add named export of `styled`

### LINKED ISSUE #4281: Feature request for v5 to ease migration to v6 of large ecosystems
# Feature request for v5 to ease migration to v6 of large ecosystems

## Problem
I work at a company where we have quite a bit of internal open source (code open sourced internally in the company and in use by all teams). We use `styled-components` for all our styling. 

I see an issue after upgrading some of our shared libs to styled v6.
We want to use the named export (code ex) to silence a lint-rule relating to use named if both named and default exists.
As of now we are not able to use the named export in shared libs as this breaks in SPAs. This is because we are stuck on v5 i SPAs until all "styled-props" are converted to transient props in all shared libs (due to breaking change in v6). Classic chicken egg problem 😅 

```TS
import { styled } from 'styled-components';
```

## Suggested feature
Add the named export of styled to v5 as well. Shouldn't cause any problems that I can think of, and it would make the migration process better, at least for me and my co-workers 😁
 
If this issue is not handled by anyone else I will submit a PR this weekend when I actually have time 😅

Kegulf:
## [v.5.4.0] - 2024-06-03

- Backport introduction of named export of `styled` to make migration to v6 easier to manage for consumers with large quantities of internal libraries.

## Related issues
resolves #4281

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ _The format is based on [Keep a Changelog](http://keepachangelog.com/) and this 
 
 ## Unreleased
 
+## [v.5.4.0] - 2024-06-03
+
+- Backport introduction of named export of `styled` to make migration to v6 easier to manage for consumers with large quantities of internal libraries.
+
 ## [v5.3.7] - 2023-03-01
 
 - fix: (React Native) passing testID as attrs property by @ku8ar (see #3857)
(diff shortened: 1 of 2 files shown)


Kegulf:
Hmm, seems that some Rollup dependency is not available anymore. Not quite sure how to debug this, I don't normally use `yarn` 😓

quantizor:
Closing — stale. The standalone build export issue is tracked in #5558.

Issue #4289, opened 2024-04-07
React SSR Streaming Bug


chin2km:
## React SSR Streaming bug
I have followed the documentation on react SSR streaming ([doc](https://styled-components.com/docs/advanced#streaming-rendering)) and endup having a broken generated html in styled-components v6. The same code works just fine in styled-components v5.3.5.

I understand that new streaming methods in React (renderToPipeableStream) is not supported yet in styled-components ([issue](https://github.com/styled-components/styled-components/issues/3658)). But `renderToNodeStream` is seemingly supported according to the docs but in practice its breaking.

### System:
 - OS: macOS 14.1
 - CPU: (10) arm64 Apple M1 Pro
 - Memory: 3.11 GB / 32.00 GB
 - Shell: 5.9 - /bin/zsh
### Binaries:
 - Node: 20.0.0 - ~/.nodenv/versions/20.0.0/bin/node
 - Yarn: 1.22.22 - ~/.nodenv/versions/20.0.0/bin/yarn
 - npm: 9.6.4 - ~/.nodenv/versions/20.0.0/bin/npm
 - pnpm: 8.15.6 - ~/.nodenv/versions/20.0.0/bin/pnpm
 - bun: 0.1.10 - ~/.bun/bin/bun
### npmPackages:
 - babel-plugin-styled-components: 2.1.4 => 2.1.4 
 - styled-components: 6.1.8 => 6.1.8 

## Reproduction
Minimal reproduction using webpack: https://github.com/chin2km/styled-components-react-streaming-bug

## Steps to reproduce
start the application using `yarn dev:ssr` and the html generated for react streaming is broken.

## Expected Behavior
it works just like when using `renderToString`

## Actual Behavior
The generated html breaks the html by injecting the <style> tags at wrong points in the html stream.

<img width="927" alt="Screenshot 2024-04-07 at 3 21 49 PM" src="https://github.com/styled-components/styled-components/assets/10056971/c131dd21-7659-4c57-9ed9-9e8d367f9dc6">

quantizor:
Closing — this reports an issue with the deprecated `renderToNodeStream` + `interleaveWithNodeStream` API. SSR streaming is now handled via React Server Components in v6.3.x. Please use the RSC approach or the `StyledComponentsRegistry` pattern with `renderToPipeableStream`.

PR #4327, opened 2024-06-08
Prevent ServerStyleSheet generating empty style tags/elements

hamidrezahanafi:
It seems like this PR not ported over to v6 and with React 18 streaming, it's possible to insert empty style tag in middle of html.
It should fix this issue: https://github.com/styled-components/styled-components/issues/4289

Ported over from: https://github.com/styled-components/styled-components/commit/b9e60f7efb66c0710002f0e133e42af1c7deb497

### DIFF
--- a/packages/styled-components/src/sheet/Rehydration.ts
+++ b/packages/styled-components/src/sheet/Rehydration.ts
@@ -16,7 +16,7 @@ export const outputSheet = (sheet: Sheet) => {
 
     const names = sheet.names.get(id);
     const rules = tag.getGroup(group);
-    if (names === undefined || rules.length === 0) continue;
+    if (names === undefined || !names.size || rules.length === 0) continue;
 
     const selector = `${SC_ATTR}.g${group}[id="${id}"]`;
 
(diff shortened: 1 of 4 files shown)


hamidrezahanafi:
@quantizor would you please take a look at this one? It's going to save much of headaches with SSR

packages/styled-components/src/sheet/Rehydration.ts
```diff
@@ -16,7 +16,7 @@ export const outputSheet = (sheet: Sheet) => {
 
     const names = sheet.names.get(id);
     const rules = tag.getGroup(group);
-    if (names === undefined || rules.length === 0) continue;
+    if (names === undefined || !names.size || rules.length === 0) continue;
```
quantizor, line 19: I'm not sure about this change w.r.t. createGlobalStyle. Could you add a test verifying that cGS usage without any other components works as expected in SSR?

hamidrezahanafi: Test added

quantizor:
Awesome, looks good. Can you settle the prettier job and then I'll approve?

hamidrezahanafi:
> Awesome, looks good. Can you settle the prettier job and then I'll approve?

Done thank you for looking into this!

quantizor (approved):

PR #4328, opened 2024-06-13
Export rehydrate from StyleSheet

hamidrezahanafi:
This will help rehydrate sheet on NextJS/React 18 streaming when initial rehydration is already done and we receive a new chunk from NextJS `useServerInsertedHTML` afterwards.
With this change we can manually get the stylesheet and call rehydration.


hamidrezahanafi:
@quantizor Do you have any objection about exporting this?

quantizor:
> @quantizor Do you have any objection about exporting this?

No should be fine 👍 might want to document it in  the website repo

quantizor (approved):

PR #4333, opened 2024-07-19
Replace deprecated global JSX namespace

Janpot:
The global `JSX` namespace is [deprecated](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#the-jsx-namespace-in-typescript). This Replaces it with `React.JSX`.

Motivation:
This will make the library compatible with `types-react@19.0.0-rc.1` (Upcoming `@types/react` version). It comes at no backwards compatibility cost, the `React.JSX` namespace is already used in some places. At MUI we need `@types/react@19` compatibility (but not necessarily `react@19` compatibility) for https://github.com/mui/material-ui/pull/42824.


quantizor (approved):
Thank you!

PR #4353, opened 2024-09-21
Support for StyleSheetManager props in collectStyles

ashokawardhan:
Upgrading to v6 from v5 for SSR apps would require everyone to add another StyleSheetManager instance if they want to add vendor prefixes.

This increases code size, and rendering time, since StyleSheetManager would need to be evaluated twice. This adds support for passing in StyleSheetManager props, and directly reusing the already initialized StyleSheetManager.

### DIFF
--- a/.changeset/rare-socks-glow.md
+++ b/.changeset/rare-socks-glow.md
@@ -1,0 +1,5 @@
+---
+'styled-components': patch
+---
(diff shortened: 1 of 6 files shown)


quantizor:
Closing — stale with no engagement.

PR #4356, opened 2024-10-05
feat: add nonce to style tag

ericsen-tsai:
# Add NONCE support and improve nonce handling

## Changes

1. Added a new `NONCE` constant:
   - Retrieves the nonce value from a `<meta name="sc-nonce">` tag in the document head when available.
   - Falls back to `undefined` if not in a browser environment or if the meta tag is not present.

2. Enhanced `ServerStyleSheet` class:
   - Added a `nonce` property to the class.
   - Modified the constructor to accept an optional `nonce` parameter.
   - Updated `_emitSheetCSS`, `getStyleElement`, and other methods to use the instance's `nonce` value.

3. Improved nonce handling in `getStyleElement`:
   - Now includes a `<meta name="sc-nonce">` tag in the returned array when a nonce is present.

4. Modified `getNonce` function:
   - Updated to use the new `NONCE` constant as a fallback.

## Purpose

These changes improve the handling of Content Security Policy (CSP) nonces in styled-components, particularly in server-side rendering scenarios. By allowing the nonce to be set explicitly and adding a meta tag for client-side retrieval, we ensure better consistency between server and client nonce usage.

## Impact

- Improves security by supporting CSP nonces more robustly.
- Enhances server-side rendering capabilities with better nonce handling.
- Provides a more flexible API for setting nonces in `ServerStyleSheet`.

## Testing

- Tested server-side rendering with and without nonces.
- Verified client-side nonce retrieval from the new meta tag.
- Ensured backward compatibility with existing usage patterns.

### DIFF
--- a/packages/styled-components/src/constants.ts
+++ b/packages/styled-components/src/constants.ts
@@ -13,7 +13,10 @@ export const SC_VERSION = __VERSION__;
 export const SPLITTER = '/*!sc*/\n';
 
 export const IS_BROWSER = typeof window !== 'undefined' && 'HTMLElement' in window;
-
+export const NONCE =
+  typeof document !== 'undefined'
+    ? document.head.querySelector('meta[name="sc-nonce"]')?.getAttribute('content') ?? undefined
+    : undefined;
 export const DISABLE_SPEEDY = Boolean(
   typeof SC_DISABLE_SPEEDY === 'boolean'
     ? SC_DISABLE_SPEEDY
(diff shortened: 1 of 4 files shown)


quantizor:
Superseded by #5693 which builds on your work with additional framework support (Vite auto-detection, StyleSheetManager nonce prop, browser .nonce property reading). Your seed commit is attributed as the author. Thank you for the contribution!

Issue #3951, opened 2023-02-14
Requesting other approaches over process.env to advanced options


hikariNTU:
## Environment

Just like these issues described:
- #3933 
- #3621
- #3166
- #3852

Any project with [Webpack 5](https://webpack.js.org/plugins/environment-plugin/), [Vite](https://vitejs.dev/config/shared-options.html#define) or any similar `process.env` patching bundler without `process` been polyfilled will not work since the source code always check `process` existence in global.

```ts
export const SC_ATTR: string =
  (typeof process !== 'undefined' && typeof process.env !== 'undefined' && (process.env.REACT_APP_SC_ATTR || process.env.SC_ATTR)) ||
  'data-styled';
```

This breaks the whole point of patching only the environment variable during the bundling phase and prevent polyfilling the `process/browser` into distributed files.
  
## Expected Behavior

- An unified options to handle these advanced options without the need of `process` object

## Actual Behavior

- Must place a `process` object into production code and potentially break other library if they use `typeof process` to determine other logic.

probablyup:
These can be added to `StyleSheetManager`

rob-swu:
Looks like this was fixed as part of https://github.com/styled-components/styled-components/releases/tag/v5.3.7-fixed

rob-swu:
@hikariNTU the pr part of the release https://github.com/styled-components/styled-components/pull/3957 fixes it so if process.env is not defined it won't crash.
and they mentioned it was already fixed for v6 in https://github.com/styled-components/styled-components/pull/3907 

I just updated my styled components install to 5.3.8 and am running vite, and now the app will run, whereas before it was failing due to the process env issue.

hikariNTU:
@rob-swu Your are talking about other issue. Read the description carefully.

The issue here is conditional check `process.env`. While using directly string replacement for `process.env.REACT_APP_SC_ATTR` will not work since `process.env` will be `undefined` and the condition fail

quantizor:
Closing — this issue is stale and likely resolved in newer releases. If still reproducible on the latest v6.x, please open a new issue.

PR #4364, opened 2024-10-25
feat: include SC_ATTR as the seed

leoc4e:
### Change
Include `SC_ATTR` as part of the seed.

### Why
In styled-components, the classname is generated using the component ID and a seed. Currently, the seed is created using only `SC_VERSION`. When multiple micro-frontends use the same styled-components version and share components, there’s a risk of classname collisions. This issue can be exacerbated when using an external library built with styled-components, where namespacing is not an option because the library’s components are already transpiled.

By including `SC_ATTR` in the seed, you can prevent classname collisions by setting a unique `SC_ATTR` value for each micro-frontend instance. This serves as an effective way to avoid collisions when multiple instances are in use.

### Caveat
To prevent rendering issues, ensure that `SC_ATTR` remains consistent between SSR and client-side rendering; otherwise, different classnames may be generated.
https://github.com/styled-components/styled-components/issues/3951

Related Issues and PR:
https://github.com/styled-components/styled-components/issues/2974
https://github.com/styled-components/styled-components/pull/3503#issuecomment-896141562


quantizor:
Closing — stale with no follow-up.

PR #5547, opened 2025-01-26
Replace deprecated JSX namespace with direct import

Efklaz:
The problem was that after the request (Replace deprecated global JSX namespace #4333), I couldn’t use styled-components in React 19. This happened because the .d.ts files were generated without using the React namespace in places where React.JSX was required. Instead, they were generated as plain JSX, which caused errors since, in React 19, this is no longer a global namespace but is scoped to React.

### DIFF
--- a/packages/styled-components/src/constructors/styled.tsx
+++ b/packages/styled-components/src/constructors/styled.tsx
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import React, { JSX } from 'react';
 import createStyledComponent from '../models/StyledComponent';
 import { BaseObject, KnownTarget, WebTarget } from '../types';
 import domElements, { SupportedHTMLElements } from '../utils/domElements';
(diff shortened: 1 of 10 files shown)


quantizor:
Has anyone else tested this change and confirmed it is necessary?

gdbd (commented):
its ok

binhpv:
I believe a simpler change is to upgrade the @types/react package to version 19. When doing this in some other projects, TypeScript infer the correct return type as React.JSX.Element.

binhpv:
With this change I have another error:
`../node_modules/.pnpm/styled-components@file+..+..+github+styled-components+packages+styled-components+package.tgz__loi6del3vqmxvl5u5pb7bprybe/node_modules/styled-components/dist/hoc/withTheme.d.ts:3:139 - error TS2694: Namespace 'globalThis' has no exported member 'JSX'.
`
The generated .d.ts file contains globalThis instead of React namespace and still cause error.

I tried to upgrade types/react to version 19 but then there are many issue with defaultProps that is also now removed from React 19, so I'm stuck at the moment :(

quantizor:
The alternate fix was merged, so I will close this one. Thank you.

PR #5566, opened 2025-03-19
fix: add documentation link for Dynamic Creation console warning

murs313:
Hi,

I added a documentation link for the Dynamic Creation console warning.
(It took me a while to understand why it must be defined outside the render method.)

I also added a line break at the end because 'Error Component Stack' was continuing without a line break.
スクリーンショット 2025-03-19 11 18 04


quantizor (approved):
Thank you

PR #5600, opened 2025-08-14
feat: use insertion effects when available (React 18)

skovhus:
Can be used with changing the `styled-components` dependency to:
```
"styled-components": "npm:@skovhus/styled-components@6.1.19-1",
```


skovhus:
Not as performant...

PR #5607, opened 2025-09-15
test sanity proposal

quantizor:
This is @stipsan's PR #4332 on top of my concurrency benchmark PR #5606


quantizor:
@stipsan I went ahead and relaid your PR changes over my updates to get the suite properly making use of React 18. At the moment I'm still seeing no real change, can you look at the benchmark and suggest any changes to make it more representative?

stipsan:
> @stipsan I went ahead and relaid your PR changes over my updates to get the suite properly making use of React 18. At the moment I'm still seeing no real change, can you look at the benchmark and suggest any changes to make it more representative?

@quantizor in order to see more representative results you need to:
- trigger suspense interrupting render
- do much more work during render, if the bench uses `.attrs((p) => ({style: {borderBottomColor: p.$color}}))` type techniques then the penalty of doing work during render that is ultimately discarded is much more apparent.

[For example](https://github.com/sanity-io/styled-components-last-resort/pull/205),

image

PR #5618, opened 2025-11-28
Update csstype dependency from 3.1.3 to 3.2.3

stevenocchipinti:
This updates the pinned `csstype` dependency from `3.1.3` to `3.2.3` to fix a type incompatibility with `@types/react`.

## Problem

`@types/react` depends on `csstype@^3.2.2`, which resolves to `3.2.3`. Since `styled-components` pins `csstype@3.1.3`, projects using both packages end up with two different versions of `csstype`.

The `3.2.x` release of `csstype` changed the type definitions for several CSS properties (e.g., `textTransform`, `alignmentBaseline`), making types from `3.1.x` incompatible with types from `3.2.x.`

This causes TypeScript errors when mixing types from both packages, for example when extending `CSSObject` from `styled-components` with properties typed using `React.CSSProperties`.

## Solution

Update the pinned `csstype` version to `3.2.3` to align with what `@types/react` resolves to.

I was going to use a `^3.2.2` to match `@types/react`, but as the existing dependencies use specific versions, I followed the convention.

I've run the tests, built the package locally and tested it with my project and this seems to resolve the error.

## Workaround

For now, I've used put this in my project to ensure the version of `csstype` matches the one `styled-components` currently depends on to prevent the errors:

```
  "overrides": {
    "csstype": "3.1.3"
  },
```

### DIFF
--- a/.changeset/tricky-flowers-mate.md
+++ b/.changeset/tricky-flowers-mate.md
@@ -1,0 +1,8 @@
+---
+'styled-components': patch
+---
+
+Update csstype dependency from 3.1.3 to 3.2.3
+
+This updates the pinned csstype dependency from 3.1.3 to 3.2.3 to fix a
+type incompatibility with @types/react.


michaelkay-wrisk:
Please could this get some attention when you have chance

stevenocchipinti:
Hello 👋 

Happy new year!
Thanks for approving and merging this PR 🙏 

I think I followed the contribution guidelines but I'm not sure how this change get's released 🤔 
Was I supposed to do something else?

stevenocchipinti:
I think I found the issue, the `yarn install --immutable` command was failing due to mismatching checksums, which I think was caused by `yarn` being updated right before this PR was merged.

I've opened a new PR here to update the `yarn.lock` file: #5622

I think this will fix the build and get this fix released.

@quantizor Please let me know if that seems right or if you need me to make any other changes.

Issue #5608, opened 2025-09-21
`domElements` missing `<search>` tag (`styled.search`)


otomad:
Cannot use `styled.search` to create a [`<search>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/search) tag. Although you can use `styled("search")`, however the typescript typing is wrong (got `never` instead of `HTMLElement`). And the DevTools console will report ```Invalid DOM property `class`. Did you mean `className`? Error Component Stack```, however I never type className to class by mistaken. 

### Solution

Add `search` to here:
https://github.com/styled-components/styled-components/blob/a21089e1cde9d2492349088787c59dd85358a337/packages/styled-components/src/utils/domElements.ts#L87-L99

Change the logic for determining whether it is a custom element.
https://github.com/styled-components/styled-components/blob/a21089e1cde9d2492349088787c59dd85358a337/packages/styled-components/src/models/StyledComponent.ts#L186-L192

The current logic just determine whether the tag is a known tag, otherwise it is a custom element tag. This is wrong, which will effect to the future native tags. The correct way is to check if the tag name has any hyphens.

It can be changed to this:
```typescript
propsForElement[
    // handle custom elements which React doesn't properly alias
    isTag(elementToBeCreated) &&
    elementToBeCreated.includes('-')
      ? 'class'
      : 'className'
  ] = classString;
```

PR #5620, opened 2025-12-21
feat: add mostly transparent RSC support

quantizor:
Adjust library to function properly under RSC by conditionally bypassing some hook usage. Normally conditional use of hooks is a no-no, but the final behavior is deterministic per runtime environment.

Also adds trusted publishing and fixes https://github.com/styled-components/styled-components/issues/5608

### DIFF
--- a/.changeset/eight-webs-cross.md
+++ b/.changeset/eight-webs-cross.md
@@ -1,0 +1,5 @@
+---
+'styled-components': patch
+---
+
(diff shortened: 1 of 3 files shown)


quantizor:
This PR is available at `styled-components@prerelease` for testing.

quantizor:
If you are interested in RSC support, please test the build and react to https://github.com/styled-components/styled-components/issues/5624

PR #5622, opened 2026-01-05
Fix the build by updating `yarn.lock`

stevenocchipinti:
Yarn was updated in c43a66cebfdb7fdaa42c0a1e4a71fa181a0e0bc3 just before pr #5618 was merged.
I think this is causing the `yarn install --immutable` command to fail due to mismatching checksums when [running the checks](https://github.com/styled-components/styled-components/actions/runs/20405357581/job/58634293229).

I updated `yarn` and re-ran `yarn install` locally to update the `yarn.lock` file and now the checksums match, which I think will fix the build.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -6175,7 +6175,7 @@ __metadata:
 "csstype@npm:3.2.3":
   version: 3.2.3
   resolution: "csstype@npm:3.2.3"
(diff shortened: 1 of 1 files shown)


stevenocchipinti:
@quantizor I haven't added a changeset to this PR as the last PR (#5618) had that and this is just to fix the build. 

Do I need one to trigger the release?

PR #5644, opened 2026-01-15
Add exports to package.json

coolassassin:
Hello!

I am trying to move on ESM modules in my project and on Vitest and I have started to catch errors like so:
```
TypeError: styled.div is not a function
```

I have this issue when I use this kind of import:
```
```

While I am writing my application over bundlers all works fine, because every bundler goes to package.json and takes "module" instruction from there.
But when I use vitest and it sees package which is marked as `"type": "module"` it tries to resolve it with a Node.js module way
By the [Node.js module resolution](https://nodejs.org/api/packages.html?utm_source=chatgpt.com#main-entry-point-export) it tries to read "exports" and if there is no information it goes to the "main" field. And there is "cjs" as you can see

Here is a scheme:
```
Bundler(App) -> package.json(any) -> "module" ✅
Vitest(Node.js) -> package.json(without "type": "module") -> "main(cjs)" ✅
Vitest(Node.js) -> package.json("type": "module") -> "main(cjs)" ❌
```
With export it will work like so:
```
Bundler(App) -> package.json(any) -> "module" ✅
Vitest(Node.js) -> package.json(without "type": "module") -> "exports.require(cjs)" ✅
Vitest(Node.js) -> package.json("type": "module") -> "exports.import(esm)" ✅
```

---

> [!NOTE]
> Improves Node/ESM resolution by defining an explicit `exports` map for `styled-components`.
> 
> - Adds `exports` in `packages/styled-components/package.json` covering `.` and `./native` with `import` (ESM), `require` (CJS), `browser`, `react-native`, and `types`, plus `./package.json`
> - Ensures correct entry selection across bundlers, Node/Vitest, browser, and React Native
> - Adds a changeset marking a patch release
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7f22eb0f8f36b9bddc77795e3f10020d0a5c8d13. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/styled-components/package.json
+++ b/packages/styled-components/package.json
@@ -16,6 +16,7 @@
   "exports": {
     ".": {
       "types": "./dist/index.d.ts",
+      "react-native": "./native/dist/styled-components.native.cjs.js",
       "browser": {
         "import": "./dist/styled-components.browser.esm.js",
         "require": "./dist/styled-components.browser.cjs.js"


quantizor (approved):
Thank you!

coolassassin:
@quantizor 
https://github.com/styled-components/styled-components/commit/788e8c05606a621dae748e2a5b36bdbb6a0c8045
Can you tell me why it was reverted? It is neccessary changes.
Can I remake it somehow to do it right?

quantizor:
> @quantizor [788e8c0](https://github.com/styled-components/styled-components/commit/788e8c05606a621dae748e2a5b36bdbb6a0c8045) Can you tell me why it was reverted? It is neccessary changes. Can I remake it somehow to do it right?

It broke a bunch of stuff unfortunately and caused issues like multiple versions of the library being loaded, type visibility problems for composite projects, etc. There's probably a different way to fix it.

quantizor:
We can change the `main` field to the ESM bundle in the next major. We're probably due for one soon.

coolassassin:
@quantizor can I reproduce these bugs somehow? Maybe I can add some tests
I need this fix 😔

coolassassin:
I have opened this PR:
https://github.com/styled-components/styled-components/pull/5662

There, I have mentioned the issue withthe composite projects has appeared before my change

Issue #5648, opened 2026-01-16
No longer works in composite typescript projects since 6.3.7


domarmstrong:
In a typescript project which uses `composite: true` the styled-components types now fail in the simplest case with:

```
TS2742: The inferred type of X cannot be named without a reference to ../node_modules/styled-components/dist/types. This is likely not portable. A type annotation is necessary.
```

This appears to be since the addition of the package exports field.

One solution appears to be to add `export * from './types'` rather than a restricted list [here](https://github.com/styled-components/styled-components/blob/main/packages/styled-components/src/index.ts#L4).

See very simple reproduction [here](https://github.com/domarmstrong/styled-components-repro).

imccausl:
FWIW, we are having the same issue, but do not use `composite: true`. Reverting to 6.3.6 fixed it.

justnewbee:
6.3.8 still breaks...

ALI-KHALED-13:
for me reverting to 6.3.5 fixed it, newer versions break

ALI-KHALED-13:
so as i mentioned reverting to 6.3.5 fixed it locally for me (windows) but it kept failing to build in my hosting service Linux environment

been trying with different solutions last couple of days, and I hope this solution fixes it for all
changing tsconfig.json options of these 2 fields to 
```
"moduleResolution": "node",  // Changed from "bundler"
 "esModuleInterop": true, 
```

did the trick

hope it helps, I use it with react 18

quantizor:
Please try 6.3.9 https://github.com/styled-components/styled-components/releases/tag/styled-components%406.3.9

ALI-KHALED-13:
> Please try 6.3.9 https://github.com/styled-components/styled-components/releases/tag/styled-components%406.3.9

version 6.3.9 fixed it for me, thanks @quantizor

PR #5660, opened 2026-02-05
Allow access to all types from types.d.ts

limarkxx:
https://github.com/styled-components/styled-components/issues/5648

The internal types may need to be available to be accessed to avoid TS2742 errors.


quantizor:
Thanks for your PR, but I'm reverting the `exports` package.json change, which should resolve this issue. Apologies for the trouble.

PR #5662, opened 2026-02-10
Add exports to package.json

coolassassin:
It is a redo for [this PR](https://github.com/styled-components/styled-components/pull/5644)

I have updated to the latest release, and my Vitest setup is failing again after this commit:
https://github.com/styled-components/styled-components/commit/788e8c05606a621dae748e2a5b36bdbb6a0c8045

I have tried to reproduce the bug, which is described in the commit, but I have no results
Maybe only the second part of the fix is really necessary?

Also, I want to add that if we talk about these issue:
https://github.com/styled-components/styled-components/issues/5648
We can see that the issue has been present since 6.3.7.
But exports were added in 6.3.8

### DIFF
--- a/packages/styled-components/package.json
+++ b/packages/styled-components/package.json
@@ -13,6 +13,31 @@
     "./dist/styled-components.cjs.js": "./dist/styled-components.browser.cjs.js",
     "./dist/styled-components.esm.js": "./dist/styled-components.browser.esm.js"
   },
+  "exports": {
+    ".": {
+      "react-native": {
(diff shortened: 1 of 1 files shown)


quantizor:
Closing — we've tried adding `exports` before and it caused TS2742 errors in composite TypeScript projects and module resolution ordering issues. The current `main`/`module`/`browser` field approach is the most compatible.

If you're hitting a specific resolution issue with a modern bundler, please open an issue with the exact error and bundler config so we can evaluate targeted fixes.

Issue #5663, opened 2026-02-14
React 19 / Next App Router RSC: styled-components hoisted <style> triggers warning: React expected the `href` prop for a <style> tag opting into hoisting semantics using the `precedence` prop to not have any spaces but ecountered spaces instead.


mobeigi:
## Current Behaviour
In my project: https://github.com/mobeigi/mobeigi.com/

I am using these [versions](https://github.com/mobeigi/mobeigi.com/blob/main/app/package.json):
```
react: 19.2.4
next: 15.4.11
node: >=24

"styled-components": "^6.3.9"
```

In `next.config.ts`:
```
compiler: {
  styledComponents: true,
},
```

I am using Next.js APP Router with React 19.

I am using the [StyledComponentsRegistry ](https://github.com/mobeigi/mobeigi.com/blob/main/app/src/lib/StyledComponentsRegistry/StyledComponentsRegistry.tsx) approach with `'use client'` in all my `styled.ts` files.

In this setup, everything is behaving as expected.

## Post RSC Behaviour

There has been recent support for styled components to use RSC.
As per the documentation I wanted to try this by removing the **StyledComponentsRegistry** and `'use client'` in all my `styled.ts` files. I have a [branch ](https://github.com/mobeigi/mobeigi.com/pull/301/changes) showing the changes. Unfortunately, I wasn't able to reproduce the issue in a small minimum example repository.

Making those changes and running `yarn dev` I notice a few things: 
* Console is spamming `React expected the `href` prop for a <style> tag opting into hoisting semantics using the `precedence` prop to not have any spaces but ecountered spaces instead. using spaces in this prop will cause hydration of this style to fail on the client. The href for the <style> where this ocurred is "sc-sc-eACynQ-uXfoQ IppAg".`
* There is some regression in styling (notable a style like [this](https://github.com/mobeigi/mobeigi.com/blob/deba0900b2c2c6f4e0985246460bc611483bfa3f/app/src/components/PersonShowcase/styled.ts#L63-L67) isn't applying all of its parent styles via `styled()` correctly, we lose the `width` and `height` CSS styles on `svg` from `IconWrapper`).
* There is a last style flash (FOUC) on page refresh affecting some styling. Although this might be because I have some `'use client'` components mixing client logic and styles in one file.

The `<head>` of my HTML contains:
```
<html>
<body>
<!--StartFragment-->
[...]

PR #5684, opened 2026-03-19
fix: prevent HMR style tag accumulation in RSC mode

### LINKED ISSUE #5670: hotloading styles doesn't unmount the previous style tag
When I update my styles, a new `style` tag is added with the new styles. Normally, this goes well because there's a new className. But:

```
export const CategoryBadge = styled.div`
  font-size: normal;
  font-weight: bold;

  & + & {
    margin-left: 4.5rem;
  }
`;
```

```
<style data-href="sc-styles__CategoryBadge-sc-4ae1dfd-0-dEiWqV" data-precedence="styled-components">.dEiWqV{font-size:normal;font-weight:bold;}/*!sc*/
.styles__CategoryBadge-sc-4ae1dfd-0+.styles__CategoryBadge-sc-4ae1dfd-0{margin-left:1.5rem;}/*!sc*/
</style>
<style data-href="sc-styles__CategoryBadge-sc-4ae1dfd-0-fxbJLw" data-precedence="styled-components">.fxbJLw{font-size:normal;font-weight:bold;}/*!sc*/
.styles__CategoryBadge-sc-4ae1dfd-0+.styles__CategoryBadge-sc-4ae1dfd-0{margin-left:4.5rem;}/*!sc*/
</style>
```

Because I use the `&` symbol, the permanent className is used. So these style tags all keep applying styles.

Image

quantizor:
## Summary
- In **development**, use stable hrefs (`sc-{componentId}` only) on RSC `<style>` tags so React 19 replaces them on hot reload instead of treating each CSS content change as a new permanent resource
- In **production**, content-aware hrefs (including name hashes) are preserved for correct dynamic variant deduplication
- Adds 3 new RSC tests validating HMR-compatible href behavior

## Background
When styles change during HMR, the CSS content hash changes, producing a new `href`. React 19 treats `<style precedence="...">` tags as permanent resources keyed by `href` — two different hrefs = two separate tags, neither removed. This causes stale style tags to accumulate, and stale `&`-based selectors keep matching because they use the stable component ID class.

## Tradeoff
Dynamic prop variants (e.g., `$bg="red"` vs `$bg="blue"`) may incorrectly dedup in dev mode since they share the same stable href. This is acceptable because:
- It's dev-only — production uses content-aware hrefs
- Both variants' CSS still renders (last one wins)
- HMR correctness is more important than variant isolation in dev

## History check
- No prior attempts at this approach found in closed issues/PRs
- Content-aware hrefs were added in #5663 (Feb 17) for dedup correctness — HMR was not considered
- `createGlobalStyle` already avoids `precedence` entirely (different mechanism)

## Test plan
- [x] All 506 tests pass (503 existing + 3 new HMR tests)
- [x] Build passes with bundlewatch
- [x] All 17 RSC tests pass
- [ ] Manual HMR test with Next.js sandbox app

Closes #5670

### DIFF
--- a/.changeset/fix-hmr-style-accumulation.md
+++ b/.changeset/fix-hmr-style-accumulation.md
@@ -2,4 +2,4 @@
 "styled-components": patch
 ---
 
-Fix HMR style tag accumulation in RSC mode. Style tags now use stable hrefs in development so React 19 replaces them on hot reload instead of accumulating new permanent resource tags.
+Fix HMR style tag accumulation in RSC mode. Style tags omit `precedence` and `href` in development so React treats them as regular elements that unmount on re-render, preventing stale tags from persisting.
--- a/packages/styled-components/src/constructors/test/styled.rsc.test.tsx
+++ b/packages/styled-components/src/constructors/test/styled.rsc.test.tsx
@@ -31,10 +31,13 @@ describe('styled RSC mode', () => {
   });
 
   describe('style tag href attribute (#5663)', () => {
-    // React 19 uses href on <style precedence="..."> for deduplication.
-    // Spaces in href cause: console warning, hydration failure, and style loss.
+    // In dev mode, precedence/href are omitted so React treats style tags as
+    // regular elements (prevents HMR stale tag accumulation, #5670).
+    // These tests verify the style content is still correct in dev mode.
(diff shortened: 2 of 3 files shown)


quantizor:
Available to test via `styled-components@6.3.13-prerelease.0`

Issue #3146, opened 2020-05-19
CreateGlobalStyles Rendered After Component Styles


pxwee5:
![image](https://user-images.githubusercontent.com/14123790/82287855-40b71900-99f5-11ea-9e7e-371484f2310d.png)

## System:
 - OS: macOS High Sierra 10.13.6
 - CPU: (8) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
 - Memory: 74.02 MB / 16.00 GB
 - Shell: 3.2.57 - /bin/bash
## Binaries:
 - Node: 10.16.0 - ~/.nvm/versions/node/v10.16.0/bin/node
 - Yarn: 1.17.3 - /usr/local/bin/yarn
 - npm: 6.9.0 - ~/.nvm/versions/node/v10.16.0/bin/npm
## npmPackages:
 - babel-plugin-styled-components: ^1.10.7 => 1.10.7 
 - styled-components: ^5.1.0 => 5.1.0 

## Expected Behavior
```
/* createGlobalStyle */
html {
  background-color: gray;
}
.cooVxd {
  padding-left: 50px;
}
body {
  background-color: yellow;
}

/* Component styles */
.cooVxd {
  padding-left: 15px;
  padding-right: var(--test);
  /* Trimmed */
}
.fDNorw {
  padding-left: 15px;
  /* Trimmed */
}
```

## Actual Behavior
Component styles loads before Global styles.

```
/* Component styles */
.cooVxd {
  padding-left: 15px;
  padding-right: var(--test);
  /* Trimmed */
}
.fDNorw {
  padding-left: 15px;
  /* Trimmed */
}

/* createGlobalStyle */
html {
  background-color: gray;
}
.cooVxd {
  padding-left: 50px;
}
body {
  background-color: yellow;
}
```

ptdede:
I also experienced this.

probablyup:
@pxwee5 need a codesandbox reproduction to be able to help you, thanks!

arteforme:
@probablyup 

I'm seeing this as well occur as well.  I put together a [codesandbox](https://codesandbox.io/s/vigorous-rgb-osztp?file=/src/App.tsx) that has the issue.  Below is what I see in Chrome:

![Screen Shot 2020-09-15 at 12 20 59 AM](https://user-images.githubusercontent.com/6117939/93178805-996b9980-f6e9-11ea-848f-67699fea3973.png)

probablyup:
[...]

Issue #3022, opened 2020-02-14
createGlobalStyle creates memory leak in Next.js


robbue:
### Binaries:
 - Node: 13.7.0 - ~/.nvm/versions/node/v13.7.0/bin/node
 - Yarn: 1.21.1 - /usr/local/bin/yarn
 - npm: 6.13.6 - ~/.nvm/versions/node/v13.7.0/bin/npm
### npmPackages:
 - babel-plugin-styled-components: ^1.10.7 => 1.10.6 
 - styled-components: ^5.0.1 => 5.0.1 

## Problem
We are using styled-components and Next.js (7.2.1) and we have experienced memory leaks in our app for some time, so I started to investigate it, in the process I found other issues here about: 
- using the css helper when composing styling partials to be interpolated into a styled component
- not to use `@import` inside `createGlobalStyle`
- not dynamically create components

So I went through the whole app making sure to avoid these. I deployed a new version but the memory leak was still present. As seen on the graph below:
<img width="942" alt="Skjermbilde 2020-02-14 19 49 28" src="https://user-images.githubusercontent.com/1470041/74558893-6a2e5780-4f63-11ea-99b6-ceb72fed1239.png">

After inspecting the node server I found out it was `createGlobalStyle` in our [`_app.js`](https://nextjs.org/docs/advanced-features/custom-app) (Next.js file) that caused the memory to rise. In the `createGlobalStyle` we used `fontFace`/`normalize` from [polished](https://github.com/styled-components/polished)), and around 50 other lines of reset css (nothing special).

Removing `createGlobalStyle` solved the memory leak  (or at least reduced to a minimum). 
See the graph below:
<img width="951" alt="Skjermbilde 2020-02-14 19 50 55" src="https://user-images.githubusercontent.com/1470041/74558899-6bf81b00-4f63-11ea-9d33-ccc43076c023.png">

Did we use `createGlobalStyle`wrong or do someone have some insight to this?

boyuan459:
in our next.js project, we can only use one createGlobalStyle, other global styles is replaced with next.js <style global jsx>

PaulACoroneos:
Yes we are having the same issue in our Next app as well. Only difference is that we are using latest Next version.

PaulACoroneos:
We confirmed the rollback resolved the issue. Our current style file is just a copy of normalize css and reset css with some font-face definitions. Something like

export const AppGlobalStyle = createGlobalStyle<IAssetsProps>`
  ${reset}
[...]

PR #5687, opened 2026-03-19
Version Packages

quantizor:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Minor Changes

-   b0f3d29: `.attrs()` improvements: props supplied via attrs are now automatically made optional on the resulting component (previously required even when attrs provided a default). Also fixes a bug where the attrs callback received a mutable props object that could be changed by subsequent attrs processing; it now receives an immutable snapshot.

-   2a973d8: Dropped IE11 support: ES2015 build target, inlined unitless CSS properties (removing @emotion/unitless dependency), removed legacy React class statics from hoist and other unnecessary code.

-   9e07d95: Add `createTheme(defaultTheme, options?)` for CSS variable theming that works across RSC and client components.

    Returns an object with the same shape where every leaf is `var(--prefix-path, fallback)`. Pass it to `ThemeProvider` for stable class name hashes across themes (no hydration mismatch on light/dark switch).

    ```ts
    const theme = createTheme({ colors: { primary: '#0070f3' } });
    // theme.colors.primary → "var(--sc-colors-primary, #0070f3)"
    // theme.raw → original object
    // theme.vars.colors.primary → "--sc-colors-primary"
    // theme.resolve(el?) → computed values from DOM (client-only)
    // theme.GlobalStyle → component that emits CSS var declarations
    ```

    `vars` exposes bare CSS custom property names (same shape as the theme) for use in `createGlobalStyle` dark mode overrides without hand-writing variable names:

    ```ts
    const { vars } = createTheme({ colors: { bg: '#fff', text: '#000' } });

    const DarkOverrides = createGlobalStyle`
      @media (prefers-color-scheme: dark) {
        :root {
          ${vars.colors.bg}: #111;
          ${vars.colors.text}: #eee;
        }
      }
    `;
    ```

    Options: `prefix` (default `"sc"`), `selector` (default `":root"`, use `":host"` for Shadow DOM).

-   79cc7b4: Add first-class CSP nonce support. Nonces can now be configured via `StyleSheetManager`'s `nonce` prop (recommended for Next.js, Remix), `ServerStyleSheet`'s constructor, `<meta property="csp-nonce">` (Vite convention), `<meta name="sc-nonce">`, or the legacy `__webpack_nonce__` global.

-   b0f3d29: Rearchitect `createGlobalStyle` to use shared stylesheet groups.

    All instances of a `createGlobalStyle` component now share a single stylesheet group, registered once at definition time. This fixes unmounting one instance removing styles needed by others (#5695), styles scattering after remount (#3146), and group ID leaks during SSR (#3022).

    CSS injection order is now fully determined at definition time (lower group ID = earlier in stylesheet). Render order no longer affects CSS order. Keyframes defined before a component correctly appear before that component's rules.

    Also fixes: O(n^2) performance regression in jsdom test environments from unbounded rule accumulation, and stale static global styles during client-side HMR (effect deps now include the `globalStyle` reference so module re-evaluation triggers re-injection).

-   b0f3d29: Significant render performance improvements via three-layer memoization and hot-path micro-optimizations. Client-only; server renders are unaffected.

    Re-renders that don't change styling now skip style resolution entirely. Components sharing the same CSS (e.g., list items) benefit from cross-sibling caching. Hot-path changes include `forEach` → `for`/`for...of`, template literal → manual concat, and reduced allocations.

    Benchmarks vs 6.3.12:

    -   **Parent re-render (most common):** 3.3x faster
    -   **First mount:** 1.7-2.5x faster
    -   **Prop cycling:** 2.3-2.4x faster
    -   **10K heavy layouts:** 1.9x faster
    -   No regressions on any benchmark

-   9ada92b: React Server Components support: inline style injection, deduplication, and a new `stylisPluginRSC` for child-index selector fixes.

    **Inline style injection:** RSC-rendered styled components emit `<style data-styled>` tags alongside their elements. CSS is deduplicated per render via `React.cache` (React 19+). Extended components use `:where()` zero-specificity wrapping on base CSS so extensions always win the cascade regardless of injection order.

    **`StyleSheetManager` works in RSC:** `stylisPlugins` and `shouldForwardProp` are now applied in server component environments where React context is unavailable.

    **`stylisPluginRSC`** — opt-in stylis plugin that fixes `:first-child`, `:last-child`, `:nth-child()`, and `:nth-last-child()` selectors broken by inline `<style>` tags shifting child indices. Rewrites them using CSS Selectors Level 4 `of S` syntax to exclude styled-components style tags from the count.

    ```jsx
    import { StyleSheetManager, stylisPluginRSC } from 'styled-components';

    <StyleSheetManager stylisPlugins={[stylisPluginRSC]}>{children}</StyleSheetManager>;
    ```

    The plugin rewrites `:first-child`, `:last-child`, `:nth-child()`, and `:nth-last-child()` using CSS Selectors Level 4 `of S` syntax to exclude injected style tags from the child count.

    Browser support: Chrome 111+, Firefox 113+, Safari 9+ (~93% global). In unsupported browsers, the entire CSS rule is dropped — only opt in if your audience supports it. Use `:first-of-type` / `:nth-of-type()` as a universally compatible alternative.

    **HMR:** Stale styles during client-side HMR are detected and invalidated when module re-evaluation creates new component instances while IDs remain stable (SWC plugin assigns IDs by file location). `createGlobalStyle` additionally clears stale sheet entries when the instance changes between renders.

    The plugin is fully tree-shakeable — zero bytes in bundles that don't import it.

### Patch Changes

-   b0f3d29: Expose `as` and `forwardedAs` props in `React.ComponentProps` extraction for styled components
-   553cbb4: Fix memory leak in long-running apps using components with free-form string interpolations (e.g. `color: ${p => p.$dynamicValue}` where the value comes from unbounded user input).
-   b0f3d29: React Native improvements: replaced postcss with a lightweight CSS declaration parser, fixing `nanoid` crashes in Expo/Metro (#5705) and improving parse speed 4-6x. Parent re-renders with unchanged children are 2.6-3.2x faster via cache-first render. Updated native component alias list (removed 5 dead components, added 4 missing). Added `react-native` as an optional peer dependency.
-   74e8b76: Smaller install footprint via unused dependency cleanup.


quantizor:
Edit: 6.4.0 RC4 is available for testing at `styled-components@test`, thank you. If all looks well, please react to this message!

quantizor:
New prerelease published: `styled-components@6.4.0-prerelease.5`

```sh
npm i styled-components@test
```

Changes since prerelease.4:
- Deduplicate RSC inline `<style>` tags across component instances via `React.cache` (React 19+) — reduces HTML bloat when multiple instances of the same styled component render with identical CSS

quantizor:
`6.4.0-prerelease.6` published to `npm i styled-components@test`:

- **#5670** — RSC style tag accumulation across HMR cycles (PR #5702)

quantizor:
`6.4.0-prerelease.7` (final RC pending any bugfixes) is out

- #5704 
- #5703 

Benchmarks vs 6.3.12:

  -   **Parent re-render cascading to children (most common pattern):** 3.3x faster for 50 children, 3.3x for 100
  -   **First mount of dynamic component trees:** 1.7x for individual components, 2.5x for sibling-heavy layouts
  -   **Prop changes cycling through many values:** 2.3x for 50 children, 2.4x for 100
  -   **Heavy, dynamic layouts (10K children cycling colors):** 1.9x without attrs, 1.6x with attrs chains

quantizor:
`6.4.0-prerelease.8` is out, major RN improvements!

- #5706 

Screenshot 2026-03-27 at 10 32 53 PM

quantizor:
`6.4.0-prerelease.9` is out with an RSC-HMR fix (needed to add a cache-busting condition)

This should be the last RC. Please 👍 if it works for you!

```
npm i styled-components@test
```

laurenskling:
So in the new version, all my RSC styles are supposed to be inline `<style>` tags, correct? With this change, the Elements tab bubbles full of duplicate style rules. Here's my:

I have a component that accepts 3 props. It outputs different css based on those props, just like we loved to do for all these years. In this new version, a style tag is created for each possible output. Meaning shared css, outside of the prop checking, is added multiple times to all those style tags. Resulting in my styles tab bubbling full of the same rules:
```
export default styled.li<{ $someprop?: boolean }>`
  position: relative;
  ${({ $someprop }) => $someprop && css`
    position: static;
  `}
`;
```

(.lhACiT have $someprop)
Scherm­afbeelding 2026-03-31 om 11 26 22

Scherm­afbeelding 2026-03-31 om 11 25 45

It's not bug, it's a feature. Do we want it?

I know you've suggested to move away from props checking styling in RSC, but until we do...

laurenskling:
Having inline `style` tag breaks `:first-child` and `:last-child` type of selectors currently used :/

quantizor:
`6.4.0-prerelease.10` is out:

```
npm i styled-components@test
```

Changes since prerelease.9:
- New `stylisPluginRSC` export — opt-in stylis plugin that rewrites `:first-child`, `:last-child`, `:nth-child()`, and `:nth-last-child()` to exclude inline `<style data-styled>` tags from the child count using CSS Selectors L4 `of S` syntax (Chrome 111+, Firefox 113+, Safari 9+). This fixes the child-index selector breakage caused by RSC inline style injection.
- `StyleSheetManager` now works in RSC via a module-level override slot (no `createContext` needed). Wrap your RSC tree to apply `stylisPlugins` and `shouldForwardProp`.

Usage:
```jsx
import { StyleSheetManager, stylisPluginRSC } from 'styled-components';

// In your RSC layout
export default function Layout({ children }) {
  return (
    <StyleSheetManager stylisPlugins={[stylisPluginRSC]}>
      {children}
    </StyleSheetManager>
  );
}
```

@laurenskling the `:first-child`/`:last-child` breakage you reported is addressed by `stylisPluginRSC` above. The most robust fix is to refactor to `:first-of-type`/`:nth-of-type()` where possible (these are naturally immune since they filter by tag name), but if that's not feasible the plugin will handle it. Give it a try and let me know how it goes!

laurenskling:
wow that was fast. The plugin does the trick. So that means I now have:
```
import { StyleSheetManager, stylisPluginRSC } from 'styled-components';
import StyledComponentsRegistry from '@lib/registry';

// In your RSC layout
export default function Layout({ children }) {
  return (
    <StyleSheetManager stylisPlugins={[stylisPluginRSC]}>
      <StyledComponentsRegistry>
        {children}
      </StyledComponentsRegistry>
    </StyleSheetManager>
  );
}
```

Where `StyledComponentsRegistry` still has a `StyleSheetManager` in use client. This is not a problem? Two managers?

quantizor:
> This is not a problem? Two managers?

It's ok, nesting is supported for StyleSheetManager. I'll make some tweaks to the documentation to make this more obvious.

quantizor:
~~`6.4.0-prerelease.11`~~ `6.4.0-prerelease.12` is out - please 👍 the comment if it works for you.

```
  npm i styled-components@test
```

  Changes since prerelease.10:
  - `stylisPluginRSC` now expands adjacent sibling combinator (`+`) to account for inline `<style
  data-styled>` tags between siblings
  - `StyleSheetManager` nesting inherits `stylisPlugins`, `shouldForwardProp`, and `nonce` from parent—no more redundant config on inner SSMs
  - RSC performance: 1.8-7.4x faster via name-based dedup, compiled CSS cache, and merged
  keyframes/component `<style>` tags
  - RSC output: ~10% smaller CSS by stripping rehydration splitter comments (unused in RSC)
  - Hot-path streamlining: ~3-8% faster renders (flatten restructure, single-pass sanitizeCSS, lazy dev
  allocations)
  - Fix: `sanitizeCSS` tail validation for unterminated strings
  - Fix: custom elements no longer trigger unknown prop warning
  - Fix: boundary-aware class name matching in RSC partial-dedup filter
  - Fix: static `createGlobalStyle` dedup in RSC (duplicate `<style>` tags eliminated)
  - Replaced `rollup-plugin-terser` with `@rollup/plugin-terser`, resolved all audit vulnerabilities (29
  → 0)

#5708 

Pretty much everything in the library is about as well-tuned as I can get it without big architectural changes to align more closely with React 19. Pending any new bugs, this will ship next week.

wojtekmaj:
Testing 6.4.0-prerelease.12.

No meaningful regressions in performance and bundle size.

I'm still facing the issue:

```
error TS7006: Parameter 'ref' implicitly has an 'any' type.
```

when `esModuleInterop` is disabled.

Repro:

```

export const Wrapper = styled.div``;

<Wrapper
  ref={(ref) => {
    console.log(ref);
  }}
/>;
```

Playground:

https://www.typescriptlang.org/play/?esModuleInterop=false&ts=5.9.3#code/JYWwDg9gTgLgBAZxgTwDYFMAmcBmUIhwDkSaWAtAMYGQB26tMCRA3AFCiSxwBUcAhgjgAldP0rw8BYlDETWbNugAeXeNVpI4AdSj8wYdFDgBeRCgyYAdJmAA3AAYP2bADy79hqGzhxZOEwBvAAp-AEpTAD44QJ9fOA0ECAwrVAgAc1D0HDD2XwBffLYAekiWIA

Oddly enough I'm not seeing this issue on 6.0.2 but only in Playground. In real codebase it's very much there.

quantizor:
> In prerelease 12 I'm still facing the issue:
> 
> ```
> error TS7006: Parameter 'ref' implicitly has an 'any' type.
> ```
> 
> when `esModuleInterop` is disabled.

Ah, sorry I intended to fix that and forgot. Will get that in the final.

quantizor:
@wojtekmaj try `styled-components@6.4.0-prerelease.13` when you get a moment please!

wojtekmaj:
@quantizor Thanks! We're getting there, but I still managed to break it. The below code works perfectly on 6.3.12 but crashes on 6.4.0-prerelease.13:

```tsx

export const Checkbox = styled.input``;

type CheckboxProps = React.ComponentPropsWithoutRef<typeof Checkbox>;

const props: CheckboxProps = {
  value: 'on',
};

<Checkbox
  ref={(ref) => {
    console.log(ref);
  }}
  // Crucial for repro: Not adding this bit causes ref to be correctly typed
  {...props}
/>;

```

```
repro.tsx:12:9 - error TS7006: Parameter 'ref' implicitly has an 'any' type.

12   ref={(ref) => {
           ~~~
```

On a positive note, I'm testing this on 300,000 LoC codebase, and that is the only error! I might have a better test suite than you do! 😅

quantizor:
> @quantizor Thanks! We're getting there, but I still managed to break it. The below code works perfectly on 6.3.12 but crashes on 6.4.0-prerelease.13:
> 
> ...
> 
> On a positive note, I'm testing this on 300,000 LoC codebase, and that is the only error! I might have a better test suite than you do! 😅

Looking, and I'm making regression tests for all of these so your help is super appreciated!

quantizor:
Fixed in 4c0b72a3. The root cause was `Substitute`'s conditional type deferring during JSX spread resolution, which blocked TypeScript's contextual typing for ref callback parameters.

The fix splits `PolymorphicComponent`'s single callable overload into three: one for `as` polymorphism, one for `forwardedAs`, and a default that uses `BaseProps` directly (bypassing `Substitute`). The default overload handles your exact scenario — ref callback + spread props without `as`/`forwardedAs`.

@wojtekmaj your repro is now a permanent type test. Thanks for the thorough report and the 300K LoC test suite!

wojtekmaj:
https://youtu.be/Ccoj5lhLmSQ 🤣

quantizor:
@wojtekmaj please try `styled-components@6.4.0-prerelease.14` when you get a chance. Typechecking should be a little faster as well.

wojtekmaj:
@quantizor Tested 15 apps, 8 web, and 7 native, all green!

quantizor:
> @quantizor Tested 15 apps, 8 web, and 7 native, all green!

@wojtekmaj Thank you!! You've seriously been so helpful with debugging this release, very much appreciated.

laurenskling:
I've had this weird bug in prerelease.10+ and I found the reason now. Styles are dropped, _after a refresh or hmr_ (which I don't know why that is), for restyles of a originally rule-less SC. For example:

```
const Button = styled.div``;

export default styled(Button)`
  background-color: green;
`;
```

This div will not have a background color, unless you put _some_ rule inside of the `Button`.

laurenskling:
> So in the new version, all my RSC styles are supposed to be inline `<style>` tags, correct? With this change, the Elements tab bubbles full of duplicate style rules. Here's my:
> 
> I have a component that accepts 3 props. It outputs different css based on those props, just like we loved to do for all these years. In this new version, a style tag is created for each possible output. Meaning shared css, outside of the prop checking, is added multiple times to all those style tags. Resulting in my styles tab bubbling full of the same rules:
> 
> ```
> export default styled.li<{ $someprop?: boolean }>`
>   position: relative;
>   ${({ $someprop }) => $someprop && css`
>     position: static;
>   `}
> `;
> ```
> 
> (.lhACiT have $someprop) Scherm­afbeelding 2026-03-31 om 11 26 22
> 
> Scherm­afbeelding 2026-03-31 om 11 25 45
> It's not bug, it's a feature. Do we want it?
> 
> I know you've suggested to move away from props checking styling in RSC, but until we do...

This seems fixed in .14. I hadn't seen any specific changelogs to it, but you did fix it? Nice :)

I see now, they result to different classNames now, so no longer pointing to the same thing over and over.

rafaell-lycan:
@quantizor could you generate a mini benchmark comparing this version with Emotion and perhaps Linaria?

Anywho, this update is great news for those who are still on Styled Components and can't migrated.

Thanks a lot 🙏

quantizor:
> @quantizor could you generate a mini benchmark comparing this version with Emotion and perhaps Linaria?
> 
> Anywho, this update is great news for those who are still on Styled Components and can't migrated.
> 
> Thanks a lot 🙏

Here is the most recent one I posted on Twitter: https://x.com/quantizor/status/2037541158222463014?s=46 (the workspace version is sc 6.4)

They're here in the repo in the benchmarks package if you want to experiment! Don't have one for Linaria but Emotion is in there.

PR #5691, opened 2026-03-19
fix: prevent attrs callback from receiving mutable context object

### LINKED ISSUE #3336: Object passed to attrs factory is mutated
I'm running

```
const Sc = styled.div.attrs((props) => { ... })`...`
```

I had some failing tests, and it took me a long time to figure out why the attrs factory (`(props) => { ... }`) was claiming to have been called with arguments it had definitely not been called with. There's (what I call) a bug in styled-components in that the object getting passed as the first parameter to the attrs factory is being mutated.

## Environment
styled-components 5.2.1 with React 17.0.1 running in codesandbox.

## Reproduction
https://codesandbox.io/s/styled-component-playground-forked-9ocfx?file=/src/index.js

## Steps to reproduce
1. Run the sandbox example
2. See its console output

## Expected Behavior
I expect the object which the factory receives not to change. In the sandbox example I want to see this in the console:
```
props sent to attrs {"id":"id-prop","theme":{}}
after render, that same object is {"id":"id-prop","theme":{}}
```
In particular, the props explicitly passed to the component plus the theme prop, i.e. the props which ended up passed to the attrs factory.

## Actual Behavior
The object passed to the attrs factory is mutated. We see this on the console:
```
props sent to attrs {"id":"id-prop","theme":{}}
after render, that same object is {"id":"overridden-id-prop","theme":{},"newProp":"new-prop"}
```
This is the final props for the component after the return value from the factory has been merged with the explicitly-passed props.

## Why is this a problem
This gives me false negatives when running tests. I am mocking the factory function, and I expect it to be called with particular values. It *is* called with those particular values, but then when I run an assertion against the mock I'm told that it was in fact called with different values, since the `myMock.mock.calls[0][0]` object has been mutated to the merged props.

quantizor:
## Summary
Pass a shallow copy of the context to function `.attrs()` callbacks so the captured reference isn't mutated by subsequent attrs processing. The context object is still mutated internally for chaining—only the snapshot passed to callbacks is frozen at the point of invocation.

## Before
```tsx
let captured: any;
const Comp = styled.div
  .attrs(props => { captured = props; return { 'data-first': 'yes' }; })
  .attrs(() => ({ 'data-second': 'yes' }))``;

render(<Comp id="test" />);
captured['data-first']  // 'yes' — mutated after callback returned!
captured['data-second'] // 'yes' — mutated by second attrs!
```

## After
```tsx
captured['data-first']  // undefined — snapshot was taken before mutation
captured['data-second'] // undefined
```

Performance note: `{ ...context }` spread is 4x faster than `Object.assign` (validated in AGENTS.md). One extra allocation per attrs level per render is negligible for the typical 1-2 attrs chain.

## Test plan
- [x] 504 tests pass (503 existing + 1 new)
- [x] Red/green verified (test fails without fix, passes with)

Closes #3336

### DIFF
--- a/.changeset/fix-attrs-mutation.md
+++ b/.changeset/fix-attrs-mutation.md
@@ -1,0 +1,5 @@
+---
+"styled-components": patch
+---
+
+Fix `.attrs()` callback receiving a mutable props object that could be changed by subsequent attrs processing. The callback now receives an immutable snapshot.
--- a/packages/styled-components/src/models/StyledComponent.ts
+++ b/packages/styled-components/src/models/StyledComponent.ts
@@ -87,7 +87,9 @@ function resolveContext<Props extends BaseObject>(
 
   for (let i = 0; i < attrs.length; i += 1) {
     attrDef = attrs[i];
-    const resolvedAttrDef = isFunction(attrDef) ? attrDef(context) : attrDef;
+    // Pass a shallow copy to function attrs so the callback's captured
+    // reference isn't mutated by subsequent attrs processing (#3336).
+    const resolvedAttrDef = isFunction(attrDef) ? attrDef({ ...context }) : attrDef;
 
     for (const key in resolvedAttrDef) {
       if (key === 'className') {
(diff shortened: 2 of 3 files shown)


tremby:
Thanks.

Issue #5674, opened 2026-02-25
Performance degradation with tests and 6.3.x


spdaley:
We recently upgraded our project to styled-components v6.3.x (currently at 6.3.9).
After doing so we noticed that our tests became drastically slower. We were able to isolate it to our dependency update to 6.3.9 and downgrading back to 6.2.0 resolved it.

Working with codex, I was able to create a smallish repo with a benchmark test that seems to confirm the performance issue.
The repo can be found at https://github.com/spdaley/styled-components-jest-repro

To run it you would do:
1. yarn install (we use yarn v1)
2. yarn set:sc:620
3. yarn bench:compare (to get a benchmark for 6.2.0)
4. yarn set:sc:639
5. yarn bench:compare (to get a benchmark for 6.3.9)

The benchmark does two things:
1. It runs the test 10 times and grabs the timings for each run and gets some stats for those timings.
2. It then re-runs the test 10 times but with `createGlobalStyle` mocked to a no-op
3. It then compares the times

On my laptop (Macbook pro running Sequoia) I got the following times (fairly consistently within a % or so each time I ran)
* For 6.2.0
```
baseline
  runs: 10
  median: 0.741s
  min: 0.726s
  max: 1.258s

SC_DISABLE_GLOBAL_STYLE_MOCK=1
  runs: 10
  median: 0.725s
  min: 0.715s
  max: 0.767s

median delta: 0.016s (2.1%)
```

* For 6.3.9
```
baseline
  runs: 10
  median: 0.976s
  min: 0.964s
  max: 1.435s

SC_DISABLE_GLOBAL_STYLE_MOCK=1
  runs: 10
  median: 0.730s
  min: 0.723s
  max: 0.748s

median delta: 0.246s (25.2%)
```

On my laptop, we are seeing about a 25% performance degradation in that test when using 6.3.9.

Any help on this would be greatly appreciated. Maybe there is something we are doing wrong that we should be doing with 6.3.x now?
Thanks.

spdaley:
We've been able to work around this by adding the following to a jest setup file:

```
if (typeof globalThis.SC_DISABLE_SPEEDY === 'undefined') {
  globalThis.SC_DISABLE_SPEEDY = false;
}
```

aelfric:
What does `SC_DISABLE_SPEEDY` do?  I noticed some tests were sporadically failing when I tried to upgrade an application to styled-components 6.3 and this workaround seems to fix the problem, but I'm curious why.

[...]

PR #5694, opened 2026-03-19
perf: skip effect re-runs for static createGlobalStyle

quantizor:
## Summary
Static `createGlobalStyle` components (no interpolations) now use a narrow deps array `[instance, styleSheet]` instead of `[instance, props, styleSheet, theme, stylis]`. This prevents the effect from re-running on every render, eliminating the `removeStyles`+`renderStyles` cycle that was causing ~25% test performance degradation.

Dynamic global styles (with interpolations) retain the full deps array.

## Why the previous fix (#5679) wasn't enough
PR #5679 eliminated double `renderStyles` (component body + effect), but the effect still fired on every render because `props` is a new object reference each time. For static global styles, `renderStyles` exits early via `hasNameForId`, but the effect setup/teardown overhead and `removeStyles` cleanup still ran unnecessarily.

## Ref
#5674 — user confirmed 6.4.0-prerelease.0 still showed 27% delta

### DIFF
--- a/.changeset/perf-static-global-style.md
+++ b/.changeset/perf-static-global-style.md
@@ -1,0 +1,5 @@
+---
+"styled-components": patch
+---
+
+Skip unnecessary effect re-runs for static `createGlobalStyle` components. Static global styles now only inject once and clean up on unmount, instead of re-running the effect on every render.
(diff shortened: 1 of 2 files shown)


quantizor:
Available to test at `styled-components@6.4.0-prerelease.1`

PR #5696, opened 2026-03-20
fix: shared-group global styles and eager keyframes registration

### LINKED ISSUE #3008: Global styles does not work in React.StrictMode

### LINKED ISSUE #3022: createGlobalStyle creates memory leak in Next.js

### LINKED ISSUE #3146: CreateGlobalStyles Rendered After Component Styles

### LINKED ISSUE #5695: createGlobalStyle stylesheet removed when a second instance unmounts, breaking styles on still-mounted instances
When multiple instances of a component that renders a createGlobalStyle-based component are mounted, unmounting any one of them removes the injected stylesheet entirely — even though other instances are still mounted and still depend on those styles.

Live reproduction: https://codesandbox.io/p/sandbox/sc-global-style-repro-forked-tw9tgn

Environnement:
- react: 18.3.1
- styled-components: 6.3.12

quantizor:
## Summary

Rearchitects `createGlobalStyle` to use shared stylesheet groups, fixing a recurring bug family spanning 7 years and 5+ issues (#5695, #3146, #3008, #3022, #2769).

**Root cause:** Each mounted instance of a `createGlobalStyle` component previously allocated its own stylesheet group (`componentId + instanceNumber`). This coupling between instance lifetime and stylesheet slot lifetime meant any mount/unmount cycle could scatter groups across the stylesheet or orphan them. The bug was "fixed" at least 3 times but kept reappearing.

**Fix:** All instances now share ONE group, registered once at definition time. An `instanceRules` cache enables selective removal via clear-and-rebuild.

### Ordering algorithm

CSS injection order is now fully deterministic:

1. When `styled()`, `createGlobalStyle()`, or `keyframes()` is called, a group ID is allocated from a global counter
2. Lower group IDs always appear earlier in the stylesheet
3. Render order does not affect CSS order — only definition order matters
4. Global style instances share their definition's group, so remounting always returns to the original position
5. Keyframes defined before a component now correctly appear before that component's rules

### Changes

- **`GlobalStyle.ts`** — shared group with `instanceRules` Map, `computeRules` helper, `rebuildGroup` (clear + re-insert survivors), rules-equal fast-path to skip CSSOM rebuild when CSS is unchanged, rehydration cache-seeding for static styles
- **`Keyframes.ts`** — eager group registration via `getGroupForId` in constructor (uses GroupIDAllocator directly to keep native builds DOM-free)
- **`createGlobalStyle.ts`** — `useRef`-based instance IDs, RSC per-instance style output from cache instead of shared group
- Bundle size: 12.35KB → 12.4KB (+50 bytes, +0.4%)

### Breaking changes

- Keyframes CSS now appears _before_ components that reference them (when keyframes are defined first). Semantically correct but changes observable order.
- SSR rehydration markers changed from `id="sc-global-X1"` to `id="sc-global-X"`. Apps upgrading with cached SSR output may see one-time style duplication until caches refresh.

## Test plan

- [x] 568 tests pass across 53 suites (+30 new tests)
- [x] Multi-instance mount/unmount in every permutation order (2→1→0, 0→1→2, 1→0→2)
- [x] Definition order permutations (A→B→C rendered as C→A→B, B→C→A)
- [x] Rehydration: static and dynamic styles surviving instance unmount + prop change after
- [x] StrictMode double-render with static and dynamic interpolations
- [x] RSC: multi-instance, themed, global+styled in same tree, independent render passes
- [x] RSC coverage for StyleSheetManager, ThemeProvider, withTheme, checkDynamicCreation
- [x] Build passes with bundlewatch size check
- [x] Native build: 0 `document.` references (DOM isolation verified)
- [x] Performance validated via microbenchmarks (rules-equal fast-path saves ~530ns CSSOM ops at 4-7ns comparison cost)

Fixes #5695. Fixes #3146. Fixes #3008. Fixes #3022.


quantizor:
This PR is testable via `styled-components@6.4.0-prerelease.2`, aliased to `styled-components@test`.

PR #5757, opened 2026-05-30
fix(native): animate multi-argument transform shorthands

quantizor:
In React Native apps (both native and react-native-web), animating a multi-argument transform shorthand inside `@keyframes` or a `transition` previously threw `Transform with key of translate must have an array as the value` and broke the component. This covers the shorthand forms `translate(x, y)`, `translate3d(x, y, z)`, and the two-argument `scale(x, y)`.

These now animate correctly, interpolating each axis independently. Single-axis forms (`translateX`, `translateY`, `scaleX`, `scaleY`) and uniform `scale(n)` were already fine and are unchanged.

```js
const Drift = styled.View`
  @keyframes drift {
    0%   { transform: translate(0px, 0px)    scale(1, 1); }
    50%  { transform: translate(26px, -12px) scale(1.4, 0.7); }
    100% { transform: translate(0px, 0px)    scale(1, 1); }
  }
  animation: drift 2.4s ease-in-out infinite;
`;
```

A matching "drift" cell was added to the native showcase's keyframe examples so the behavior is verifiable across iOS, Android, and react-native-web.

### DIFF
--- a/.changeset/native-animated-translate-shorthand.md
+++ b/.changeset/native-animated-translate-shorthand.md
@@ -1,0 +1,5 @@
+---
+'styled-components': patch
+---
+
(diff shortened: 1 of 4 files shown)


quantizor:
Good catch on the `scale(1,)` NaN path. Fixed: a missing or malformed axis in a 2-argument `scale` now falls back to the other axis (uniform) for `y` and to identity for `x`, with regression tests for both `scale(1,)` and `scale(,2)`.

PR #5761, opened 2026-06-08
fix: call useDebugValue unconditionally past the render-cache branch

alelode:
Fixes a rules-of-hooks violation introduced in #5704. `useDebugValue(generatedClassName)` was inside `useInjectedStyle`, which is only called on a cache miss. On a cache-hit re-render the call was skipped, producing an inconsistent hook count between renders. Moved it unconditionally after the cache if/else block in `useStyledComponentImpl`.


quantizor (approved):
Thank you!

PR #5768, opened 2026-07-18
fix(types): cut type-instantiation cost for polymorphic styled components

### LINKED ISSUE #5767: 6.4.3: `PolymorphicCallProps` rewrite causes 57× more `ComponentPropsWithRef` instantiations, +1.6 GB tsc heap in our app
Hi, thanks for maintaining styled-components. 

Filing this because in our monorepo the 6.4.2 → 6.4.3 bump added enough tsc pressure to start OOMing our dev-server type checker. Feel free to close if you knew about the trade-off and decided it was worthwhile so this is WAI. 

  We bisected on the same commit and same lockfile, swapping only styled-components between 6.4.2 and 6.4.3, running `tsc --extendedDiagnostics --noEmit` cold each time (deleted `.tsbuildinfo` before every run, alternated the order A→B→A→B to rule out caching artifacts):

  | Metric | 6.4.2 (run 1) | 6.4.3 (run 1) | 6.4.2 (run 2) | 6.4.3 (run 2) |
  |---|---:|---:|---:|---:|
  | Types | 1,487,074 | 1,892,984 | 1,487,074 | 1,892,984 |
  | Instantiations | 33,223,024 | 47,784,078 | 33,223,024 | 47,784,078 |
  | Symbols | 3,701,255 | 3,703,629 | 3,701,255 | 3,703,629 |
  | Memory used | 3,427 MB | 5,244 MB | 3,555 MB | 5,133 MB |
  | Check time | 70.0s | 91.4s | 70.9s | 103.2s |

  Structural metrics are bit-identical per-version; memory/time jitter run-to-run as expected. Net effect of the 6.4.2 → 6.4.3 bump:

  - **+406,282 types (+27%)**
  - **+14,561,054 instantiations (+44%)**
  - **+1.6 GB heap (+45%)**
  - **+22–33s check time (+30–45%)**

  We used `--generateTrace` to attribute the growth. It is not spread across the codebase — it is concentrated in a single symbol:

  | Symbol (from `@types/react`) | 6.4.2 | 6.4.3 | Delta |
  |---|---:|---:|---:|
  | `ComponentPropsWithRef` | 4,428 | 252,248 | **+247,820 (57×)** |

  By flag: **+258,286 new `Conditional` types** and **+146,454 new `Intersection` types**, essentially entirely rooted in `@types/react` instantiations from styled-components.

  Comparing `styled-components/dist/types.d.ts` between the two versions, the change is the new `PolymorphicCallProps` alias and the widened `AsTarget` bound in `PolymorphicComponentProps` (referenced in your own comment as ~6% more instantiations across call sites):

  ```ts
  // 6.4.3 additions
  export type PolymorphicComponentProps<R, BaseProps,
    AsTarget extends StyledTarget<R> | (BaseProps extends { as?: infer A } ? A : never) | void,
    ForwardedAsTarget extends StyledTarget<R> | void,
    AsTargetProps extends BaseObject = AsTarget extends KnownTarget ? React.ComponentPropsWithRef<AsTarget> : {},
    ForwardedAsTargetProps extends BaseObject = ForwardedAsTarget extends KnownTarget ? React.ComponentPropsWithRef<ForwardedAsTarget> : {},
  > = ...

  type PolymorphicCallProps<R, BaseProps, AsTarget, ForwardedAsTarget> =
    [AsTarget] extends [string | AnyComponent] ? PolymorphicComponentProps<...> & { as: AsTarget }
    : [ForwardedAsTarget] extends [void] ? OverrideStyle<...>
    : PolymorphicComponentProps<R, BaseProps, void, ForwardedAsTarget> & { forwardedAs: ForwardedAsTarget };
  ```

  Every JSX usage of a styled component now goes through the three-branch conditional, and each branch that lands on the target-props branch re-instantiates `React.ComponentPropsWithRef` (already a known TS hotspot). The comment on the type acknowledges the tradeoff for
  `KnownTarget` narrowing but the multiplier we see in a real app is much larger than 6%.

  **Scale of usage on our side:** a large SPA, heavy `styled(Component)` and `styled.div` usage across ~2000 components.

  **Repro:** any large codebase with heavy `styled()` usage; alternate installs of 6.4.2 and 6.4.3 with `.tsbuildinfo` deleted each time, run `tsc --extendedDiagnostics`, and diff the numbers. `ComponentPropsWithRef` count via `--generateTrace` should show the jump.

  **Mitigation on our side:** pinned to 6.4.2 for now.

quantizor:
Fixes #5767.

6.4.3 introduced a TypeScript type-checking regression: every `styled(...)` call site began routing through a single three-way conditional whose `as`-target branch is instantiated even for plain, non-polymorphic usage. In codebases heavy on `styled(Component)` wrappers and polymorphic `as`, this inflated `tsc` instantiation counts and memory, up to out-of-memory in the worst cases.

This restructures the polymorphic call-props resolution so plain call sites stop paying for the `as`-target branch, with no change to the emitted types, the public API, or runtime behavior.

## Impact (synthetic fixtures, TS 5.9.3)

On a workload representative of the report (reused `styled(Component)` over wide discriminated-union props, plus `as` usage):

| | Instantiations | Memory |
| --- | --- | --- |
| 6.4.3 | 4.36M | 1022 MB |
| this PR | 3.05M (-30%) | 940 MB |
| 6.4.2 (pre-regression) | 2.96M | 782 MB |

That recovers ~94% of the regression on that shape. A plain-`styled.tag`-heavy workload shows a smaller -4.4%, since those sites barely regressed to begin with.

## How

The three-way conditional is split into two intersected conditionals whose untaken side is `unknown`. While the `as`/`forwardedAs` type parameters are still uninferred, the checker probes each conditional through its constraint union, and `union(branch, unknown)` absorbs to `unknown`, so plain sites no longer instantiate the heavy `as`-target branch on every probe. After inference the result is identical, because `X & unknown` reduces to `X`. A flat `{ as?; forwardedAs? }` member preserves per-attribute contextual typing (literal `as="video"` inference), and both discriminants are kept positive so spreading a styled component's own props stays branch-safe for ref-callback inference. The structure is load-bearing and documented inline so it is not "simplified" back into the slower form.

## Verification

No test changes were needed: behavior is identical. The full `test:types` suite, the ref-callback and polymorphic `as` type contracts, and JSX attribute autocompletion (including mid-typing `as`-target completion) all pass unchanged. The package's own type-check is clean.

### DIFF
--- a/.changeset/polymorphic-callprops-type-perf.md
+++ b/.changeset/polymorphic-callprops-type-perf.md
@@ -1,0 +1,5 @@
+---
+"styled-components": patch
+---
+
+Reduce TypeScript type-checking cost for styled components, most noticeably `styled(Component)` wrappers and polymorphic `as` usage. Large codebases that saw elevated `tsc` memory and type-instantiation counts get lower type-check memory and time, with no change to the emitted types or runtime behavior.
--- a/packages/styled-components/src/types.ts
+++ b/packages/styled-components/src/types.ts
@@ -179,6 +179,11 @@ export interface IStyledStatics<
     | undefined;
 }
 
+/** ExecutionProps sans as/forwardedAs, pre-resolved so call sites relate against a concrete interface. */
+interface ThemedExecutionProps {
+  theme?: DefaultTheme | undefined;
+}
+
 /**
  * Used by PolymorphicComponent to define prop override cascading order.
  */
@@ -206,7 +211,7 @@ export type PolymorphicComponentProps<
       keyof ExecutionProps
     >
   > &
-    FastOmit<ExecutionProps, 'as' | 'forwardedAs'> & {
+    ThemedExecutionProps & {
       as?: AsTarget;
       forwardedAs?: ForwardedAsTarget;
     }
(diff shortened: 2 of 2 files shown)


eMerzh:
Thanks for the fix .. here the cost of the "as" prop is still pretty high, as we don't really use it we've patch style-component ... something looking like

```diff
diff --git a/dist/types.d.ts b/dist/types.d.ts
index 98a2c4d0f5bb111ba6523c9193028f7a69512179..46f7524b2c3346ad659227ac1480f375210cb3a9 100644
--- a/dist/types.d.ts
+++ b/dist/types.d.ts
@@ -174,10 +174,7 @@ export interface PolymorphicComponent<out R extends Runtime, in out BaseProps ex
     as?: StyledTarget<R> | undefined;
     forwardedAs?: StyledTarget<R> | undefined;
 }> {
-    <AsTarget extends StyledTarget<R> | (BaseProps extends {
-        as?: infer A;
-    } ? A : never) | void = void, ForwardedAsTarget extends StyledTarget<R> | void = void>(props: PolymorphicCallProps<R, BaseProps, AsTarget, ForwardedAsTarget>): React.JSX.Element;
-}
+} // patch: generic `as`/`forwardedAs` call signature removed, it made tsc ~12x slower
 /**
  * Some wrapped targets can't be statically introspected and their props
  * collapse to `{}` -- most notably polymorphic-factory components (e.g. Mantine
```

as the comment says, tsc then goes from 60s ish to 5s ish for our code base...

i was wondering is a solution with a global type flag to enable/disable it wouldn't be interesting.

quantizor:
> Thanks for the fix .. here the cost of the "as" prop is still pretty high, as we don't really use it we've patch style-component ... something looking like
> 
> ```diff
> diff --git a/dist/types.d.ts b/dist/types.d.ts
> index 98a2c4d0f5bb111ba6523c9193028f7a69512179..46f7524b2c3346ad659227ac1480f375210cb3a9 100644
> --- a/dist/types.d.ts
> +++ b/dist/types.d.ts
> @@ -174,10 +174,7 @@ export interface PolymorphicComponent<out R extends Runtime, in out BaseProps ex
>      as?: StyledTarget<R> | undefined;
>      forwardedAs?: StyledTarget<R> | undefined;
>  }> {
> -    <AsTarget extends StyledTarget<R> | (BaseProps extends {
> -        as?: infer A;
> -    } ? A : never) | void = void, ForwardedAsTarget extends StyledTarget<R> | void = void>(props: PolymorphicCallProps<R, BaseProps, AsTarget, ForwardedAsTarget>): React.JSX.Element;
> -}
> +} // patch: generic `as`/`forwardedAs` call signature removed, it made tsc ~12x slower
>  /**
>   * Some wrapped targets can't be statically introspected and their props
>   * collapse to `{}` -- most notably polymorphic-factory components (e.g. Mantine
> ```
> 
> as the comment says, tsc then goes from 60s ish to 5s ish for our code base...
> 
> i was wondering is a solution with a global type flag to enable/disable it wouldn't be interesting.

I'll look into it 👍

PR #5771, opened 2026-07-18
docs: distill v7 agent instructions and materialize stack knowledge

quantizor:
AGENTS.md carried the project's own mechanism and invariants well, but the universal engineering rules that bind anyone touching the repo lived only in a maintainer's private config. Cloud review agents, headless runs, and outside contributors read only what's in the tree, so they never saw them. This ports those rules into the repo and materializes the external-stack knowledge a maintainer of this stack actually needs.

## AGENTS.md

Folded the universal rules into the existing sections (no second, drift-prone copies) plus two new sections:

- Mandates now state types-are-law (no `as any` / non-null `!` / `@ts-expect-error` except a marked negative type test), backward-compat only on public API surfaces, snapshot discipline (never snapshot class strings; normalize volatile bytes), and the coverage/fast-suite bar.
- New `Method` section: fix at the cause not the symptom (with the suppression forms enumerated), adversarial self-play before calling work done, ground claims in fresh evidence, one-fact-one-home.
- New `Security posture` section: the parser trust model (three structural guards against interpolation escaping) and the ReDoS regex discipline for the parser and native transforms.
- Basics/Agent Rules gained prettier-as-authority, git safety (push consent, no stash, lockfile), no attribution footer, block-comments-for-hover-docs, and field alphabetization.

## New reference docs

External-stack knowledge, one file per area, wired into Topical references:

- `docs/react.md` — React 19 behaviors that hit the render hot path and SSR/RSC output (ref-as-prop allocation, `renderToString` no longer encoding CSS characters, `act()`, the RSC `:nth-child` hazard).
- `docs/react-native.md` — RN's CSS capability matrix by version, the Yoga `flex-shrink` default divergence, and the in-house declaration transform.
- `docs/jest.md` — jsdom's modern-CSS ceiling (the suite runs jsdom 26.x) and how the web harness reads injected CSS back off the live CSSOM sheet.
- `docs/bundling.md` — per-bundle build-constant substitution, the terser module-scope DCE trap, and the `.d.ts`/TS2742 packaging pitfall.
- `docs/typescript-performance.md` — added a section on testing the type surface (`@ts-expect-error` negatives, project-wide `declare module` isolation, src-vs-dist).

## Drift fixes

Both verified against current code:

- The repo-root `src/utils/errors.md` rule claimed it mirrors the canonical file and to edit it in place. It is actually a frozen pre-monorepo stub that self-declares `DO NOT EDIT` and that `generateErrors` never reads. The rule now points at the real canonical and warns off the dead file.
- `IS_BROWSER` was documented as `typeof window !== 'undefined'`; it is `window && document`. Corrected to state the invariant and point to the bundling doc rather than keep a drifting copy of the expression.


quantizor:
Folding this into #5769 so it lands as one PR.

Issue #5812, opened 2026-09-08
[v7 native] env(safe-area-inset-*) always resolves to 0; SafeAreaProvider is not wired


midasstack:
## Environment

- styled-components: v7 (current main / `@test` dist-tag)
- react-native: >= 0.85
- Platform: iOS / Android (Hermes native entry)
- Related docs: https://styled-components.com/docs/compatibility (`env(safe-area-inset-*)` row)

`npx envinfo` omitted: this is a library-source bug reproducible from the native entry without an app-specific setup.

## Reproduction

Minimal:

```tsx
import styled from 'styled-components/native';
import { SafeAreaProvider } from 'react-native-safe-area-context';

const Pad = styled.View`
  padding-top: env(safe-area-inset-top);
  padding-bottom: env(safe-area-inset-bottom);
`;

export function Screen() {
  return (
    <SafeAreaProvider>
      <Pad>{/* content under the notch */}</Pad>
    </SafeAreaProvider>
  );
}

PR #5813, opened 2026-09-08
fix(native): resolve env(safe-area-inset-*) from SafeAreaProvider

midasstack:
Fix related to https://github.com/styled-components/styled-components/issues/5812


quantizor:
Thanks for this, it's a clean implementation and the resolver wiring is exactly where it needed to go.

I opened #5814 as a small stack on top of this branch. It adds a `SafeAreaInsetsProvider` (exported from `styled-components/native`) so insets can come from any source, a custom native module, a web host, or tests, and it works even when `react-native-safe-area-context` isn't installed. Precedence in `useSafeAreaInsets` becomes explicit provider, then the peer context, then zeros. It also folds the peer-missing dev warning into the hook so an explicit provider suppresses it, and bumps the peer floor to `>= 5.0.0` to match the React 19 / New Architecture baseline v7 targets.

Plan is to land this one first, then #5814 follows. Appreciate the thorough work here.

quantizor (approved):

